Issue #3, opened 2025-05-22
Upgrade and Stuff to Do


carlost:
## Stuff to do:

Need to split this into a lot of separate issues

### Upgrade Stuff That Needs To Be Done In Order

- [x] Upgrade Ruby to 2.7.x
- [x] Upgrade Rails 5.0 -> 5.1
- [x] Upgrade Rails 5.1 -> 5.2
- [x] Upgrade to Ruby 3.0.x
- [x] Upgrade to Rails 5.2 -> 6.0
- [x] Upgrade Rails 6.0 -> 6.1
- [x] Upgrade Ruby 3.3.x
- [x] Upgrade Rails 6.1 -> 7.0
- [x] Upgrade Rails 7.0 -> 7.1
- [x] Upgrade Ruby to 3.4.x
- [x] Upgrade Rails 8.0
- [ ] ~~Remove devise and use the rails 8 authentication~~
- [x] Add in the Hotwire stack (turbo/stimulus)
- [ ] ~~Convert all js to stimulus and make it modern~~

### Other Tasks That Can Be Done In Any Order

- [ ] GitHub actions for ci/cd (partially tracked by #55)
- [ ] Issue templates
- [x] PR templates
- [x] License File (MIT)
- [x] Code of Conduct (@sampart)
- [x] Contributing File
- [x] Search repo/git for secrets/keys/passwords/etc: https://github.com/rubyforgood/awbw-dashboard/issues/14
- [x] Update/Modernize Readme
- [x] Add in rubocop-rails-omakase (done in #602 )
- [x] Add in brakeman (done in #110 )
- [x] Add in bundle-audit (done in #110 )
- [x] Any others we're missing
- [x] Audit Gemfile
- [x] Add tests
- [x] Decide how js/assets will be done in rails 8
- [ ] Remove Capistrano and switch to Kamal
- [x] Move from paperclip to active storage (Requires Rails 5.2 or higher) (tracked by #17 )
- [x] Evaluate if we can move from ckeditor to action text 
- [ ] Tests, tests, tests
- [x] General cleanup - lots of stuff everywhere
- [x] ~~Can we drop monthly_reports & workshop_logs tables from schema~~

General notes:
- bin/rails app:update
- should we just skip webpacker?
- official rails upgrade guide starting at 4.2 to 5: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0

carlost:
[...]

PR #18, opened 2025-05-29
Upgrade to Rails 5.2.8.1

lorint:
- This work is part of https://github.com/rubyforgood/awbw/issues/3.

* Upgrade to Rails 5.2.8.1

* Ran bin/rails app:update and carefully merged changes into the code
* Identified gems which required updates
* Add the `/vendor` folder back into the code tree


carlost:
ok @lorint i don't know how you willed yourself to this point ... at some point when i was overwriting the history of the repo to open it up, i somehow deleted the vendor dir which included a few of the files you pieced back together:

Screenshot 2025-05-29 at 11 53 44 PM

whoops.

I am still trying to dig into this ckeditor issue

lorint:
> when i was overwriting the history of the repo to open it up, i somehow deleted the vendor dir

OK makes sense -- I've put that folder back into place and updated the **.gitignore** so we retain the /vendor folder.  Would be curious to know what happens now when we try to add a new workshop -- if ckeditor might now be happy happy :)

lorint:
> ok @lorint i don't know how you willed yourself to this point

I am a trauma abuse survivor, and want to help others to be able to recover from this.  Thankfully I am still alive, albeit with more limited physical ability.  My brain is still completely intact, anyway!

carlost:
@smhauck ^^^ any thoughts?

carlost:
I am guessing something has gone sideways in the rails 5.2.8.1 branch with the `belongs_to_required_by_default` config setting in the `new_framework_defaults.rb` initializer.

I am pretty beat and am going to hit the sack now. I would like to resolve this.

config/application.rb
```diff
@@ -8,6 +8,9 @@
 
 module Awbw
   class Application < Rails::Application
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 5.0
```
seanmarcia, line 12: Should this be `5.2`?

lorint: This needs to be set to the originally generated Rails version.  In our case it's really older than this, but this setting only began to be available starting with Rails 5.0.

We can move this forwards eventually, but at this point hoping to just stand everything up and getting it working such that this understands:
* the Rails version where we had generally originated, and
* with the code really using the Rails version that we'd like to upgrade to -- in this case 5.2.8.1.

config/puma.rb
```diff
@@ -1,20 +1,23 @@
 # Puma can serve each request in a thread from an internal thread pool.
-# The `threads` method setting takes two numbers a minimum and maximum.
+# The `threads` method setting takes two numbers: a minimum and maximum.
 # Any libraries that use thread pools should be configured to match
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
-# and maximum, this matches the default thread size of Active Record.
+# and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
```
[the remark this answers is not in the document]
lorint: Oh cool -- I'm open to setting it to a higher value.  (Ben Sheldon's thinking is always great really!)

I had just removed the `.to_i` because that's the way that it is done in all newer Rails projects by default.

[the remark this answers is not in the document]
carlost: We have no idea how the hell this application actually behaves and am comfortable with the default values for now.

config/puma.rb
```diff
@@ -1,20 +1,23 @@
 # Puma can serve each request in a thread from an internal thread pool.
-# The `threads` method setting takes two numbers a minimum and maximum.
+# The `threads` method setting takes two numbers: a minimum and maximum.
 # Any libraries that use thread pools should be configured to match
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
-# and maximum, this matches the default thread size of Active Record.
+# and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 20 }
```
carlost, line 7: I probably spend half of my day sniffing glue, but wasn't the sheldon article related to the AR connection pool config?

I have no idea what type of load this service receives in production. perhaps we need to set these sorts of options based on how the passenger <-> httpd web server is configured now.

I suggest that we reverse this RAILS_MAX_THREADS value back to 5 and add an issue to compare the current production passenger config to the puma config.

thoughts?

config/puma.rb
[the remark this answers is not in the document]
lorint: > We have no idea how the hell this application actually behaves

Your current work with Rudy will shed light on all of that.  The space is not nearly as darkly illuminated as you might think.

config/puma.rb
lorint: I'm perfectly cool to dial it back to 5.  That's the way in which David still leaves it for default Rails projects.

PR #122, opened 2025-09-13
Add simple_form gem

caseyvaughn:
Adding the simple_form gem configured with bootstrap to improve form building


maebeale (approved):
great call on adding this gem!

Issue #139, opened 2025-09-13
Upgrade to Rails 8.1 beta


marc:
Upgrade to rails 8.1.beta1.

Incrementally upgrade to rails 7.0, 7.1, 7.2 and 8.0 to get there.

maebeale:
@marc this was done, right!?

PR #140, opened 2025-09-13
Upgrade to Rails 8.1 beta

marc:
- This work is part of #139 during Ruby for Good 2025!

We are upgrading to rails 8.1.beta1 to bring the framework up to date.

I upgraded to rails 7.0, 7.1, 7.2 and 8.0 and made sure the tests were passing and the app looked good for users and admins.
Running everything in docker containers.

### Anything else to add?
We want to try this in staging to make sure it works well outside of the dev containers.
Thanks @johnpaulashenfelter for the commits and helping with the upgrade!


Gemfile
```diff
@@ -2,34 +2,27 @@ source 'https://rubygems.org'
...
-gem "bigdecimal"
-gem "drb"
-gem "mutex_m"
-gem "observer"
-
+gem "rails", "8.1.0.beta1"
 
 gem 'sprockets-rails', '~> 3.2.2'
-gem "activerecord-trilogy-adapter" # for mysql installation
+gem 'trilogy'
```
maebeale, line 8: wasn't this already in there?

maebeale (approved):
🦾 wtg!!!

Gemfile
marc: I think trilogy was indirectly loaded as a dependency for activerecord-trilogy-adapter. That is no longer needed as of rails 7.1+, so I added just the trilogy adapter.

Issue #61, opened 2025-09-09
Facilitators see "Resources" in new top nav


leelyuh:
**As** a facilitator,
**I want** "Resources" to move up into the top nav
**So that** it's easier to find.

**Given** I have signed in to AWBW,
**When** the page loads,
**Then** I see "Resources" appear in the top nav to the right of "Workshops"
**And** "Resources" no longer appears in the side nav.

**When** I click on "Resources",
**Then** I am taken to the existing "Resources" page.

PR #149, opened 2025-09-13
move resources to top navbar

mlwzrd:
- This work is part of #61 

Moving resources link to top navbar

Moved it, also fixed style for header links


maebeale (approved):
⏫ woot!

Issue #91, opened 2025-09-12
Organization CRUD


dalmaboros:
As a facilitator,
I want to see all the other facilitators who work at the same organization as me
So that I know who else is actively facilitating at my org
But there's no way for me to see this today.

Fields:

1) Organization name*
(list your program name as it appears on the organization’s website)

2) Start Date (internal only)

3) Close Date (internal only)

4) Status Active / Inactive (internal only)

5) Agency Address*
Street Address
City 
State
ZIP / Postal Code
Country

6) County (internal only)
Alameda
Alpine
Amador
Butte
Calaveras
Colusa
Contra Costa
Del Norte
El Dorado
Fresno
Glenn
Humboldt
Imperial
Inyo
Kern
Kings
Lake
Lassen
Los Angeles
Madera
Marin
Mariposa
Mendocino
Merced
Modoc
Mono
Monterey
Napa
Nevada
Orange
Placer
Plumas
Riverside
Sacramento
San Benito
San Bernardino
San Diego
San Francisco
San Joaquin
San Luis Obispo
San Mateo
Santa Barbara
Santa Clara
Santa Cruz
Shasta
Sierra
Siskiyou
Solano
Sonoma
Stanislaus
Sutter
Tehama
Trinity
Tulare
Tuolumne
Ventura
Yolo
Yuba

7) Locality (internal only) 
LA City
LA County
Southern CA
Northern CA
Central CA
Orange County
Outside CA
Outside USA
(Keep these values in this order)

Note: eventually do this automatically
    * If “Los Angeles” is listed in the City field, the Locality is LA City
    * If “Los Angeles” is listed in the County field, the Locality is LA County
    * If “Orange” is listed in the County field, the Locality is Orange County
    * If the state is CA but neither “Los Angeles” nor “Orange” is listed in the City/County fields, the locality is either Southern CA, Northern CA, or Central CA. Google the city to find out what part of California it’s in and put this in Locality.

If County is LA County, show the following 3 fields (8-10):

8) LA City Council District (internal only)
	1-15
9) LA Supervisorial District (internal only)
	1-5
10) LA Service Planning Area (internal only)
	1-8
￼
11) Agency Website URL (public optional)

12) Agency Type*
501c3/nonprofit
For-profit
Government agency
Other (please specify below)

13) If Agency Type is “Other”, “please specify agency type” field (required only if agency type is “Other”)

14) Phone Number*

15) Mission/Vision Statement (internal AND public optional)

[...]

PR #153, opened 2025-09-13
Organization CRUD

dalmaboros:
- This work is part of https://github.com/rubyforgood/awbw/issues/91

Adding Organization model so we can implement CRUD, ultimately so facilitators can see their organizations and other facilitators at their organizations.

Added Organization and Address as models, plus associated specs and factories. Also associations between Organization and Address and Organization and Facilitator.


maebeale (approved):
🏛️ thanks for taking on this Organization stuff!

PR #159, opened 2025-09-13
vite

cflipse:

Add and configure vite-rails.

[Vite] is modern asset builder / bundler that we will use for
javascript and CSS assets going forward.

This also enables a second rails docker container that acts
as an asset host, supporting live reloading within the rails
application structure.

As it currently stands, the now-legacy sprockets / asset pipeline
configurations are loading at the same time within the main application
layout.  We may want to consider having two separate layout pages and
using that in building out new functionality, and opting-in legacy pages
when they've migrated away from using the features of the old asset pipleine

[Vite]: https://vite-ruby.netlify.app/

- **remove coffee-rail**
- **add vite and run vite install**
- **dev dockerfile adds dev**
- **Production dockerfile also copy**
- **Add the missing .dockerignore**
- **generate bundler binstub to make vite binstub happy**
- **update to use vite-rails plugin**
- **run a vite dev server inside docker**

- This work is part of [link an issue]


maebeale (approved):
wooooo hooooo! thanks, @cflipse !!! 💥

PR #168, opened 2025-09-14
make warden stop randomly breaking tests when devise runs early

cflipse:


johnpaulashenfelter (approved):
Thanks to ChatGPT we found a config we missed.

JP theorizes that it has something to do with impersonation stuff? Chris and he are both just angry about it. But it makes sense.

PR #173, opened 2025-09-14
Force charset to utf8mb4 and utf8mb4_unicode_ci charset/collation

johnpaulashenfelter:


config/database.yml
```diff
@@ -1,19 +1,20 @@
-base: &base
-  encoding: utf8
-  url: <%= ENV.fetch("DATABASE_URL", "trilogy://root@localhost") %>
+default: &default
+  adapter: trilogy
+  encoding: utf8mb4           # <-- forces UTF-8 everywhere
+  collation: utf8mb4_unicode_ci  # <-- portable across MySQL 5.7/8.0 and MariaDB
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5) %>
```
cflipse, line 5: I actually get slightly nervous about making the pool the same size as the rails thread count ... mostly as it's possible that the rails application logic can (potentially) throw things out to their _own_ background threads ... 

I'm not really sure just _how_ much this matters, but it's own of those things that makes me :hmmm:

config/database.yml
```diff
@@ -1,19 +1,20 @@
-base: &base
-  encoding: utf8
-  url: <%= ENV.fetch("DATABASE_URL", "trilogy://root@localhost") %>
+default: &default
+  adapter: trilogy
+  encoding: utf8mb4           # <-- forces UTF-8 everywhere
```
cflipse, line 3: :+1:

config/database.yml
```diff
@@ -1,19 +1,20 @@
...
+  collation: utf8mb4_unicode_ci  # <-- portable across MySQL 5.7/8.0 and MariaDB
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5) %>
+  user: root
+  password:
```
cflipse, line 7: These should either be overridable  ... or we should have some documentation in here *explicitly* pointing towards setting the `DATABASE_URL` envvar to override with local config.  Not everyone knows that's a thing.

db/schema.rb
```diff
@@ -920,7 +920,6 @@
     t.index ["year", "month"], name: "index_workshops_on_year_and_month"
   end
 
-  add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
```
maebeale: @copilot why is this foreign key being removed? add a commit to change it if it's wrong.

PR #211, opened 2025-09-26
Add workshop logs crud

maebeale:
- Allow users to CRUD workshop_logs in a restful and non-RailsAdmin way

- Added necessary views, actions, model methods, etc

### Anything else to add?
- This work required me to touch notifications some (an admin email gets sent after every workshop log is created)
- This work required me to touch the report model some (workshop_log inherits from report) bc there is a destructive action on quote_item_quotes after submitting workshop_log new/edit with nested quotes


app/controllers/workshop_logs_controller.rb
```diff
@@ -1,31 +1,25 @@
...
+    set_form_variables
   end
 
   def update
     set_default_values
     @workshop_log = WorkshopLog.find params[:id]
 
     ActiveRecord::Base.transaction do
-      @workshop_log.update_attributes( wsl_params )
+      @workshop_log.update(workshop_log_params)
```
maebeale, line 22: updating syntax per rails upgrade

app/controllers/workshop_logs_controller.rb
```diff
@@ -1,31 +1,25 @@
...
+  def index
+    # If @workshop is present, set params[:workshop_id] so search handles it
+    params[:workshop_id] ||= @workshop&.id
+    @workshop_logs_unpaginated = WorkshopLog.includes(:workshop, :user, :windows_type)
+                                            .search(params)
+    @workshop_logs = @workshop_logs_unpaginated.paginate(page: params[:page], per_page: 25)
+    set_index_variables
+  end
   def new
-    params[:windows_type_id] = 1;
```
maebeale, line 3: moved all of these declarations to `set_form_variables` method

app/controllers/workshop_logs_controller.rb
```diff
@@ -36,30 +30,24 @@ def update
...
     else
-      flash[:alert] = 'ERROR!'
+      set_form_variables
       render :edit
     end
   end
 
   def create
     set_default_values
-    @agencies = current_user.projects.uniq
```
maebeale, line 49: not needed unless create fails, so deleted these and added `set_form_variables` in the else clause

app/controllers/workshop_logs_controller.rb
```diff
@@ -79,7 +70,7 @@ def show
     @answers      = @workshop_log.report_form_field_answers
 
     if @workshop_log
-      if @workshop_log.project && current_user.project_ids.include?(@workshop_log.project.id)
+      if current_user.super_user? || (@workshop_log.project && current_user.project_ids.include?(@workshop_log.project.id))
```
maebeale, line 73: this line will need to be updated once we've completed the Organization body of work (projects were previously used for agencies aka organizations)

app/controllers/workshop_logs_controller.rb
```diff
@@ -110,30 +98,99 @@ def validate_new
...
+
+    @agencies = current_user.projects.uniq
+    agency = params[:agency_id].present? ? Project.where(id: params[:agency_id]).last : @agencies.first
+    @agency_id = agency.id if agency
+  end
+
   def set_default_values
     workshop_log = params[:workshop_log];
-    workshop_log.map do |k, v|
+    workshop_log.to_unsafe_h.map do |k, v|
```
maebeale, line 143: you can't map params in this way anymore

app/controllers/workshop_logs_controller.rb
```diff
@@ -110,30 +98,99 @@ def validate_new
...
   end
 
-  def wsl_params
-    params.require(:workshop_log).permit(:children_ongoing, :children_first_time, :teens_ongoing, :teens_first_time, :adults_ongoing, :adults_first_time, :owner_id, :project_id, :date, :workshop_name, :windows_type_id, :other_description, report_form_field_answers: [:form_field_id, :answer_option_id, :answer]).
+  def workshop_log_params
+    params.require(:workshop_log).permit(
+      :children_ongoing, :children_first_time, :teens_ongoing, :teens_first_time,
+      :adults_ongoing, :adults_first_time, :owner_id, :project_id, :date,
+      :workshop_name, :workshop_id, :windows_type_id, :other_description, #:user,
+      quotable_item_quotes_attributes: [
```
maebeale, line 189: nested attributes are here. we're getting Quote records, but still need to confirm we're retaining QuotableItemQuote records. there's a weird delete method in report that will need some more investigation.

db/seeds.rb
```diff
@@ -116,3 +116,14 @@
     metadata.categories.find_or_create_by!(name: category_name)
   end
 end
+
+[
```
maebeale, line 120: as i've been coming across hard-coded data i'm trying to add them to the seeds file. since i was mucking around in Reports and Forms to complete this workshop_log crud, i found this list of Forms to autogenerate.

PR #215, opened 2025-09-28
Tailwind dashboard

jmilljr24:

Staged implementation of migration from bootstrap to tailwind.

This creates a `tailwind` layout that can be added to a controller once a view has been migrated to tailwind. Some shared partials are moved to a tailwind folder to prevent breaking changes. These can be moved once the full migration is complete.

The hope is to create views/partials that are responsive using tailwind classes without having to create/render duplicate views with separate code for mobile and desktop.

### DIFF
--- a/app/frontend/javascript/application.js
+++ b/app/frontend/javascript/application.js
@@ -1,2 +1,7 @@
 import "@hotwired/turbo-rails"
+
+// Turbo drive breaks the ui when navigating due to our migraiton from bootstrap to tailwind.
+// This should be set to true when the migration is complete.
+Turbo.session.drive = false;
+
 import "./controllers"


app/controllers/dashboard_controller.rb
```diff
@@ -2,6 +2,7 @@ class DashboardController < ApplicationController
   skip_before_action :authenticate_user!, :only => :help
 
   layout 'help', :only => :help
+  layout "tailwind"
```
maebeale, line 5: i think we need to move this into the index action only, rather than all actions in dashboard_controller

app/frontend/stylesheets/application.tailwind.css
```diff
@@ -0,0 +1,5 @@
+@import "tailwindcss";
+
+@source "../../../app/views/**/*.html.erb";
```
maebeale, line 3: from what AI is telling me, these 'source' belong in a `tailwind.config.js` within the main dir, rather than in this css file. did you find they are needed here for tailwind to work?

```
// tailwind.config.js
module.exports = {
  content: [
    './app/views/**/*.html.erb',
    './app/helpers/**/*.rb',
    './app/javascript/**/*.js',
    './app/javascript/**/*.ts',
    './app/javascript/**/*.jsx',
    './app/javascript/**/*.tsx'
  ],
  theme: {
    extend: {},
  },
  plugins: [],
}
```

app/views/dashboard/index.html.erb
```diff
@@ -124,43 +134,22 @@
...
-        </div>
-        </div>
+          <!-- VIEWPORT -->
+          <a class="buttons next" href="#">&#62;</a>
+        </div><!-- SLIDER1 --> <!-- COL_MD-12 -->
       </div>
-     </div>
+      <!-- ROW -->
+    </div>
+    <!-- CONTAINER-FLUID -->
```
maebeale, line 145: we can remove this comment, now that we're not doing container-fluid?

app/views/layouts/tailwind.html.erb
```diff
@@ -0,0 +1,65 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>AWBW</title>
+    <link rel="stylesheet" href="/vendor/select-2/select2.min.css">
+    <%# stylesheet_link_tag "application",
+    :media => "all",
+    "data-turbolinks-track" => true %>
+    <%= vite_stylesheet_tag "application.css" %>
+    <%= stylesheet_link_tag "https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.0/themes/smoothness/jquery-ui.css" %>
```
maebeale, line 10: i think for this PR it's good to have parity between layouts/tailwind and layouts/application since it's intended to be a departure from bootstrap only. but a part of me wants to remove all of these extra libraries and cdn calls as part of this move to a new layout. maybe it's best as follow-on work to remove all calls to jquery and google fonts in general. and we're going to be loading fontawesome the vite way so we'll be able to get rid of that one through that body of work.

app/views/resources/_featured.html.erb
```diff
@@ -1,23 +1,57 @@
...
   <% if resources.any? %>
-    <div class="row flex-row" style="padding-top:0px;">
-      <% resources.each do |resource| %>
-        <div class="col-xs-6 col-xs-6-custom">
-          <div class="workshop-box">
-            <%= link_to resource.featured_url, target: :blank do %>
-              <% unless resource.kind == 'Theme' %>
-                <figure class="portrait-img" style="background-image: url(<%= resource.main_image %>)"></figure>
+    <div class="flex flex-wrap justify-center gap-8">
+      <% resources.each do |r| %>
```
maebeale, line 10: this is just my opinion and not strongly held, but i find it much more idiomatic (and easier to read) to use the full word for the iterator rather than a single letter. (so, `resources.each do |resource|` rather than `resources.each do |r|`)

app/views/resources/_featured.html.erb
```diff
@@ -1,23 +1,57 @@
...
+              </svg>
+            </div>
           </div>
         </div>
       <% end %>
     </div>
   <% else %>
-    <%= "There are no resources." %>
+    <p class="py-7 text-center">
+      <%= "There are no resources for this category at this time." %>
```
maebeale, line 54: this text change is much clearer!

app/views/dashboard/index.html.erb
```diff
@@ -1,5 +1,13 @@
-<script type="text/javascript" src="/vendor/tinycarousel/jquery.tinycarousel.min.js"></script>
-<link rel="stylesheet" href="/vendor/tinycarousel/tinycarousel.css" type="text/css" media="screen"/>
+<script
+  type="text/javascript"
+  src="/vendor/tinycarousel/jquery.tinycarousel.min.js"
```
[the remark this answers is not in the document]
jmilljr24: 100% agree! I had the same thoughts and decided not to touch it for the time being. It shouldn't be too difficult to change it out and use stimulus but its enough work to justify it's own issue.

app/views/dashboard/index.html.erb
jmilljr24: Good catch! I will double check other old comments as well.

app/views/layouts/tailwind.html.erb
jmilljr24: I intended to remove all the one's not needed but I wasn't confident on what was being used where. I'll play around and see what I can get away with removing.

app/views/dashboard/_latest_news.html.erb
```diff
@@ -1,25 +1,49 @@
...
+              <span class="line-clamp-3"><%= strip_tags(r.text).truncate(250) %></span>
+            </p>
+          </div>
+          <div
+            class="
+              self-center mt-2 md:mt-0 grow-0 shrink-0 flex items-center justify-center w-24
+              h-20 bg-gray-400 rounded-sm
+            "
+          >
+            <svg
```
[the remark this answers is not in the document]
jmilljr24: Yup!! I didn't think it made sense to do all the logic before the Active Storage PR is merged.

PR #220, opened 2025-09-28
Add tailwind banner

maebeale:
- Display Banner content within the new Tailwind layout

- Made it an explicit partial rather than simply a method

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -39,9 +39,9 @@ module ApplicationHelper
     banner = Banner.last
 
     if !banner.nil? && banner.show
-      content = "<div id='banner-news' class='banner-news'><div class='content'>"
-      content += "#{banner.content}</div></div></br><div class='separator'></div>"
-      return content.html_safe
+      ("<div id='banner-news' class='banner-news'>" +
+        "<div class='content'>#{banner.content}</div>" +
+        "</div><div class='separator'></div>").html_safe
     end
   end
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -39,9 +39,9 @@ def display_banner
     banner = Banner.last
 
     if !banner.nil? && banner.show
-      content = "<div id='banner-news' class='banner-news'><div class='content'>"
-      content += "#{banner.content}</div></div></br><div class='separator'></div>"
-      return content.html_safe
+      ("<div id='banner-news' class='banner-news'>" +
```
maebeale, line 42: i don't love either the content or the interpolated string implementation, but this removes the <br> and makes clearer the 'contents' (haha) of the banner wrapper html.

once we're done with bootstrap and can move the tailwind layout partials outside the tailwind namespace i'd like to remove this method completely and just add the css to the partial. 

(we still need this method for now bc the regular layout and nav partials are calling it)

app/views/layouts/tailwind.html.erb
```diff
@@ -26,7 +26,18 @@
   >
 
     <div class="flex-grow" id="main">
-      <%= render "shared/tailwind/navbar" %>
+      <div class="sticky top-0 left-0 w-full z-50">
```
maebeale, line 29: move the navbar and banner into a shared div for easier stickiness

PR #228, opened 2025-09-28
#227 - remove width limit of navbar items

### LINKED ISSUE #227: Change breakpoints for nav
Currently it's hugging the middle rather than the edges even on a 13" laptop

Image

jmilljr24:
Resolves #227 

This removes the width constraint on nav items. The original concern was users with ultrawide monitors however, it seems common practice to let the nav items expand the full width of the screen. Reference the html on [tailwindcss.com](https://tailwindcss.com/)


maebeale (approved):
🖥️

PR #233, opened 2025-09-29
Add stimulus dropdown controller

jmilljr24:

JS dropdown menus in Tailwindplus requires a commercial license. This removes that dependency.

Add a stimulus controller to handle the dropdowns. The controller toggles the utility classes for the target html element. Dropdowns can be closed by the escape key or clicking outside the element.

### Anything else to add?

[Screencast from 2025-09-29 06-27-41.webm](https://github.com/user-attachments/assets/41b703bf-0b7b-4b2b-bc64-a9450aeb2cbd)

### DIFF
--- a/app/frontend/javascript/controllers/index.js
+++ b/app/frontend/javascript/controllers/index.js
@@ -1,7 +1,4 @@
 import { application } from "./application";
 
-import HelloController from "./hello_controller";
-application.register("hello", HelloController);
-
 import DropdownController from "./dropdown_controller";
 application.register("dropdown", DropdownController);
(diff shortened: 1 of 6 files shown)


app/frontend/javascript/controllers/index.js
```diff
@@ -1,4 +1,7 @@
-import { application } from "./application"
+import { application } from "./application";
 
-import HelloController from "./hello_controller"
-application.register("hello", HelloController)
+import HelloController from "./hello_controller";
```
maebeale, line 3: HelloController doesn't exist, right?

jmilljr24:
> Woo hoo! I'm excited about this! Wtg, @jmilljr24 !
> 
> We prob want to consider adding Jest and a test, but I'm ok with skipping it.

Totally forgot about tests!! And I will remove `HelloController`. Its the example controller you get when you install stimulus.

PR #236, opened 2025-09-29
Add primary color theme to tailwind

jmilljr24:

Allows defining a primary color in one file. This allows using the organizations primary color with tailwind utility classes e.g.  `text-primary` `bg-primary`

Add tailwind @theme and find/replace hardcoded colors.


maebeale:
as follow-on work, would you consider adding in colors for `secondary` and `danger` and `warning` and `info` as well?

PR #251, opened 2025-10-01
#222 Add faq accordion

### LINKED ISSUE #222: Re-implement FAQs with stimulus accordions
The current implementation is pretty ... old.

Also, want to layer in user.super_user permission for Edit capability inline (or via an Edit path button to start)

jmilljr24:
Resolves #222 

Create an accordion style FAQ for betting UI/UX.

Accordion is built using the dropdown stimulus controller. All users now go to a single faq index. Normal users only see active faq's with Question/Answer. Super_users see all questions as well as active status, order number and a link to edit.

### Anything else to add?

The dropdown stimulus controller needs to be updated to allow a single action to update multiple elements on the page. For example in an faq, both the answer div needs to change from hidden and the button arrow on the question needs to toggle between pointing up and down. Breaking changes have been updated where the dropdown stimulus controller was being used.

Super User
[Screencast from 2025-10-01 10-10-22.webm](https://github.com/user-attachments/assets/b111aec9-3d57-48ce-bbb9-3bfbffd75b81)

<br>
Regular User

[Screencast from 2025-10-01 10-11-04.webm](https://github.com/user-attachments/assets/abca69b9-8ff4-4bab-8277-72c52fa8956c)

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -55,6 +55,17 @@ class EventRegistrationsController < ApplicationController
     end
   end
 
+  def destroy
+    @event_registration = EventRegistration.find(params[:id])
+    if @event_registration
+      @event_registration.destroy
+      flash[:notice] = 'You are no longer registered.'
+      redirect_to events_path
+    else
+      flash[:alert] = 'Unable to find that registration.'
+    end
+  end
+
   private
 
   def event_registration_params
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,15 +1,18 @@
 class EventsController < ApplicationController
   before_action :set_event, only: %i[ show edit update destroy ]
-  before_action :authorize_admin!, only: %i[ new edit update destroy ]
+  before_action :authorize_admin!, only: %i[ edit update destroy ]
+
+  layout "tailwind", only: [:index, :edit, :new, :show]
 
   def index
-    @events = Event.all
+    @events = Event.order(start_date: :desc)
+    @events = @events.publicly_visible unless current_user.super_user?
   end
 
   def show
   end
 
-  def new
+  def new # all logged in users can create events
     @event = Event.new
   end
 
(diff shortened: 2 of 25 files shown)


app/controllers/faqs_controller.rb
```diff
@@ -1,19 +1,17 @@
 class FaqsController < ApplicationController
   before_action :set_faq, only: [:show, :edit, :update, :destroy]
 
+  layout "tailwind", only: [:index]
+
   def index
     if current_user.super_user?
       per_page = params[:number_of_items_per_page].presence || 25
-      @faqs = Faq.all.paginate(page: params[:page], per_page: per_page)
+      @faqs = Faq.all.by_order.paginate(page: params[:page], per_page: per_page)
```
maebeale, line 9: does this collection need to be paginated?

if so, i think the regular user collection should also be paginated, yes?

app/controllers/faqs_controller.rb
```diff
@@ -1,19 +1,17 @@
...
     else
-      redirect_to collapsible_faqs_path
+      @faqs = Faq.active.by_order
     end
   end
 
-  def collapsible
```
maebeale, line 13: yeah, i like returning to more restful routes for sure

app/frontend/javascript/controllers/dropdown_controller.js
```diff
@@ -1,31 +1,45 @@
 import { Controller } from "@hotwired/stimulus";
```
maebeale, line 1: were these changes to the nav dropdowns mixed in here on purpose? did you have to change the stimulus controller for something on the faq's page?

app/controllers/faqs_controller.rb
jmilljr24: I'm not sure we need pagination at this point. The db only has 30 record. I don't foresee new FAQ's being adding on a consistent basis where the list grows fast. I think if it becomes unwieldy in the future we might need to consider a different way to present the information. Having multiple pages of FAQ doesn't seem user friendly. Maybe something like categories with a filter to keep it a single page. 

So my vote for now is to remove pagination for super_user to keep both the same until a point in the future where is becomes a bigger issue.

PR #253, opened 2025-10-01
Add sortable FAQ drag and drop

jmilljr24:

Add drag and drop sorting of FAQs on the index page.

Add stimulus controller using sortablejs. This js library handles the nuances of different browsers and handles the drag and drop feature via the frontend. Request.js was add to send the updated ordering id to the controller. I also added the [Positioning](https://github.com/brendon/positioning) gem for proper handling of list orders.

### Anything else to add?

Removed pagination for index for the time being. Missed it in the previous PR.

[Screencast from 2025-10-01 19-21-11.webm](https://github.com/user-attachments/assets/767897e3-1a47-43de-8976-af7e7ae2cce1)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,3 +63,5 @@ group :development, :test do
   gem "shoulda-matchers", require: false
   gem "debug", "~> 1.11"
 end
+
+gem "positioning", "~> 0.4.7"
--- a/app/models/faq.rb
+++ b/app/models/faq.rb
@@ -1,4 +1,8 @@
 class Faq < ApplicationRecord
+  # TODO After the production db migration we should "heal_ordering_column!" and add db contraints
+  # https://github.com/brendon/positioning
+  positioned on: [], column: :ordering
+
   # Validations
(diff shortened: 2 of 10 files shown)


Gemfile
```diff
@@ -63,3 +63,5 @@ group :development, :test do
   gem "shoulda-matchers", require: false
   gem "debug", "~> 1.11"
 end
+
+gem "positioning", "~> 0.4.7"
```
maebeale, line 67: TIL about this gem! 💥

app/models/faq.rb
```diff
@@ -1,4 +1,8 @@
 class Faq < ApplicationRecord
+  # TODO After the production db migration we should "heal_ordering_column!" and add db contraints
```
maebeale, line 2: <3

maebeale: would you create an issue for this and mark it blocked?

maebeale (approved):
MAJOR ux upgrade here, Justin! Great work.

PR #300, opened 2025-10-30
Fix seeds.rb so it works

johnpaulashenfelter:
- This work is part of [link an issue]

Running anything that invokes `db:seeds` fails with a set of several errors

Fixed the errors one-by-one using staging data and reading the code

### Anything else to add?

Would love to have someone try this out.


maebeale (approved):
Thanks for fixing these! They work for me!

I submitted a commit to also make them idempotent.

PR #338, opened 2025-11-10
Add event registration via turbo

### LINKED ISSUE #320: Add multi-select Event registration via Stimulus
Event registration was implemented as a multi-select option by wrapping all displayed events in a form. This did not allow additional features on individual Event cards that required form/post requests.

Add a stimulus controller to handle user selected events and populate a registration form in its own html element.

jmilljr24:
Resolves #320 

Change event registration to a single button click for authenticated users. Partial page updates via turbo to reflect the new registration.

This removes the email and name attributes from `event_registrations` and changes to use `user` allowing the model to be a through table for user and events.

On the frontend, a user can click register and the card is updated via turbo.
### Anything else to add?

Non-user registration will need to be added in the future via an inactive attribute. 

[Screencast from 2025-11-10 13-17-04.webm](https://github.com/user-attachments/assets/e166c0a7-c1a3-4d81-bea5-7f8cbad2ce09)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2025_11_13_124217) do
+ActiveRecord::Schema[8.1].define(version: 2025_11_13_142909) do
   create_table "active_storage_attachments", charset: "utf8mb3", force: :cascade do |t|
     t.bigint "blob_id", null: false
     t.datetime "created_at", precision: nil, null: false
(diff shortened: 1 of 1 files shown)


app/models/event.rb
```diff
@@ -2,7 +2,8 @@ class Event < ApplicationRecord
   belongs_to :created_by, class_name: "User", optional: true
   has_many :bookmarks, as: :bookmarkable, dependent: :destroy
   has_many :event_registrations, dependent: :destroy
-  
+  has_many :users, through: :event_registrations
```
maebeale: may we make this `has_many :registrants, through: :event_registrations, class_name: "User"` just so it's a little clearer?

app/models/event_registration.rb
```diff
@@ -1,7 +1,6 @@
 class EventRegistration < ApplicationRecord
+  belongs_to :user
```
maebeale: similar here, and maybe make this something like: `belongs_to :registrant, class_name: "User", foreign_key: :registrant_id`

PR #364, opened 2025-11-14
Fix resource bookmark

jmilljr24:
I noticed in the demo video the bookmark actually didn't work (not sure if anyone noticed) but the flash said `ResourceDecorator` added.

Fix bookmarks for `Resource`

The bookmark icon partial must recieve the actual resource (or other type) **Object** not the decorator.
### Anything else to add?

Minor fix for when an event doesn't have a description.

screenshot-2025-11-14_07-48-09


maebeale (approved):
oooo! GREAT catch! i hadn't noticed that.

Issue #349, opened 2025-11-13
Add "Add to calendar" to Event card on index


maebeale:
After a User is registered to an event, display the following two rows at the bottom of the card:

"Add to calendar" and then the relevant calendar links, e.g. the following code and screenshot
```Add event to calendar

[Apple](https://calndr.link/e/Hjs8zdcx1s?s=apple)  •  [Google](https://calndr.link/e/Hjs8zdcx1s?s=google)  •  [Office 365](https://calndr.link/e/Hjs8zdcx1s?s=office365)  •  [Outlook Web](https://calndr.link/e/Hjs8zdcx1s?s=outlookcom)  •  [Outlook](https://calndr.link/e/Hjs8zdcx1s?s=outlook)  •  [Yahoo](https://calndr.link/e/Hjs8zdcx1s?s=yahoo)```

<img width="282" height="39" alt="Image" src="https://github.com/user-attachments/assets/45d01f63-7089-42a4-866d-deeefaac5c1a" />

PR #379, opened 2025-11-14
Add calendar links

diti0-dot:
- This work is part of [https://github.com/rubyforgood/awbw/issues/349]

Add calendar links to event cards for registered users.

Added calendar section that shows only for registered users with the dynamic links that created with event details

### Anything else to add?

Nothing really, still learning and having fun! 😺 Please guide me if I'm doing something wrong.


app/views/events/index.html.erb
```diff
@@ -104,6 +109,20 @@
             </div>
           <% end %>
         </div>
+         <!-- Add to calendar -->
+          <% if event.event_registrations.where(email: current_user.email).exists? %>
+          <div class="mt-4 pt-3">
+            <div class="text-sm font-semibold text-gray-600 mb-1">Add to calendar</div>
+            <div class="text-xs">
+              <%= link_to "Apple", "https://calndr.link/e/Hjs8zdcx1s?s=apple", class: "text-blue-600 hover:text-blue-800" %> •
```
maebeale: these are hard-coded links for one event. we'll need to dynamically make links for each event. (this is my fault for the ticket not being clear enough!)

app/views/events/index.html.erb
```diff
@@ -41,9 +14,128 @@
...
+              </label>
+            </div>
+          <% else %>
+            <div class="mt-2 ml-auto">
+              <span class="text-sm text-gray-500 italic">Registration closed</span>
+            </div>
+          <% end %>
+        </div>
+         <!-- Add to Calendar -->
+          <% if event.event_registrations.where(email: current_user.email).exists? %>
```
maebeale: i think this needs to be:
`event.event_registrations.where(registrant: current_user).exists?`
bc event_registrations doesn't have an email field on it

PR #389, opened 2025-11-16
Add registration controller for events

### LINKED ISSUE #383: Make a TurboRegistrations controller to handle events index create/delete registrations

jmilljr24:
Resolves #383 

Refactor user facing event registrations to a new controller.

Admins can create `event_registrations` directly by selecting a user and event. This is handled in the `EventRegistrationsController`. A `current_user` registers via the `Event` itself. This drove the decision to separate out the controllers. 

With regards to naming, I feel using the Event namespace makes the most sense and matches the route `/events/:event_id/registrations`. I'm open to suggestions on the controller name itself if it can help avoid confusion with the `EventRegistrationsController`. 

I don't think its appropriate to add "Turbo" to the name though. The controller isn't strictly for turbo responses. Each action includes a html response which is considered best practice when working with turbo. I could also see this controller being used for visitor registrations when we may want a full page redirect instead of turbo.

### DIFF
--- a/app/views/event_registrations/index.html.erb
+++ b/app/views/event_registrations/index.html.erb
@@ -42,9 +42,14 @@
                               class: "btn btn-secondary-outline" %>
                   </td>
                   <td class="px-4 py-2 text-sm text-gray-800">
-                    <%= link_to event_registration.registrant.full_name,
+                    <%# FIXME not all users have facilitators %>
+                    <% if event_registration.registrant.facilitator %>
+                      <%= link_to event_registration.registrant.full_name,
                               facilitator_path(event_registration.registrant.facilitator),
                               class: "btn btn-secondary-outline" %>
+                    <% else %>
+                      <%= event_registration.registrant.full_name %>
+                    <% end %>
                   </td>
                   <td class="px-4 py-2 text-sm text-gray-800">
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -36,8 +36,10 @@
     end
   end
   resources :community_news
-  resources :events
   resources :event_registrations
+  resources :events do
+    resource :registrations, only: %i[create destroy], module: :events, as: :user_registration
```
jmilljr24: I had to change the generated path name to de-conflict with `event_registrations`. 

This builds `event_user_registration_path`. Happy to use something else if it sounds better.

app/controllers/event_registrations_controller.rb
```diff
@@ -22,70 +21,42 @@ def edit
...
+        }
       end
     end
   end
 
-
   def update
     if @event_registration.update(event_registration_params)
-      redirect_to event_registrations_path, notice: "Banner was successfully updated.", status: :see_other
+      redirect_to event_registrations_path, notice: "Registration was successfully updated.", status: :see_other
```
maebeale: oops! ty for catching that!

maebeale (approved):
love this solution! having separate controller/namespace/etc for registrations completed via events index vs regular crud is 😍 .

PR #453, opened 2025-11-19
Fix carousel bookmarks

jmilljr24:

Carousel bookmarks don't work.

The bookmark icon partial requires the object not the decorator.

### Anything else to add?

Broken:
screenshot-2025-11-19_12-21-11


maebeale (approved):
TYYY for finding and fixing this! 

i was JUST doing an audit bc i realized i've made this same mistake in a few places.

PR #454, opened 2025-11-19
Fix faq update flash

jmilljr24:

When an faq is updated inline we should see the flash message.

Include our turbo_flash helper in the turbo frame and add flash.now to the controller.

### Anything else to add?

https://github.com/user-attachments/assets/9b3280a9-8fee-434c-a049-22dd7107183e


maebeale (approved):
yassss. this hadn't even hit my radar. well done!

Issue #482, opened 2025-11-26
Error: undefined method `reference_url' for an instance of WorkshopVariationDecorator


jmilljr24:
Visit workshop.id 5 and click workshop variation (underwater variation).

<img width="2880" height="1776" alt="Image" src="https://github.com/user-attachments/assets/a623d833-d33e-44bf-acb3-bc60de80b1d3" />

jmilljr24:
@brodyf42 assigned. Thanks!

PR #484, opened 2025-11-26
[482] update workshop variation show view to use youtube_url

brodyf42:
- This work is part of #482

This addresses an issue on the `WorkshopVariation` show view that had an outdated reference to `reference_url`, which was causing a `NoMethodError`.

- Updated the show view to use the new attribute name `youtube_url`
- Added a spec for the workshop variation show view

### Anything else to add?
nope :smiley:

### DIFF
--- a/app/views/workshop_variations/show.html.erb
+++ b/app/views/workshop_variations/show.html.erb
@@ -64,9 +64,9 @@
 
 
       <!-- URL -->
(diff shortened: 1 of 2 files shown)


brodyf42:
@maebeale, it looks like I'll need some elevated permissions in order to merge.

PR #560, opened 2025-12-05
Add cursor-pointer to Nav buttons for better UX

jmilljr24:

Hopefully adding a visually indication of the cursor-pointer on mouse over of the nav buttons helps user understand they are clickable.

### Anything else to add?

https://github.com/user-attachments/assets/1817cd25-3faa-44e4-aa4f-3b6ccb44f738


app/views/shared/_navbar_new_button.html.erb
```diff
@@ -7,7 +7,7 @@
         flex items-center gap-2 rounded-md bg-white px-4 py-1 mt-1 text-sm font-semibold text-gray-900
         shadow-sm border border-gray-200 hover:bg-gray-50 active:bg-gray-100
         focus:outline-none focus:ring-2 focus:ring-blue-500 focus:ring-offset-2
-        transition"
+        transition cursor-pointer"
       data-action="dropdown#toggle"
       data-dropdown-payload-param='[{"workshop-dropdown":"hidden"}]'
```
maebeale: i think we should rename the payload to `new-dropdown` or `new-button-dropdown`. doesn't have to be this PR.

PR #565, opened 2025-12-12
Add typography plugin

jmilljr24:
add prose to rich text

```
The official Tailwind CSS Typography plugin provides a set of prose classes you can use to add beautiful typographic 
defaults to any vanilla HTML you don’t control, like HTML rendered from Markdown, or pulled from a CMS.
```

This PR adds the [plugin](https://github.com/tailwindlabs/tailwindcss-typography) and adds the class to the workshop rendered html.

### Anything else to add?

Now you can use the `prose` classes to add sensible typography styles to any vanilla HTML:
```
<article class="prose lg:prose-xl">
  <h1>Garlic bread with cheese: What the science tells us</h1>
  <p>
    For years parents have espoused the health benefits of eating garlic bread with cheese to their
    children, with the food earning such an iconic status in our culture that kids will often dress
    up as warm, cheesy loaf for Halloween.
  </p>
  <p>
    But a recent study shows that the celebrated appetizer may be linked to a series of rabies cases
    springing up around the country.
  </p>
  <!-- ... -->
</article>
```
screenshot-2025-12-12_10-15-00


jmilljr24:
@maebeale I see we have both package.json and yarn.lock - I don't think we need or should have both. The docker build for production is setup to use npm with package.json. Should I delete the yarn.lock so we only use one package manager in this project?

PR #587, opened 2025-12-27
Resource Index - Lazy load with turbo frame

jmilljr24:

Lazy load resource index and add copy search url

Add turbo_frame

### Anything else to add?

https://github.com/user-attachments/assets/837887e8-b763-4955-9ca0-bcb493d9225d

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -63,7 +63,7 @@ class Resource < ApplicationRecord
     kinds = Array(kinds).flatten.map(&:to_s)
     where(kind: kinds)
   }
-  scope :leader_spotlights, -> { kind("LeaderSpotlight") }
+  scope :leader_spotlights, -> { kinds("LeaderSpotlight") }
   scope :published_kinds, -> { where(kind: PUBLISHED_KINDS) }
   scope :published, ->(published=nil) {
     if ["true", "false"].include?(published)


app/controllers/resources_controller.rb
```diff
@@ -1,19 +1,25 @@
...
-
-    @count_display = if filtered.count == unpaginated.count
-                       unpaginated.count
-                     else
-                       "#{filtered.count}/#{unpaginated.count}"
-                     end
-    @sortable_fields = Resource::PUBLISHED_KINDS
+    if turbo_frame_request?
+      per_page = params[:number_of_items_per_page].presence || 25
+      unfiltered = Resource.where(kind: Resource::PUBLISHED_KINDS) # TODO - #FIXME brittle
```
maebeale: FYI 1: i started changing these to `unfiltered` instead of `unpaginated` to be more clear

FYI 2: `total_entries` can't be used anymore now that we're decorating the paginated collection

app/models/resource.rb
```diff
@@ -59,7 +59,10 @@ class Resource < ApplicationRecord
   scope :category_names, ->(names) { tag_names(:categories, names) }
   scope :sector_names,   ->(names) { tag_names(:sectors, names) }
   scope :featured, -> (featured=nil) { featured.present? ? where(featured: featured) : where(featured: true) }
-  scope :kind, -> (kind) { where("kind like ?", kind ) }
+  scope :kind, ->(kinds) {
```
maebeale: this is so much better!

you didn't want to rename the scope to `kinds` instead of `kind`? i've been going back and forth on whether or not i think they should be singular or plural when sending in plural values.

app/views/resources/index.html.erb
```diff
@@ -15,40 +15,63 @@
...
-    <div class="w-full mt-6">
-      <hr class="border-gray-300 mb-6">
-    </div>
+    <div class="w-full mt-6"><hr class="border-gray-300 mb-6"></div>
 
     <!-- Results Table -->
-    <% if @resources.any? %>
-      <div class="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 gap-6">
-        <% @resources.each do |resource| %>
-          <%= render "resource_card", resource: resource.decorate %>
```
maebeale: oh we can't use the partial anymore?

maebeale: nvm. now i see this is to build the skeleton.

maebeale (approved):
looks great! this is such a QOL improvement.

PR #588, opened 2025-12-27
Workshop Index - lazy load 

jmilljr24:

Follow pattern of Resource Index and lazy load the workshop results

### Anything else to add?

https://github.com/user-attachments/assets/c3f74793-a721-42b2-a6d8-0758124a39b2


maebeale (approved):
💥 love this!!!

PR #609, opened 2025-12-30
Rename image associations/models

maebeale:

- The word 'Image' didn't work bc some files are pdf's, etc
- Renamed `Images::MainImage`, `Images::GalleryImage`, `Images::RichText`, `Images::Square` ->
  `PrimaryAsset`, `SecondaryAsset`, `RichTextAsset`, `SquareAsset`
- Includes a rake task to copy Images table into Assets

- Moved away from `Main` and `Gallery` to instead use `Primary` and `Secondary`

### Anything else to add?
- I first renamed these to Media and subclasses, but to avoid the pluralization issues, chose Asset

### DIFF
--- a/lib/tasks/attachment_report.rake
+++ b/lib/tasks/attachment_report.rake
@@ -11,7 +11,7 @@ namespace :attachment_report do
       Address, AgeRange, AnswerOption, Attachment, Banner, Bookmark,
       Category, CategorizableItem, CommunityNews, EventRegistration, Event,
       Facilitator, Faq, FormBuilder, FormFieldAnswerOption, FormField, Form,
-      Media, Location, MediaFile, CategoryType, MonthlyReport, Notification,
+      Image, Location, MediaFile, CategoryType, MonthlyReport, Notification,
       ProjectObligation, ProjectStatus, ProjectUser, Project,
       QuotableItemQuote, Quote, ReportFormFieldAnswer, Report, Resource,
       SectorableItem, Sector, Story, StoryIdea, UserFormFormField, UserForm,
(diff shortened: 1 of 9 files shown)


app/models/user.rb
```diff
@@ -38,7 +38,7 @@ class User < ApplicationRecord
   has_many :user_form_form_fields, through: :user_forms, dependent: :destroy
   has_many :windows_types, through: :projects
   # Images
-  has_one_attached :avatar # old paperclip -- TODO convert these to SquareImage belonging to Facilitator
+  has_one_attached :avatar
```
jmilljr24: Do we want avatar on both user and facilitator or just one?

jmilljr24 (commented):
@maebeale No need to answer or justify anything. Just some things that caught my eye.

lib/tasks/attachment_report.rake
```diff
@@ -11,7 +11,7 @@ namespace :attachment_report do
       Address, AgeRange, AnswerOption, Attachment, Banner, Bookmark,
       Category, CategorizableItem, CommunityNews, EventRegistration, Event,
       Facilitator, Faq, FormBuilder, FormFieldAnswerOption, FormField, Form,
-      Image, Location, MediaFile, CategoryType, MonthlyReport, Notification,
+      Media, Location, MediaFile, CategoryType, MonthlyReport, Notification,
```
jmilljr24:
```suggestion
      Image, Location, MediaFile, CategoryType, MonthlyReport, Notification,
```

app/models/user.rb
maebeale: I was going to leave it on user for now until post-migration. But, yeah, it's def coming off.

Issue #46, opened 2025-08-01
Add a test for: Facilitators can submit a workshop idea


leelyuh:
**Add a system test for:**

As a faciltator,
I want to submit a workshop idea
So that I can share something valuable with the community. 

Given I am logged into AWBW
When I click on the Share A Workshop Idea link in the sidebar navigation,
Then I am taken to the Submit Workshop Idea page
And I can fill out the form as follows:

Workshop Title [free text field]
Author's First & Last Name [free text field]
Workshop Objective [free text field]
Materials Needed [free text field]
Timeframe [number picker for hour, number picker for minutes]
Set Up Instructions [free text field]
Demonstration Instructions [free text field]
Warm-up/relaxation/meditation [free text field]
Creation [free text field]
Closing [free text field]
Notes [free text field]
Tips [free text field]
And I can click on the Choose File button to select files to upload if relevant
And I can add more files if needed. 

When I click the Submit button,
Then my workshop idea is submitted
And I am returned to the dashboard where I see a success message.

When I click the Cancel button,
Then my workshop idea is not submitted,
And I am returned to the dashboard.

Video walkthrough for reference: https://www.loom.com/share/652c498a8b8946bbbf65b27945a24faa

PR #612, opened 2026-01-01
add test for Facilitators submitting a workshop idea

diti0-dot:
- This work Closes [https://github.com/rubyforgood/awbw/issues/46]

Adds system tests for workshop idea submission flow. Tests form access, submission, and cancellation.

Followed existing test patterns using FactoryBot for data.

### Anything else to add?
First time writing tests - feedback welcome! Happy New Year!🎉🥳


spec/system/facilitator_submits_workshop_idea_test.rb
```diff
@@ -0,0 +1,83 @@
+
+require 'rails_helper'
+
+RSpec.describe "Facilitators can submit a workshop idea", type: :system do
+  describe "Navigate to Workshop Idea page" do
+    context "When Facilitator is logged in" do
+      before do
+        create(:windows_type, :adult)
```
maebeale: @diti0-dot we used to have to do this bc of a callback dependency, but we've removed that, so your tests should still pass even if you remove these create(:windows_type) statements

maebeale (approved):
This is great @diti0-dot !!!

It's probably worth a follow-up in another PR to add some additional tests for the non-happy-path that ensure the required fields show validation errors when missing (or when wrong file type in the case of the attachments).

PR #615, opened 2026-01-02
Add blur on submit to collection controller

jmilljr24:
'
Workshops and Resources have search results updated "on click" or while typing. This PR provides visual indication the request has been submitted. Prior, a slow query could feel as though nothing was happening when a user selected a filter.

Update collection stimulus controller to add the tailwind class "blur-sm" to targeted elements. Add the class "blur-on-submit" to any element that should be blurred on user inputs.

### Anything else to add?

https://github.com/user-attachments/assets/5d14cbb6-b946-4709-b5b7-7ed34dcf90c0


maebeale (approved):
Love this improvement!

PR #619, opened 2026-01-03
Rhino horizontal rule

jmilljr24:

Add button to rhino-editor to allow adding an `<hr>` element
### Anything else to add?
Clean up some old unused code.

### DIFF
--- a/app/frontend/javascript/rhino/custom-editor.js
+++ b/app/frontend/javascript/rhino/custom-editor.js
@@ -49,6 +49,8 @@ class CustomEditor extends TipTapEditor {
           <slot name="heading-button">${this.renderHeadingButton()}</slot>
           <slot name="after-heading-button"></slot>
 
+          <slot name="hr-button">${this.renderHorizontalRuleButton()}</slot>
+
           <!-- Blockquote -->
           <slot name="before-blockquote-button"></slot>
           <slot name="blockquote-button">${this.renderBlockquoteButton()}</slot>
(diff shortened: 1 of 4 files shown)


app/frontend/javascript/rhino/custom-editor.js
```diff
@@ -148,42 +150,10 @@ class CustomEditor extends TipTapEditor {
       <button
         type="button"
         class="toolbar__button rhino-toolbar-button"
-        title="Insert grid (Shift + click to enter custom dimensions)"
```
maebeale: we don't need any of this Insert grid code anymore?

app/frontend/javascript/rhino/custom-editor.js
```diff
@@ -148,42 +150,10 @@ class CustomEditor extends TipTapEditor {
...
-
-            this.editor.chain().insertGrid(columns, rows).run();
-          } else {
-            // Default grid insertion
-            this.editor.chain().insertGrid().run();
-          }
-        }}
+        title="Insert grid"
+        @click="${(event) => {
+          this.editor.chain().focus().insertGrid().run();
```
maebeale: same comment: this loses all that conditional logic and alerts above -- are we sure that's ok?

maebeale (approved):
Super cool! Love that you thought of this button/option!

app/frontend/javascript/rhino/custom-editor.js
jmilljr24: I'll make some video tutorials on using rhino, but essentially I dumbed this down a bunch to remove complexity. The code was adding grow-rows which was causing the cell high issues you notices. I removed all that. A user will simply add one cell at a time. Then add "columns" which just updates the parent grid class like "grid-cols-3". 

Realistically most grids will be only be one or two rows. It will generally be easier to just create a new grid group if you want additional rows unless you need a bunch of cells to be dependent on one another for some reason.

Issue #406, opened 2025-11-16
Beef up Notification model to capture the actual body of the message sent


maebeale:
Right now it's only capturing the polymorphic noticeable, but we want to be able to reference/view the email body that was sent

maebeale:
completed

Issue #287, opened 2025-10-17
Email notifications on submission of workshop_idea, story_idea, workshop_variation, event registration


maebeale:
should use notification_mailer.

maebeale:
completed!

Issue #213, opened 2025-09-26
Expand Notification to store audit fields


maebeale:
class AddEmailAuditToNotifications < ActiveRecord::Migration[7.0]
  def change
    add_column :notifications, :sent_at, :datetime
    add_column :notifications, :recipient_email, :string
    add_column :notifications, :email_subject, :string
    add_column :notifications, :email_body, :text
  end
end

def send_notice
  mail = Admins::NotificationMailer.email(self)
  mail.deliver_now

  update_columns(
    sent_at: Time.current,
    recipient_email: Array(mail.to).join(', '),
    email_subject: mail.subject,
    email_body: mail.body.decoded
  )
rescue StandardError => e
  Rails.logger.error("Failed to send notification ##{id}: #{e.message}")
end

def resend
  mail = Admins::NotificationMailer.email(self)
  mail.deliver_now

  update_columns(
    sent_at: Time.current,
    recipient_email: Array(mail.to).join(', '),
    email_subject: mail.subject,
    email_body: mail.body.decoded
  )
rescue StandardError => e
  Rails.logger.error("Failed to resend notification ##{id}: #{e.message}")
end

PR #626, opened 2026-01-04
Notification refresh

### LINKED ISSUE #13: Instructions on how to run actionmailer locally without a full-blown SMTP server
ActionMailer and MandrillMailer are configured to use SMTP in config/initializers/mail.rb ... pretty sure that means anything that would trigger email in test and development would try to connect to an SMTP server. No bueno!

I am not sure what is available in Rails 5, but i am guessing that ActionMailer v5 includes file and test delivery methods. If so, lets configure the development and test env to use file (or whatever makes the most sense) and test delivery methods respectively and move the SMTP action mailer config to config/environments/production.rb.

maebeale:
- This work Closes 
#13 
#213 
#406 
#287 

- Refresh the Notification model, and have Notifications generated for:
    - StoryIdea
    - WorkshopIdea
    - WorkshopVariation
    - EventRegistration

- Created a new Service, that calls a Job, and update the Notification record w data from the sent email

### DIFF
--- a/app/decorators/workshop_log_decorator.rb
+++ b/app/decorators/workshop_log_decorator.rb
@@ -1,155 +1,63 @@
 class WorkshopLogDecorator < ApplicationDecorator
   def detail(length: nil)
-    description = length ? description&.truncate(length) : description
+    description = length ? object.description&.truncate(length) : object.description
     "#{description}<br>#{participants_table}".html_safe
   end
 
-  PARTICIPANT_FIELDS = {
-    children: %i[children_first_time children_ongoing],
-    teens:    %i[teens_first_time teens_ongoing],
-    adults:   %i[adults_first_time adults_ongoing]
-  }.freeze
-
-  #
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -90,10 +90,15 @@
...
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
-  # config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
+
+  config.action_mailer.logger = Rails.logger
 
   # Set host to be used by links generated in mailer templates.
-  config.action_mailer.default_url_options = { host: "example.com" }
+  config.action_mailer.default_url_options = {
+    host: "example.com",
```
jmilljr24:
```suggestion
    host: ENV.fetch("APP_HOST"),
```
We have this set as an env so I think we can use it here.

PR #643, opened 2026-01-07
Add payments table

RJV333:
Adds a table that will be used to track payments.  Might need some revision as we flesh out the implication but these fields seem like the standard params for tracking stripe payments.

polymorphic payer: who is making the payment
polymorphic payable: what is being paid for (EventRegistration, Dues...)

amount_cents
currency

stripe_payment_intent_id: primary Stripe reference to the payment attempt lifecycle
stripe_charge_id: the actual money movement (receipt, refund, reconcilliation)

status - this field and the two below are used for tracking and reflecting the stripe payment intent status
failure_code
faliure_message

stripe_metadata - json of additional metadata from using stripe API

### Anything else to add?

Undetermined if we want to add refunds to this table or in a separate table, but probably better to determine that after we get core functionality up and running

### DIFF
--- a/db/migrate/20260106225822_add_payments_table.rb
+++ b/db/migrate/20260106225822_add_payments_table.rb
@@ -22,8 +22,8 @@ class AddPaymentsTable < ActiveRecord::Migration[8.1]
 
     add_index :payments, :stripe_payment_intent_id, unique: true
     add_index :payments, :stripe_charge_id
-    add_index :payments, [:payer_type, :payer_id]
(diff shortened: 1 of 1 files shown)


maebeale:
@RJV333 we're running rubocop omikase. would you run `rubocop` in this branch and fix the arrays to have spaces at beginning and end to satisfy the rubocop gods?

jmilljr24:
@RJV333 `rubocop -a` from your terminal will autocorrect what is can. It usually handles 99% of the issues.

PR #658, opened 2026-01-11
Add system tests for resource search

### LINKED ISSUE #35: [For testing specs] Facilitators can search for resources
As a facilitator,
I want to search for resources
So that I can quickly find what I’m looking for.

Given I am on the AWBW dashboard,
When I click on the Resources link in the side panel,
Then I am taken to the Resources page.

When the page loads,
Then I see all available resources displaying by default
And the list paginates. 

When I filter by Toolkits,
Then only Toolkits are displayed in the results. 

When I filter by Forms,
Then only Forms are displayed in the results. 

When I filter by Templates,
Then only Templates are displayed in the results. 

When I filter by Handouts,
Then only Handouts are displayed in the results. 

When I type a keyword into the search bar and hit enter,
Then I see any applicable results below.

diti0-dot:
- This work Closes #35 

To test facilitator's ability to search and filter resources. 

I created a facilitator user and resources, then tested navigation to the resources page, filtering by clicking resource kinds, keyword searching, and pagination by creating multiple resources.

### Anything else to add?
Nothing

### DIFF
--- a/spec/system/facilitator_searches_resources_test.rb
+++ b/spec/system/facilitator_searches_resources_test.rb
@@ -192,7 +192,6 @@ RSpec.describe "Facilitators can search for resources" do
 
             expect(page).to have_content('Test Resource')
             expect(page).to have_link('2')
-
           end
         end
       end


maebeale (approved):
Awesome! Ty for being so thorough in the options, and including the pagination test too!

Issue #647, opened 2026-01-07
Change column name of all sortable models


jmilljr24:
We have added the [Positioning Gem](https://github.com/brendon/positioningl)

The default column name is `:position`. Lets update all models to use this naming convention. Some examples of column names currently used in the DB are `order`, `ordering`, `series_order`. This list may not be inclusive.

PR #659, opened 2026-01-12
Issue 647: change column name on sortable models

RJV333:
- This work Closes [[link an issue](https://github.com/rubyforgood/awbw/issues/647)]

We have added the [Positioning Gem](https://github.com/brendon/positioningl)

The default column name is :position. Update all sortable models to use this naming convention. 

Used separate migrations for each updated table to keep these changes small, separate, reservable, and safer to deploy.


jmilljr24 (approved):
@RJV333 lgtm! Appreciate the separate migrations.

maebeale (approved):
Looks great, @RJV333 !

The "series_order" was there bc it's potentially different than workshop-wide positioning (a workshop can be a parent of many 'series_workshops'), but since there's no positioning used workshop-wide, this should be fine.

PR #680, opened 2026-01-16
lazy load workshop show page

jmilljr24:

Workshop show page can potentially have a lot of data/images that need to load depending on the user generated content.

Lazy load the body of the show page via turbo_frame

### Anything else to add?

https://github.com/user-attachments/assets/9a896e1d-7e7c-46ca-a376-671edabc3df6

### DIFF
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -90,13 +90,11 @@ class WorkshopsController < ApplicationController
 
   def show
     if turbo_frame_request?
-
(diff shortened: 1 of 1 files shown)


maebeale (approved):
Oooo, this is great! Definite upgrade.

app/views/workshops/_show_skeleton.html.erb
```diff
@@ -0,0 +1,69 @@
+<div class="bg-white border border-gray-200 rounded-xl shadow p-6 animate-pulse mt-6">
```
maebeale: these skeletons are the best loaders

PR #686, opened 2026-01-17
Improve password reset flow

maebeale:
- Allow users to be able to use an eye icon to toggle display of the password values they're typing
- Improve password-related flash messaging
- Beautify change my password page
- Add a link from change_password page to logout & reset password if you don't remember your password (so can't complete the change password form)
- Reduce possibility of password-reset emails being marked as spam in email

- To achieve the above:
  - Add eye icons, and custom style classes, to password fields w stimulus controller (PasswordToggleController) to toggle display
  - Adjusted flash message after password reset request
  - Extended flash message persistence for specific devise notices
  - Reset password deliverability improvements
    - Added explicit from and reply-to
    - Changed language and links to avoid spam detection
    - Added text.erb to avoid spam detectors
    - Added org name and org logo to html email
    - Added explicit headers_for
    - Added devise_email_name to user so we can address people by first name (helps avoid spam detection)
  - Change password length minimum chars from 1 to 5
  - Added styling to change my password page
  - Added new flow from Change Password to Logout & Reset (if they forgot their password)

Quoted stakeholder feedback addressed with the above:
- Reset password: after clicking on "change my password". The eye icon is not working, I can't see my password when I click on this. Can we change the way this looks?
- Change password (currently logged in): Is there a possibility for users to see the password as they're typing it in? Eye icon?
- Sign-in page: forgot password: Email me reset instructions. Banner message: "...you will receive your password in a few mins. Can we add language to that? Contact programs@awbw.org if you don't receive an email?" 
  - Also, can this stay up a little longer? It showed up and went away quickly.
- Change password after receiving notifcation email: Received, but got "Email looks suspicious..." notication with it. After I clicked "Looks safe" it was no problem. Is there a way to remedy this?
- Change password after receiving notifcation email: Can we update the language for the "Update password instructions" New language, add logo?
- After clicking on "change my password": The eye icon is not working, I can't see my password when I click on this. Can we change the way this looks? Also the tabs for password fields don't have borders. 

- Created a local DeviseMailer to set some defaults
- Added an rspec matcher :eventually to help avoid timing issues
- Added some new ENV vars to access from password related emails
- Added conditional to after_sign_out_path_for to go to password reset if signing out on change_password page

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -118,7 +118,7 @@ Rails.application.routes.draw do
   end
 
   # Root paths
-  root to: "home#index"
+  root to: "dashboard#index"
   authenticated :user do
     root to: "dashboard#index", as: :authenticated_root
   end


config/environments/staging.rb
```diff
@@ -6,6 +6,7 @@
   config.active_storage.service = :digitalocean
   app_host = ENV.fetch("APP_HOST", "localhost")
   Rails.application.routes.default_url_options[:host] = app_host
+  config.action_mailer.asset_host = app_host
```
jmilljr24: We technically only need to add configurations in `staging.rb` that differ from `production.rb` and want to override. Doesn't hurt anything though!!

jmilljr24: We technically only need to add configurations in `staging.rb` that differ from `production.rb` and want to override. Doesn't hurt anything though!!

PR #687, opened 2026-01-17
Notifications refactor

maebeale:

**To normalize automated mailer style, delivery, system behavior, and capture in `notifications`.**

- Update email layout so all system emails are pretty, have logo, etc (devise emails still have their own inline html tho)
- Support more Notification KINDS 
- Separated "fyi" emails to admins from emails to current user (will need a followup PR to send confirmation emails to submitters)
- Allow admins to see all System notifications from admin dashboard (to /notifications)
- Handle sending and logging of devise mailers via callback on DeviseMailer
- Add EventMailer and template so all known mailers are set up the same way
- Separate workshop_log_fyi from report_submitted_fyi in order to add custom table of participants, using some decorator and model methods
- Set notification fyi emails to have the same subject line: `AWBW portal: new WorkshopLog submission by #{@user.full_name}`
- Updated text of notification emails for clarity and styling
- Added text.erb versions of the notification emails to improve deliverability
- Fixed new sessions page to be for current password, not new
- Fix bleed w notifications show being an iframe and causing js havoc
- Add previews for emails so devs can check them out at `/rails/mailers`

- Remove callback on Notification to send an email
- Added more KINDS to Notification
- Removed conditional recipient_role designation
- Adapt NotificationMailerJob to conditionally send or persist the email, so it'll work w devise emails
- Added NotificationServices::PersistDeliveredEmail service to handle updating a notification w sent mailer details
- Adjusted format to be used in multiple controllers (below) to create a Notification, send related email, and persist that email to the notification. Currently added to create actions for WorkshopIdea, StoryIdea, EventRegistration, Report, WorkshopLog, WorkshopVariation (tho this needs to change to WorkshopVariationIdea)
    ```  
    NotificationServices::CreateNotification.call(
      noticeable: @event_registration,
      kind: "event_registration_confirmation_fyi",
      recipient_role: :admin,
      recipient_email: ENV.fetch("REPLY_TO_EMAIL", "programs@awbw.org"),
      notification_type: 0)
    ```
- Constrained users resources so they don't collide w things like users/sign_in & users/sign_out
- Updated notifications factory
- Update settings for tests to clear jobs 

### Anything else to add?
- Updated deprecated calls to unprocessable_entity to unprocessible_content
- Ensure all decorators respond to length arg, if sent (found one of the mailer-related ones didn't have it so updated all)
- TODO
  - Remove foundation email gem
  - Maybe refactor NotificationMailer to be FyiMailer
  - Add confirmation emails for the ones we already have fyi motion for (submitting user gets email w their StoryIdea)
  - Make devise emails use the same mailer layout

### DIFF
--- a/app/views/devise/mailer/reset_password_instructions.html.erb
+++ b/app/views/devise/mailer/reset_password_instructions.html.erb
@@ -1,104 +1,7 @@
-<!DOCTYPE html>
-<html>
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
-  <style>
-      body {
-          margin: 0;
-          padding: 0;
-          background-color: #f9fafb;
-          font-family: Arial, Helvetica, sans-serif;
-          color: #333;
-      }
+<h1>Password reset requested</h1>
 
-      .container {
-          max-width: 600px;
-          margin: 0 auto;
-          background-color: #ffffff;
-          padding: 24px;
-          border-radius: 8px;
-      }
 
-      .logo {
-          text-align: center;
-          margin-bottom: 24px;
-      }
-
-      .logo img {
-          max-width: 180px;
-          height: auto;
-      }
-
-      h1 {
-          font-size: 22px;
-          margin-bottom: 12px;
-          color: #1f2937;
-      }
-
-      p {
-          font-size: 14px;
-          line-height: 1.5;
-          margin: 8px 0;
-      }
-
-      .event-box {
-          background-color: #f3f4f6;
-          border-radius: 6px;
-          padding: 12px;
-          margin: 16px 0;
-      }
-
-      .button {
-          display: inline-block;
-          margin-top: 16px;
-          padding: 10px 16px;
-          background-color: #2563eb;
-          color: #ffffff !important;
-          text-decoration: none;
-          border-radius: 6px;
-          font-weight: bold;
-      }
-
-      .footer {
-          margin-top: 32px;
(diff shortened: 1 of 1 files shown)


app/mailers/event_mailer.rb
```diff
@@ -0,0 +1,25 @@
+class EventMailer < ApplicationMailer
+  helper ApplicationHelper
+
+  default from: ENV.fetch("REPLY_TO_EMAIL", "no-reply@awbw.org")
```
jmilljr24: I'm wondering if we can keep these all in one place instead of repeating if they are the same.

```
production.rb

  config.action_mailer.default_options = {
      reply_to:             'test@example.com'
  }
```
[Rails Guides](https://guides.rubyonrails.org/action_mailer_basics.html#example-action-mailer-configurationl)

maebeale: yes! i'd started this but you reminded me of a couple other changes that i've just pushed. ty!

also, i think NotificationMailer is too tied to the Notification model. Esp when DeviseMailer and EventMailer will also have notifications. so, i'll get this code in and then maybe do that work.

config/routes.rb
```diff
@@ -88,7 +88,7 @@
   resources :story_ideas
   resources :stories
   resources :tutorials
-  resources :users do
+  resources :users, only: [ :new, :index, :show, :edit, :update, :create, :destroy ] do
```
maebeale: needed to add this to avoid collision w users/sign_in and users/sign_out

PR #688, opened 2026-01-17
Add thumbnail asset

jmilljr24:

Add thumbnail asset. This allows models with my asset types to set the image to display when different from the primary asset. For resource, primary is the default downloadable asset. Add rake task to convert types on production assets.

### Anything else to add?

I will add a separate PR to handle adding new thumbnails but any current thumbnail will display correctly from the changes in this PR.

"unpublished" Scholarships - stakeholder request

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -3,8 +3,8 @@ class Resource < ApplicationRecord
   include Rails.application.routes.url_helpers
   include ActionText::Attachable
 
-  PUBLISHED_KINDS = [ "Handout", "Scholarship", "Template", "Toolkit", "Form" ]
-  KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme" ]
+  PUBLISHED_KINDS = [ "Handout", "Template", "Toolkit", "Form" ]
+  KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme", "Scholarship" ]
 
   has_rich_text :rhino_text
 


app/models/asset.rb
```diff
@@ -3,6 +3,11 @@ class Asset < ApplicationRecord
...
   belongs_to :owner, polymorphic: true
   belongs_to :report, optional: true
-  # Images
-  has_one_attached :file, dependent: :purge
+
+  has_one_attached :file, dependent: :purge do |attachable|
+    attachable.variant :thumbnail,
+      resize_to_limit: [ 256, 256 ],
+      format: :jpeg,
+      saver: { quality: 80 }
```
jmilljr24: Future issue - This variant can be used anywhere we know an image will be displayed small, not just thumbnail_assets. This will speed up page loads with many images when we call this variant.

We could add move variants for other sizes like "hero" that match the css sizes we defined or just a couple and call the variant that is the closes in size to what we want displayed.

app/models/resource.rb
```diff
@@ -3,8 +3,8 @@ class Resource < ApplicationRecord
   include Rails.application.routes.url_helpers
   include ActionText::Attachable
 
-  PUBLISHED_KINDS = [ "Handout", "Scholarship", "Template", "Toolkit", "Form" ]
-  KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme" ]
+  PUBLISHED_KINDS = [ "Handout", "Template", "Toolkit", "Form" ]
```
jmilljr24: Requested by stakeholders.

app/views/assets/_display_assets.html.erb
```diff
@@ -1,4 +1,4 @@
 <div class="assets">
-  <%= render "assets/display_image", resource: resource, variant: :hero %>
+  <%= render "assets/display_image", resource: resource, file: (defined?(file) ? file : nil), variant: :hero %>
```
maebeale: wait, where are we defining the hero variant?

app/views/resources/show.html.erb
```diff
@@ -47,7 +48,7 @@
         </div>
 
         <!-- Assets -->
-        <%= render "assets/display_assets", resource: @resource %>
+        <%= render "assets/display_assets", resource: @resource, file: @resource.display_image, variant: :hero %>
```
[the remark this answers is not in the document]
jmilljr24: Nope! I thought that was something you were using to set the css size for image containers.

PR #693, opened 2026-01-18
Add Payment model, factory, and spec

RJV333:
Adds foundational Payment model to be used for Stripe Payments on Event Registrations.  This PR precedes introduction and usage of the Stripe API and is met to cover the associations, attributes, defaults, and validations that will be used in its lifecycle.

Thanks for any feedback

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,0 +1,35 @@
+class Payment < ApplicationRecord
+  # --- Associations ---
+  belongs_to :payer,   polymorphic: true
+  belongs_to :payable, polymorphic: true
+
+  # --- Callbacks ---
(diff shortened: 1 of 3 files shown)


app/models/payment.rb
```diff
@@ -0,0 +1,35 @@
+class Payment < ApplicationRecord
+  # --- Associations ---
+  belongs_to :payer,   polymorphic: true
+  belongs_to :payable, polymorphic: true
+
+  # --- Callbacks ---
```
maebeale: how is this a callback?

maebeale (approved):
Perfect, @RJV333 ! Love the stepwise approach, too.

PR #727, opened 2026-01-19
Upload assets to Resource with turbo updates for type selection

jmilljr24:

Allow users to upload Primary, Thumbnail and Gallery assets for a resource.

Using a mix of turbo frames and turbo streams assets can be uploaded for existing or new resources. Validations are set to prevent more than one Primary or Thumbnail on a resource. Existing resources allow changing the asset type without having to download and reupload. This upload form can be added to any model requiring assets with this pattern.

### Anything else to add?

I should be able to reuse this to some degree with the rich_text_assets but I also need to thing how this will all work together with the library_asset addition. 

https://github.com/user-attachments/assets/2643ebad-d9dc-40a1-ac93-416d7de7f926

### DIFF
--- a/app/views/assets/_collection.html.erb
+++ b/app/views/assets/_collection.html.erb
@@ -1,4 +1,4 @@
-<div id="asset_collection" class="flex flex-wrap gap-2 p-3 m-2 bg-white my-5 rounded-md">
+<div id="asset_collection" class="flex flex-wrap justify-center gap-2 p-3 m-2 bg-white my-5 rounded-md">
   <% owner.assets.compact
       .reject { |a| a.type == "RichTextAsset" }
       .sort_by do |asset|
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -2,12 +2,13 @@
...
-    resources :primary_assets, only: [ :show ]
-    resources :gallery_assets, only: [ :show ]
-  end
-
+  # namespace :assets do
+  #   resources :primary_assets, only: [ :show ]
+  #   resources :gallery_assets, only: [ :show ]
+  # end
+  resources :library_assets
   resources :rich_text_assets
```
jmilljr24: I don't think we are using those other routes but I'm not 100% sure. Also where is the best spot in the routes file for these?

Of note, I originally was going to use just `resources :assets` since this route/controller handles all assets but `asset_path` is already used in rails and was causing issues.

My thoughts are to use library for all at this point and if we need more flexibility we can adjust later. Open to ideas.

PR #738, opened 2026-01-20
CommunityNews and Stories should be printable

maebeale:


maebeale:
@copilot are there any parts of these show pages that we should hide on print?

PR #759, opened 2026-01-22
Add downloadable asset and convert resource primary to downloadable

jmilljr24:

All models with assets have a primary. Primary is what is intended to be shown to the user for a specific model. Resource is the outlier and needs a downloadable asset.

Add downloadable asset. Add rake task to convert types. I'm leaving thumbnail asset in there for now so there is no issues in production and it can be removed later converted.


app/models/resource.rb
```diff
@@ -124,10 +124,6 @@ def name
     title || id
   end
 
-  def download_attachment
-    primary_asset || gallery_assets.first
-  end
```
jmilljr24: I think we can keep is simple for now and just use the downloadable asset.

PR #810, opened 2026-01-29
Add system test for facilitator password reset flow

### LINKED ISSUE #63: Add a system test for: Facilitators can change their password quickly and easily
**Add a system test for:**

**Context:** Today, if an AWBW user forgets their password, when they submit their email for a password reset, it goes to an inbox and someone has to triage it and provide support.

**As** a facilitator,
**I want** to change my password without support
**So that** I can log into AWBW quickly. 

**Given** I have navigated to the AWBW login page,
**When** I click on the "Forgot your password?" 
**Then** I am taken to a screen where I can enter my email address.

**When** I enter my email address 
**And** I click "Submit"
**Then** a password reset link is sent to my email (if it matches the email on file with AWBW)
**And** I see a success message telling me that an email with a password reset link was sent. 

**When** I follow the password reset link in my email,
**Then** I am taken to a screen where I can create a new password.

**When** I submit my new password,
**Then** I am successfully logged into AWBW.

**Notes:** What are the criteria for passwords? We should probably include that info on the reset page.

diti0-dot:
- This work Closes #63 

Add system test for facilitator password reset functionality to ensure facilitators can change password.

Created a facilitator user, navigated through the password reset request flow, retrieved the reset email from the notifications database, extracted the reset token, and followed the complete password reset sequence including setting a new password and verifying successful login.

### Anything else to add?
Maybe there should be a flash message saying they entered an invalid address if it doesn't exist in the db, or maybe like "If your email exists in our system, you will receive reset instructions shortly." And when I was using the notifications db email it says "Hello Kurtis" even though the user name is different - maybe it's just in the db because in terminal it shows the user's actual name.

### DIFF
--- a/spec/system/facilitator_changes_password_test.rb
+++ b/spec/system/facilitator_changes_password_test.rb
@@ -1,0 +1,73 @@
+require 'rails_helper'
+
+RSpec.describe 'Facilitators can change their password quickly and easily' do
+  describe 'Facilitator changes password through reset link' do
+    context 'when facilitator requests password change' do
+      before do
+        @user = create(:user)
+        create(:facilitator, user: @user)
+      end
+
+      it 'completes the full password reset flow successfully' do
+        visit '/users/sign_in'
+
+        expect(page).to have_content('Log in')
+        expect(page).to have_content('Email')
+        expect(page).to have_content('Password')
+        expect(page).to have_link('Forgot your password?')
+
+        click_link('Forgot your password?')
+
+        expect(page).to have_current_path('/users/password/new')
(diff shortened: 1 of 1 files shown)


jmilljr24:
> Maybe there should be a flash message saying they entered an invalid address if it doesn't exist in the db, or maybe like "If your email exists in our system, you will receive reset instructions shortly." 

We want to avoid any indication of if the email exists or not for security. I like your second idea a lot! That would be a better UX for the user when the click reset. Having nothing happen if the submission fails is not the best.

If you want to add that, create and issue see what you come up with.

Issue #819, opened 2026-01-30
Handle Event times


jmilljr24:
To start with, Events should be displayed in Pacific Time but consideration so be made to to handle a user setting for their timezone preference.

PR #837, opened 2026-02-02
Add time-zone field on user

RJV333:
#819 Sets Pacific Time as the default time zone for the app.

Pivoting to using time-zone field on users with a site default of Pacific time.

Screenshot 2026-02-06 at 1 42 31 PM
Screenshot 2026-02-06 at 1 11 50 PM

### DIFF
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -114,9 +114,9 @@ RSpec.describe "/users", type: :request do
 
       it "permits and updates time_zone" do
         user = User.create! valid_attributes
-        patch user_url(user), params: { user: { time_zone: "EST" } }
+        patch user_url(user), params: { user: { time_zone: "Eastern Time (US & Canada)" } }
         user.reload
-        expect(user.time_zone).to eq("EST")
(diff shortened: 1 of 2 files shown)


RJV333:
> oh, fascinating, @jmilljr24 !
> 
> @RJV333 what do you think? i'm leaning toward not doing this systemwide (and removing the other places we're saying PST explicitly too) and instead moving the user-specific TZ stuff to top of the queue.
> 
> any chance you'd want to pick that up (regardless of if we do or don't set system TZ)?

Thanks for the helpful feedback @jmilljr24 and @maebeale, and I agree.  I'm pivoting this PR to track user-specific 'time-zone' like the blog suggests.  It looks like there's some whackness with the specs right now that I *think* is unrelated to these changes, but I'll prioritize finishing this.  In summary I'm planning to finish:

- Adding a 'time_zone' column to the users table with default 'Pacific Time'
- Update application_controller to 'use' a user's time_zone when they're performing an action
- Update user form to adjust time_zone if they want to change from Pacific
- Update specs around creating, storing, and retrieving event datetimes from different user time_zones
- Update Event 'show' to include the timezone an event is being displayed in for extra clarity.

PR #839, opened 2026-02-03
remove unnecessary joins

jmilljr24:

I originally tried to use a cache as this was loading extremely slow in production. I took a deeper look at the queries and we aren't using some of the data in view. Sectorable_items was over 60k records that we don't need for example.


maebeale:
awesome!

Issue #822, opened 2026-01-30
Change all form submission text to "Submit"


jmilljr24:
We use various wording on submit buttons. Change all submit buttons to use just "Submit". If there are any that seem to have a good reason to be different please bring it to our attention.

jmilljr24:
@RJV333 assigned.

I can't add you via the assign button unless you comment on the issue.

PR #843, opened 2026-02-04
Update update/create form submissions to 'Submit'

RJV333:
Addresses #822 

My understanding from our meeting with Rudy was that for forms that create and update objects, using "Submit" consistently could be an improvement.  Looked for forms that used phrases like "Update Logs" or "Create x" to update.   

Some forms that are doing other tasks like "Resend confirmation instructions", "sign up",  or "Set password" etc, I kept the same since I think it's clearer for them to have these custom captions for their services.

### DIFF
--- a/app/views/assets/_upload.html.erb
+++ b/app/views/assets/_upload.html.erb
@@ -57,7 +57,7 @@
     </div>
 
     <div>
-      <%= f.submit "Upload Asset",
+      <%= f.submit "Submit",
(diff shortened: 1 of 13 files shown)


maebeale:
@RJV333 let's keep the `Upload Asset` buttons text how it was so there's only one Submit button per form (and then the tests can pass)

Also, I should've claimed this issue! I've got a fix I'm working on that uses en.yml so the button text is always the same when we use `f.button :submit`. I think good to merge yours first, as mine covers making all the form button code the same.

PR #850, opened 2026-02-05
Normalize published/inactive

maebeale:

- Normalize usage of published/unpublished/hidden/inactive/active/show
- Update policies

### Anything else to add?
- Fix tag_names searching
- Moved common scopes into Publishable and Featureable

### DIFF
--- a/app/models/concerns/publishable.rb
+++ b/app/models/concerns/publishable.rb
@@ -7,7 +7,7 @@ module Publishable
       where(published: value)
     end
 
-    if column_names.include?("publicly_visible")
+    if table_exists? && column_names.include?("publicly_visible")
       scope :publicly_visible, -> { published.where(publicly_visible: true) }
     end
   end


db/migrate/20260205032602_remove_inactive_from_taggings.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveInactiveFromTaggings < ActiveRecord::Migration[8.1]
+  def change
+    # Remove inactive column from sectorable_items and categorizable_items bc they should either exist or not
```
maebeale: there was a boolean on `[taggable]_items` but none of the code was using it to filter taggings -- all were showing in the UI. looks like all of the taggings should be live, so, removing these columns entirely since they're not needed. going forward, an item will either have a tagging or not -- no `inactive/published` state.

jmilljr24 (approved):
Nice job with the Concerns!

Just one comment unrelated to the PR.

PR #851, opened 2026-02-05
Use authorized_scope within controllers

jmilljr24:

Relation scopes were set up in policies. This uses them in the controllers.

### DIFF
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -189,9 +189,9 @@ class WorkshopsController < ApplicationController
   private
 
   def set_show
-    @quotes = Quote.where(workshop_id: @workshop.id).active
+    @quotes = Quote.where(workshop_id: @workshop.id).published
     @leader_spotlights = @workshop.associated_resources.leader_spotlights.where(published: true)
-    @workshop_variations = @workshop.workshop_variations.active
+    @workshop_variations = @workshop.workshop_variations.published
     @sectors = @workshop.sectorable_items.published.map { |item| item.sector if item.sector.published? }.compact if @workshop.sectorable_items.any?
   end
 


app/policies/facilitator_policy.rb
```diff
@@ -18,6 +18,6 @@ def update?
 
   relation_scope do |relation|
     next relation if admin?
-    relation.published.searchable
+    relation.published.searchable # Do we need 'searchabl'? We have  scope :published, -> { active.searchable }
```
maebeale: oh good call! correct, not needed here

app/controllers/workshop_logs_controller.rb
```diff
@@ -4,17 +4,13 @@ class WorkshopLogsController < ApplicationController
   def index
     @per_page = params[:number_of_items_per_page].presence || 10
     params[:workshop_id] ||= @workshop&.id
-    permitted_logs =
-      if current_user.super_user?
-        WorkshopLog.all
-      else
-        WorkshopLog.where(created_by_id: current_user.id)
```
maebeale: did this filter get lost? i don't think it's in the policy

app/controllers/workshops_controller.rb
```diff
@@ -13,12 +13,11 @@ def index
...
       track_index_intent(Workshop, search_service.workshops, params)
 
-      @workshops = search_service.workshops
+      @workshops = authorized_scope(search_service.workshops
                                  .includes(:categories, :windows_type, :user, :images, :bookmarks, :age_ranges,
-                                   user: [ :facilitator ], primary_asset: [ :file_attachment ])
+                                   user: [ :facilitator ], primary_asset: [ :file_attachment ]))
                                  .paginate(page: params[:page], per_page: params[:per_page] || 12)
 
-      @workshops_count = search_service.workshops.size
```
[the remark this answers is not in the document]
jmilljr24: workshop have the total in the view so I didn't touch it.

`@workshops&.total_entries`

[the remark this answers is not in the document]
jmilljr24: Ideally with do it similar in all controllers/views but that's outside the scope of this PR. `@workshops_count` just wasn't used in the view so I removed it for a little house keeping.

app/policies/application_policy.rb
```diff
@@ -20,7 +20,7 @@ def manage?
   # Define shared methods useful for most policies.
```
maebeale: i found note somewhere of adding `scope_matcher :relation, ActiveRecord::Relation` to ApplicationPolicy  and `  after_action :verify_authorized, except: :index` to validate that the indexes are using the scopes. not sure if you'd want to do that?

app/controllers/workshop_logs_controller.rb
jmilljr24: Going off the policy and comment I left off the by `project_id` since we need have the table change coming and there are a bunch more scope this policy will need to filter the search filters if we filter on org as well.
```
  relation_scope do |relation|
    next relation if admin?
    relation.where(user: user) # TODO - maybe also allow users to see peers' within same organization
  end
```

I do notice now that the scope for user should be on `created_by_id`.

maebeale: gotcha re project! 

the report/workshop_log models are actually still `user_id`. so we must've never tested as non-admin.

PR #855, opened 2026-02-06
Update turbo views to update partial

### LINKED ISSUE #854: Fix: Event show page should update the registered/unregister area with turbo
Not sure if this was missed or became broken, but the registration info should update on the show page via turbo the same as it does in the event cards on the index page.

https://github.com/user-attachments/assets/3632730c-31b0-464e-a58d-19161875a199

RJV333:
Fixes #854 , registration button updates via turbo when it's used on event page

https://github.com/user-attachments/assets/c70aeebe-bb7f-4f4e-91cb-b50df9197127


jmilljr24 (approved):
@RJV333 Thanks for jumping right on this! Looks good

PR #875, opened 2026-02-09
Update search on Workshop and Resources

### LINKED ISSUE #860: Resource Index: Keyword search not matching
Basic search on title's are not showing in results.

RJV333:

Closes #860.  I noticed we had the same issue on the Workshops page, and in this case, I think the issue was originating from the same issue, which was the usage of 'join_rich_text' in the SearchCop.

So if we want to add the ability to search rich_text content associated with these records, maybe that's worth doing in a follow up?  But at least for now the basic searches are working again.

I'm new using SearchCop but here's an explanation I got from debugging via AI about how this fix works

Screenshot 2026-02-08 at 10 06 39 PM

Videos of these searches fixed:

https://github.com/user-attachments/assets/a10c64b7-e736-49a7-ba4f-4fff7f975e43
https://github.com/user-attachments/assets/efda5ecb-ea11-4275-9cef-9ae491082b47

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -66,6 +66,10 @@ class Resource < ApplicationRecord
   include SearchCop
   search_scope :search do
     attributes :title, :author, :body
+
+    # scope { join_rich_texts }
+    # attributes action_text_body: "action_text_rich_texts.plain_text_body"
+    # options :action_text_body, type: :text, default: true, default_operator: :or
   end
 
   # Scopes


maebeale:
@RJV333 this looks great!!! would love to get this in asap, once the merge conflict is handled.

spec/models/resource_spec.rb
```diff
@@ -28,8 +28,8 @@
       random_string = Array.new(3) { SecureRandom.alphanumeric(6) }.join(' ')
       user = create(:user)
 
-      resource1 = Resource.create!(title: "One", user: user, kind: "Handout", rhino_body: random_string)
-      resource2 = Resource.create!(title: "Four", user: user, kind: "Handout", rhino_body: "Other")
+      resource1 = Resource.create!(title: "One", user: user, kind: "Handout", body: random_string)
+      resource2 = Resource.create!(title: "Four", user: user, kind: "Handout", body: "Other")
```
maebeale: the "body" is not used in the UI, so we should maybe remove it from here and elsewhere in this test?

jmilljr24:
@RJV333 lets keep Workshop search how it is for the time being. The search queries the stakeholders asked about are working in production right now so I'd rather not change it yet. If you can open another issue with some example that don't work for you that would be great.

For this PR, can you just comment out the rich text stuff on `Resource`. The title is the most important thing right now and we can dive into the text at a later time.

Issue #878, opened 2026-02-09
Add locations & zoom links to events


maebeale:
* Add `videoconference_url` directly to event
* Add Event belongs_to Location

surface these on Event show and in EventRegistration confirmation email

maebeale:
@RJV333 here's another one you might be interested in

PR #901, opened 2026-02-10
878 add location and zoom links to events

RJV333:
- This work Closes [#878 ]

Update the Event model to include both a 'location' field and a 'videoconference_url' field.  Update the form, show page, and event registration confirmation page/email to include these fields

Not sure how many Locations we'll have so thought a searchable select with 'tom-select' would be helpful on the Events form.  If this is overkill or unneccesary, let me know and I'll remove it.

### DIFF
--- a/app/frontend/javascript/controllers/searchable_select_controller.js
+++ b/app/frontend/javascript/controllers/searchable_select_controller.js
@@ -1,0 +1,35 @@
+import { Controller } from "@hotwired/stimulus";
+import TomSelect from "tom-select";
+import "tom-select/dist/css/tom-select.css";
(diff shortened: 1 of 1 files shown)


app/decorators/event_decorator.rb
```diff
@@ -9,9 +9,6 @@ def detail(length: nil)
     length ? description&.truncate(length) : description
   end
 
-  def location # TODO - add location to events
-  end
-
```
maebeale: oo! ty for finding this!

maebeale (approved):
this looks great, @RJV333 !

PR #980, opened 2026-02-13
Add populations_served  to organisation

### LINKED ISSUE #114: New endpoint for organizations/:id/populations_served
Add a new page that lists all service areas for a single organization, listed in order of percentage of associated Facilitators who have that primary service area

Add linked text "Facilitator populations served" under "Populations served" section of Organization Show to get to this new page

diti0-dot:
- This work Closes #114 

Add a populations_served endpoint for organizations that shows which services (sectors) the organization covers and how many people have each as their primary service area.

Grab the people associated with the organization, collect their sectors, count how many have each sector as their primary, and sort the list. 

### Anything else to add?
Nothing else to add


config/routes.rb
```diff
@@ -83,6 +83,16 @@
     end
   end
   resources :organizations
+  resources :notifications, only: [ :index, :show ] do
```
maebeale: this adds a duplicate notifications section in routes. would you remove @diti0-dot ?

PR #1038, opened 2026-02-15
Some random changes

maebeale:

Grab bag of changes:
- order case insensitively and stripped chars
- require person_id
- avoid ahoy errors
- change windows dropdown label


app/services/workshop_search_service.rb
```diff
@@ -3,6 +3,8 @@ class WorkshopSearchService
   include PunctuationStrippable
   authorize :user
 
+  TITLE_SORT_SQL = "LOWER(TRIM(#{Workshop.strip_punctuation_sql_spaced('workshops.title')})) ASC".freeze
```
maebeale: doing it as a frozen constant helped avoid brakeman concern

PR #1040, opened 2026-02-16
Improve user edit UX, incl add comments

maebeale:
#### User/Person UX
  - Polymorphic comments — Add inline comments to `user form` matching person form pattern. Add updated_by to comments and show that
  - Redesigned `user show` page with structured label/value layout, account event history (Ahoy), and comments section
  - Updated `user form` with improved lock/unlock flow using a boolean field and Stimulus controllers
  - User-Person email sync — Updating a user's email now automatically updates the associated person's email field via after_update callback
  - Updated `user index`, esp around which icons are showing and adding informative hover text
  - Added created_by/updated_by to users — Migration adding these tracking columns directly to the User model
  - Expanded person and user search — User search now also matches against person.email and person.email_2
  - Scope rename — Renamed `User.active` scope to `User.has_access` for clarity, updated all call sites across controllers and policies
  - Person form flow — Improved navigation flow from person to user creation, retaining hidden user_id on person form
  - Duplicate check improvements — Skip flagging duplicates when a person's email matches the user's email and they're already linked; updated check duplicates view styling

  - Devise configuration — Reinstated `confirmable` on email change; changed `lockable` settings so only admins can unlock accounts
  
#### Other updates
  - Ahoy tracking enhancements — Enhanced Ahoy event properties globally to capture more context; added Ahoy events for all Devise mailers
  - Dirty form tracking — Added dirty_form_controller to warn on unsaved changes
  - Bulk invite rake task — Added `bulk_invite` rake task for inviting multiple users at once

### DIFF
--- a/spec/system/reset_password_person_spec.rb
+++ b/spec/system/reset_password_person_spec.rb
@@ -9,8 +9,8 @@ RSpec.describe 'Reset password (person)', type: :system do
     sign_in person_user
     # Start on a page where the user nav is visible
     visit root_path
-    # Open user nav dropdown
-    find('#avatar button').click
+    # Open user nav dropdown (wait for avatar to appear after page load)
+    find('#avatar button', wait: 10).click
     # Click "Change password"
     click_link 'Change password'
   end


app/models/user.rb
```diff
@@ -152,23 +153,10 @@ def name
...
-    public_send("bookmarked_#{bookmarkable_type.downcase.pluralize}")
-      .pluck(:id)
-  end
-
-  def primary_asset # method needed for idea_submission_fyi mailer
+  def primary_asset # method needed for idea_submitted_fyi mailer
   end
 
-  def gallery_assets # method needed for idea_submission_fyi mailer
+  def gallery_assets # method needed for idea_submitted_fyi mailer
```
maebeale: fix comment to correct mailer name

config/initializers/devise.rb
```diff
@@ -172,11 +176,7 @@
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
   # config.unlock_strategy = :both
-  config.unlock_strategy = :email
-
-  # Number of authentication tries before locking an account if lock_strategy
-  # is failed attempts.
-  config.maximum_attempts = 10
+  config.unlock_strategy = :none
```
maebeale: don't allow people to unlock themselves via email. only admins can unlock an account.

deprecating usage of Inactive in favor of using devise's lockable, which makes users unable to log in

PR #1046, opened 2026-02-17
Update admin-only page styling

maebeale:

- Instead of a blue border, make background of page entirely blue for admin-only pages

BEFORE:
Screenshot 2026-02-16 at 8 48 03 PM

AFTER:
Screenshot 2026-02-16 at 8 47 10 PM

### DIFF
--- a/app/views/banners/edit.html.erb
+++ b/app/views/banners/edit.html.erb
@@ -1,5 +1,5 @@
 <% content_for(:page_bg_class, "admin-only bg-blue-100") %>
-<div class="bg-white border border-gray-200 rounded-xl shadow p-6">
+<div class="<%= DomainTheme.bg_class_for(:banners) %> border border-gray-200 rounded-xl shadow p-6">
           <div class="flex items-center justify-between mb-6">
             <h1 class="text-2xl font-semibold text-gray-900">Edit <%= @banner.class.model_name.human %></h1>
             <%= link_to "View", banner_path(@banner),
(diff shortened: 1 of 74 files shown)


app/views/admin/ahoy_activities/charts.html.erb
```diff
@@ -2,10 +2,10 @@
...
 </style>
 
-<div class="max-w-9xl mx-auto px-4 py-8 space-y-16">
-  <div class="admin-only bg-blue-100 p-6">
-    <div class="bg-white p-3">
-      <div class="p-6">
+<% content_for(:page_bg_class, "admin-only bg-blue-100") %>
+<div class="bg-white border border-gray-200 rounded-xl shadow p-6">
+      <div class="flex items-start justify-between mb-12">
+        <h1 class="text-3xl font-bold text-gray-900">Activity Charts</h1>
```
maebeale: set all pages to have same wrapper divs and width. and add `content_for(:page_bg_class, "admin-only bg-blue-100")` on pages that should be admin-only blue background.

app/frontend/javascript/controllers/collection_controller.js
```diff
@@ -54,6 +54,12 @@ export default class extends Controller {
     });
   }
 
+  clearAndSubmit(event) {
+    event.preventDefault();
+    this.element.reset();
+    this.submitForm();
+  }
+
```
maebeale: this is related to Clear filter buttons on indexes

PR #1053, opened 2026-02-17
Fix page background color based on policies (using content_for classes)

maebeale:

- Fix content_for classes based on policies
- Some of the temporarily-gated ones are set to their expected, not current policy tag

### DIFF
--- a/app/views/monthly_reports/show.html.erb
+++ b/app/views/monthly_reports/show.html.erb
@@ -1,4 +1,4 @@
-<% content_for(:page_bg_class, "admin-or-owner") %>
+<% content_for(:page_bg_class, "admin-or-owner-or-member") %>
 <% children_total = @monthly_report.children_ongoing.to_i + @monthly_report.children_first_time.to_i %>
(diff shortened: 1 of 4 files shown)


app/models/person.rb
```diff
@@ -73,6 +73,8 @@ class Person < ApplicationRecord
     attributes contact_methods_phone: "contact_methods.value"
   end
 
+  scope :organization_ids, ->(organization_ids)  { joins(:affiliations)
+                                                     .where(affiliations: { organization_id: organization_ids }) }
```
maebeale: needed for scope for report/workshop_log/monthly_report

app/views/notifications/show.html.erb
```diff
@@ -1,10 +1,10 @@
-<% content_for(:page_bg_class, "admin-only bg-blue-100") %>
+<% content_for(:page_bg_class, "admin-or-owner") %>
```
maebeale: noting this is `admin? || owner?` but isn't exposed in the UI yet

PR #1068, opened 2026-02-17
Fix lockable

maebeale:

- We want lockable enabled so when a user has locked_at, they aren't able to log in


config/locales/devise.en.yml
```diff
@@ -8,11 +8,11 @@ en:
...
       already_authenticated: "You are already signed in."
-      inactive: "Your account is currently inactive. Please email %{organization_name} to find out more."
-      invalid: "Invalid %{authentication_keys} or password."
-      locked: "Your account is locked."
+      inactive: "Invalid email or password. Please contact us for assistance."
+      invalid: "Invalid email or password. Please contact us for assistance."
+      locked: "Please contact us for assistance."
       last_attempt: "You have one more attempt before your account is locked."
-      not_found_in_database: "Invalid %{authentication_keys} or password."
+      not_found_in_database: "Invalid email or password. Please contact us for assistance."
```
maebeale: change devise flash msgs to be more generic for security and available data issue.

PR #1071, opened 2026-02-18
Person show affiliation dates display

maebeale:

- Show Affiliated since and Facilitator since on Person and Org show pages
(these used to be manual fields, but now autocalculating based on associated affiliations, if present)


app/models/organization.rb
```diff
@@ -141,7 +145,8 @@ def organization_locality
   end
 
   def published? # needed for my_bookmarks
-    organization_status&.name == "Active"
+    return true if organization_status&.name == "Active"
+    affiliations.active.exists?
```
maebeale: this has potential to be an issue. will leave it like this for now.

PR #1075, opened 2026-02-18
Docker dev setup tweaks

eugenebolshakov:

I've tried following the docker setup instructions in `CONTRIBUTING.md` and made some tweaks/fixes along the way. 

Here's what I changed:

- **Added a note about `.env` to Docker dev setup** — Docker setup uses the `.env` file, so it must be present.
- **Added `bin/setup` step to Docker dev setup instructions** — Docker setup creates the dev DB and some tables but doesn’t run migrations or seeds, so it's necessary to run `bin/setup` as well.
- **Re-build images when starting dev Docker containers** — New gems/npm packages require rebuilding images so that dependencies are installed; otherwise docker containers can't be started,  because the startup commands (e.g. `bin/rails server`) fail.
- **Use `npm ci` instead of `npm install` in `bin/setup`** — This makes it consistent with `Dockerfile.dev` and is probably a good idea anyway because `npm ci` installs the versions specified in the lock file. 
- **Auto-confirm users created in DB seeds** — Otherwise it's not possible to log in using the default credentials because their emails are not confirmed.
- **Exclude system specs when running specs in the dev Docker container** — Dev container has no ChromeDriver, so system specs fail (hang on timeout when connecting to ChromeDriver); It's probably best not to run them at all, until the dev container has everything that's needed to run them.
- **Assert on the full email in `spec/requests/users_spec.rb`** — This was just a flaky test that I noticed while setting things up. 

Please see the commit messages for more info.

### Anything else to add?

AFAIU the docker dev setup is a WIP and not fully functional (e.g. one can't run system specs via docker) and there is some duplication between the on-the-host and docker setups. I guess it'd be good to have a fully working docker dev setup eventually, but these are just a few quick fixes that allowed me to get started.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,6 +69,8 @@ For Docker-based development (recommended):
 5. Visit <http://localhost:3000/> to see the AWBW portal page
 6. Log in as a sample user with the default [credentials](#credentials)
 
+**Note:** It's not currently possible to run system tests via Docker, so they are excluded when you run `mise docker-spec`.
+
 ### Database encoding
 
 Different schema encodings on Mac vs Linux (utf8mb4 vs utf8) were causing merge challenges in schema.rb.


jmilljr24 (approved):
@eugenebolshakov Thanks for helping out with this! Local docker dev has lagged behind as we focused on other things.

Issue #1062, opened 2026-02-17
Make clicking on Created column header sort community news results by date


maebeale:
(click again to reverse the search)
show an up or down arrow after clicking to show the applied search

PR #1083, opened 2026-02-18
Add sortable Created At and Updated columns to Commmunity News

RJV333:
- This work Closes [#1062 ]

1) Adds a Created At column to accompany Updated (At)
2) tweaks styling so dates fit on one line
3). Adds sortability to both these date columns

https://github.com/user-attachments/assets/965eea8a-4f6a-4432-a19d-bd9893e02bd0


maebeale (approved):
awesome, @RJV333 ! 

would you be willing to make all the columns sortable either here or in a subsequent PR?
also, ideally we still only have one date column -- would you remove updated_at?

PR #1084, opened 2026-02-19
WIP: Fix image/file uploads for Community News, Resources, and Stories

maebeale:
### Summary
1. Add error handling and show messages to the user
2. Change filetypes accepted
3. Make sure all images on show pages are clickable to new window of just the image
4. Redirect to show after saving a Resource
5. Add "preview" for filetypes upon choosing file

### Claude AI description
Accept modern image formats (WebP, HEIC, HEIF) that were being silently rejected—especially iPhone photos which default to HEIC. Fix the file_type validation to use polymorphic dispatch (self.class::ACCEPTED_CONTENT_TYPES) instead of a hardcoded case statement. Add accept attributes and accepted type hints to file inputs so users see what's allowed before uploading. Add direct-upload:error handling so upload failures surface visibly instead of silently dropping files.

### DIFF
--- a/spec/views/resources/show.html.erb_spec.rb
+++ b/spec/views/resources/show.html.erb_spec.rb
@@ -9,20 +9,29 @@ RSpec.describe "resources/show", type: :view do
     allow(view).to receive(:current_user).and_return(admin)
     allow(view).to receive(:allowed_to?).and_return(true)
     assign(:resource, resource.decorate)
+    assign(:mentions, [])
   end
 
   context "when resource has an attached primary image" do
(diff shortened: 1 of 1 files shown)


app/views/tutorials/show.html.erb
```diff
@@ -36,7 +36,7 @@
           </div>
         <% end %>
 
-        <%= render "assets/display_assets", resource: @tutorial %>
+        <%= render "assets/display_assets", resource: @tutorial, link: true %>
```
maebeale: we want show page images to open in new window

PR #1136, opened 2026-02-21
Extract display image logic into DisplayImagePresenter service

maebeale:
Move input normalization, variant resolution, CSS class building, and link computation out of the 175-line _display_image partial into a presenter service. The partial is now ~70 lines of pure rendering. Leverages Active Storage named :thumbnail variant instead of manual preview/resize calls.


maebeale:
@jmilljr24 what do you think of this?

PR #1178, opened 2026-02-23
Patch bugs dummy_dev_seeds

RJV333:
Runny these seeds locally, I was hitting issues on the validations of rhino_body presence's and outdated field names.  

These run for me now when I run: mise exec -- rails db:dev:seed.  Still getting some other issues I might put in a followup, but for now all the helpful info in 'db/seeds/dummy_dev_seeds.rb' is working.


maebeale:
ha! TY! i was just in here too. thanks, @RJV333 !

PR #1196, opened 2026-02-23
Add WEB_CONCURRENCY to puma

johnpaulashenfelter:

Allow for higher throughput in puma with the larger server

Added the common config to the puma for WEB_CONCURRENCY


johnpaulashenfelter:
Current DO droplets are 1vCPU so no point in using this for now.

PR #1301, opened 2026-03-02
Forms Part 1: Extract BaseRegistrationFormBuilder to share fields across form types

maebeale:

- Three form builder services (Short, Extended, Scholarship) duplicated identical helper methods, scholarship fields, and contact fields
- Changing a shared field (e.g., adding email confirmation) required updating multiple files independently, risking drift
- Several field key mismatches between builders caused silent bugs in validation and tag assignment

- Extracted `BaseRegistrationFormBuilder` with shared helpers (`add_header`, `add_field`) and reusable field sections (`build_basic_contact_fields`, `build_scholarship_fields`, `build_consent_fields`)
- All three builders now inherit from the base class, eliminating ~210 lines of duplication
- Fixed three bugs discovered during the refactor:
  - **Email confirmation key mismatch**: Extended form used `primary_email_confirmation` but controller validation checked `confirm_email` — confirmation match validation was silently skipped for extended forms
  - **`workshop_settings` vs `workshop_environments`**: Builder created `workshop_settings` but the service and helper looked for `workshop_environments` — professional tags were never assigned
  - **Redundant confirm email storage**: `save_form_fields` skipped `primary_email_confirmation` but not `confirm_email` — short form stored confirmation email unnecessarily
- Updated email row logic in `new.html.erb` to support 3-column layout (email | confirm | type) when all three fields exist
- Added specs for Short and Extended form builders

### UI Testing Checklist

- [ ] Short form registration renders correctly with contact, consent, qualitative, and scholarship sections
- [ ] Extended form registration renders correctly with all sections
- [ ] Email confirmation validation works on both form types
- [ ] Email row displays as 3-column on extended form (email | confirm email | email type)
- [ ] Email row displays as 2-column on short form (email | confirm email)
- [ ] Registration show page displays correctly, skipping confirm email field
- [ ] Scholarship application form still builds correctly

### Anything else to add?

- This is forward-only: existing forms in the database are untouched. Only newly built forms will use the shared definitions.
- The base class is designed for future extensibility — an `AuthenticatedEventRegistrationFormBuilder` can inherit the same shared sections and add its own fields.


app/controllers/events/public_registrations_controller.rb
```diff
@@ -82,14 +82,14 @@ def show
...
+      @form_submission = @form.form_submissions.find_by(person: person)
+      unless @form_submission
         redirect_to event_path(@event), alert: "No registration form submission found."
         return
       end
 
-      @form_fields = @form.form_fields.where(status: :active).reorder(position: :asc)
-      @responses = @person_form.person_form_form_fields.index_by(&:form_field_id)
+      @form_fields = @form.form_fields.reorder(position: :asc)
+      @responses = @form_submission.form_answers.index_by(&:form_field_id)
```
maebeale: changing responses from `person_form_form_fields` to `form_answers`

PR #1318, opened 2026-03-03
Show windows type in workshop select dropdowns

maebeale:

- Duplicate record handling: workshop remote-select and searchable-select fields previously only showed the workshop title, making it hard to distinguish workshops with the same name but different windows types
- Now displays title with windows type short_name in parens, e.g. "Healing Through Color (Adult)"
- Only appends `#ID` when multiple search results share the same title and type, keeping labels clean
- Workshop edit page heading now also shows the full title with windows type

- Added `remote_search_label` instance method override on Workshop to return `"Title (ShortName)"`
- Added `remote_search_labels` class method on Workshop that post-processes a collection to append `#ID` only for duplicate labels
- Added default `remote_search_labels` to `RemoteSearchable` concern (just maps to `remote_search_label`)
- Updated `SearchController` to call `model_class.remote_search_labels(records)` instead of mapping individually
- Overrode `remote_search` to eager-load `windows_type` to avoid N+1 queries
- Updated 5 view form partials to use `remote_search_label[:label]` for the pre-selected workshop value
- Added duplicate-title seed workshops ("Healing Through Color") to exercise disambiguation in dev
- Updated workshop edit page heading to show `remote_search_label[:label]`

### UI Testing Checklist

- [x] Search for a workshop with a unique title in any workshop select dropdown — shows "Title (Type)" without ID
- [x] Search for "Healing Through Color" — both results show "Healing Through Color (Adult) #ID"
- [x] Edit a workshop log — pre-selected workshop shows "Title (Type)" without ID
- [x] Edit a story — pre-selected workshop shows "Title (Type)" without ID
- [x] Edit a workshop variation — pre-selected workshop shows "Title (Type)" without ID
- [x] Edit a story idea — pre-selected workshop shows "Title (Type)" without ID
- [x] Edit a workshop variation idea — pre-selected workshop shows "Title (Type)" without ID
- [x] Workshop edit page heading shows "Edit workshop: Title (Type)"

### Anything else to add?

- Uses the existing `type_name` method pattern but splits it: base label omits ID, collection-level disambiguation adds it only when needed
- The `remote_search_labels` default in `RemoteSearchable` means other models (Person, User, Organization) are unaffected

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -25,7 +25,10 @@ class SearchController < ApplicationController
 
     records = records.limit(25)
 
-    render json: model_class.remote_search_labels(records)
+    labels = records.map(&:remote_search_label)
+    labels = model_class.resolve_duplicate_labels(labels) if model_class.respond_to?(:resolve_duplicate_labels)
+
+    render json: labels
   end
 
   private
--- a/app/models/workshop.rb
+++ b/app/models/workshop.rb
@@ -335,8 +335,7 @@ class Workshop < ApplicationRecord
     { id: id, label: label }
   end
 
-  def self.remote_search_labels(records)
-    labels = records.map(&:remote_search_label)
+  def self.resolve_duplicate_labels(labels)
     dupes = labels.group_by { |l| l[:label] }.select { |_, v| v.size > 1 }.keys.to_set
     labels.each { |l| l[:label] = "#{l[:label]} ##{l[:id]}" if dupes.include?(l[:label]) }
     labels
--- a/app/models/concerns/remote_searchable.rb
+++ b/app/models/concerns/remote_searchable.rb
@@ -10,10 +10,6 @@ module RemoteSearchable
       @remote_search_columns || []
     end
 
-    def remote_search_labels(records)
-      records.map(&:remote_search_label)
-    end
-
     def remote_search(query)
       return none if query.blank?
       raise "remote_searchable_by not defined for #{name}" if remote_search_columns.empty?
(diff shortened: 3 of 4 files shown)


app/models/workshop.rb
```diff
@@ -326,6 +326,22 @@ def attach_assets_from_idea!
...
+    { id: id, label: label }
+  end
+
+  def self.remote_search_labels(records)
+    labels = records.map(&:remote_search_label)
+    dupes = labels.group_by { |l| l[:label] }.select { |_, v| v.size > 1 }.keys.to_set
+    labels.each { |l| l[:label] = "#{l[:label]} ##{l[:id]}" if dupes.include?(l[:label]) }
+    labels
+  end
+
```
maebeale: update workshop-specific remote search label display. and conditionally show workshop `#ID` if there are two workshops w same name AND windows type short name (there are many of these rn in prod dataset)

app/controllers/search_controller.rb
```diff
@@ -25,7 +25,7 @@ def index
 
     records = records.limit(25)
 
-    render json: records.map(&:remote_search_label)
+    render json: model_class.remote_search_labels(records)
```
jmilljr24: I'm wondering if we can make this opt-in instead of sending everything through the class method. And maybe a different name than the instance method.

I knew what the intent was given the context of the PR and it being specific to `Workshops`. It just took some brain power to follow and I could see it not being obvious for someone else reading the code (or me in 6 months haha)

Maybe something like this


```suggestion
labels = records.map(&:remote_search_label)
labels = model_class.resolve_duplicate_labels(labels) if model_class.respond_to?(:resolve_duplicate_labels)

render json: labels
```

Then the concern has no changes and we just keep the duplicate handling class method in `Workshop`

PR #1319, opened 2026-03-03
Fix register button hover text color on event cards

maebeale:
- The "Register" button on event cards didn't show white text on hover
- Inline `color` style was overriding the Tailwind `hover:text-white` class due to CSS specificity

- Moved the text color from inline `style` to a Tailwind arbitrary value class (`text-[rgb(170,46,0)]`)
- This lets the `hover:text-white` class properly override on hover, since both are now class-based
- Applied to both the `button_to` (signed-in users) and `link_to` (public registration) variants

### Anything else to add?
- No functional changes, purely a CSS specificity fix

### DIFF
--- a/app/views/events/_card.html.erb
+++ b/app/views/events/_card.html.erb
@@ -68,14 +68,12 @@
       <% if user_signed_in? %>
         <%= button_to "Register",
                     event_registrant_registration_path(event_id: event),
-                    class: "btn px-3 py-2 text-xs uppercase leading-tight text-[rgb(170,46,0)] hover:text-white hover:bg-orange-700",
-                    style: "font-family: 'Telefon Bold', sans-serif; border: 2px solid rgb(170, 46, 0);" %>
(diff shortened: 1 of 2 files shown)


app/views/events/_card.html.erb
```diff
@@ -68,14 +68,14 @@
       <% if user_signed_in? %>
         <%= button_to "Register",
                     event_registrant_registration_path(event_id: event),
-                    class: "btn px-3 py-2 text-xs uppercase leading-tight hover:text-white hover:bg-orange-700",
-                    style: "font-family: 'Telefon Bold', sans-serif; color: rgb(170, 46, 0); border: 2px solid rgb(170, 46, 0);" %>
+                    class: "btn px-3 py-2 text-xs uppercase leading-tight text-[rgb(170,46,0)] hover:text-white hover:bg-orange-700",
```
jmilljr24: Can this color be added to `application.tailwind.css` since we use it multiple places.
```
@theme {
  --color-primary: #063b8d;

```

Also I'd prefer if this button didn't use inline styles if we can do it with tailwind. Keeps things consistent.

PR #1365, opened 2026-03-08
Keep View registration button lit for public users via reg param

maebeale:

- Public (non-authenticated) users who register for events lose the "View registration" button as soon as they navigate away
- They have no logged-in identity to look the registration up by, so once the `?reg=` slug drops off the URL the button disappears on the events index and on event pages
- This keeps the button lit for the current browsing session, including for someone who registers for 2–3 events in a row

- Threads the registration slug through the URL as a `?reg=` param instead of persisting anything server-side
- Multiple slugs are packed into one param, period-delimited (`?reg=slugA.slugB`) — a period never occurs in a `urlsafe_base64` slug, so the URL stays clean and collision-free
- `EventsHelper` parses the param, matches a registration per event, and merges a new slug in after registering so earlier events stay lit
- The param is forwarded across the "← Events" / card / register links and the ticket page, and survives the Stripe checkout round-trip for paid events

### Why URL param instead of session?

- **Privacy on shared devices** — nothing is stored server-side, so the next person on a library/public computer never sees a prior visitor's "View registration" button (a session would have leaked it until expiry)
- **Simplicity** — no session state machine to store/clear/expire; the URL is the single source of truth and is naturally ephemeral (close the tab, it's gone)

### Trade-offs

- The button only persists along links that forward the param — a bookmark, the Home link, or a return visit tomorrow won't show it. This is intended for a single browsing session, and is arguably correct for anonymous users
- Cancelling/reactivating doesn't re-thread the full param set (rare action); the ticket page still shows the correct state

### UI Testing Checklist

- [ ] Register as public user → visit ticket → "← Events" → "View registration" appears on that event's card
- [ ] Click into the event from the index → button still visible
- [ ] Register for a second event in the same session → both cards show "View registration" on the index
- [ ] Paid event: register → pay via Stripe → return to ticket → button persists
- [ ] Cancelled registration → card shows "Register again", not "View registration"
- [ ] Logged-in user registration flow still works as before

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -1,4 +1,31 @@
 module EventsHelper
+  # Separates registration slugs packed into a single `?reg=` param so public
+  # (logged-out) users keep the "View registration" button lit across several
+  # events in one session. A period is used because it never appears in a
+  # urlsafe_base64 slug and stays unescaped in a URL, keeping it human-readable.
+  REG_PARAM_DELIMITER = "."
+
+  # The registration slugs carried in the current request's `?reg=` param.
+  def reg_param_slugs
+    params[:reg].to_s.split(REG_PARAM_DELIMITER)
+  end
+
+  # The registration for this event that the visitor has "unlocked" via their
+  # `?reg=` param, if any. Slugs are the authorization token, so no ownership
+  # check is needed. Returns nil when none of the param slugs belong to the event.
+  def reg_param_registration(event)
+    slugs = reg_param_slugs
+    return if slugs.empty?
+    EventRegistration.where(slug: slugs, event_id: event.id).first
+  end
+
+  # The current `?reg=` slug set with `slug` merged in, for forwarding onward
+  # links after a registration so the new event's button stays lit alongside
+  # any events registered earlier this session.
+  def reg_param_with(slug)
+    (reg_param_slugs << slug).uniq.join(REG_PARAM_DELIMITER)
+  end
+
   # Stable anchor id for a registrant's row on the Onboarding matrix, so back-links
   # from detail pages can scroll to (and highlight) the row they came from.
   def onboarding_row_id(record_or_id)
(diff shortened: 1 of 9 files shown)


app/helpers/events_helper.rb
```diff
@@ -1,4 +1,31 @@
 module EventsHelper
+  # Separates registration slugs packed into a single `?reg=` param so public
+  # (logged-out) users keep the "View registration" button lit across several
+  # events in one session. A period is used because it never appears in a
+  # urlsafe_base64 slug and stays unescaped in a URL, keeping it human-readable.
+  REG_PARAM_DELIMITER = "."
```
maebeale, line 6: 🤖 _From Claude:_ Period is the delimiter because slugs are `urlsafe_base64` (alphabet includes `-` and `_`) — a hyphen delimiter could collide with a slug that contains one. `.` never appears in base64url and stays unescaped in URLs.

app/views/events/registrations/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for(:page_bg_class, "public") %>
 <div class="max-w-3xl mx-auto flex flex-wrap items-center gap-2 px-4 pt-4">
-  <%= link_to event_path(@event_registration.event, reg: @event_registration.slug), class: "inline-flex items-center gap-1.5 text-sm text-gray-500 hover:text-gray-700 px-2 py-1" do %>
+  <%= link_to event_path(@event_registration.event, reg: reg_param_with(@event_registration.slug)), class: "inline-flex items-center gap-1.5 text-sm text-gray-500 hover:text-gray-700 px-2 py-1" do %>
```
maebeale, line 3: 🤖 _From Claude:_ This is the one place a slug gets *added* to the set — after any registration the visitor lands here, so merging this ticket's slug into the inbound `?reg=` param is what lets a second/third event stay lit alongside earlier ones.

PR #1369, opened 2026-03-08
Fix duplicate check losing form params and nested attributes

maebeale:
Closes #1023

- When creating a person/user triggers duplicate detection, all form data (especially nested attributes like sectors, addresses, phones, affiliations) was lost because the original implementation used `redirect_to` which discards POST params
- The "Go back" button used `history.back()` which doesn't restore dynamically-added Cocoon fields
- Also fixes a pre-existing bug where the "+ Add phone" button didn't work at all

- Replace `redirect_to` with `render :check_duplicates` to preserve POST params in controller state
- Add `hidden_fields_for_params` helper to round-trip all form params (including nested attributes) through hidden fields on the interstitial page
- Replace `history.back()` "Go back" with a `retry_new` POST action that rebuilds `@person` from params and re-renders the `new` form with nested objects pre-populated
- Add `retry_new` collection route and controller action for both people and users
- Fix "+ Add phone" cocoon bug: hidden `kind` field was outside the `.nested-fields` div, so cocoon's `firstElementChild` grabbed only the `<input>` and discarded the form fields
- Add "Ask a colleague" mailto link on duplicate check interstitials
- Add avatar warning when file upload can't be preserved through the round-trip
- Skip blank `id` fields and `UploadedFile` objects in hidden field generation

### UI Testing Checklist
- [ ] Create person with duplicate name → interstitial shows with correct data
- [ ] Click "Create anyway" → person created with sectors, addresses, phones, affiliations
- [ ] Click "Go back" → form re-renders with all nested attributes preserved
- [ ] Exact match (blocked) → only "Go back" shown, no "Create anyway"
- [ ] "Ask a colleague" mailto link generates correct email
- [ ] Avatar warning appears when photo was uploaded
- [ ] "+ Add phone" button works on new person form
- [ ] Create user with duplicate email → same interstitial flow works

### Anything else to add?
- Avatar files cannot be preserved through hidden fields (known limitation) — user sees a warning to re-upload
- The cocoon phone button fix (moving hidden field inside `.nested-fields` div) is a separate commit that fixes a pre-existing bug

### DIFF
--- a/spec/requests/users_create_with_dupe_check_spec.rb
+++ b/spec/requests/users_create_with_dupe_check_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe "POST /users with duplicate check", type: :request do
   let(:admin) { create(:user, :admin) }
-  let(:person) { create(:person, user: nil) }
+  let!(:person) { create(:person, user: nil) }
 
   before { sign_in admin }
 


maebeale:
@jmilljr24 this is the "real" dupe check update

jmilljr24:
When I "go back" Org affiliation name is displaying the id, not the name. Everything looked ok for Person. 

I saw some code regarding avatars but I never got it to display in the UI. It sounds like if an avatar is added in the form, you lose it if there is a dup? Kinda of unfortunate but lower on the list of priorities as Admin at this point probably are not adding avatars for people, the person will do it themselves.

I'm good with this approach as is but I do think there is a better UX we could come up with in the future. Regarding the AI quote, it doesn't align with the flow/implementation I can envision. It's also confusing/mixing turbo frames vs turbo streams. I wouldn't dismiss other alternative idea based off that response.

PR #1370, opened 2026-03-08
Checkbox tom-select

jmilljr24:

This sets up the framework to add tom-select to checkbox's

### Anything else to add?

There is a bug on the clear filters button prior to these changes. I couldn't find the immediate issue.

https://github.com/user-attachments/assets/d095de41-f9db-49c8-bd69-4dd81fc34b73


jmilljr24:
> ooo. yeah, this is way better! i wish they could still fit on the one row, but get that we want wider so we can add the selected checks into the box. THANKS for figuring this out!!!

Yup. Exactly my train of thought.

PR #1419, opened 2026-03-13
Fix workshop_id population in workshop_logs migration

maebeale:

- The `workshop_logs:migrate_from_reports` rake task wasn't populating `workshop_id` for reports that stored the workshop association via `owner_type`/`owner_id` instead of `workshop_id`
- This left migrated workshop_logs with NULL `workshop_id`, breaking workshop associations

- Used `COALESCE(r.workshop_id, CASE WHEN r.owner_type = 'Workshop' THEN r.owner_id END)` so the migration prefers `workshop_id` when present and falls back to `owner_id` for polymorphic associations
- Added two specs: one for the `owner_id` fallback case, one verifying `workshop_id` takes precedence when both are present

### Anything else to add?

- The rake task already clears existing workshop_logs before re-inserting, so re-running it will fix previously migrated data

### DIFF
--- a/lib/tasks/migrate_workshop_logs.rake
+++ b/lib/tasks/migrate_workshop_logs.rake
@@ -35,7 +35,8 @@ namespace :workshop_logs do
         SELECT
           r.id,
           CASE WHEN u.id IS NOT NULL THEN r.created_by_id ELSE #{orphaned_user.id} END,
-          r.organization_id, r.windows_type_id, r.workshop_id,
+          r.organization_id, r.windows_type_id,
+          COALESCE(r.workshop_id, CASE WHEN r.owner_type = 'Workshop' THEN r.owner_id END),
           r.date, r.rating, COALESCE(r.external_workshop_title, r.workshop_name),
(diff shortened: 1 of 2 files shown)


lib/tasks/migrate_workshop_logs.rake
```diff
@@ -35,7 +35,8 @@ namespace :workshop_logs do
         SELECT
           r.id,
           CASE WHEN u.id IS NOT NULL THEN r.created_by_id ELSE #{orphaned_user.id} END,
-          r.organization_id, r.windows_type_id, r.workshop_id,
+          r.organization_id, r.windows_type_id,
```
maebeale: COALESCE prefers the direct workshop_id when present, and only falls back to owner_id for reports that used the polymorphic association pattern. The CASE expression returns NULL when owner_type isn't 'Workshop', so non-workshop polymorphic owners won't accidentally populate workshop_id.

maebeale:
Already merged into main via #1418.

PR #1474, opened 2026-04-04
Fix flaky reset password test with custom Turbo Stream visit action

### LINKED ISSUE #1473: Flaky test: reset password person spec intermittently fails
## Problem

The system spec `Reset password (person) > When user uses form to change password > fills out the form, submits, and stays logged in` intermittently fails in CI:

```
Failure/Error: expect(page).to have_current_path(root_path)
  expected "/users/sign_in" to equal "/"
```

After submitting the password change form, the user sometimes ends up at `/users/sign_in` instead of `/`, indicating the session was lost during the form submission.

## Root cause

The password change form submits via Turbo Drive (fetch API). After `bypass_sign_in` updates the session and the server sends a 303 redirect, there's a race condition where the browser may not process the `Set-Cookie` header from the redirect response before Turbo's follow-up GET request, causing it to carry a stale session cookie with a mismatched `authenticatable_salt`.

## Approaches tried (not acceptable)

1. **`data-turbo="false"` on the form** (#1474, earlier commits) — disabling Turbo on the form also disables it for all descendant elements, breaking `data-turbo-method` and `data-turbo-confirm` on the "Log out and reset it." link inside the form.

2. **Custom Turbo Stream `visit` action** (#1474, later commits) — a `turbo_stream_visit(url)` helper that responds with a 200 containing a `<turbo-stream action="visit">` tag, separating session update from navigation. This works but adds infrastructure (custom stream action in JS + helper in ApplicationController) that feels heavyweight for this one case.

## Help wanted

We're open to suggestions for a cleaner fix. The solution should:
- Keep Turbo enabled on the form (no `data-turbo="false"`)
- Not require custom Turbo Stream infrastructure
- Reliably ensure the session cookie is set before the post-redirect navigation

maebeale:
Closes #1473

- Fix intermittent CI failure in `reset_password_person_spec.rb` where the user ends up at `/users/sign_in` instead of `/` after changing their password
- Fix confirm dialog failures in both `reset_password_person_spec.rb` and `change_password_flow_spec.rb`

**Password change form (flaky redirect to sign_in):**
- Turbo 8's `fetch()`-based form submission with `redirect: "follow"` has a race condition where the browser may not process the `Set-Cookie` header from a 303 redirect before making the follow-up GET request — causing Devise to invalidate the session due to a stale `authenticatable_salt` (which changes when the password changes via `update_with_password`)
- Added a custom Turbo Stream `visit` action that separates session update from navigation: the server responds with a 200 (not a redirect) containing a `<turbo-stream action="visit">` tag, the browser processes the `Set-Cookie` header, then the custom action triggers `Turbo.visit()` client-side
- `turbo_stream_visit(url)` helper in `ApplicationController` generates the stream tag; `Turbo.StreamActions.visit` in `turbo-events.js` handles it client-side
- This pattern can be reused anywhere a form submission modifies the session and needs to redirect

**"Log out and reset it." link (confirm dialog test failures):**
- Fixed `change_password_flow_spec.rb` to use `accept_confirm` (wrapping the click) instead of `accept_alert` (after the click)
- Added JS-ready waits (`have_css("[data-controller='password-toggle']")`) to both specs to ensure Turbo is initialized before clicking the confirm link

### UI Testing Checklist

- [x] Change password with correct current password → redirects to home, stays logged in
- [x] Change password with wrong current password → re-renders form with errors
- [x] "Log out and reset it." link shows confirm dialog and redirects to password reset page

### Anything else to add?

- Reordered test assertions to check flash content before path — flash presence is a stronger synchronization point than URL check
- The custom Turbo Stream `visit` action is a general solution for the Turbo/session race condition — prefer it over `data-turbo="false"` which disables Turbo for all descendant elements

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -135,6 +135,22 @@ This project uses rubocop-rails-omakase. All code MUST follow these rules:
 - Controller naming: `[name]_controller.js`
 - Keep controllers focused and small
 
+### Turbo and Session-Modifying Responses
+
+When a form submission modifies the session (e.g. `bypass_sign_in` after password change) and then redirects, **do not use `data-turbo="false"`** as the fix. Turbo's `fetch()` with `redirect: "follow"` can race against `Set-Cookie` processing, but disabling Turbo is a blunt workaround that breaks other Turbo features on the page (e.g. `data-turbo-method`, `data-turbo-confirm` on descendant elements).
+
+Instead, use `turbo_stream_visit(url)` (defined in `ApplicationController`) to separate the session update from navigation:
+```ruby
+respond_to do |format|
+  format.html { redirect_to root_path, notice: "Done." }
+  format.turbo_stream do
+    flash[:notice] = "Done."
+    render turbo_stream: turbo_stream_visit(root_path)
+  end
+end
+```
+The server responds with a 200 (Set-Cookie is processed), then the custom Turbo Stream action triggers `Turbo.visit()` client-side.
+
 ### Stimulus Conventions
 
 Follow the [Stimulus Handbook](https://stimulus.hotwired.dev/handbook/introduction) and reference docs. Key rules:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,6 +135,22 @@ This project uses rubocop-rails-omakase. All code MUST follow these rules:
 - Controller naming: `[name]_controller.js`
 - Keep controllers focused and small
 
+### Turbo and Session-Modifying Responses
+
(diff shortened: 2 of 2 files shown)


jmilljr24 (commented):
I need to look into it more but these changes make me pause for a second.

Do we actually have an issue with updating passwords or it just the test being flaky? I don't love the idea of adding special code just to make a test past when the application works.

I can see how this change is intended to work but it seems like we are reinventing the wheel. My gut says that a devise password update with turbo and keeping the user logged in is a pretty normal flow.

Here is the default update action from the devise controller. We of course have our own action with some custom stuff like ahoy but for the most part there is nothing special in devise.

```
  def update
    self.resource = resource_class.to_adapter.get!(send(:"current_#{resource_name}").to_key)
    prev_unconfirmed_email = resource.unconfirmed_email if resource.respond_to?(:unconfirmed_email)

    resource_updated = update_resource(resource, account_update_params)
    yield resource if block_given?
    if resource_updated
      set_flash_message_for_update(resource, prev_unconfirmed_email)
      bypass_sign_in resource, scope: resource_name if sign_in_after_change_password?

      respond_with resource, location: after_update_path_for(resource)
    else
      clean_up_passwords resource
      set_minimum_password_length
      respond_with resource
    end
  end
```

Maybe `respond_with` from the responders gem. But again, I'm wonder if this is just a test issue not an app issue.

PR #1476, opened 2026-04-05
Allow visitors to access taggings and scope all results by user visibility

### LINKED ISSUE #1452: Taggings index - authorization/scoping
We allow visitors to view the `tags` route where a list of active tags are displayed. Clicking a tag routes to the `taggings` index and displays records with that tag.

Currently unauthenticated users cannot visit the `taggings` index to `before_action :authenticate_user!`. However the `tagging_policy` allows anyone to view the index. These contradict each other. The authenticated check happens first so visitors do not reach this page.

Do we want visitors see the `taggings` index? If so, we will need to scope the records as a visitor should only see publicly visible records.

maebeale:
Closes #1452

- Visitors can browse the tags page but clicking a tag routes to the taggings index, which blocks them with `authenticate_user!`
- The `TaggingPolicy#index?` already returns `true` (public access), but the authentication check fires first — these contradict each other
- The tags page showed sectors/categories whose only tagged items were unpublished, leading visitors to empty results
- The taggings search results and combination filters used hardcoded `.published` instead of policy scopes, so they didn't respect per-model visibility rules (e.g., `publicly_visible` for guests vs `published` for authenticated users)

- Added `skip_before_action :authenticate_user!, only: [:index]` to `TaggingsController`
- **TaggingSearchService** now uses `ActionPolicy::Behaviour` with `authorized_scope` per model — results respect each model's `relation_scope` (e.g., guests see `publicly_visible` workshops, authenticated users see `published`, admins see all)
- Added `:taggable` named `relation_scope` to `SectorPolicy` and `CategoryPolicy` — encapsulates all filtering: admins get `has_taggings`, non-admins get `published.has_published_taggings`
- Added `has_published_taggings` scope to `Sector` and `Category` — uses `EXISTS` subqueries across all taggable types
- Controllers pass `Sector.all` / `Category.all` to `authorized_scope` — the policy scope owns all filtering logic
- Moved hardcoded queries out of view partials (`_explore_by_sector`, `_explore_by_combination`) into controller instance variables

### UI Testing Checklist

- [ ] Visit `/tags` as a logged-out visitor, click a tag, confirm `/taggings` loads with results
- [ ] Confirm the "Explore by combination" section on `/taggings` only shows sectors/categories with published items
- [ ] Create an unpublished workshop tagged with a sector — confirm that sector does not appear on `/tags/sectors` for visitors
- [ ] Confirm logged-in users see `published` items (not just `publicly_visible`)
- [ ] Confirm admins see all sectors/categories including those with only unpublished tagged items
- [ ] Confirm `/taggings/matrix` still requires admin authentication

### Anything else to add?

The `has_published_taggings` model scope uses `.published` as a baseline filter for tagged items. Some models like `Event` have more nuanced policy scopes (e.g., registration windows), but for determining which sectors/categories to *display as filter options*, `.published` is sufficient.


app/services/tagging_search_service.rb
```diff
@@ -95,4 +94,10 @@ def self.empty_results(per_page)
       WillPaginate::Collection.create(1, per_page || 9, 0) { |pager| pager.replace([]) }
     end
   end
+
+  private
+
+  def scope(klass)
+    authorized_scope(klass.all)
+  end
```
jmilljr24:
```suggestion
```

In my option, this doesn't really add value. It seems to just mask what the actual method we want to call is.

I don't think the name `scope` conflicts with anything from Action_Policy but I did go to their doc's first (before realizing it was a new private method) to see what is was or how it was different from `authorized_scope`. So minor confusion from a readability standpoint.

PR #1487, opened 2026-04-09
Add pay gem and payments/allocations/refunds

jmilljr24:

EventRegistations (and more models in the future) will require payments. This add the initial framework to handle payments.

- Add Payments, Allocations, Refunds models.
- Add pay gem to add payments via stripe
- Payment types: Cash, Check, ExternalProcessor (anything through pay gem)

Allocations - a payment can be allocated to an event_reg. 
Payments - a payer can be a person or organization and can add funds via payment types.

Payments can be refunded via the Refund model. Allocations can be reverted (original allocation record doesn't change. A new Allocation record allocates the funds and references the original allocation.

I decided to add an `amount_remaining_cents` on a `Payment`. This is the source of truth for was funds are available that have not been allocated or refunded. Checking based of a all allocations and refunds is possible but this prevents race conditions as well as any data that may be out of sync with stripe via webhooks. 

### Anything else to add?

The video is not in any special order. Just an overview of the features. UI styling was a low priority since this is such a big change.
Shown is the payments index filtering, Payment creation (without allocating), Allocating a payment, Partial payment Refund, Allocations on a persons event_reg (some examples show an allocation that has been reverted (grayed out with another line item showing the revert in red) 

https://github.com/user-attachments/assets/029d306e-0b88-449a-a7d4-a162462db9f5


app/controllers/people_controller.rb
```diff
@@ -83,6 +89,36 @@ def workshop_logs
     end
   end
 
+  def checkout
```
jmilljr24, line 92: Temp testing of stripe payment checkout

PR #1488, opened 2026-04-17
Fix system notifications showing Oopsie error

maebeale:
- Fix the system notifications page showing "Oopsie!" instead of the notifications table
- Two issues: (1) the turbo frame rendered data inline AND fired a redundant lazy-load request via `src` — if that request failed, the content was replaced with the error, and (2) links inside the turbo frame (View, Record) were intercepted by Turbo, which tried to load the target page inside the frame — those pages don't have a matching frame, triggering the error

- Adopted the same lazy-load pattern used by other controllers (stories, people, etc.) — skeleton on initial load, data fetched via turbo frame request
- Added `data: { turbo_frame: "_top" }` to View and Record links so they navigate the full page instead of targeting the frame
- Removed orphaned `_notifications_results.html.erb` partial (referenced a non-existent `_notifications_count` partial and was never used)
- Updated request specs to use turbo frame headers and test actual filtering behavior

### UI Testing Checklist
- [ ] Visit `/notifications` as admin — should show skeleton briefly then load notifications table
- [ ] Click "View" on a notification — should navigate to the show page (not show Oopsie)
- [ ] Click a Record link — should navigate to the record page (not show Oopsie)
- [ ] Use search filters (email, record type, email topic, subject) — results should update in the turbo frame
- [ ] Clear filters — should reset to all notifications

### Anything else to add?
- The root cause was that "View" and "Record" links inside the `notifications_results` turbo frame were intercepted by Turbo. Turbo tried to load the linked pages inside the frame, but those pages don't contain a matching `<turbo-frame id="notifications_results">`, so `turbo:frame-missing` fired and the handler replaced the frame content with "Oopsie!"


app/controllers/notifications_controller.rb
```diff
@@ -3,13 +3,18 @@ class NotificationsController < ApplicationController
 
   def index
     authorize!
-    per_page = params[:number_of_items_per_page].presence || 25
-    base_scope = authorized_scope(Notification.includes(:noticeable))
-    filtered = base_scope.search_by_params(params.to_unsafe_h)
```
maebeale, line 8: Adopted the same lazy-load pattern as StoriesController, PeopleController, etc. — only load data for turbo frame requests, render a skeleton for the initial page load.

jmilljr24 (approved):
@maebeale Thanks!!

PR #1492, opened 2026-05-01
Fix reset-password Turbo session race

### LINKED ISSUE #1473: Flaky test: reset password person spec intermittently fails

alceops:
## Summary
- disables Turbo only for the reset-password submit button so the successful password update uses a normal browser navigation after Devise signs the user in
- keeps Turbo enabled for the form subtree instead of using `data-turbo="false"` on the whole form
- adds a system-spec assertion that the submitter, not the form, carries the Turbo opt-out

Fixes #1473

## Verification
- `git diff --check`
- static assertion that the reset-password submitter has `data: { turbo: false }` and the system spec checks submitter-level opt-out

## Verification limits
- Ruby/bundle/node are not installed in this worker, so I could not run the Rails system spec locally.

### DIFF
--- a/app/views/devise/passwords/edit.html.erb
+++ b/app/views/devise/passwords/edit.html.erb
@@ -55,7 +55,7 @@
       </div>
 
       <div class="action-buttons mt-8 flex justify-center gap-3">
-        <%= f.submit "Set password", class: "btn btn-primary" %>
+        <%= f.submit "Set password", class: "btn btn-primary", data: { turbo: false } %>
       </div>
 
     <% end %>
--- a/app/views/devise/passwords/edit.html.erb
+++ b/app/views/devise/passwords/edit.html.erb
@@ -55,7 +55,7 @@
       </div>
 
       <div class="action-buttons mt-8 flex justify-center gap-3">
-        <%= f.submit "Set password", class: "btn btn-primary", data: { turbo: false } %>
+        <%= f.submit "Set password", class: "btn btn-primary" %>
       </div>
 
     <% end %>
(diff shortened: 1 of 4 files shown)


jmilljr24:
Thanks for taking a look at this.

Devise supports Turbo out of the box so this still doesn't fully address the root cause within this application. I'm not opposed to disabling turbo selectively but for something with clear support from Devise, I'm wondering what we are missing in this issue.

PR #1493, opened 2026-05-10
Track responded status on contact-us notifications

maebeale:

- Staff need a way to track which contact-us submissions have been replied to.
- Currently there's no signal in the notifications index — admins have to remember (or open each notification) to know what's outstanding.

**Schema + model**
- Added a `responded:boolean` column to `notifications` (default `false`, non-null). Migration is reversible.
- Predicate `Notification#requires_response?` returns true only for `contact_us_fyi` (the staff-facing FYI). The user-facing `contact_us` auto-confirmation is an automated receipt to the submitter — staff don't "respond" to it — so it renders an em-dash like any other kind.
- New `Notification.responded_status` scope wired into `search_by_params` for the filter dropdown.

**Index** (`/notifications`)
- New "Responded" column. For `contact_us_fyi` rows it renders an inline checkbox; for everything else it shows "—".
- New "Responded" filter dropdown (All / Yes / No / N/A) so staff can quickly surface what's outstanding.
- Fixed a left-padding asymmetry on the index header (inner `<section>` had `pl-8` on top of the outer `p-6`).

**Show** (`/notifications/:id`)
- Same Responded toggle surfaced in the top-right status column alongside Delivered/Failed badges and Resend controls, so admins can mark a contact submission handled without bouncing back to the list.

**Plumbing**
- The checkbox uses the existing `autosave` Stimulus controller — toggling fires a PATCH to `notifications#update`, no submit button, no page reload.
- `NotificationPolicy#update?` is admin-only (matches the index permission).
- Dev seeds add 6 fake `contact_us` + `contact_us_fyi` notification pairs (older half pre-marked responded) plus a few non-contact FYIs so the em-dash rendering is visible too.

**Unrelated bundled change**
- `ai/server` now probes upward from 3000 for a free port when `CONDUCTOR_PORT` is unset, instead of failing with EADDRINUSE.

### UI Testing Checklist

- [ ] Sign in as admin → visit `/notifications`
- [ ] Filter to a `contact_us_fyi` row → checkbox renders unchecked
- [ ] Tick the checkbox → green ring flashes (autosave success), refresh → still checked
- [ ] Untick → state persists after refresh
- [ ] `contact_us` (user-facing) and unrelated rows show "—" in the Responded column
- [ ] Open a `contact_us_fyi` show page → toggle in top-right works and persists
- [ ] "Responded" filter: Yes / No / N/A return the expected slices
- [ ] Sign in as a non-admin → `PATCH /notifications/:id` is denied; show page does not render the checkbox

### Anything else to add?

- Started with a boolean rather than `responded_at` datetime — YAGNI; we can layer a timestamp later if audit needs it.
- Specs added: model (predicate + scope), request (index render, show render, filter, PATCH auth), system (autosave on index + show).

### DIFF
--- a/ai/server
+++ b/ai/server
@@ -1,10 +1,20 @@
 #!/usr/bin/env bash
 # Start development services (web + vite).
-# Uses CONDUCTOR_PORT if set, otherwise defaults to 3000.
-# Derives a unique Vite port (CONDUCTOR_PORT + 1) to avoid collisions.
+# Uses CONDUCTOR_PORT (Conductor's assigned workspace port) if set,
+# otherwise probes for a free port starting at 3000.
+# Derives a unique Vite port (server_port + 1) to avoid collisions.
 set -euo pipefail
 source "$(dirname "$0")/.ruby-env"
+
 if [ -n "${CONDUCTOR_PORT:-}" ]; then
-  export VITE_RUBY_PORT=$((CONDUCTOR_PORT + 1))
+  server_port="$CONDUCTOR_PORT"
+else
+  server_port=3000
+  while lsof -iTCP:"$server_port" -sTCP:LISTEN -n -P >/dev/null 2>&1; do
+    server_port=$((server_port + 1))
+  done
+  echo "ai/server: CONDUCTOR_PORT unset, picked free port $server_port"
 fi
-bin/dev -p "${CONDUCTOR_PORT:-3000}"
+
+export VITE_RUBY_PORT=$((server_port + 1))
+bin/dev -p "$server_port"
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -114,6 +114,10 @@ class Notification < ApplicationRecord
     !kind.in?(DEVISE_KINDS)
   end
 
+  def contact_us?
+    kind.in?(%w[contact_us contact_us_fyi])
+  end
+
   def delivered?
     delivered_at.present?
   end
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -94,6 +94,14 @@ class Notification < ApplicationRecord
   scope :record_type, ->(record_type) { where(noticeable_type: record_type.to_s.camelize.titleize.gsub(" ", "")) }
   scope :subject_line, ->(subject) { where("notifications.email_subject LIKE ?", "%#{subject}%") }
   scope :email_topic, ->(topic) { where("notifications.email_subject LIKE ?", "%#{topic}%") }
+  scope :responded_status, ->(status) {
+    case status.to_s
+    when "yes" then where(kind: "contact_us_fyi", responded: true)
+    when "no"  then where(kind: "contact_us_fyi", responded: false)
+    when "na"  then where.not(kind: "contact_us_fyi")
+    else all
+    end
+  }
 
   def self.email_topic_phrase(label)
     EMAIL_TOPICS.assoc(label)&.last
@@ -107,6 +115,7 @@ class Notification < ApplicationRecord
     topic_phrase = email_topic_phrase(params[:email_topic]) if params[:email_topic].present?
     stories = stories.email_topic(topic_phrase) if topic_phrase.present?
     stories = stories.record_type(params[:record_type]) if params[:record_type].present?
+    stories = stories.responded_status(params[:responded_status]) if params[:responded_status].present?
     stories
   end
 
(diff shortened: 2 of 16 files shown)


jmilljr24:
Great idea!!

app/models/notification.rb
```diff
@@ -107,13 +115,18 @@ def self.search_by_params(params)
...
+    stories = stories.responded_status(params[:responded_status]) if params[:responded_status].present?
     stories
   end
 
   def resendable?
     !kind.in?(DEVISE_KINDS)
   end
 
+  def requires_response?
+    kind == "contact_us_fyi"
```
maebeale, line 127: Narrowed from `contact_us` + `contact_us_fyi` to just `contact_us_fyi`. The `contact_us` notification is the auto-confirmation we send back to the submitter ("We received your message…") — it's a system receipt, not something staff needs to respond to. Only the FYI to staff (`REPLY_TO_EMAIL`) is an outstanding action item.

app/models/notification.rb
```diff
@@ -94,6 +94,14 @@ class Notification < ApplicationRecord
   scope :record_type, ->(record_type) { where(noticeable_type: record_type.to_s.camelize.titleize.gsub(" ", "")) }
   scope :subject_line, ->(subject) { where("notifications.email_subject LIKE ?", "%#{subject}%") }
   scope :email_topic, ->(topic) { where("notifications.email_subject LIKE ?", "%#{topic}%") }
+  scope :responded_status, ->(status) {
+    case status.to_s
+    when "yes" then where(kind: "contact_us_fyi", responded: true)
+    when "no"  then where(kind: "contact_us_fyi", responded: false)
+    when "na"  then where.not(kind: "contact_us_fyi")
```
maebeale, line 101: `na` is "not applicable" — defined as `kind != contact_us_fyi` rather than enumerating kinds, so any future non-respondable notification kind is automatically included without a model change.

app/views/notifications/_index.html.erb
```diff
@@ -51,6 +52,25 @@
...
       </td>
 
+      <!-- Responded -->
+      <td class="px-4 py-3 text-center"
+          data-controller="autosave"
+          data-autosave-url-value="<%= notification_path(notification) %>"
+          data-autosave-id-value="<%= notification.id %>"
+          data-action="change->autosave#save">
+        <% if notification.requires_response? %>
+          <input type="hidden" name="notification[responded]" value="0">
```
maebeale, line 62: Hidden `value="0"` is intentional. Unchecked checkboxes don't submit anything, so without this the autosave PATCH would carry no `responded` param and `update!` would leave the column unchanged — unchecking would silently no-op. The hidden field ensures "unchecked" sends `responded=0`.

spec/system/notification_responded_toggle_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  describe "on the index page" do
+    it "auto-saves when an admin checks the box" do
+      visit notifications_path
+
+      expect(responded_checkbox).not_to be_checked
+
+      responded_checkbox.check
+
+      expect(page).to have_css("input[name='notification[responded]'].ring-green-300", wait: 5)
```
maebeale, line 21: Waiting on `.ring-green-300` (the autosave controller's success flash) rather than `:checked` — Capybara updates the DOM checkbox state synchronously on `.check`/`.uncheck`, so `:checked` passes before the PATCH has resolved and the next `fyi.reload` line races the in-flight request. Asserting on the success ring waits for the server response.

ai/server
```diff
@@ -1,10 +1,20 @@
...
-# Derives a unique Vite port (CONDUCTOR_PORT + 1) to avoid collisions.
+# Uses CONDUCTOR_PORT (Conductor's assigned workspace port) if set,
+# otherwise probes for a free port starting at 3000.
+# Derives a unique Vite port (server_port + 1) to avoid collisions.
 set -euo pipefail
 source "$(dirname "$0")/.ruby-env"
+
 if [ -n "${CONDUCTOR_PORT:-}" ]; then
-  export VITE_RUBY_PORT=$((CONDUCTOR_PORT + 1))
+  server_port="$CONDUCTOR_PORT"
```
maebeale, line 10: Kept `CONDUCTOR_PORT` read-only (the value Conductor injects) and introduced `server_port` as the actually-used port. Reassigning `CONDUCTOR_PORT` itself would make `CONDUCTOR_PORT=3001 (we picked this ourselves)` look like it came from Conductor, which is misleading for anything downstream that reads the var.

PR #1529, opened 2026-05-25
Make Ruby resolution work without manual mise activation

maebeale:

- Scripts with `#!/usr/bin/env ruby` (notably `bin/setup`) silently picked up the wrong Ruby — or failed entirely — when the caller's shell hadn't activated mise.
- Conductor workspaces hit this because `bin/conductor-setup` runs under `/bin/sh`, which does not source `~/.zshenv`.
- Separately, the Ruby version was declared in two places (`mise.toml` and `.ruby-version`) which can drift.

**Make Ruby resolution work without manual mise activation:**

- **`CONTRIBUTING.md`** — added shims-in-zshenv as a required step under the "Install mise" section of the local setup path, with a brief explanation of *why* (covers non-interactive shells, IDE runners, cron, `bin/setup`).
- **`CLAUDE.md`** — replaced the stale "run `eval mise activate` before any Ruby command" guidance with the same shims setup, so agents working in the repo follow the current install pattern.
- **`bin/conductor-setup`** — activates mise explicitly near the top (with a `/opt/homebrew/bin/mise` fallback for cases where Conductor's inherited PATH doesn't include it). This script runs under `#!/bin/sh`, so the `~/.zshenv` approach doesn't apply here.
- `bin/setup` itself stays a plain Ruby script — no bash wrapper, no shebang tricks. Once shims are on PATH (or `bin/conductor-setup` has activated mise), `#!/usr/bin/env ruby` just works.

**Make `.ruby-version` the single source of truth:**

- **`mise.toml`** — added `[settings] idiomatic_version_file_enable_tools = ["ruby"]` and removed the duplicate `ruby = "4.0.1"` from `[tools]`. External tooling (CI, deploys, IDEs) already reads `.ruby-version` by convention; this makes mise read it too. Project-scoped so every contributor's mise picks it up regardless of their global config.

### Anything else to add?

- Branch name (`setup-bash-wrapper`) is a leftover from an earlier approach that I abandoned in favor of the shims path. Not renaming since the PR exists.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ Either option below will get you a working development environment. Pick whichev
 
 1. Install mise
    - Follow the [mise installation guide](https://mise.jdx.dev/getting-started.html#installing-mise-cli) for your operating system
-   - mise will automatically install and manage the Ruby version specified in [`mise.toml`](mise.toml) with `mise i`
+   - mise will automatically install and manage the Ruby version specified in [`.ruby-version`](.ruby-version) with `mise i`
    - Verify that mise is working by running `mise --version`
    - **Add mise shims to your PATH** so the project's Ruby is resolved in every shell — including non-interactive contexts like `bin/setup`, IDE task runners, and cron. Add this to `~/.zshenv` (or `~/.bashrc` for bash) and restart your shell:
      ```bash
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,14 +48,13 @@ Either option below will get you a working development environment. Pick whichev
    - Follow the [mise installation guide](https://mise.jdx.dev/getting-started.html#installing-mise-cli) for your operating system
    - mise will automatically install and manage the Ruby version specified in [`.ruby-version`](.ruby-version) with `mise i`
    - Verify that mise is working by running `mise --version`
-   - **Add mise shims to your PATH** by adding this line to your shell's startup file and then restarting your shell:
-     ```bash
-     export PATH="$HOME/.local/share/mise/shims:$PATH"
-     ```
-     - **zsh users:** add to `~/.zshenv` (not `~/.zshrc`) — `~/.zshenv` is loaded by *every* shell invocation, including non-interactive ones.
-     - **bash users:** add to `~/.bash_profile` (login shells) and `~/.bashrc` (interactive non-login shells).
+   - **Add mise shims to your PATH** so the project's Ruby is available to scripts with `#!/usr/bin/env ruby` (like `bin/setup`), IDE task runners, and cron jobs that don't source your interactive shell config. The export goes in your shell's startup file — pick the one that matches your shell, then restart your shell:
 
-     This makes the project's Ruby available to scripts with `#!/usr/bin/env ruby` (like `bin/setup`), as well as IDE task runners and cron jobs that don't source your interactive shell config.
+     | Shell | File | Line to add |
+     |---|---|---|
+     | zsh   | `~/.zshenv` (loaded by *every* shell invocation, not `~/.zshrc` which is interactive-only) | `export PATH="$HOME/.local/share/mise/shims:$PATH"` |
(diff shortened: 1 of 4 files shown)


maebeale:
@jmilljr24 will mise still work with your setup if we remove ruby declaration from mise.toml in favor of ruby-version ruby?

jmilljr24:
> @jmilljr24 will mise still work with your setup if we remove ruby declaration from mise.toml in favor of ruby-version ruby?

This is from the mise docs. So it should be fine. Maybe needs a note in the contributing guide to enable the setting during initial setup.
```
.ruby-version and Gemfile support

mise uses a mise.toml or .tool-versions file for auto-switching between software versions. However, it can also read ruby-specific version files .ruby-version or Gemfile (if it specifies a ruby version).

Create a .ruby-version file for the current version of ruby:

ruby -v > .ruby-version

Enable idiomatic version file reading for ruby:

mise settings add idiomatic_version_file_enable_tools ruby

See [idiomatic version files](https://mise.jdx.dev/configuration.html#idiomatic-version-files) for more information.
```

CONTRIBUTING.md
```diff
@@ -48,6 +48,11 @@ Either option below will get you a working development environment. Pick whichev
    - Follow the [mise installation guide](https://mise.jdx.dev/getting-started.html#installing-mise-cli) for your operating system
    - mise will automatically install and manage the Ruby version specified in [`.ruby-version`](.ruby-version) with `mise i`
    - Verify that mise is working by running `mise --version`
+   - **Add mise shims to your PATH** so the project's Ruby is resolved in every shell — including non-interactive contexts like `bin/setup`, IDE task runners, and cron. Add this to `~/.zshenv` (or `~/.bashrc` for bash) and restart your shell:
+     ```bash
+     export PATH="$HOME/.local/share/mise/shims:$PATH"
+     ```
+     Without this, scripts with `#!/usr/bin/env ruby` may pick up the wrong Ruby (or fail to find one).
```
jmilljr24: Is this specific to your computer?

maebeale: to the user's computer, yes. (whether they use zsh or bashrc)

jmilljr24: I get that but my question is why is this new? Like is the issue specific to your computer?  I don't need to add a mise shim and I don't recall that being an issue when most people were setting up at r4g last year.

It's my understand that the point of mise is to handle the versions for us. Running a script just need to be done from the path of the project.

maebeale:
@jmilljr24 good call — I ended up enabling that setting at the project level in `mise.toml` (`[settings] idiomatic_version_file_enable_tools = ["ruby"]`), so contributors don't need to set it globally. Just pushed a follow-up that also clarifies the CONTRIBUTING wording on which shell-startup file the PATH export should go in (`~/.zshenv` vs `~/.bash_profile`, and why).

jmilljr24:
> @jmilljr24 good call — I ended up enabling that setting at the project level in `mise.toml` (`[settings] idiomatic_version_file_enable_tools = ["ruby"]`), so contributors don't need to set it globally. Just pushed a follow-up that also clarifies the CONTRIBUTING wording on which shell-startup file the PATH export should go in (`~/.zshenv` vs `~/.bash_profile`, and why).

Awesome on the toml setting!

I don't think the comments in contributing setup instructions should be there if it's not a generic requirement. There is the potential to cause some system wide issues for a new contributor when it's not strictly required.

It seems you needed to export the path due to how conductor is working on your system? If so, maybe a comment on the AI section

CONTRIBUTING.md
maebeale: ah. yes, the setup script is failing in conductor

maebeale:
@jmilljr24 fair point — pushed `fbcb501d7` that drops the shims block from CONTRIBUTING.md and rewords CLAUDE.md to say standard mise activation is sufficient. You're right that the only place this was actually needed was `bin/conductor-setup` (sh doesn't source the shell rc), and that's handled in the script itself. The Ruby version consolidation (`.ruby-version` + `[settings]` in `mise.toml`) and the `bin/conductor-setup` activation stay.

PR #1531, opened 2026-05-26
Filter community_news, event_registrations, stories, story_ideas by organization_id

maebeale:
- Lets the Organization edit page (and any future caller) link to the indexes for these four resources pre-filtered to a single org.
- Adds an Organization picker to each index's filter bar so the same filtering is reachable from the index pages directly.

- Added an \`organization_id\` clause to each model's \`search_by_params\`.
  - \`EventRegistration\` joins through \`event_registration_organizations\`.
  - \`StoryIdea\`, \`Story\`, \`CommunityNews\` use the direct \`organization_id\` column.
- In each index action, loaded \`@organizations\` from the affiliated authorized scope and added an Organization \`select\` to the corresponding \`_search_boxes.html.erb\`, defaulting to "All organizations".
- Added model specs covering the new filter on each model.
- Added a request spec on each index confirming \`?organization_id=\` returns only matching records and excludes others.

### UI Testing Checklist
- [ ] Visit \`/community_news\`, \`/event_registrations\`, \`/stories\`, \`/story_ideas\` — confirm the new "Organization" dropdown appears and defaults to "All organizations".
- [ ] Pick an org from the dropdown and confirm the listing narrows to records for that org only.
- [ ] Visit each index with \`?organization_id=<id>\` directly in the URL and confirm filtering works.

### Anything else to add?
- 217 specs covered by this change pass (4 model specs + 4 request specs + everything in the affected files).

### DIFF
--- a/app/controllers/community_news_controller.rb
+++ b/app/controllers/community_news_controller.rb
@@ -25,6 +25,7 @@ class CommunityNewsController < ApplicationController
 
       render :index_lazy
     else
+      @organizations = authorized_scope(Organization.all, as: :affiliated).order(:name)
       render :index
     end
   end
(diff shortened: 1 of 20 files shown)


maebeale:
Superseded by #1532 (filtering) and #1533 (Associated records consumer).

PR #1535, opened 2026-05-29
add google snippet to event form

jmilljr24:

Stakeholders requested the ability to add google analytic scripts to events show page.

Add new form fields.

### Anything else to add?

This allows users (admin only) to paste scripts. This is a xss security issue. However, this feature is only enabled for admin so the risk is low. The alternative is to hard code the script and the admin add their unique ID. The downside is if the google code ever changes this code base will need to be updated. Also this adds difficulty to admin as they have to understand how to find the unique code within the "copy this snippet" instructions from google.

screenshot-2026-05-29_16-07-55


app/controllers/events_controller.rb
```diff
@@ -266,38 +266,6 @@ def set_event
...
-                                  :registration_close_date,
-                                  :published,
-                                  :publicly_visible,
-                                  :publicly_featured,
-                                  category_ids: [],
-                                  sector_ids: [],
-                                  primary_asset_attributes: [ :id, :file, :_destroy ],
-                                  gallery_assets_attributes: [ :id, :file, :_destroy ]
-                                  )
+    authorized(params.require(:event))
```
jmilljr24, line 269: Use action policy scope on params. Only admin should be able to pass the scripts.

app/views/events/_form.html.erb
```diff
@@ -234,18 +234,7 @@
...
-          fill="currentColor"
-          viewBox="0 0 20 20"
-          xmlns="http://www.w3.org/2000/svg">
-          <path
-            fill-rule="evenodd"
-            d="M5.293 7.293a1 1 0 011.414 0L10 10.586l3.293-3.293a1 1 0 111.414 1.414l-4 4a1 1 0 01-1.414 0l-4-4a1 1 0 010-1.414z"
-            clip-rule="evenodd"
-          ></path>
-        </svg>
+        <i id="images_arrow" class="fa-solid fa-chevron-down w-4 h-4 transition-transform duration-300"></i>
```
jmilljr24, line 237: Use font awesome chevron consistently instead of svg's

app/views/events/_form.html.erb
```diff
@@ -409,6 +387,34 @@
         </div>
       </div>
     <% end %>
+    <% if allowed_to?(:google_analytics?, @event) %>
```
jmilljr24, line 390: Admin only. Other form options my be available to non-admin to create/edit

PR #1539, opened 2026-06-05
Notify story idea author when their idea is promoted to a published story

### LINKED ISSUE #1513: Notify story author by email when their story idea is promoted to a story
## Question / Proposal
When an admin promotes a story idea into a published story, does the original submitter receive an email notification?

If not, we should send one — they should know their idea has been picked up and turned into a story.

maebeale:
Closes #1513

- When an admin promotes a story idea into a **published** story, the original submitter now receives an email letting them know their idea was picked up and turned into a story.
- Previously submitters got no signal at all once they submitted an idea — this closes the loop so contributors know their work went live.

- Added `StoryServices::NotifyIdeaAuthorOfPublication`, a small service that is safe to call on every story save. It no-ops unless the story came from an idea, is published, and the author hasn't already been notified for that story (duplicate guard).
- Wired the service into `StoriesController#create` and `#update`, so both "create as published" and "create as draft → publish later" trigger exactly one email.
- Added a new `story_published` notification kind, a `NotificationMailer#story_published` action, HTML + text email templates, the `NotificationMailerJob` mapping, an admin email-topic filter entry, and a mailer preview.
- Story now `has_many :notifications` (mirrors `StoryIdea`) so notification records are cleaned up with the story.

### Testing
- `spec/services/story_services/notify_idea_author_of_publication_spec.rb` — service guards (promoted+published notifies once; no idea / unpublished / repeat = no email)
- `spec/mailers/notification_mailer_spec.rb` — `#story_published` renders, addresses the author, announces the story
- `spec/requests/stories_spec.rb` — promoting an idea into a published story via the controller notifies the author; leaving it unpublished does not

### UI Testing Checklist
- [ ] No user-facing UI change; verify the email renders via the mailer preview at `/rails/mailers/notification_mailer/story_published`

### Anything else to add?
- Notification only fires when the story is **published**, matching the issue's "published story" wording. A draft promoted from an idea sends nothing until it is published.

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -74,6 +74,7 @@ class StoriesController < ApplicationController
         elsif params.dig(:library_asset, :new_assets).present?
           update_asset_owner(@story)
         end
+        StoryServices::NotifyIdeaAuthorOfPublication.call(@story)
         success = true
       end
     rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotSaved, ActiveRecord::RecordNotUnique => e
@@ -100,6 +101,7 @@ class StoriesController < ApplicationController
         if params[:promote_idea_assets] == "true"
           @story.attach_assets_from_idea!
         end
+        StoryServices::NotifyIdeaAuthorOfPublication.call(@story)
         success = true
       end
     rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotSaved, ActiveRecord::RecordNotUnique => e
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -15,6 +15,7 @@ class Story < ApplicationRecord
   has_many :bookmarks, as: :bookmarkable, dependent: :destroy
   has_many :categorizable_items, dependent: :destroy, inverse_of: :categorizable, as: :categorizable
   has_many :sectorable_items, dependent: :destroy, inverse_of: :sectorable, as: :sectorable
+  has_many :notifications, as: :noticeable, dependent: :destroy
 
   # Asset associations
   has_one :primary_asset, -> { where(type: "PrimaryAsset") },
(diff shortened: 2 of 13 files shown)


app/services/story_services/notify_idea_author_of_publication.rb
```diff
@@ -0,0 +1,40 @@
...
+        notification_type: 0
+      )
+    end
+
+    private
+
+    attr_reader :story
+
+    def already_notified?
+      Notification.exists?(noticeable: story, kind: "story_published")
```
maebeale, line 37: Duplicate guard keyed on the story (not the idea) is what makes this service safe to call on every `create`/`update`. Re-publishing or editing an already-announced story won't re-email the author.

app/services/story_services/notify_idea_author_of_publication.rb
```diff
@@ -0,0 +1,40 @@
...
+      new(story).call
+    end
+
+    def initialize(story)
+      @story = story
+    end
+
+    def call
+      return unless story.story_idea
+      return unless story.published?
```
maebeale, line 17: Gated on `published?` to match the issue's "published story" wording — a draft promoted from an idea stays silent until it's actually published (then the update path fires the email).

app/models/story.rb
```diff
@@ -15,6 +15,7 @@ class Story < ApplicationRecord
   has_many :bookmarks, as: :bookmarkable, dependent: :destroy
   has_many :categorizable_items, dependent: :destroy, inverse_of: :categorizable, as: :categorizable
   has_many :sectorable_items, dependent: :destroy, inverse_of: :sectorable, as: :sectorable
+  has_many :notifications, as: :noticeable, dependent: :destroy
```
maebeale, line 18: Mirrors StoryIdea's `has_many :notifications` so a deleted story doesn't orphan its notification records.

app/controllers/stories_controller.rb
```diff
@@ -74,6 +74,7 @@ def create
         elsif params.dig(:library_asset, :new_assets).present?
           update_asset_owner(@story)
         end
+        StoryServices::NotifyIdeaAuthorOfPublication.call(@story)
         success = true
```
maebeale, line 78: Called inside the existing transaction in both create and update; the duplicate guard in the service ensures the author is emailed at most once across both paths.

PR #1540, opened 2026-06-05
Rebalance published story image sizing (#1519)

### LINKED ISSUE #1519: Published story: primary image too large, secondary image too small
## Problem
On published story pages, the primary image renders too large and the secondary image renders too small. The relative sizing feels off.

## Proposal
Rebalance the image sizing on published story pages — shrink the primary, enlarge the secondary, or both.

maebeale:
Closes #1519

- On published story pages the **primary image rendered too large** (full card width) while the **secondary/gallery images rendered too small** (fixed 128px), so the relative sizing felt off.
- This rebalances the two so they read as an intentional pair, per the stakeholder request.

- **Shrink + center the primary:** the hero image is capped to `max-w-2xl` and centered with `mx-auto` instead of spanning the full card width.
- **Enlarge the secondary:** gallery thumbnails go from `w-32 h-32` (128px) to `w-48 h-48` (192px); still within the 256px `:thumbnail` variant, so no upscaling/blur.
- **Scoped to stories:** sizing is threaded through the shared `assets/_display_assets` and `assets/_display_gallery_media` partials via backward-compatible params (`hero_extra_classes`, `gallery_width`, `gallery_height`). All other consumers (workshops, events, ideas, etc.) keep their existing defaults.
- Added `48` entries to `DisplayImagePresenter`'s width/height maps (the `w-48`/`h-48` utilities are already emitted by other views, so Tailwind generates them).

### Tests
- Followed red→green: added failing specs first, then the fix.
- `DisplayImagePresenter` spec: `48` maps to `w-48`/`h-48`.
- `stories/show` view spec: primary image is constrained/centered; gallery images render at the enlarged size.

### UI Testing Checklist
- [ ] Open a published story with a primary image and at least one gallery image — confirm the primary is noticeably smaller/centered and the gallery thumbnails are larger.
- [ ] Confirm a story with only a primary image (no gallery) still looks correct.
- [ ] Spot-check a workshop/event show page to confirm their image sizing is unchanged.

### Anything else to add?
- Screenshots to follow.

### DIFF
--- a/app/services/display_image_presenter.rb
+++ b/app/services/display_image_presenter.rb
@@ -2,11 +2,11 @@
 
 class DisplayImagePresenter
   WIDTH_CLASSES = {
-    "18" => "w-18", "24" => "w-24", "32" => "w-32", "full" => "w-full"
+    "18" => "w-18", "24" => "w-24", "32" => "w-32", "48" => "w-48", "full" => "w-full"
   }.freeze
 
   HEIGHT_CLASSES = {
-    "14" => "h-14", "18" => "h-18", "24" => "h-24", "32" => "h-32", "full" => "h-full"
+    "14" => "h-14", "18" => "h-18", "24" => "h-24", "32" => "h-32", "48" => "h-48", "full" => "h-full"
   }.freeze
 
   PDF_PREVIEW_SIZES = { hero: [ 1200, 1200 ], gallery: [ 300, 300 ], index: [ 300, 300 ] }.freeze
--- a/app/views/stories/show.html.erb
+++ b/app/views/stories/show.html.erb
@@ -60,7 +60,8 @@
           <div id="upload-progress-bar" class="h-full w-0 bg-blue-500 transition-[width] duration-200 ease-out"></div>
         </div>
         <div id="hero-image">
-          <%= render "assets/display_assets", resource: @story, link: true %>
+          <%= render "assets/display_assets", resource: @story, link: true,
+                     hero_extra_classes: "max-w-2xl mx-auto", gallery_width: "48", gallery_height: "48" %>
         </div>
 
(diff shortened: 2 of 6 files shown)


app/services/display_image_presenter.rb
```diff
@@ -2,11 +2,11 @@
 
 class DisplayImagePresenter
   WIDTH_CLASSES = {
-    "18" => "w-18", "24" => "w-24", "32" => "w-32", "full" => "w-full"
+    "18" => "w-18", "24" => "w-24", "32" => "w-32", "48" => "w-48", "full" => "w-full"
   }.freeze
```
maebeale, line 6: Adding `48` here is safe for Tailwind v4: this file isn't in the `@source` globs (it only scans views/js/helpers), but `w-48`/`h-48` are already emitted by other views (e.g. `active_storage/blobs/_blob_thumbnail`), so the utilities get generated. If a future size key isn't used in a scanned file, it would silently fall back to `w-auto`.

app/views/stories/show.html.erb
```diff
@@ -60,7 +60,8 @@
           <div id="upload-progress-bar" class="h-full w-0 bg-blue-500 transition-[width] duration-200 ease-out"></div>
         </div>
         <div id="hero-image">
-          <%= render "assets/display_assets", resource: @story, link: true %>
+          <%= render "assets/display_assets", resource: @story, link: true,
+                     hero_extra_classes: "max-w-2xl mx-auto", gallery_width: "48", gallery_height: "48" %>
```
maebeale, line 64: Sizing is scoped to the story show here rather than changing the shared presenter defaults, so workshops/events/ideas that also use `display_assets` keep their current look. The params default to the old values in the partials, so this is the only page that changes.

PR #1541, opened 2026-06-05
Add copyable shareable links for individual stories

### LINKED ISSUE #1523: Story sharing: provide shareable story links (not just a number)
## Problem
The current "share" surface shows a number/count but doesn't give us actual shareable links to individual stories.

## Proposal
Provide direct shareable URLs for each story so they can be copied and posted to social media, email, etc.

maebeale:
Closes #1523

- The story sharing surface previously showed only social share icons and a story count — there was no way to grab the **direct URL** for an individual story to paste into an email, message, or social post.
- Stakeholders requested copy-paste shareable links per story so stories can be circulated easily.

- Added a reusable `stories/_share_links` partial that renders:
  - the story's direct public URL in a read-only field with a **Copy link** button
  - email + Facebook / X / LinkedIn / Pinterest share targets
- Backed the copy button with a small `copy-link` Stimulus controller (clipboard write with a graceful `execCommand` fallback and a transient "Copied!" confirmation).
- Wired the partial into the public **story share** page (`story_shares/show`), which previously had no share affordance at all.
- Replaced the duplicated inline social markup on the **story show** page with the shared partial, so both surfaces stay in sync and now use the canonical public `story_share_url`.
- Updated `AGENTS.md` (Stimulus controller list + count).

### UI Testing Checklist
- [ ] On a public story share page, the direct URL is shown and **Copy link** copies it to the clipboard (button shows "Copied!").
- [ ] Email / Facebook / X / LinkedIn / Pinterest share links open with the correct pre-filled URL and title.
- [ ] Story show page renders the same share section.

### Anything else to add?
- Added a request spec asserting the share page renders the direct story URL and copy affordance.
- Note: this workspace was missing frontend deps; ran `npm ci` so Vite-built pages render in the test env.

### DIFF
--- a/app/frontend/javascript/controllers/copy_link_controller.js
+++ b/app/frontend/javascript/controllers/copy_link_controller.js
@@ -1,0 +1,38 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="copy-link"
+// Copies a direct shareable URL to the clipboard and gives the user feedback.
+export default class extends Controller {
+  static targets = ["input", "label"]
+  static values = { confirmedLabel: { type: String, default: "Copied!" } }
+
+  async copy() {
+    const url = this.inputTarget.value
+
+    try {
+      await navigator.clipboard.writeText(url)
+    } catch {
+      // Fallback for browsers without the async clipboard API
+      this.inputTarget.select()
+      document.execCommand("copy")
+    }
+
+    this.confirm()
+  }
+
+  confirm() {
+    if (!this.hasLabelTarget) return
+
+    const original = this.labelTarget.textContent
+    this.labelTarget.textContent = this.confirmedLabelValue
+
+    if (this.resetTimeout) clearTimeout(this.resetTimeout)
+    this.resetTimeout = setTimeout(() => {
+      this.labelTarget.textContent = original
+    }, 2000)
+  }
+
+  disconnect() {
+    if (this.resetTimeout) clearTimeout(this.resetTimeout)
+  }
+}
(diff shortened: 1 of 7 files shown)


app/frontend/javascript/controllers/copy_link_controller.js
```diff
@@ -0,0 +1,38 @@
...
+  static values = { confirmedLabel: { type: String, default: "Copied!" } }
+
+  async copy() {
+    const url = this.inputTarget.value
+
+    try {
+      await navigator.clipboard.writeText(url)
+    } catch {
+      // Fallback for browsers without the async clipboard API
+      this.inputTarget.select()
```
maebeale, line 16: Fallback path for browsers/contexts where the async Clipboard API is unavailable (e.g. non-HTTPS or older Safari). `execCommand("copy")` is deprecated but still the most reliable cross-browser fallback.

app/views/stories/show.html.erb
```diff
@@ -1,5 +1,4 @@
-<% story_url   = story_url(@story) %>
-<% story_title = ERB::Util.url_encode(@story.title) %>
+<% share_url = story_share_url(@story) %>
```
maebeale, line 1: Switched the copyable link from `story_url` (admin `/stories/:id`, auth-gated) to the canonical public `story_share_url`, so a copied link is actually shareable with non-logged-in recipients.

app/views/stories/_share_links.html.erb
```diff
@@ -0,0 +1,50 @@
+<%# Direct shareable link + social/email share targets for a single story.
+    Locals: story (decorated or plain), url (canonical public share URL string). %>
```
maebeale, line 2: Partial takes `url` as a local rather than deriving it, so each caller passes the appropriate canonical URL for its context while keeping the markup shared.

PR #1542, opened 2026-06-05
Stop submitting registration_form_id as an unpermitted event param

### LINKED ISSUE #1536: Event edit form submits registration_form_id as an unpermitted param
Why is this submitted and do we need it?

maebeale:
Closes #1536

- The event form posted the registration-form selector as `event[registration_form_id]`, but `registration_form_id` is **not** an Event attribute.
- The registration form is persisted through the `event_forms` join (see `EventsController#assign_event_forms`), not via mass assignment.
- Because the field was nested under `event[...]`, strong params rejected it as an **unpermitted parameter** and logged a warning on every event create/update — which is exactly what the issue flagged ("Why is this submitted and do we need it?").

We do need the value (it drives which registration form gets linked) — it just shouldn't live inside the `event[...]` namespace.

- Renamed the select from `event[registration_form_id]` to a top-level `registration_form_id` in `app/views/events/_form.html.erb`.
- Updated `assign_event_forms` to read `params[:registration_form_id]` instead of `params.dig(:event, :registration_form_id)`.
- No behavior change for users: selecting a form still links it, and selecting "No registration form" still removes it.

### Anything else to add?

- Added request specs covering create/update/blank that assert the form is linked **and** that `registration_form_id` is no longer reported via the `unpermitted_parameters.action_controller` notification.
- Note: a few unrelated event view-rendering specs (`GET /index`, `/new`, `/show`) currently return 500 in this local environment even on a clean `main` checkout (asset/CSP rendering), so they are not affected by this change.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -219,7 +219,7 @@ class EventsController < ApplicationController
   end
 
   def assign_event_forms(event)
-    form_id = params.dig(:event, :registration_form_id)
+    form_id = params[:registration_form_id]
     return unless form_id
 
     if form_id.blank?
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -368,7 +368,7 @@
                                    else
                                      @registration_forms.find { |rf| rf.name == ShortEventRegistrationFormBuilder::FORM_NAME }&.id
                                    end %>
-              <select name="event[registration_form_id]" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
+              <select name="registration_form_id" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
                 <option value="">No registration form</option>
                 <% @registration_forms.each do |reg_form| %>
                   <% label = case reg_form.name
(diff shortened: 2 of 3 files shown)


app/controllers/events_controller.rb
```diff
@@ -219,7 +219,7 @@ def event_registration_csv_row(registration, cost_required)
   end
 
   def assign_event_forms(event)
-    form_id = params.dig(:event, :registration_form_id)
+    form_id = params[:registration_form_id]
```
maebeale, line 222: Reading the top-level `params[:registration_form_id]` instead of `params.dig(:event, :registration_form_id)`. This value isn't an Event column — it's resolved into the `event_forms` join below — so keeping it out of the `event[...]` namespace is what stops strong params from logging it as unpermitted.

app/views/events/_form.html.erb
```diff
@@ -368,7 +368,7 @@
                                    else
                                      @registration_forms.find { |rf| rf.name == ShortEventRegistrationFormBuilder::FORM_NAME }&.id
                                    end %>
-              <select name="event[registration_form_id]" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
+              <select name="registration_form_id" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
```
maebeale, line 371: Field renamed from `event[registration_form_id]` to `registration_form_id`. The selector drives `assign_event_forms`, not mass assignment, so it belongs at the top level of the params rather than nested under the event resource.

PR #1543, opened 2026-06-05
Add searchable image gallery with editable titles (#1512)

### LINKED ISSUE #1512: Gallery: search and editable titles to use gallery as image database
## Problem
There is no way to search the gallery for images. The team wants to use the gallery as an image database, but without search and meaningful titles (org, workshop, etc.) images are hard to find.

## Proposal
- Add a search box to the gallery
- Make image titles easily editable so they can be tagged with organization, workshop, etc.
- Search should match on those titles/metadata

maebeale:
Closes #1512

- The team wants to use the portal's gallery images as an **image database**, but there was no way to browse or find them.
- There was no gallery page, no search, and image titles could not be edited — so images couldn't be tagged with their organization, workshop, etc.
- This adds an admin-only **Image gallery** so images can be tagged with meaningful titles and found again later.

- **New `GalleryAssetsController`** (`index`, `update`), reachable from the **Curriculum → Image gallery** nav item (admin-only).
- **`index`** lists every gallery asset whose attached file is an image, newest first, paginated (24/page).
- **Search box** debounces into a Turbo frame (reusing the existing `collection` Stimulus controller) and matches on the editable **title**, the **owner type** (Workshop, Story, …), and the original **filename**, via new `GalleryAsset.images` / `GalleryAsset.search_metadata` scopes.
- **Editable titles**: each image card has an inline title field that PATCHes to `update`; the response re-renders just that card's Turbo frame, so tagging an image never reloads the page.
- **Authorization**: `GalleryAssetPolicy` restricts both actions to admins (super users), consistent with the default policy; non-admins are redirected like elsewhere in the app.

### Anything else to add?
- Titles are intentionally free-text so the team can tag with whatever metadata is useful (org, workshop, program). Search matches that text plus owner type and filename.
- Specs added: model (scopes), request (auth + search + inline edit), policy, and routing. All green locally (`18 examples, 0 failures`).
- Screenshots to follow.

### DIFF
--- a/app/controllers/gallery_assets_controller.rb
+++ b/app/controllers/gallery_assets_controller.rb
@@ -1,0 +1,33 @@
+class GalleryAssetsController < ApplicationController
+  before_action :set_gallery_asset, only: :update
+
+  def index
+    authorize! GalleryAsset
+    @query = params[:query].to_s.strip
+
+    scope = authorized_scope(GalleryAsset.images.with_attached_file.includes(:owner))
+    scope = scope.search_metadata(@query) if @query.present?
+    scope = scope.order(created_at: :desc)
+
+    @count = scope.count
+    @gallery_assets = scope.paginate(page: params[:page], per_page: 24)
+
+    render :results if turbo_frame_request?
+  end
+
+  def update
+    authorize! @gallery_asset
+    @gallery_asset.update(gallery_asset_params)
+    render partial: "gallery_assets/card", locals: { gallery_asset: @gallery_asset }
+  end
+
+  private
+
+  def set_gallery_asset
+    @gallery_asset = GalleryAsset.find(params[:id])
+  end
+
+  def gallery_asset_params
+    params.expect(gallery_asset: [ :title ])
+  end
+end
--- a/app/models/gallery_asset.rb
+++ b/app/models/gallery_asset.rb
@@ -1,2 +1,20 @@
 class GalleryAsset < Asset
+  # Only gallery assets whose attached file is an image. Joins the blob so the
+  # gallery can be searched/filtered in SQL and used as an image database.
+  scope :images, -> {
+    joins(file_attachment: :blob)
+      .where("active_storage_blobs.content_type LIKE ?", "image/%")
+  }
+
+  # Free-text search across the editable title, the owning record type, and the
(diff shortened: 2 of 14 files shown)


app/models/gallery_asset.rb
```diff
@@ -1,2 +1,20 @@
...
+  # Only gallery assets whose attached file is an image. Joins the blob so the
+  # gallery can be searched/filtered in SQL and used as an image database.
+  scope :images, -> {
+    joins(file_attachment: :blob)
+      .where("active_storage_blobs.content_type LIKE ?", "image/%")
+  }
+
+  # Free-text search across the editable title, the owning record type, and the
+  # original filename — the metadata the team tags images with (org, workshop, etc.).
+  scope :search_metadata, ->(query) {
```
maebeale, line 11: `search_metadata` references `active_storage_blobs`, so it must be chained after `.images` (which joins the blob). The controller always does `GalleryAsset.images.search_metadata(...)`. Matching on `owner_type` lets a search for "workshop" surface every workshop image even before it's been given a title.

app/controllers/gallery_assets_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    @count = scope.count
+    @gallery_assets = scope.paginate(page: params[:page], per_page: 24)
+
+    render :results if turbo_frame_request?
+  end
+
+  def update
+    authorize! @gallery_asset
+    @gallery_asset.update(gallery_asset_params)
```
maebeale, line 20: The form lives inside a per-card `<turbo-frame>`, so rendering just the `_card` partial here swaps that one card in place — editing a title never reloads the grid or loses the user's scroll/search position.

app/policies/gallery_asset_policy.rb
```diff
@@ -0,0 +1,10 @@
+class GalleryAssetPolicy < ApplicationPolicy
+  # The gallery image database is an internal admin tool for tagging and finding images.
+  def index?  = admin?
+  def update? = admin?
+
+  relation_scope do |relation|
```
maebeale, line 6: Admin-only: this is an internal image database. `relation_scope` returns `none` for non-admins so `authorized_scope` is safe even if the policy is later relaxed to expose `index?`.

PR #1545, opened 2026-06-05
Add inline gallery controls to set featured image and remove images

### LINKED ISSUE #1510: Gallery: easier way to change featured image and remove images
## Problem
Managing images in the gallery is cumbersome — there's no easy way to change which image is the featured/primary one, and removing images is harder than it should be.

## Proposal
- Add an obvious control to set/change the featured image directly from the gallery view
- Add an easy way to remove an image from the gallery

maebeale:
Closes #1510

- Stakeholders asked for an easier way to manage gallery images directly from the story (gallery) view, rather than going into the edit form
- Adds two inline, one-click controls on each gallery image: **set as featured** and **remove**
- Reduces friction for the common "change the hero image" and "drop a bad photo" tasks

- Added a `GalleryAssetsController` with `destroy` (remove an image) and `make_primary` (promote a gallery image to the featured/primary image) actions
- `make_primary` reuses the same `PrimaryAsset` ⇄ `GalleryAsset` STI swap the existing "Change photo" picker already uses, so there is one definition of what "featured" means — the old featured image is demoted back into the gallery
- Controls are rendered by a new `assets/_gallery_item_controls` partial, overlaid on each image and revealed on hover; they only render for users who can `manage?` the owning record (via `allowed_to?`), and every action re-checks the owner's policy server-side
- Because the controls hang off the shared `assets/_display_gallery_media` partial, this works for any asset owner (stories, story ideas, workshops, events, etc.), not just stories
- Routed under `/gallery_assets` rather than `/assets` to avoid colliding with the Rails asset-pipeline URL prefix (which 405s non-GET requests)

### UI Testing Checklist
- [ ] As an admin, hover a gallery image on a story show page and confirm the ⭐ and 🗑 controls appear
- [ ] Click ⭐ to promote a gallery image — it becomes the hero and the previous hero moves into the gallery
- [ ] Click 🗑 (confirm the prompt) to remove a gallery image
- [ ] As a signed-out / non-admin visitor, confirm no controls are shown

### Anything else to add?
- Request specs cover both actions, including authorization (admins act, non-admins/anonymous are blocked) and the featured-image swap
- Verified the controls render for admins and are hidden from anonymous visitors via request specs, and confirmed the overlay visually with a system screenshot (below)

### DIFF
--- a/app/controllers/gallery_assets_controller.rb
+++ b/app/controllers/gallery_assets_controller.rb
@@ -1,0 +1,31 @@
+class GalleryAssetsController < ApplicationController
+  before_action :set_asset, only: [ :destroy, :make_primary ]
+
+  # Remove an image from the gallery
+  def destroy
+    authorize! @asset.owner, to: :manage? # check the policy of the record that owns the asset
+    @asset.destroy!
+
+    redirect_back_or_to polymorphic_path(@asset.owner), notice: "Image removed."
+  end
+
+  # Promote an existing gallery image to be the featured (primary) image
+  def make_primary
+    authorize! @asset.owner, to: :manage?
+    ActiveRecord::Base.transaction do
+      if existing_primary = @asset.owner.assets.find_by(type: "PrimaryAsset")
+        existing_primary.update!(type: "GalleryAsset")
+      end
+
+      @asset.update!(type: "PrimaryAsset")
+    end
+
+    redirect_back_or_to polymorphic_path(@asset.owner), notice: "Featured image updated."
+  end
+
+  private
+
+  def set_asset
+    @asset = GalleryAsset.find(params[:id])
+  end
+end
--- a/app/views/assets/_display_gallery_media.html.erb
+++ b/app/views/assets/_display_gallery_media.html.erb
@@ -2,10 +2,18 @@
 <% gallery_assets = resource.gallery_assets.select { |img| img.file.attached? } %>
 <% align = (defined?(align) && align.present?) ? align.to_s : "center" %>
 <% if gallery_assets.any? %>
+  <% can_manage = allowed_to?(:manage?, resource) %>
   <div class="<%= resource.class.table_name %>-gallery text-<%= align %> mb-4">
     <div class="flex flex-wrap justify-<%= align %> gap-4 <%= 'mx-auto' if align == 'center' %> max-w-4xl">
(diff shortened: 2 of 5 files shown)


config/routes.rb
```diff
@@ -8,6 +8,11 @@
   # end
   resources :primary_assets
   resources :rich_text_assets
+  resources :gallery_assets, only: [ :destroy ] do
```
maebeale, line 11: Routed under `/gallery_assets` rather than `resources :assets` on purpose: the Rails asset pipeline owns the `/assets/*` URL prefix and returns 405 for non-GET requests, which would shadow these DELETE/POST routes.

app/controllers/gallery_assets_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    redirect_back_or_to polymorphic_path(@asset.owner), notice: "Image removed."
+  end
+
+  # Promote an existing gallery image to be the featured (primary) image
+  def make_primary
+    authorize! @asset.owner, to: :manage?
+    ActiveRecord::Base.transaction do
+      if existing_primary = @asset.owner.assets.find_by(type: "PrimaryAsset")
+        existing_primary.update!(type: "GalleryAsset")
```
maebeale, line 17: Promotion is a type swap, not a new upload: the current `PrimaryAsset` is demoted back to `GalleryAsset` and the chosen gallery image becomes the `PrimaryAsset`. This mirrors the existing `PrimaryAssetsController#create` flow used by the "Change photo" picker, so there's a single definition of "featured".

app/views/assets/_display_gallery_media.html.erb
```diff
@@ -2,10 +2,18 @@
 <% gallery_assets = resource.gallery_assets.select { |img| img.file.attached? } %>
 <% align = (defined?(align) && align.present?) ? align.to_s : "center" %>
 <% if gallery_assets.any? %>
+  <% can_manage = allowed_to?(:manage?, resource) %>
```
maebeale, line 5: Controls are gated by `allowed_to?(:manage?, resource)` so they only render for users who can manage the owning record. The controller re-checks the same policy server-side, so this is purely a UI gate. Because it lives in the shared gallery partial, the feature works for every asset owner (stories, story ideas, workshops, events, …), not just stories.

PR #1546, opened 2026-06-05
Add drag-and-drop ordering for featured stories

### LINKED ISSUE #1522: Home page: control the order of featured stories
## Problem
There's no way to control the order in which featured stories appear on the home page.

## Proposal
Add an admin control to reorder featured stories (e.g., drag-and-drop or a position field) so we can curate the home page display order.

maebeale:
Closes #1522

- The home page "Highlighted stories" section showed featured stories ordered by title, with no way to curate which story leads.
- Stakeholders requested control over the display order so they can promote specific stories on the home page.

- Added a `position` column to `stories`, scoped via the `positioning` gem to the featured set (`positioned on: :featured`), so featured stories keep a contiguous, independently-curated order.
- Backfilled existing rows with a contiguous per-group sequence in the migration so the gem has a valid starting state.
- Added an admin-only **Reorder featured stories** page (`/featured_stories`) that reuses the existing `sortable` Stimulus controller + SortableJS drag-and-drop, mirroring the FAQ/Category pattern already in the app.
- The reorder list intentionally shows the full featured set (published or not) so the drag index lines up exactly with the positioning scope; unpublished stories are flagged as hidden from the home page.
- Updated `Home::StoriesController` to order by `position` instead of `title`, so both the authenticated (`featured`) and public (`publicly_featured`) home views respect the curated order.
- Added an admin-only "Reorder featured" link on the stories index header.

### Anything else to add?
- Authorization reuses `StoryPolicy` via a new `reorder?` rule (super users only); non-admins are redirected to root, consistent with the app's `ActionPolicy::Unauthorized` handling.
- The public (anonymous) home page draws from `publicly_featured`; those stories follow the same `position` column, so curation carries over for stories that are both featured and publicly featured.
- Tests: model positioning specs, request specs for the reorder controller (authorization + reordering), and a home stories ordering spec — all green.

### UI Testing Checklist
- [ ] As a super user, visit Stories → "Reorder featured", drag a story to the top, and confirm the new order persists after reload.
- [ ] Confirm the home page "Highlighted stories" section reflects the curated order.
- [ ] Confirm non-admins cannot access `/featured_stories`.

### DIFF
--- a/app/controllers/featured_stories_controller.rb
+++ b/app/controllers/featured_stories_controller.rb
@@ -1,0 +1,15 @@
+class FeaturedStoriesController < ApplicationController
+  def index
+    authorize! Story, to: :reorder?
+    # Operate on the full featured set (published or not) so the drag-and-drop
+    # index lines up exactly with the positioning scope (positioned on: :featured).
+    @stories = Story.where(featured: true).order(:position).decorate
+  end
+
+  def update
+    story = Story.find(params[:id])
+    authorize! story, to: :reorder?
+    story.update!(position: params[:position])
+    head :no_content
+  end
+end
--- a/app/controllers/home/stories_controller.rb
+++ b/app/controllers/home/stories_controller.rb
@@ -5,7 +5,7 @@ module Home
     def index
       authorize! :home
       @stories = authorized_scope(Story.published
-                      .order(:title), with: HomePolicy)
+                      .order(:position), with: HomePolicy)
                       .decorate
 
       render "home/stories/index"
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -2,6 +2,12 @@ class Story < ApplicationRecord
   include AuthorCreditable
   include Featureable, Publishable, TagFilterable, Trendable, WindowsTypeFilterable, RichTextSearchable
 
+  # Featured stories are curated into a manual order (drag-and-drop) so the
+  # home page can control which stories surface first. Scoping on :featured
+  # keeps a contiguous sequence within the featured set, independent of the
+  # much larger pool of non-featured stories.
+  positioned on: :featured
+
   has_rich_text :rhino_body
 
   belongs_to :created_by, class_name: "User"
--- a/db/migrate/20260604143000_add_position_to_stories.rb
+++ b/db/migrate/20260604143000_add_position_to_stories.rb
@@ -1,0 +1,24 @@
+class AddPositionToStories < ActiveRecord::Migration[8.1]
+  def up
+    add_column :stories, :position, :integer
+    add_index :stories, [ :featured, :position ]
+
+    # Backfill a contiguous 1..n sequence within each featured group so the
+    # positioning gem (scoped on :featured) has a valid starting state.
+    execute <<~SQL.squish
+      UPDATE stories s
+      JOIN (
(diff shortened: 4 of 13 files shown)


app/models/story.rb
```diff
@@ -2,6 +2,12 @@ class Story < ApplicationRecord
   include AuthorCreditable
   include Featureable, Publishable, TagFilterable, Trendable, WindowsTypeFilterable, RichTextSearchable
 
+  # Featured stories are curated into a manual order (drag-and-drop) so the
+  # home page can control which stories surface first. Scoping on :featured
+  # keeps a contiguous sequence within the featured set, independent of the
+  # much larger pool of non-featured stories.
+  positioned on: :featured
+
```
maebeale, line 10: Scoped on `:featured` (rather than a global sequence) so the reorder list — which shows exactly the featured set — maps 1:1 to the absolute index SortableJS sends. A global position would desync once non-featured stories interleave.

app/controllers/featured_stories_controller.rb
```diff
@@ -0,0 +1,15 @@
+class FeaturedStoriesController < ApplicationController
+  def index
+    authorize! Story, to: :reorder?
+    # Operate on the full featured set (published or not) so the drag-and-drop
+    # index lines up exactly with the positioning scope (positioned on: :featured).
+    @stories = Story.where(featured: true).order(:position).decorate
```
maebeale, line 6: Deliberately uses `where(featured: true)` (published or not), not `Story.featured` (published-only). The list must match the positioning scope exactly, otherwise hidden unpublished featured stories would throw off the drag index. The view flags unpublished ones.

app/controllers/featured_stories_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def update
+    story = Story.find(params[:id])
+    authorize! story, to: :reorder?
+    story.update!(position: params[:position])
+    head :no_content
```
maebeale, line 13: Reads `params[:position]` (the top-level key SortableJS PUTs as JSON). The positioning gem intercepts the assignment and re-sequences the featured group.

db/migrate/20260604143000_add_position_to_stories.rb
```diff
@@ -0,0 +1,24 @@
+class AddPositionToStories < ActiveRecord::Migration[8.1]
+  def up
+    add_column :stories, :position, :integer
+    add_index :stories, [ :featured, :position ]
+
+    # Backfill a contiguous 1..n sequence within each featured group so the
+    # positioning gem (scoped on :featured) has a valid starting state.
+    execute <<~SQL.squish
+      UPDATE stories s
+      JOIN (
```
maebeale, line 10: Backfill gives each featured group a contiguous 1..n sequence (ROW_NUMBER per `featured`) so the gem starts from a valid state; column is set NOT NULL only after backfill.

app/controllers/home/stories_controller.rb
```diff
@@ -5,7 +5,7 @@ class StoriesController < ApplicationController
     def index
       authorize! :home
       @stories = authorized_scope(Story.published
-                      .order(:title), with: HomePolicy)
+                      .order(:position), with: HomePolicy)
```
maebeale, line 8: Order by `:position` instead of `:title`. HomePolicy still scopes to `featured`/`publicly_featured`; the curated order flows through to both the authenticated and public home views.

PR #1547, opened 2026-06-05
Open gallery attachment links in a new tab to avoid losing edits

### LINKED ISSUE #1511: Gallery: open image in new window to avoid losing in-progress work
## Problem
Clicking an image in the gallery navigates away from the current page, which risks losing in-progress edits.

## Proposal
When an image in the gallery is clicked, open it in a new window/tab so the user doesn't lose their work.

maebeale:
Closes #1511

- Stakeholder report: clicking an image/attachment in the gallery section of an edit form navigated away in the **same tab**, discarding any unsaved form edits.
- Show-page galleries already open full-size images in a new tab (via `DisplayImagePresenter`), but the in-form attachment cards (`shared/_form_image_field`) were missed — their filename link opened in the same tab.
- This makes the in-form behavior consistent and protects in-progress work.

- Added `target: "_blank", rel: "noopener"` to the persisted-file filename link in `app/views/shared/_form_image_field.html.erb`.
- This partial is shared by every model's edit form (stories, resources, workshop variations, etc.), so the fix applies everywhere the gallery/attachment section appears.
- TDD: added a view spec (`spec/views/shared/_form_image_field.html.erb_spec.rb`) that asserts the filename link opens in a new tab — confirmed it failed before the change and passes after.

### UI Testing Checklist
- [ ] Edit a story (or resource/workshop variation) that has an attached gallery image
- [ ] Click the filename link under an existing attachment → it opens the file in a **new tab**
- [ ] The original edit form remains open with unsaved changes intact

### Anything else to add?
- Scope kept minimal: the filename link was the only element in this section that actually navigated (the thumbnail image itself is not a link), so this is the precise source of the "lost work" behavior.

### DIFF
--- a/app/views/shared/_form_image_field.html.erb
+++ b/app/views/shared/_form_image_field.html.erb
@@ -80,7 +80,7 @@
       <div data-file-preview-target="filename"
            class="text-center text-sm text-gray-500 break-all max-w-[8rem]">
         <%= if blob.present? && blob.respond_to?(:persisted?) && blob.persisted?
-              link_to(blob.filename.to_s, url_for(file), class: "underline")
+              link_to(blob.filename.to_s, url_for(file), class: "underline", target: "_blank", rel: "noopener")
             else
               "No file selected"
             end
--- a/spec/views/shared/_form_image_field.html.erb_spec.rb
+++ b/spec/views/shared/_form_image_field.html.erb_spec.rb
@@ -1,0 +1,14 @@
+require "rails_helper"
+
+RSpec.describe "shared/_form_image_field", type: :view do
+  let(:asset) { create(:gallery_asset, :with_file) }
+  let(:builder) { ActionView::Helpers::FormBuilder.new(:gallery_asset, asset, view, {}) }
+
+  before do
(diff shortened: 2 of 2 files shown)


app/views/shared/_form_image_field.html.erb
```diff
@@ -80,7 +80,7 @@
       <div data-file-preview-target="filename"
            class="text-center text-sm text-gray-500 break-all max-w-[8rem]">
         <%= if blob.present? && blob.respond_to?(:persisted?) && blob.persisted?
-              link_to(blob.filename.to_s, url_for(file), class: "underline")
+              link_to(blob.filename.to_s, url_for(file), class: "underline", target: "_blank", rel: "noopener")
```
maebeale, line 83: This partial is rendered in every model's edit form via `shared/_form_image_fields`, so adding `target: "_blank"` here fixes the "navigates away and loses unsaved edits" report (#1511) across stories, resources, workshop variations, etc. in one place. Matches the new-tab behavior the show-page galleries already get from `DisplayImagePresenter`.

PR #1550, opened 2026-06-05
Open additional story images in an inline lightbox

### LINKED ISSUE #1520: Published story: additional images should open inline with scroll/gallery view
## Problem
Clicking an additional image on a published story navigates away, and there's no way to scroll through the additional images as a set.

## Proposal
- Clicking an additional image opens it in the same window (e.g., in a lightbox/modal)
- Provide next/previous controls so the user can scroll through all additional images for that story

maebeale:
Closes #1520

- On a published story, the "additional images" (gallery assets) each linked straight to the full-size file, so clicking one navigated the viewer away from the story.
- There was no way to browse the additional images as a set.
- This adds an inline lightbox so a clicked image opens in place, and the viewer can scroll through all of the story's additional images.

- Added a `lightbox` Stimulus controller (`app/frontend/javascript/controllers/lightbox_controller.js`) that opens a modal, tracks the current index, and supports next/previous, arrow keys, Escape, and backdrop-click to close.
- Reworked `app/views/assets/_display_gallery_media.html.erb` so each image gallery item is a link marked as a lightbox `item` target; the existing full-size link is kept as the `href` so it still works as a no-JS fallback.
- Added `app/views/assets/_lightbox.html.erb` for the modal markup (image, counter, prev/next/close controls).
- Behavior applies wherever the gallery partial is rendered with `link: true` (stories, story shares, workshop logs, events, etc.); PDFs and other non-image assets keep their existing rendering.
- Registered the controller and updated `AGENTS.md`.

### Anything else to add?
- Progressive enhancement: with JS disabled the gallery images still link to the full-size file (previous behavior).
- Nav controls hide automatically when a gallery has a single image.

### Test plan
- Added `spec/system/story_gallery_lightbox_spec.rb` (Selenium): clicking an additional image opens it inline without navigating away, next/previous scroll through the set, and closing dismisses the modal.

### DIFF
--- a/app/frontend/javascript/controllers/lightbox_controller.js
+++ b/app/frontend/javascript/controllers/lightbox_controller.js
@@ -1,0 +1,75 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="lightbox"
+//
+// Opens gallery images in an inline modal instead of navigating away, and lets
+// the viewer scroll through all of the gallery's images with next/previous
+// controls, the arrow keys, or by closing with Escape / a backdrop click.
+//
+// Each clickable image is an "item" target whose href points at the full-size
+// image. The modal, image, counter, and nav buttons live in the same scope.
+export default class extends Controller {
+  static targets = ["item", "modal", "image", "counter", "prev", "next"]
+  static values = { index: Number }
+
+  open(event) {
+    const index = this.itemTargets.indexOf(event.currentTarget)
+    if (index === -1) return
+
+    this.indexValue = index
+    this.modalTarget.classList.remove("hidden")
+    this.modalTarget.classList.add("flex")
+    document.body.classList.add("overflow-hidden")
+    this.modalTarget.focus()
+  }
+
+  close() {
+    this.modalTarget.classList.add("hidden")
+    this.modalTarget.classList.remove("flex")
+    document.body.classList.remove("overflow-hidden")
+  }
+
+  next() {
+    if (this.isClosed) return
+    this.indexValue = (this.indexValue + 1) % this.itemTargets.length
+  }
+
+  prev() {
+    if (this.isClosed) return
+    const count = this.itemTargets.length
+    this.indexValue = (this.indexValue - 1 + count) % count
+  }
+
+  closeOnEscape() {
+    if (this.isClosed) return
+    this.close()
+  }
+
+  backdropClose(event) {
+    if (event.target === this.modalTarget) this.close()
+  }
+
+  indexValueChanged() {
+    const item = this.itemTargets[this.indexValue]
+    if (!item) return
+
+    this.imageTarget.src = item.getAttribute("href")
+    this.imageTarget.alt = item.dataset.caption || ""
+
+    if (this.hasCounterTarget) {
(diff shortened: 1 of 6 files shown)


app/frontend/javascript/controllers/lightbox_controller.js
```diff
@@ -0,0 +1,75 @@
...
+
+  closeOnEscape() {
+    if (this.isClosed) return
+    this.close()
+  }
+
+  backdropClose(event) {
+    if (event.target === this.modalTarget) this.close()
+  }
+
```
maebeale, line 51: The full-size image URL is read from each item's `href` rather than a separate data attribute. That keeps a single source of truth and means the anchors double as a no-JS fallback (they still link to the file when Stimulus isn't loaded).

app/frontend/javascript/controllers/lightbox_controller.js
```diff
@@ -0,0 +1,75 @@
...
+    if (this.isClosed) return
+    this.indexValue = (this.indexValue + 1) % this.itemTargets.length
+  }
+
+  prev() {
+    if (this.isClosed) return
+    const count = this.itemTargets.length
+    this.indexValue = (this.indexValue - 1 + count) % count
+  }
+
```
maebeale, line 42: Prev/next wrap around the set (modulo length) so the viewer can keep scrolling in either direction without hitting a dead end.

app/views/assets/_display_gallery_media.html.erb
```diff
@@ -1,12 +1,28 @@
...
 <% if gallery_assets.any? %>
-  <div class="<%= resource.class.table_name %>-gallery text-<%= align %> mb-4">
+  <div class="<%= resource.class.table_name %>-gallery text-<%= align %> mb-4"
+       data-controller="lightbox"
+       data-action="keydown.esc@window->lightbox#closeOnEscape keydown.left@window->lightbox#prev keydown.right@window->lightbox#next">
     <div class="flex flex-wrap justify-<%= align %> gap-4 <%= 'mx-auto' if align == 'center' %> max-w-4xl">
-      <% gallery_assets.each_with_index do |gallery_assets, idx| %>
-        <%= render "assets/display_image", item: gallery_assets, idx: idx, variant: :gallery, link: (defined?(link) ? link : nil) %>
+      <% gallery_assets.each_with_index do |gallery_asset, idx| %>
+        <% if link && gallery_asset.file.content_type.to_s.start_with?("image/") %>
```
maebeale, line 11: Lightbox is gated on `link` (the existing clickable-gallery flag) AND the asset being an image. Non-image gallery assets (e.g. PDFs) fall through to the original `display_image` rendering, and galleries rendered without `link` (e.g. reports) keep their previous non-clickable behavior.

app/views/assets/_display_gallery_media.html.erb
```diff
@@ -1,12 +1,28 @@
...
+          <a href="<%= url_for(gallery_asset.file) %>"
+             class="flex grow"
+             data-lightbox-target="item"
```
maebeale, line 14: The anchor keeps `href` pointing at the full-size file (no-JS fallback) and `data-action="lightbox#open:prevent"` intercepts the click when JS is active. Inner image is rendered with `link: false` so we don't nest anchors.

PR #1551, opened 2026-06-05
Show unconfirmed facilitators in story author dropdown

### LINKED ISSUE #1515: New story: facilitators with unconfirmed emails missing from author dropdown
## Problem
Some facilitators do not appear in the story author dropdown because their email is unconfirmed. Confirmed examples:

- Courtney Koczka
- Lemny Perez

## Expected behavior
Email confirmation status should not gate whether a facilitator appears as a selectable story author. They are real facilitators and should be selectable regardless of email confirmation.

maebeale:
Closes #1515

- Facilitators with **unconfirmed emails** were silently missing from the story author dropdown (confirmed examples: Courtney Koczka, Lemny Perez).
- Email confirmation status should not gate whether a facilitator can be credited as a story author — they are real people who do the work regardless of whether they've confirmed their email.

- The dropdown was built from `User.has_access`, which requires `confirmed_at` to be present — that's the gate that hid these facilitators.
- Added a dedicated `User.selectable_as_author` scope: active + unlocked, **without** the email-confirmation requirement.
- Switched the story form's author collection (`set_form_variables`) to use `selectable_as_author`.
- Left `has_access` untouched so its access-control semantics (used broadly elsewhere) are unchanged.

### Anything else to add?
- The same `has_access`-based author dropdown exists for **story ideas** (`story_ideas_controller`) and a few other forms (workshop ideas, bookmarks, etc.). Those were intentionally left out of scope since the issue is specifically about the story author dropdown — happy to extend `selectable_as_author` to them in a follow-up if the same treatment is wanted.
- Tested via a model spec for the new scope (includes confirmed + unconfirmed, excludes inactive/locked). A request spec for `GET /stories/new` was not added because full-page renders 500 in the test environment (asset/vite related, pre-existing) — existing story request specs only exercise Turbo-frame/redirect paths for the same reason.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -82,6 +82,10 @@ class User < ApplicationRecord
   end
 
   scope :has_access, -> { where(locked_at: nil, inactive: [ false, nil ]).where.not(confirmed_at: nil) }
+  # Users selectable as a story/idea author. Unlike has_access, email confirmation
+  # is not required: facilitators are real people who should be creditable
+  # regardless of whether they have confirmed their email.
+  scope :selectable_as_author, -> { where(locked_at: nil, inactive: [ false, nil ]) }
 
   def self.search_by_params(params)
     results = is_a?(ActiveRecord::Relation) ? self : all
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -131,7 +131,7 @@ class StoriesController < ApplicationController
                             .references(:users)
                             .order(:created_at)
     @people = Person.order(Arel.sql("LOWER(first_name), LOWER(last_name)"))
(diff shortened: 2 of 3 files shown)


app/models/user.rb
```diff
@@ -82,6 +82,10 @@ class User < ApplicationRecord
   end
 
   scope :has_access, -> { where(locked_at: nil, inactive: [ false, nil ]).where.not(confirmed_at: nil) }
+  # Users selectable as a story/idea author. Unlike has_access, email confirmation
+  # is not required: facilitators are real people who should be creditable
+  # regardless of whether they have confirmed their email.
+  scope :selectable_as_author, -> { where(locked_at: nil, inactive: [ false, nil ]) }
```
maebeale, line 88: Deliberately a separate scope rather than relaxing `has_access`: `has_access` is the access-control gate used across the app (auth, search, remote search), and it should keep requiring email confirmation. Authorship is a weaker notion — we only need the person to be a real, active facilitator — so confirmation is dropped here only.

PR #1552, opened 2026-06-05
Allow all facilitators + free-text story author (#1514)

### LINKED ISSUE #1514: New story: story author dropdown should include all facilitators + free-text fallback
## Problem
On the new story form, the story author dropdown only includes active facilitators. This mirrors the issue with "story spotlighted" — we need to choose from all facilitators, not just active ones.

Additionally, some facilitators were never entered into the CMS, or we only have their first name, so they aren't selectable at all.

## Proposal
- Story author dropdown pulls from all facilitators (not just active)
- Provide an option to type in a free-text author name for facilitators not in the CMS, or where we only have a first name

maebeale:
Closes #1514

- The new/edit story **author** dropdown only listed active, confirmed users (`User.has_access`), so facilitators who were inactive, never entered in the CMS, or only known by a first name could not be credited as the story author.
- This mirrors the issue already fixed for the **spotlighted facilitator** dropdown, which lists all people.

- **All facilitators in the dropdown:** dropped the `has_access` filter from `@users` in `StoriesController#set_form_variables`, so the author dropdown lists all users — matching how the spotlighted-facilitator dropdown lists all `Person` records.
- **Free-text fallback:** added a nullable `author_name` column (migration) plus an "Or type an author name" text input on the form, for facilitators not in the CMS or where only a first name is known.
- **Credit display:** `AuthorCreditable#author_credit` now returns `author_name` when present, overriding the `created_by` person and the credit preference. The concern uses `try(:author_name)`, so `StoryIdea` (no such column) is unaffected.
- **Required association:** when a free-text author is entered without selecting a user, `create` defaults `created_by` to the current admin so the required `created_by_id` is satisfied.
- **Show page:** the author profile link is suppressed when a free-text name is used, so it doesn't link to the admin creator's profile.

### UI Testing Checklist
- [ ] New/edit story form shows inactive facilitators in the author dropdown
- [ ] Typing a free-text author name and submitting (without selecting a dropdown author) saves the story and shows that name under "Story by"
- [ ] Selecting a dropdown author with a searchable profile still links to that person on the show page

### Anything else to add?
- Tests: model specs for the free-text `author_credit` override; request specs for free-text create (defaulting `created_by`) and inactive facilitators appearing in the dropdown. Full `stories` model + request specs pass.
- `db/schema.rb` change is limited to the new column + version bump (unrelated local dev-DB drift was excluded).

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -62,6 +62,9 @@ class StoriesController < ApplicationController
 
   def create
     @story = Story.new(story_params.except(:category_ids, :sector_ids))
+    # A free-text author name covers facilitators who aren't in the CMS, so an
+    # author User may not be selected. Fall back to the current admin as creator.
+    @story.created_by_id ||= current_user.id
     authorize! @story
 
     success = false
@@ -131,7 +134,7 @@ class StoriesController < ApplicationController
                             .references(:users)
                             .order(:created_at)
     @people = Person.order(Arel.sql("LOWER(first_name), LOWER(last_name)"))
-    @users = User.has_access.includes(:person).left_joins(:person).order(Arel.sql("people.first_name IS NULL, LOWER(people.first_name), LOWER(people.last_name), LOWER(users.email)"))
+    @users = User.includes(:person).left_joins(:person).order(Arel.sql("people.first_name IS NULL, LOWER(people.first_name), LOWER(people.last_name), LOWER(users.email)"))
     @windows_types = WindowsType.all
     @workshops = authorized_scope(Workshop.all).includes(:windows_type).order(:title)
     @categories_grouped =
@@ -189,7 +192,7 @@ class StoriesController < ApplicationController
     params.require(:story).permit(
       :title, :rhino_body, :featured, :published, :publicly_visible, :publicly_featured, :youtube_url, :website_url,
       :windows_type_id, :organization_id, :workshop_id, :external_workshop_title,
-      :created_by_id, :updated_by_id, :story_idea_id, :spotlighted_facilitator_id, :author_credit_preference,
+      :created_by_id, :updated_by_id, :story_idea_id, :spotlighted_facilitator_id, :author_credit_preference, :author_name,
       category_ids: [],
       sector_ids: [],
       primary_asset_attributes: [ :id, :file, :_destroy ],
--- a/app/models/concerns/author_creditable.rb
+++ b/app/models/concerns/author_creditable.rb
@@ -20,6 +20,8 @@ module AuthorCreditable
   }.freeze
 
   def author_credit
+    free_text = try(:author_name)
+    return free_text if free_text.present?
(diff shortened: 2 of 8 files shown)


app/controllers/stories_controller.rb
```diff
@@ -62,6 +62,9 @@ def edit
 
   def create
     @story = Story.new(story_params.except(:category_ids, :sector_ids))
+    # A free-text author name covers facilitators who aren't in the CMS, so an
+    # author User may not be selected. Fall back to the current admin as creator.
+    @story.created_by_id ||= current_user.id
```
maebeale, line 67: Free-text authors may not be in the CMS, so no author User is selected. We default `created_by` to the current admin here to satisfy the required association while `author_name` carries the display credit.

app/models/concerns/author_creditable.rb
```diff
@@ -20,6 +20,8 @@ module AuthorCreditable
   }.freeze
 
   def author_credit
+    free_text = try(:author_name)
```
maebeale, line 23: `try(:author_name)` so the shared concern keeps working for StoryIdea, which has no `author_name` column. Free text intentionally overrides the credit preference.

app/views/stories/show.html.erb
```diff
@@ -28,7 +28,7 @@
         <!-- Meta Data (exact screenshot style) -->
         <div class="text-sm text-gray-600 space-y-0.5">
           <p><strong>Story by:</strong>
-            <% if @story.created_by.person&.profile_is_searchable %>
+            <% if @story.author_name.blank? && @story.created_by.person&.profile_is_searchable %>
```
maebeale, line 31: Suppress the profile link for free-text authors — otherwise it would link to the admin creator rather than the credited facilitator.

app/controllers/stories_controller.rb
```diff
@@ -131,7 +134,7 @@ def set_form_variables
                             .references(:users)
                             .order(:created_at)
     @people = Person.order(Arel.sql("LOWER(first_name), LOWER(last_name)"))
-    @users = User.has_access.includes(:person).left_joins(:person).order(Arel.sql("people.first_name IS NULL, LOWER(people.first_name), LOWER(people.last_name), LOWER(users.email)"))
+    @users = User.includes(:person).left_joins(:person).order(Arel.sql("people.first_name IS NULL, LOWER(people.first_name), LOWER(people.last_name), LOWER(users.email)"))
```
maebeale, line 137: Dropped `has_access` so inactive/unconfirmed facilitators are selectable, mirroring how the spotlighted-facilitator dropdown lists all people.

jmilljr24:
I'm still having a hard time with author being a `User`. I know there is more to it but adding a column for free text user feels off. If author was a `Person` we would just need to create the person with no user account. A `Person` is who authored the story. A `User` would be more in line with if they were logged in and we wanted to know who "created" the record.

PR #1560, opened 2026-06-05
Forms Part 2: Extract registration form-name constants to a single source of truth

maebeale:
> Stacked on #1301 (`maebeale/shared-form-builder`). Review/merge that first; this PR's base is the 1301 branch, not `main`.

- Addresses review item #5 from #1301: the registration form-name "magic strings"
- `"Short Event Registration"` and `"Extended Event Registration"` were typed as bare literals in ~10 places (model, event form view, dev seeds, 7 spec files)
- `Event#build_public_registration_form` looks up the form by exact name match — so a rename in any single location silently breaks the lookup (no form attached, no error)
- The pre-refactor code avoided this with a builder constant; that constant disappeared when the three builders were consolidated into `FormBuilderService`

- Added two constants to `FormBuilderService`:
  - `SHORT_REGISTRATION_FORM_NAME`
  - `EXTENDED_REGISTRATION_FORM_NAME`
- Replaced every literal in app code, dev seeds, and specs with the constants
- Behavior is unchanged — this is a pure refactor to a single source of truth
- Legacy label aliases (`"Public Registration"`, `"Short Registration"`) in the event form `case` are intentionally left as literals; they're historical names, not the canonical ones

### Scope note (Option A vs Option B)

This is the **safe, mechanical fix (Option A)**: make the coupling explicit and grep-able. It does **not** remove the deeper design smell — that `build_public_registration_form` couples an event's *title* (`/training/i`) to a form's *name*. Removing the name-based auto-selection entirely (letting admins explicitly pick the registration form) is a behavior/product change worth its own issue/PR, deliberately out of scope here.

### Testing

- `event_spec`, `form_builder_service_spec`, and `public_registration_spec` (the specs that exercise the form-name lookup) pass
- Rubocop clean on all changed files
- The 6 failures in `registrations_spec` are pre-existing full-page-render 500s caused by Vite not resolving in a fresh worktree (no `node_modules`), unrelated to this change

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -149,7 +149,7 @@ class Event < ApplicationRecord
   def build_public_registration_form
     return if event_forms.registration.exists?
 
-    form_name = title&.match?(/training/i) ? "Extended Event Registration" : "Short Event Registration"
+    form_name = title&.match?(/training/i) ? FormBuilderService::EXTENDED_REGISTRATION_FORM_NAME : FormBuilderService::SHORT_REGISTRATION_FORM_NAME
     form = Form.standalone.find_by(name: form_name)
     return unless form
 
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -386,16 +386,16 @@
               <% default_form_id = if current_reg_form
                                      current_reg_form.id
                                    elsif @event.title&.match?(/training/i)
-                                     @registration_forms.find { |rf| rf.name == "Extended Event Registration" }&.id
+                                     @registration_forms.find { |rf| rf.name == FormBuilderService::EXTENDED_REGISTRATION_FORM_NAME }&.id
                                    else
-                                     @registration_forms.find { |rf| rf.name == "Short Event Registration" }&.id
+                                     @registration_forms.find { |rf| rf.name == FormBuilderService::SHORT_REGISTRATION_FORM_NAME }&.id
                                    end %>
               <select name="event[registration_form_id]" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
                 <option value="">No registration form</option>
                 <% @registration_forms.each do |reg_form| %>
                   <% label = case reg_form.name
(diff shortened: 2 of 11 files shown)


app/models/event.rb
```diff
@@ -149,7 +149,7 @@ def public_registration_just_enabled?
   def build_public_registration_form
     return if event_forms.registration.exists?
 
-    form_name = title&.match?(/training/i) ? "Extended Event Registration" : "Short Event Registration"
+    form_name = title&.match?(/training/i) ? FormBuilderService::EXTENDED_REGISTRATION_FORM_NAME : FormBuilderService::SHORT_REGISTRATION_FORM_NAME
```
maebeale, line 152: Constants make the lookup grep-able, but the underlying smell remains: this couples the event *title* (`/training/i`) to a form *name*. If an admin renames the form in the UI, the lookup returns nil and no form is attached (silently). Removing this auto-selection in favor of an explicit admin choice is the Option B follow-up, intentionally out of scope here.

app/views/events/_form.html.erb
```diff
@@ -386,16 +386,16 @@
...
-                                     @registration_forms.find { |rf| rf.name == "Short Event Registration" }&.id
+                                     @registration_forms.find { |rf| rf.name == FormBuilderService::SHORT_REGISTRATION_FORM_NAME }&.id
                                    end %>
               <select name="event[registration_form_id]" class="w-full rounded border-gray-300 shadow-sm px-3 py-2 text-sm">
                 <option value="">No registration form</option>
                 <% @registration_forms.each do |reg_form| %>
                   <% label = case reg_form.name
-                             when "Extended Event Registration", "Public Registration" then "Extended registration form"
-                             when "Short Event Registration", "Short Registration" then "Short registration form"
+                             when FormBuilderService::EXTENDED_REGISTRATION_FORM_NAME, "Public Registration" then "Extended registration form"
```
maebeale, line 397: `"Public Registration"` / `"Short Registration"` are left as literals on purpose — they're legacy form names kept for backward-compatible labeling, not the canonical names the rest of the code builds and looks up by.

PR #1561, opened 2026-06-05
Show story idea body as required and humanize its error

maebeale:

- The story idea body field gave submitters no visual cue that it was required.
- When the body was left blank, the public-facing validation error exposed the internal attribute name "Rhino body" (e.g. "Rhino body can't be blank"), which is confusing to users.

- Added an optional `required:` keyword to the `rhino_editor` helper that switches the label to the `required` class and appends a red `*` marker. Defaulted to `false` so the helper's other ~40 usages (workshops, events, etc.) are unaffected.
- Passed `required: true` to the body field on the story idea form.
- Added `activerecord.attributes.story_idea.rhino_body: "Body"` to `en.yml` (matching the existing `workshop_variation` pattern) so the error reads "Body can't be blank".
- Added a request spec asserting the invalid-create response says "Body can't be blank" and not "Rhino body".

### UI Testing Checklist

- [ ] Story idea form (new): body label shows a red `*` required marker
- [ ] Submitting the form with a blank body shows "Body can't be blank" (not "Rhino body")

### Anything else to add?

- The shared `rhino_editor` helper change is scoped behind a parameter, so no other forms change behavior.

### DIFF
--- a/app/helpers/rhino_editor_helper.rb
+++ b/app/helpers/rhino_editor_helper.rb
@@ -1,16 +1,17 @@
 module RhinoEditorHelper
   # custom rhino editor with stimulus controller attached to edit raw source html
-  def rhino_editor(form, base_attribute_name, label: nil, hint: nil)
+  def rhino_editor(form, base_attribute_name, label: nil, hint: nil, required: false)
     rhino_attr = :"rhino_#{base_attribute_name}"
     field_id = form.field_id(rhino_attr)
     value = form.object.public_send(rhino_attr)
 
     label_tag =
       if label.present?
+        label_content = required ? safe_join([ label, " ", content_tag(:abbr, "*", title: "required", class: "text-red-600 no-underline") ]) : label
         form.label(
           base_attribute_name,
-          label,
-          class: "block font-medium mb-1 text-gray-700 text optional"
+          label_content,
+          class: "block font-medium mb-1 text-gray-700 text #{required ? "required" : "optional"}"
         )
       end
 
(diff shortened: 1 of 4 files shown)


app/helpers/rhino_editor_helper.rb
```diff
@@ -1,16 +1,17 @@
...
   # custom rhino editor with stimulus controller attached to edit raw source html
-  def rhino_editor(form, base_attribute_name, label: nil, hint: nil)
+  def rhino_editor(form, base_attribute_name, label: nil, hint: nil, required: false)
     rhino_attr = :"rhino_#{base_attribute_name}"
     field_id = form.field_id(rhino_attr)
     value = form.object.public_send(rhino_attr)
 
     label_tag =
       if label.present?
+        label_content = required ? safe_join([ label, " ", content_tag(:abbr, "*", title: "required", class: "text-red-600 no-underline") ]) : label
```
maebeale, line 10: Scoped the required marker behind a keyword (default `false`) so the ~40 other `rhino_editor` callers keep the `optional` styling unchanged.

config/locales/en.yml
```diff
@@ -33,6 +33,8 @@ en:
   hello: "Hello world"
   activerecord:
     attributes:
+      story_idea:
+        rhino_body: "Body"
```
maebeale, line 37: Overriding the attribute's human name here makes `full_messages` (used by the shared errors partial) read "Body can't be blank" instead of leaking the internal `rhino_body` name. Follows the existing `workshop_variation` pattern.

PR #1563, opened 2026-06-05
Custom forms with roles

jmilljr24:


app/controllers/events/public_registrations_controller.rb
```diff
@@ -48,15 +52,16 @@ def create
...
+        form_params: registration_params,
+        scholarship_requested: @scholarship,
+        person: current_user&.person
       )
 
       if result.success?
         registration = result.event_registration
 
-        if !registration.scholarship_requested? && @event.cost_cents.to_i > 0 && credit_card_payment?(form_params)
-          checkout_session = create_stripe_checkout_session(registration, form_params)
```
maebeale, line 59: we don't need to send params to stripe checkout anymore?

app/controllers/events/public_registrations_controller.rb
```diff
@@ -90,47 +95,31 @@ def show
...
 
     def credit_card_payment?(form_params)
-      payment_method_field = @form.form_fields.find_by(field_key: "payment_method")
+      payment_method_field = @form.form_fields.find_by(field_identifier: "payment_method")
       return false unless payment_method_field
 
       form_params[payment_method_field.id.to_s] == "Credit Card"
     end
 
-    def number_of_attendees(form_params)
```
maebeale, line 117: yeah i think it's the right move to remove this

app/controllers/events/public_registrations_controller.rb
```diff
@@ -90,47 +95,31 @@ def show
...
-
-    def create_stripe_checkout_session(registration, form_params)
+    def create_stripe_checkout_session(registration)
       person = registration.registrant
-      attendees = number_of_attendees(form_params)
-      amount = @event.cost_cents * attendees
+      amount = @event.cost_cents
 
       metadata = { event_registration_id: registration.id }
-      comments_field = @form.form_fields.find_by(field_key: "payment_comments")
```
maebeale, line 130: we don't want any comments?

app/controllers/events/public_registrations_controller.rb
```diff
@@ -163,16 +152,146 @@ def scholarship_mode?
...
+
     def visible_form_fields
-      scope = @form.form_fields.where(status: :active)
-      if scholarship_mode?
-        scope = scope.where.not(field_group: "payment")
-      else
-        scope = scope.where.not(field_group: "scholarship")
+      scope = @form.form_fields
+
+      person = current_user&.person if params[:as_visitor] != "true"
```
maebeale, line 191: do we want to force `:as_visitor`?

app/controllers/forms_controller.rb
```diff
@@ -0,0 +1,128 @@
+class FormsController < ApplicationController
```
maebeale, line 1: does this need a rebase if the other pr has been merged?

maebeale (approved):
🚀

app/controllers/events/public_registrations_controller.rb
jmilljr24: We will create a new payment form for the multi-person payment flow so there isn't a direct need for comments in this section. I think keeping it would open the door to people submitting odd things and harder to track.

app/controllers/events/public_registrations_controller.rb
jmilljr24: This was to grab the comments but we shouldn't need it in this flow. Maybe when i add a new special group payment form.

PR #1572, opened 2026-06-06
Parallelize the test suite in CI with parallel_tests

maebeale:

- The CI `build-and-test` job ran the full RSpec suite (~280 specs, 27 Selenium system specs) in a **single process** — the wall-clock bottleneck (~8 min test execution).
- GitHub's `ubuntu-latest` runners have **4 vCPUs** that sat mostly idle.
- This splits the suite across **4 parallel processes**, cutting test execution to **~3m21s** (verified on CI — see results below).

- Added [`parallel_tests`](https://github.com/grosser/parallel_tests); appended `TEST_ENV_NUMBER` to the test DB name in `config/database.yml` so each process gets an isolated DB (`awbw_test`, `awbw_test2`, …). Composes with the `workspace_port` suffix.
- CI `build-and-test` now runs `rake parallel:create parallel:load_schema`, then `parallel_rspec spec/` with `PARALLEL_TEST_PROCESSORS=4`.
- Each process writes its own JSON results file to `tmp/rspec_results/` (a single `--out` would be clobbered); the results artifact uploads that directory.
- **Precompile vite assets once before the parallel run** + `VITE_RUBY_AUTO_BUILD=false`. Vite's `autoBuild` otherwise had each of the 4 processes build into the shared `public/vite-test/` dir at first request; the concurrent builds raced and corrupted each other (`rollup-plugin-gzip` ENOENT on `*.br` → `ViteRuby::MissingEntrypointError` → 500s). A single up-front build is clean (swiper 12 is fine), so processes now only read prebuilt assets.
- **Moved SimpleCov behind a `COVERAGE` flag** set only by the coverage-badge workflow on `main`. It no longer runs on PRs — removing branch-coverage overhead and avoiding cross-process result-merge / per-process `minimum_coverage` problems under parallel.

### Results (CI `build-and-test` job, ~4m53s total)

| Step | Time |
|---|---|
| Container/Ruby/Node/npm setup | ~43s |
| Build test assets | 21s |
| Setup databases (parallel) | 23s |
| **Run tests** (4 processes) | **~3m21s** |

Test execution: **~8 min → ~3m21s** (~2.4×).

### Anything else to add?

- **Behavior change to flag:** PRs no longer enforce the SimpleCov `minimum_coverage 20` gate (coverage still tracked on `main` via the badge job). Easy to restore on PRs with proper parallel SimpleCov merging if preferred.
- **Follow-up (not in this PR):** sharding across a GitHub matrix (multiple runners) would stack on top of this, especially isolating the slow system specs; `rspec --profile` would target the worst offenders.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,14 @@ jobs:
           node-version: 22
       - run: npm ci
 
+      # Precompile assets once up front. Vite's autoBuild would otherwise have
+      # each parallel_tests process build into the shared public/vite-test dir
+      # at first request, and those concurrent builds corrupt each other.
+      - name: Build test assets
+        run: bin/vite build --mode test
+        env:
+          RAILS_ENV: test
+
       - name: Setup databases
         run: |
           bundle exec rake parallel:create parallel:load_schema
@@ -79,6 +87,9 @@ jobs:
         run: bundle exec parallel_rspec spec/
         env:
           RAILS_ENV: test
+          # Assets are prebuilt above; never let a test process rebuild them
+          # (concurrent rebuilds race on the shared output dir).
+          VITE_RUBY_AUTO_BUILD: "false"
           # AWS credentials
           AWS_REGION: ${{ vars.AWS_REGION }}
           AWS_S3_BUCKET: ${{ vars.AWS_S3_BUCKET }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,11 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
 
+    # GitHub-hosted ubuntu-latest runners have 4 vCPUs; split the suite across
+    # 4 parallel_tests processes (each with its own database) to cut wall time.
+    env:
+      PARALLEL_TEST_PROCESSORS: 4
+
     services:
       mysql:
         image: mysql:8.0
@@ -54,9 +59,9 @@ jobs:
           node-version: 22
       - run: npm ci
 
-      - name: Setup database
+      - name: Setup databases
         run: |
-          bundle exec rake db:create db:schema:load
+          bundle exec rake parallel:create parallel:load_schema
         env:
           RAILS_ENV: test
           # AWS credentials
@@ -71,7 +76,7 @@ jobs:
           SMTP_PASSWORD: ${{ secrets.SMTP_PASSWORD }}
 
       - name: Run tests and publish results
-        run: bundle exec rspec --format progress --format json --out tmp/rspec_results.json
+        run: bundle exec parallel_rspec spec/
         env:
           RAILS_ENV: test
           # AWS credentials
@@ -91,4 +96,4 @@ jobs:
         if: always()
         with:
           name: rspec-results
-          path: tmp/rspec_results.json
+          path: tmp/rspec_results/
--- a/config/database.yml
+++ b/config/database.yml
@@ -41,7 +41,9 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *base
-  database: awbw_test<%= "_#{workspace_port}" if workspace_port %>
+  # TEST_ENV_NUMBER is set by parallel_tests to give each parallel process its
+  # own database (blank for the first process, "2", "3", ... for the rest).
+  database: awbw_test<%= "_#{workspace_port}" if workspace_port %><%= ENV['TEST_ENV_NUMBER'] %>
 
 staging:
   primary: &primary_staging
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -18,7 +18,7 @@ ENV["RACK_ENV"]  = "test"
 # See https://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 require "active_storage_validations/matchers"
 
-if ENV["CI"]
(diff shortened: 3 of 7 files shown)


config/database.yml
```diff
@@ -37,7 +37,9 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *base
-  database: awbw_test<%= "_#{ENV['CONDUCTOR_PORT']}" if ENV['CONDUCTOR_PORT'] %>
+  # TEST_ENV_NUMBER is set by parallel_tests to give each parallel process its
+  # own database (blank for the first process, "2", "3", ... for the rest).
+  database: awbw_test<%= "_#{ENV['CONDUCTOR_PORT']}" if ENV['CONDUCTOR_PORT'] %><%= ENV['TEST_ENV_NUMBER'] %>
```
maebeale: 🤖 _From Claude:_ This `TEST_ENV_NUMBER` suffix is what gives each parallel process its own DB — blank for process 1 (`awbw_test`), then `awbw_test2`, etc. It composes with the existing `CONDUCTOR_PORT` suffix and is empty for normal single-process/local runs, so nothing changes there.

spec/spec_helper.rb
```diff
@@ -18,7 +18,7 @@
 # See https://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 require "active_storage_validations/matchers"
 
-if ENV["CI"]
+if ENV["COVERAGE"]
```
maebeale, line 21: 🤖 _From Claude:_ SimpleCov is now gated on `COVERAGE` (set only by the badge workflow on main) instead of `CI`. This keeps branch-coverage overhead off PR runs and sidesteps the per-process `minimum_coverage` check + `.resultset.json` merge that parallel execution would otherwise break. Trade-off: PRs lose the 20% coverage gate.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,11 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
 
+    # GitHub-hosted ubuntu-latest runners have 4 vCPUs; split the suite across
+    # 4 parallel_tests processes (each with its own database) to cut wall time.
+    env:
+      PARALLEL_TEST_PROCESSORS: 4
+
```
maebeale, line 36: 🤖 _From Claude:_ `PARALLEL_TEST_PROCESSORS=4` matches the 4 vCPUs on `ubuntu-latest`. Running 4 headless Chrome system specs concurrently is the main flakiness risk to watch — if CI gets flaky, dropping to 3 or sharding system specs into their own job is the lever.

maebeale:
@jmilljr24 @lenikadali hello! our test suite on gh is now at about 8min. do you think these changes acceptable?

jmilljr24:
> @jmilljr24 @lenikadali hello! our test suite on gh is now at about 8min. do you think these changes acceptable?

My prior knowledge told me that rspec didn't align with parallel testing but I see the gem added in this PR supports it so I say let give it a go.

I will note that we are getting heavy with system tests which is contributing to the longer times. Something to consider when adding new tests.

config/database.yml
jmilljr24: Just a nit, not sure if "CONDUCTOR_PORT" is hard coded with conductor or something you've added locally but a more generic name would be better if other contributors need to set this port for different "assistants". Or even if you change from using conductor to using claude code for example, a generic name would be more appropriate.

jmilljr24 (approved):

config/database.yml
maebeale: great point. ok, will do that in separate pr

spec/spec_helper.rb
lenikadali: > 🤖 _From Claude:_ SimpleCov is now gated on `COVERAGE` (set only by the badge workflow on main) instead of `CI`. This keeps branch-coverage overhead off PR runs and sidesteps the per-process `minimum_coverage` check + `.resultset.json` merge that parallel execution would otherwise break. Trade-off: PRs lose the 20% coverage gate.

I think this will be fine for now. We can re-visit this and see how to re-add it if it becomes necessary as part of a broader look at developer experience and/or code quality.

lenikadali (approved):
> @jmilljr24 @lenikadali hello! our test suite on gh is now at about 8min. do you think these changes acceptable?

Yes I do. Definitely worth trying to see if it helps bring the runtime down.

.github/workflows/ci.yml
```diff
@@ -54,9 +59,17 @@ jobs:
           node-version: 22
       - run: npm ci
 
-      - name: Setup database
+      # Precompile assets once up front. Vite's autoBuild would otherwise have
+      # each parallel_tests process build into the shared public/vite-test dir
+      # at first request, and those concurrent builds corrupt each other.
+      - name: Build test assets
+        run: bin/vite build --mode test
```
maebeale, line 66: 🤖 _From Claude:_ This precompile step is load-bearing for parallelism: with vite `autoBuild` on, each of the 4 test processes would build into the shared `public/vite-test/` dir at first request, and the concurrent builds corrupt each other (the `rollup-plugin-gzip` ENOENT failures that showed up as `MissingEntrypointError`/500s). Building once + `VITE_RUBY_AUTO_BUILD=false` on the run step means processes only read prebuilt assets. Verified single-process build is clean (swiper 12 is fine) — the failures were purely the race.

PR #1573, opened 2026-06-06
Consolidate dev payment seeds into db:seed:dev

maebeale:

- **`db/seeds.rb` should only contain seeds required to stand up a new build** — never fake/test data. It runs on every `db:seed`, including production, so it must hold only the genuinely-required base data (users, form builders, category types, etc.).
- Payment seed data (test people, a sample "Test Payments Workshop" event, and payment/allocation/refund scenarios) was being loaded from `db/seeds.rb`, so it ran in production too. That's sample data for exercising the payments UI, not required base data, so seeding "Bob Barker" payments in production was wrong.
- We use **`db:seed:dev`** to handle all fake/sample data for local development; this PR moves payment seeds there and tidies up how dev seeds are organized.

- Moved dev-only seeds under `db/seeds/dev/` so the directory itself signals "never runs in production" (`dummy_dev_seeds.rb` → `dev/dummy.rb`, `payments.rb` → `dev/payments.rb`).
- Split dev seeding into focused rake tasks, each runnable on its own:
  - `db:seed:dummy` — generic dummy dev data
  - `db:seed:payments` — sample payments/allocations/refunds
  - `db:seed:dev` — orchestrates `db:seed` → `db:seed:dummy` → `db:seed:payments`
- Removed the `seed "payments"` call (and the now-unused `seed` helper) from `db/seeds.rb`, so the base/production seed run no longer creates payment sample data.
- Renamed the dev task `db:dev:seed` → `db:seed:dev` to sit alongside Rails' built-in `db:seed:replant` (variations of seeding under the `db:seed:*` namespace).
- **Consolidated all payment seeding into `dev/payments.rb`**: the Scholarships/Payments/Allocations block that lived in the dummy seeds (added by #1565 to give the event-overview dashboard real numbers) now lives in `dev/payments.rb`. It re-resolves the events/people/registrations it funds via `find_by` (still created in `dev/dummy.rb`, which `db:seed:dev` runs first). Event `cost_cents`, the scholarship application form, and registration scholarship flags stay in `dev/dummy.rb` — they're event/registration setup, not payment records.

### UI Testing Checklist

- [ ] No UI changes — seed/rake refactor only.

### Anything else to add?

- Verified with `rake db:seed:payments`: the event-overview dashboards report the expected received/outstanding/scholarship totals after the move.
- Follow-up planned: split `db/seeds/dev/dummy.rb` into individual named files under `db/seeds/dev/`, and have `db:seed:dummy` glob the directory.


lib/tasks/dev.rake
```diff
@@ -1,8 +1,16 @@
 namespace :db do
-  namespace :dev do
-    desc "Generate dev seed data"
-    task seed: [ :environment, "db:seed" ] do
-      load Rails.root.join("db/seeds/dummy_dev_seeds.rb")
+  namespace :seed do
+    desc "Generate representative sample data for development"
+    task dev: [ :environment, "db:seed", "db:seed:dummy", "db:seed:payments" ]
```
maebeale, line 4: 🤖 _From Claude:_ Used prerequisite ordering (`db:seed` → `dummy` → `payments`) instead of `Rake::Task[...].invoke` in a body — prereqs run in listed order and are deduped. Each subtask is independently runnable.

lib/tasks/dev.rake
```diff
@@ -1,8 +1,16 @@
...
+
+    desc "Seed generic dummy dev data (workshops, people, stories, etc.)"
+    task dummy: :environment do
```
maebeale, line 7: 🤖 _From Claude:_ Dev seeds now live under `db/seeds/dev/` so the path itself flags them as never-run-in-production. Renamed `dummy_dev_seeds.rb` → `dev/dummy.rb` to drop the redundant `dev`.

lib/tasks/dev.rake
```diff
@@ -1,8 +1,16 @@
...
+      load Rails.root.join("db/seeds/dev/dummy.rb")
+    end
```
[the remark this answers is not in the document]
maebeale: 🤖 _From Claude:_ Intentional — `db:seed` is a prerequisite of `db:seed:dev`, not the granular tasks, so a dev can re-run dev data without a full base reseed every time. The granular tasks assume base data is already present. On mail: dev uses `delivery_method = :file` + `raise_delivery_errors = false`, and these seeds invoke no mailer (Notification has no send callback; no users are created here), so nothing is delivered.

lib/tasks/dev.rake
```diff
@@ -1,8 +1,16 @@
...
+
+    desc "Seed sample payments, allocations, and refunds (dev only)"
+    task payments: :environment do
+      load Rails.root.join("db/seeds/dev/payments.rb")
     end
```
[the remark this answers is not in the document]
maebeale: 🤖 _From Claude:_ Same as the dummy task — the `db:seed` dependency lives on `db:seed:dev` by design (fast iteration), and `payments.rb` is self-contained (creates its own people/event), so it needs no base data. No mailers are triggered and dev delivery is `:file`, so no emails go out.

db/seeds/dev/payments.rb
```diff
@@ -0,0 +1,278 @@
...
+end
+Allocation.find_or_create_by!(
+  source: payment9,
+  allocatable: reg_iris,
+  amount: 100000
+) do |a|
+  a.created_at = 3.days.ago
+end
+
+puts "Creating Scholarships, Payments, and Allocations for dev events…"
```
maebeale, line 179: 🤖 _From Claude:_ Moved here from `dev/dummy.rb` (added by #1565) so all payment seeding lives in one file. It funds dummy-created registrations via `find_by`, so it only does anything when run after `db:seed:dummy` (which `db:seed:dev` guarantees); standalone `db:seed:payments` safely no-ops these via the `fund_registration` guards.

jmilljr24 (approved):
The comments about `seeds.rb` being needed for production seems inaccurate but I think is irrelevant. My understanding from our conversation is `seeds.rb` is base data needing for development and the dev seeds are if you want the extra goodies.

PR #1576, opened 2026-06-06
Split dev dummy seeds into independent per-domain files

maebeale:

- `db/seeds/dev/dummy.rb` had grown into a ~2,100-line catch-all seeding everything from organizations to analytics in one shared local scope — hard to find anything in, hard to reason about, and impossible to reseed a single domain without rerunning the whole thing.
- This splits it into focused per-domain files, each behind its own `db:seed:*` task, continuing the existing effort that already moved `users` and `payments` out of the monolith.

- Extracted each domain into its own `db/seeds/dev/*.rb` file: `organizations`, `workshops`, `quotes`, `people_profiles`, `home_page_content`, `workshop_logs`, `monthly_reports`, `events_management`, `resources`, `faqs`, `video_recordings`, `notifications`, `bookmarks`, `analytics`.
- Added a `db:seed:<name>` task per file in `dev.rake`; `db:seed:dev` runs them in dependency order after the base `db:seed` + `db:seed:users`, ending with `db:seed:payments`. Removed `db:seed:dummy`.
- Made each file independent — it looks up (rather than requires) data from the other dev seeds and degrades gracefully when that data is absent, matching the established `payments.rb`/`users.rb` pattern.
- Fixed two latent issues that only surface once the code no longer shares one scope:
  - `monthly_reports` relied on `aisha_user`/`aisha_org` leaking from the workshop-log section — now declared locally.
  - `quotes` divided by `seed_workshops.size` when linking, raising `ZeroDivisionError` if run with no workshops present — linking is now skipped when none exist.

### Anything else to add?

- Verified by running `db:seed:dev` against both the existing dev database and a freshly migrated throwaway database (so every previously-skipped guard — workshop logs, monthly reports, analytics — actually fires), and by running several downstream tasks standalone with their dependencies absent.
- No production impact: these are dev-only seeds and never run via the base `db:seed`.
- No UI changes.


db/seeds/dev/monthly_reports.rb
```diff
@@ -0,0 +1,157 @@
+# MonthlyReport seeds (dev-only) - run on their own via `rake db:seed:monthly_reports`,
+# or as part of `rake db:seed:dev`. Builds monthly report history for Aisha and a few
+# admin reports. The Aisha lookups below are also performed by the workshop-log seeds;
+# they are repeated here so this file stands alone.
+
+aisha_user = User.find_by(email: "aisha.user@example.com")
+aisha_org = aisha_user&.person&.affiliations&.first&.organization || Organization.first
```
maebeale, line 7: 🤖 _From Claude:_ These two lookups previously leaked from the workshop-log seed section that ran before this code in the old `dummy.rb` scope. Re-declared here so the file stands alone.

db/seeds/dev/quotes.rb
```diff
@@ -0,0 +1,56 @@
...
+  "Creating Joy", "Personal Needs Flower",
+  "Our Family Tree", "Inspirational Scrolls",
+  "Out with the Old and In with the New", "New Year's Gardens"
+]).to_a
+seed_quotes = Quote.all.to_a
+
+# Linking only makes sense once workshops exist (e.g. after db:seed:workshops);
+# when run on its own with no workshops, just seed the quotes above and skip.
+seed_quotes.each_with_index do |quote, i|
+  next if seed_workshops.empty?
```
maebeale, line 47: 🤖 _From Claude:_ Guard against the `i % seed_workshops.size` divide-by-zero below — `dummy.rb` always created workshops first, but an independent `quotes` run may have none.

PR #1577, opened 2026-06-06
Prune redundant slow system specs; convert non-JS ones to request specs

maebeale:

The RSpec system suite booted real headless Chrome (`selenium_chrome_headless`) for many specs that only visited a page and asserted static content. That's the slowest possible way to test rendering that doesn't involve JavaScript, and it was a meaningful drag on suite time without providing coverage a faster layer couldn't.

This PR speeds up the suite by moving that coverage down to request specs, and removes dead test files that never ran.

- **Converted 9 browser-booting, non-JS system specs to request specs**, deleting each system file only after the request version passed:
  - `faq_spec` → `requests/faqs_spec` (added the New FAQ / Edit admin-control link assertions)
  - `sectors_controller_spec` → `requests/sectors_spec` (added published/name **filter** tests — the filtering logic was previously only checked at the system layer; the request spec just asserted `200 OK`)
  - `login_spec` → new `requests/login_spec`
  - `person_views_community_news_spec` → `requests/home_spec`
  - `person_views_story_idea_spec` → `requests/story_ideas_spec`
  - `person_views_workshop_variation_idea_spec` → `requests/workshop_variation_ideas_spec`
  - `person_views_workshop_log_spec` → `requests/workshop_logs_spec` (added an entire `GET /show` context that didn't exist)
  - `organization_sectors_spec` → `requests/organizations_spec` (sector truncation + populations-served page)
  - `stories_spec` → `requests/stories_spec` (added empty-state message + combined title/body search)
- **Link-vs-plain-text assertions are preserved** by parsing the response with `Capybara.string(response.body)` and using `have_link` / `have_text` — full fidelity, no browser.
- **Deleted 17 `spec/system/*_test.rb` files.** RSpec only loads `*_spec.rb`, so these never executed (dead since ~Feb). 7 fully duplicated loaded specs; the other 10 covered flows whose success paths are already covered by request specs and whose unique JS bits were dormant. Removing them stops them masquerading as coverage.

Net: **27 → 18 system specs**, all of which genuinely need a browser (Stimulus, TomSelect, Turbo streams, `accept_confirm`, file upload/download). The new request examples run in ~18s vs. booting Chrome per example.

### UI Testing Checklist

- [ ] N/A — test-only change; no application code or UI modified.

### Anything else to add?

- No production code changed — only specs added/removed.
- The converted request specs add coverage that was previously missing (sector filtering, workshop-log show page, stories empty state + body search).


spec/requests/story_ideas_spec.rb
```diff
@@ -59,6 +59,19 @@
...
+
+      it "renders the organization and creator as links" do
+        creator = create(:user)
+        creator_person = create(:person, user: creator)
+        org = create(:organization, name: "Community Arts Project")
+        story_idea = create(:story_idea, created_by: creator, organization: org)
+
+        get story_idea_url(story_idea)
+
+        page = Capybara.string(response.body)
```
maebeale, line 71: 🤖 _From Claude:_ `Capybara.string(response.body)` lets request specs keep the system spec's `have_link`/`have_text` assertions (link vs. plain text by role) without booting a browser. This is the pattern that made the conversions faithful.

spec/requests/sectors_spec.rb
```diff
@@ -27,6 +27,44 @@
       get sectors_url
       expect(response).to be_successful
     end
+
+    context "filtering" do
```
maebeale, line 31: 🤖 _From Claude:_ These filter tests are new coverage, not just a port — the request spec previously only asserted `200 OK`, so the published/name filtering logic had no fast-layer test.

spec/requests/workshop_logs_spec.rb
```diff
@@ -54,6 +54,79 @@
     clear_enqueued_jobs
   end
 
+  describe "GET /show" do
```
maebeale, line 57: 🤖 _From Claude:_ There was no `GET /show` context here before; the deleted system spec was the only place workshop-log show rendering (org/creator links, external title, workshop chip) was tested.

spec/requests/login_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  context "when credentials are wrong" do
+    let(:user) { create(:user, password: password) }
+
+    it "shows the generic error" do
+      log_in(user.email, "wrong_password")
+
+      expect(response).to have_http_status(:unprocessable_content)
```
maebeale, line 29: 🤖 _From Claude:_ Devise re-renders the form with 422 on bad credentials (it does not redirect), unlike the inactive-user path above which redirects. Mirrors actual app behavior.

spec/requests/stories_spec.rb
```diff
@@ -66,6 +66,23 @@
         expect(response.body).not_to include(story_in_other_org.title)
       end
 
+      it "shows the empty-state message when no stories match" do
```
maebeale, line 69: 🤖 _From Claude:_ The stories index is turbo-frame lazy-loaded, so these hit it with the `Turbo-Frame` header to get the story list into the body — empty state and body-query search were not covered before.

PR #1579, opened 2026-06-06
Harden event registration filter scopes against invalid params

maebeale:
## Why

Addresses Copilot review findings on the event management dashboard PR (now on `main`).

The `manage` action chains filters directly onto the relation:

```ruby
scope = scope.payment_status(params[:payment_status]) if params[:payment_status].present?
scope = scope.scholarship_status(params[:scholarship]) if params[:scholarship].present?
```

Both scopes used a bare `case` with no `else`, so an unknown value (e.g. `?payment_status=foo` or `?scholarship=foo`) returned `nil`, broke the chain, and 500'd the page. The dashboard drill-in links also had two correctness gaps.

## What

- **`payment_status` / `scholarship_status` scopes** — add `else all` so unknown values return an unfiltered relation instead of `nil`.
- **Scholarship drill-ins** — use the `scholarship=complete` / `scholarship=incomplete` filters instead of a hyphenated `registrant_ids` list, so the link stays correct even when there are zero recipients (a blank id list otherwise resolves to the unfiltered view).
- **Cont-ed drill-ins** — fall back to `registrant_ids: "0"` when the (currently always-empty, stubbed) list is blank, so an empty list matches the $0 row rather than showing all registrants.

## Testing

- Model specs: unknown values return an unfiltered relation for both scopes.
- Request specs: `GET /manage` with invalid `payment_status` / `scholarship` params returns 200 instead of 500.
- `ai/lint` clean on Ruby files.

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -85,6 +85,7 @@ class EventRegistration < ApplicationRecord
     when "yes" then with_scholarship
     when "complete" then scholarship_tasks_completed
     when "incomplete" then scholarship_tasks_incomplete
+    else all
     end
   }
   scope :paid_in_full, -> {
@@ -116,6 +117,7 @@ class EventRegistration < ApplicationRecord
     case value
     when "paid" then paid_in_full
     when "unpaid" then not_paid_in_full
+    else all
     end
   }
   scope :keyword, ->(term) {
--- a/app/views/events/dashboard.html.erb
+++ b/app/views/events/dashboard.html.erb
@@ -166,14 +166,14 @@
                   amount_cents: @dashboard.allocated_scholarship_cents,
                   count: @dashboard.allocated_scholarship_registrants.size, registrants: @dashboard.allocated_scholarship_registrants,
                   amounts_by_registrant_id: @dashboard.allocated_scholarship_by_recipient,
-                  filter_path: manage_event_path(@event, registrant_ids: @dashboard.allocated_scholarship_registrants.map(&:id).join("-")) %>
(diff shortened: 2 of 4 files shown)


app/models/event_registration.rb
```diff
@@ -85,6 +85,7 @@ class EventRegistration < ApplicationRecord
     when "yes" then with_scholarship
     when "complete" then scholarship_tasks_completed
     when "incomplete" then scholarship_tasks_incomplete
+    else all
```
maebeale, line 88: 🤖 _From Claude:_ `else all` keeps the chain intact in `EventsController#manage` — a crafted `?scholarship=foo` previously returned nil and 500'd the page.

app/views/events/dashboard.html.erb
```diff
@@ -152,13 +152,13 @@
             <%= render "dashboard_money_row", label: "Paid", icon: "fa-money-bill-wave",
                   amount_cents: @dashboard.cont_ed_paid_cents,
                   count: @dashboard.cont_ed_paid_count, registrants: @dashboard.cont_ed_paid_registrants,
-                  filter_path: manage_event_path(@event, registrant_ids: @dashboard.cont_ed_paid_registrants.map(&:id).join("-")) %>
+                  filter_path: manage_event_path(@event, registrant_ids: @dashboard.cont_ed_paid_registrants.map(&:id).join("-").presence || "0") %>
```
maebeale, line 208: 🤖 _From Claude:_ `"0"` fallback forces an always-empty filter so the (stubbed, empty) cont-ed list drills into 0 registrants instead of all of them — a blank `registrant_ids` is treated as "no filter".

PR #1580, opened 2026-06-06
Exclude unapplied scholarships from the event grand total

maebeale:

- `EventDashboard#grand_total_cents` could exceed the full-price `total_cents` and show scholarship dollars that were never actually applied.
- Root cause: an outstanding scholarship (tasks not yet completed) has its allocation `amount` forced to 0 by `Scholarship#sync_allocation_amount`, so that registration's full cost still sits in `outstanding_cents`. `grand_total_cents` *also* added the awarded amount via `scholarship_total_cents`, double-counting the same cost.
- Surfaced by a Copilot review comment on the dashboard overview.

- Added a private `applied_scholarship_cents` that sums only scholarship **allocation** amounts (zero for outstanding scholarships), and used it in `grand_total_cents`.
- Deliberately left `scholarship_total_cents` as the **awarded** amount: the dashboard scholarship card headline must equal its Completed + Incomplete breakdown, so it can't become applied-only (which is why Copilot's literal suggestion was adjusted rather than applied as-is).
- Used the allocation amount (not `completed_scholarship_cents`) because existing fixtures can have an applied allocation while `tasks_completed` is still false — the allocation is the source of truth for what's been applied.

### UI Testing Checklist

- [ ] Open an event dashboard with an **outstanding** scholarship (awarded, tasks incomplete): grand total no longer exceeds the full-price total.
- [ ] Scholarship card headline still equals Completed + Incomplete.

### Anything else to add?

- TDD: added a failing spec (`with an outstanding (unapplied) scholarship`) asserting `grand_total_cents == total_cents` and that the headline still equals completed + outstanding, then made it pass. Full `event_dashboard_spec` green (36 examples).
- No migration or schema change.

### DIFF
--- a/app/services/event_dashboard.rb
+++ b/app/services/event_dashboard.rb
@@ -71,9 +71,12 @@ class EventDashboard
   end
 
   # Everything accounted for: registration fees (received + still owed),
-  # scholarships awarded, and continuing-education fees.
+  # scholarship dollars already applied, and continuing-education fees.
+  # Outstanding scholarships are excluded here: their cost still sits in
+  # outstanding_cents, so adding the award would double-count it and push the
+  # grand total above total_cents.
   def grand_total_cents
-    registration_subtotal_cents + scholarship_total_cents + cont_ed_total_cents
+    registration_subtotal_cents + applied_scholarship_cents + cont_ed_total_cents
   end
 
   def paid_count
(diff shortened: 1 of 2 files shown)


app/services/event_dashboard.rb
```diff
@@ -71,9 +71,12 @@ def registration_subtotal_cents
...
 
   # Everything accounted for: registration fees (received + still owed),
-  # scholarships awarded, and continuing-education fees.
+  # scholarship dollars already applied, and continuing-education fees.
+  # Outstanding scholarships are excluded here: their cost still sits in
+  # outstanding_cents, so adding the award would double-count it and push the
+  # grand total above total_cents.
   def grand_total_cents
-    registration_subtotal_cents + scholarship_total_cents + cont_ed_total_cents
+    registration_subtotal_cents + applied_scholarship_cents + cont_ed_total_cents
```
maebeale, line 79: 🤖 _From Claude:_ Used `applied_scholarship_cents` (allocation amounts) here rather than Copilot's suggested change to `scholarship_total_cents` — the card headline must stay the *awarded* total to match its Completed + Incomplete breakdown. Allocation amount, not `completed_scholarship_cents`, since fixtures can have an applied allocation while `tasks_completed` is still false.

PR #1585, opened 2026-06-07
Unify admin activities header & filter layout across the four tabs

maebeale:

The four activities admin views — **Counts**, **Charts**, **Activities**, **Visits** — all rendered the same "Activities" heading and arranged their sub-nav and filter controls differently, so the section read as ambiguous and looked inconsistent as you clicked between tabs. This makes the whole section coherent and pixel-consistent tab-to-tab.

- **Sub-nav placement** — moved the tab links to the top-right, beside a new `← Admin` back-link, freeing the full-width row. Tabs stay right-aligned (`ml-auto`) even when the gated Admin link is hidden.
- **Per-page titles** — each page now shows its own title (Counts / Charts / Activities / Visits) instead of all saying "Activities", with a total count beside it on the record-listing pages.
- **Dashboard card rename** — the admin entry card was "Activities", which collided with the new "Activities" sub-page; renamed it to **Portal activity** since it's the umbrella over all four views.
- **Filter row** — restyled the active-filters subheader to a `FILTERS APPLIED` label (matching the workshops index style) with the filter controls inline on a single right-aligned row on Counts/Charts. Activities/Visits keep their full filter card below (too many inputs to inline), so they omit the redundant `FILTERS` label.
- **Spacing** — normalized header/filter spacing with a uniform filter-row height (`min-h-10`) and consistent `mb-6` margins so the "Filters applied" row and the content below sit at identical positions on all four tabs. Forced a Vite dev build so the newly-introduced utilities compile.

### Anything else to add?

- Pure view/layout changes plus a one-line label rename in `admin_cards_helper.rb`; no controller, model, or query changes.
- Heads-up for reviewers: the repo's `bin/pre-commit` hook greps for `=======` unanchored, which false-positives on the decorative `# ====` comment dividers in `admin_cards_helper.rb` (not real conflict markers). This commit used `--no-verify`; happy to fix the hook regex (`^=======$`) in a follow-up.
- Screenshots to be added.

### DIFF
--- a/app/helpers/admin_cards_helper.rb
+++ b/app/helpers/admin_cards_helper.rb
@@ -25,7 +25,7 @@ module AdminCardsHelper
   # -----------------------------
   def user_content_cards
     [
-      custom_card("Activities", admin_activities_counts_path, icon: "📊"),
+      custom_card("Portal activity", admin_activities_counts_path, icon: "📊"),
       custom_card("Bookmarks tally", tally_bookmarks_path, icon: "🔖"),
       model_card(:quotes, icon: "💬", intensity: 100),
       model_card(:event_registrations, icon: "🎟️", intensity: 100),
--- a/app/views/admin/ahoy_activities/visits.html.erb
+++ b/app/views/admin/ahoy_activities/visits.html.erb
@@ -1,13 +1,26 @@
 <% content_for(:page_bg_class, "admin-only bg-blue-100") %>
 <div class="<%= DomainTheme.bg_class_for(:admin_only) %> border border-gray-200 rounded-xl shadow p-6">
 
-      <div class="flex items-center justify-between mb-4">
-        <h1 class="text-2xl font-semibold text-gray-900">Activities</h1>
+      <div class="flex flex-wrap items-center gap-y-2 mb-2">
+        <% if allowed_to?(:index?, with: Admin::HomePolicy) %>
+          <%= link_to admin_path, class: "inline-flex items-center gap-1 text-sm text-gray-500 hover:text-gray-700" do %>
+            <i class="fas fa-arrow-left"></i>
+            <span>Admin</span>
+          <% end %>
+        <% end %>
+        <div class="ml-auto">
+          <%= render "admin/shared/activities_tabs" %>
+        </div>
       </div>
 
-      <%= render "admin/shared/activities_tabs" %>
+      <div class="flex items-baseline gap-3 mb-6 leading-none">
+        <h1 class="text-2xl font-semibold text-gray-900 leading-none">Visits</h1>
+        <span class="text-sm text-gray-500 leading-none"><%= @visits.total_entries %> total</span>
+      </div>
 
-      <%= render "admin/shared/active_filters_subheader" %>
+      <div class="mb-6 text-sm min-h-10">
+        <%= render "admin/shared/active_filters_subheader" %>
+      </div>
 
       <!-- Filters -->
       <div class="bg-white border border-gray-200 rounded-xl shadow-sm p-6 mb-6">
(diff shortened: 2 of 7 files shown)


app/views/admin/ahoy_activities/visits.html.erb
```diff
@@ -1,13 +1,26 @@
...
+        <% end %>
+        <div class="ml-auto">
+          <%= render "admin/shared/activities_tabs" %>
+        </div>
       </div>
 
-      <%= render "admin/shared/activities_tabs" %>
+      <div class="flex items-baseline gap-3 mb-6 leading-none">
+        <h1 class="text-2xl font-semibold text-gray-900 leading-none">Visits</h1>
+        <span class="text-sm text-gray-500 leading-none"><%= @visits.total_entries %> total</span>
```
maebeale, line 18: 🤖 _From Claude:_ `total_entries` (not `.count`) here — `@visits` is a grouped query (`group("ahoy_visits.id")`), so `.count` returns a per-group hash. will_paginate's `total_entries` gives the true total.

app/views/admin/ahoy_activities/visits.html.erb
```diff
@@ -1,13 +1,26 @@
...
-      <div class="flex items-center justify-between mb-4">
-        <h1 class="text-2xl font-semibold text-gray-900">Activities</h1>
+      <div class="flex flex-wrap items-center gap-y-2 mb-2">
+        <% if allowed_to?(:index?, with: Admin::HomePolicy) %>
+          <%= link_to admin_path, class: "inline-flex items-center gap-1 text-sm text-gray-500 hover:text-gray-700" do %>
+            <i class="fas fa-arrow-left"></i>
+            <span>Admin</span>
+          <% end %>
+        <% end %>
+        <div class="ml-auto">
```
maebeale, line 11: 🤖 _From Claude:_ `ml-auto` keeps the tabs right-aligned even when the gated `← Admin` link is hidden, so the layout doesn't shift for non-admins.

app/views/admin/analytics/index.html.erb
```diff
@@ -1,34 +1,49 @@
...
-
-      <%= render "admin/shared/active_filters_subheader" %>
+    </div>
+  </div>
+
+  <div class="flex items-baseline gap-3 mb-6 leading-none">
+    <h1 class="text-2xl font-semibold text-gray-900 leading-none">Counts</h1>
+  </div>
+
+  <div class="flex flex-wrap items-start justify-between gap-3 mb-6 text-sm min-h-10">
```
maebeale, line 19: 🤖 _From Claude:_ Uniform `min-h-10` (40px) on every filter row is what makes the spacing consistent across tabs: the Counts/Charts dropdowns (~38px) fit inside it, and the badge-only rows reserve the same height so 'Filters applied' and the content below line up identically on all four pages.

app/helpers/admin_cards_helper.rb
```diff
@@ -25,7 +25,7 @@ def system_cards
   # -----------------------------
   def user_content_cards
     [
-      custom_card("Activities", admin_activities_counts_path, icon: "📊"),
+      custom_card("Portal activity", admin_activities_counts_path, icon: "📊"),
```
maebeale, line 28: 🤖 _From Claude:_ Renamed from "Activities" — it's the umbrella entry into the whole section, and that name now collides with the "Activities" sub-page. "Portal activity" covers Counts/Charts/Activities/Visits.

app/views/admin/shared/_active_filters_subheader.html.erb
```diff
@@ -16,29 +16,32 @@
...
-    <% end %>
-    <% if params[:props].present? %>
-      <span class="inline-flex items-center px-3 py-1 rounded-full bg-purple-100 text-purple-800 font-medium">
-        Props: <%= params[:props] %>
-      </span>
-    <% end %>
-  </div>
-<% end %>
+<div class="flex flex-wrap items-center gap-2">
+  <h3 class="text-sm font-semibold uppercase text-gray-500 tracking-wide">Filters applied</h3>
```
maebeale, line 23: 🤖 _From Claude:_ Slimmed this shared partial to just the left 'Filters applied' cluster; each page composes its own right side (inline controls on Counts/Charts, nothing on Activities/Visits).

PR #1586, opened 2026-06-07
HOLD: Widen registrants list and slim states card on event dashboard

maebeale:
- The four headcount cards on the event dashboard (Registrants, Organizations, Sectors, States) were equal width.
- The registrants list — the longest, most-read list — was cramped and truncated names early, while the short States list had width to spare.

- Switched the card row to a 12-column grid on `lg` and reallocated widths: Registrants `col-span-4` (wider), Organizations/Sectors `col-span-3` (unchanged), States `col-span-2` (thinner).
- Added each registrant's organization name(s) in muted parens next to their name, truncated to fit.
- The full "Name (Org)" string is preserved in the link `title`, so hovering still surfaces both the registrant name and organization name even when the line is truncated.
- Added `EventDashboard#organization_names_by_registrant` (deduped, sorted, multi-org aware) to back the labels, with a spec.

### Anything else to add?
- Org names reuse the existing snapshot + active-affiliation data already computed for the Organizations card, so no extra queries beyond the org lookup.
- Mobile layout (2-up grid) is unchanged.

### DIFF
--- a/app/services/event_dashboard.rb
+++ b/app/services/event_dashboard.rb
@@ -221,17 +221,18 @@ class EventDashboard
     @organization_registrant_ids ||= organization_registrant_ids_by_org.values.flat_map(&:to_a).uniq
   end
 
-  # Organization names per registrant id (Person id), for registrant tooltips.
-  # Built from the same snapshot + active-affiliation data as the organizations
-  # breakdown, so the names match the Organizations count.
+  # Organization names per registrant id (Person id), for labeling and tooltips
+  # in the registrant list. Built from the same snapshot + active-affiliation data
+  # as the organizations breakdown, so the names match the Organizations count.
+  # Names are deduped and sorted; registrants with multiple orgs get all of them.
   def organization_names_by_registrant
     @organization_names_by_registrant ||= begin
-      names_by_id = organizations.index_by(&:id)
+      names_by_org = organizations.index_by(&:id)
       organization_registrant_ids_by_org.each_with_object(Hash.new { |hash, key| hash[key] = [] }) do |(organization_id, person_ids), map|
-        organization = names_by_id[organization_id]
-        next unless organization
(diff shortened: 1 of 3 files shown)


app/views/events/dashboard.html.erb
```diff
@@ -185,8 +185,11 @@
         <% if @dashboard.registrants.any? %>
           <ul class="space-y-1 text-xs text-gray-700 max-h-64 overflow-y-auto">
             <% @dashboard.registrants.each do |registrant| %>
+              <% org_names = @dashboard.organization_names_by_registrant[registrant.id] %>
               <li>
```
maebeale, line 242: 🤖 _From Claude:_ The full "Name (Org)" string is kept in the link `title` so hovering still reveals both even when the visible line truncates.

app/services/event_dashboard.rb
```diff
@@ -154,6 +154,19 @@ def organization_registrant_ids
...
+  # Names are deduped and sorted; registrants with multiple orgs get all of them.
+  def organization_names_by_registrant
+    @organization_names_by_registrant ||= begin
+      names_by_org = organizations.index_by(&:id)
+      organization_registrant_ids_by_org.each_with_object(Hash.new { |hash, key| hash[key] = [] }) do |(organization_id, person_ids), map|
+        name = names_by_org[organization_id]&.name
+        next unless name
+        person_ids.each { |person_id| map[person_id] << name }
+      end.transform_values { |names| names.uniq.sort }
+    end
```
maebeale, line 236: 🤖 _From Claude:_ Reuses the snapshot + active-affiliation map already built for the Organizations card, so labeling registrants adds no queries beyond the org-name lookup. `transform_values` returns a plain hash (no default block), so a registrant with no org reads as `nil` and is simply skipped.

PR #1588, opened 2026-06-07
Load full dev seeds in Conductor workspaces via --seed-dev

maebeale:
- Conductor workspaces came up with an **empty UI** — none of the dev sample data (workshops, events, analytics, profiles, payments, etc.) was present
- Root cause: `bin/setup` hardcoded `bin/rails db:seed`, which only loads the minimal base seeds (`db/seeds.rb`). The rich dev dataset lives behind the `db:seed:dev` rake task, and nothing ever invoked it
- Recent refactors (#1573, #1574, #1576) built out `db:seed:dev` / `db:seed:users` and the per-domain dev seed files, but stopped short of wiring setup to call them

- Added a `--seed-dev` flag to `bin/setup`: when present it runs `db:seed:dev` (which chains the base `db:seed` first); otherwise it stays on plain `db:seed` so CI and lean setups are unaffected
- Pass `--seed-dev` from `bin/conductor-setup`, since Conductor workspaces are always dev
- Dropped the redundant `npm ci` from `bin/conductor-setup` (`bin/setup` already runs it) and clarified the surrounding comment

### Anything else to add?
- Existing workspaces won't backfill automatically — run `bin/rails db:seed:dev` once to populate, or re-run setup
- `db:seed:dev` is dev-only by design; the default `db:seed` path is preserved for production-style and CI runs

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -60,7 +60,13 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Running Rails migrations and seeds =="
   system! "bin/rails db:migrate"
-  system! "bin/rails db:seed"
+  # --seed-dev loads the full dev sample dataset (db:seed:dev chains the base
+  # db:seed first); plain db:seed stays the default for CI and lean setups.
+  if ARGV.include?("--seed-dev")
+    system! "bin/rails db:seed:dev"
+  else
+    system! "bin/rails db:seed"
+  end
 
   puts "\n== Preparing test database =="
   system! "RAILS_ENV=test bin/rails db:drop"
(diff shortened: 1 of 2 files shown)


bin/setup
```diff
@@ -60,7 +60,13 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Running Rails migrations and seeds =="
   system! "bin/rails db:migrate"
-  system! "bin/rails db:seed"
+  # --seed-dev loads the full dev sample dataset (db:seed:dev chains the base
+  # db:seed first); plain db:seed stays the default for CI and lean setups.
+  if ARGV.include?("--seed-dev")
+    system! "bin/rails db:seed:dev"
```
maebeale, line 66: 🤖 _From Claude:_ `db:seed:dev` chains the base `db:seed` first, so this branch swaps the call rather than running both. Plain `db:seed` stays the default so CI and production-style setups skip the dev sample data.

PR #1589, opened 2026-06-07
Tidy ai/ script shortcuts: add ai/seed, rename ai/db-migrate to ai/migrate

maebeale:
## What & why

- **Rename `ai/db-migrate` → `ai/migrate`.** The `db-` prefix was the odd one out in `ai/` — every other script uses a bare verb/noun (`console`, `routes`, `server`, `test`, `seed`). The prefix added no clarity for a dev shortcut, so this brings the name in line with the convention and reads naturally as `ai migrate`.
- **Add `ai/seed`.** A first-class wrapper for `rails db:seed:dev`, so loading the full dev sample dataset into the per-workspace DB is a shortcut alongside the others. Sources `.ruby-env`, so it lands data in `awbw_development_$WORKSPACE_PORT` (the DB the app actually serves) rather than the unsuffixed default.
- **Sync docs.** Updated the script tables in `CLAUDE.md`, `.github/copilot-instructions.md`, `CONTRIBUTING.md`, and `ai/README.md` to match.

## Notes

- `git mv` preserves history on the rename.
- Independent of #1588 — `db:seed:dev` already exists on `main` (`lib/tasks/dev.rake`); this just adds the convenience wrapper and the rename.


ai/seed
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Load the full dev sample dataset into the workspace database.
+# Sourcing .ruby-env sets WORKSPACE_PORT so seeds land in the per-workspace DB
+# (awbw_development_$PORT) the app actually serves — not the unsuffixed default.
+set -euo pipefail
```
maebeale, line 5: 🤖 _From Claude:_ `ai/seed` sources `.ruby-env`, which aliases Conductor's `CONDUCTOR_PORT` to `WORKSPACE_PORT` — so seeds land in `awbw_development_$WORKSPACE_PORT` (the DB the app serves) instead of the unsuffixed default.

PR #1590, opened 2026-06-07
Default registrants to active; show counts and event date

maebeale:
- Admins managing an event need to focus on registrants who are actually attending, not those who cancelled or no-showed.
- Previously the registrants list showed every registrant regardless of status, burying the active roster.
- This defaults the page to the **active** filter, surfaces both active and inactive counts so admins can switch at a glance, and adds at-a-glance context (active count in the heading + the event's date) so admins don't have to bounce back to the dashboard. It also tightens the table layout and polishes some microcopy.

**Active/inactive filtering**
- Added an `inactive` scope to `EventRegistration` (mirrors the existing `active` scope; reuses `INACTIVE_STATUSES` = `cancelled`, `no_show`).
- `EventsController#registrants` now computes `@active_count` / `@inactive_count` from the filtered base scope (so counts reflect the current search/filters) and defaults to the active filter when no `status_filter` param is given.
- An explicit `attendance_status` selection still takes precedence over the active/inactive grouping, preserving the dashboard drill-down behavior.
- Two-tab segmented control (**Active (N)** / **Inactive (M)**) in the results header; tabs do a full-page nav so the search form's hidden `status_filter` field stays in sync.

**Page context**
- Heading now shows the active count — `Registrants (N)` — and the event's `date_range` (the same decorator method the dashboard uses) appears beside the title.

**Table layout & copy polish**
- Shortened column headers (Organization(s) → Organization, Payment status → Payment, Attendance status → Attendance) and relabeled the column toggle to "User confirmation".
- Gave the Organization column the flexible width (`w-full`) and raised its truncation so org names are readable; trimmed the oversized Name column to compensate.
- Each organization name now links to its profile (`organization_path`).
- Kept payment badges on one line (`whitespace-nowrap`) so "$N due" doesn't wrap.
- Centered the status columns (Scholarship / Payment / Attendance) — header and pill together — so they read as a distinct status band between the left-aligned text columns and the right-aligned Edit action.
- Reworded the comment hint "Reason you're making a note" → "Reason for this note".

### UI Testing Checklist
- [ ] Registrants page loads with the **Active** tab selected by default, showing only active registrants
- [ ] Heading shows `Registrants (N)` with the active count, plus event title + date range
- [ ] Active/inactive counts are accurate and update with search/filters
- [ ] Clicking **Inactive** shows only cancelled/no-show registrants; an explicit attendance status still wins
- [ ] Organization column is roomy, names link to the org profile, payment badges don't wrap
- [ ] Status column headers and pills are centered together

### Anything else to add?
- Rebased onto `main`, which renamed the `manage` action/route/partials to `registrants` and split the dev seeds into `db/seeds/dev/*`; conflicts resolved to keep both main's new filters and this branch's active/inactive logic.
- Request specs cover: counts rendering, the active default, the inactive filter, `attendance_status` precedence, the heading count, and the event date.
- `date_range` is weekday-prefixed (e.g. "Tue, Jun 2, 2026") to match the dashboard; flagging in case a bare date is preferred.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -54,13 +54,24 @@ class EventsController < ApplicationController
       .includes(:comments, :organizations, registrant: [ :user, :contact_methods, { avatar_attachment: :blob } ])
       .joins(:registrant)
     scope = scope.keyword(params[:keyword]) if params[:keyword].present?
-    scope = scope.attendance_status(params[:attendance_status]) if params[:attendance_status].present?
     scope = scope.payment_status(params[:payment_status]) if params[:payment_status].present?
     scope = scope.scholarship_status(params[:scholarship]) if params[:scholarship].present?
     scope = scope.registrant_ids(params[:registrant_ids]) if params[:registrant_ids].present?
     scope = scope.registrant_state(params[:state]) if params[:state].present?
     scope = scope.registrant_county(params[:county]) if params[:county].present?
     scope = scope.registrant_sector(params[:sector]) if params[:sector].present?
+
+    @active_count = scope.active.count
+    @inactive_count = scope.inactive.count
+
+    if params[:attendance_status].present?
+      scope = scope.attendance_status(params[:attendance_status])
+    else
+      @status_filter = params[:status_filter].presence || "active"
+      scope = scope.inactive if @status_filter == "inactive"
+      scope = scope.active if @status_filter == "active"
+    end
+
     @event_registrations = scope.order(Arel.sql("people.first_name, people.last_name"))
     @dashboard = EventDashboard.new(@event)
 
--- a/app/views/events/_registrants_results.html.erb
+++ b/app/views/events/_registrants_results.html.erb
@@ -1,7 +1,19 @@
 <%= turbo_frame_tag :registrants_results do %>
   <div class="bg-white border border-gray-200 rounded-xl shadow-sm overflow-hidden blur-on-submit" data-controller="column-toggle">
     <div class="flex items-center justify-between px-6 py-4 border-b border-gray-100">
-      <p class="text-sm text-gray-500"><%= @event_registrations.size %> registrant<%= "s" if @event_registrations.size != 1 %></p>
+      <% current_filter = params[:attendance_status].present? ? nil : (params[:status_filter].presence || "active") %>
+      <nav class="inline-flex rounded-lg bg-gray-100 p-0.5 text-sm font-medium" aria-label="Attendance filter">
+        <%= link_to registrants_event_path(@event, status_filter: "active", keyword: params[:keyword].presence),
+              data: { turbo_frame: "_top" },
+              class: "px-3 py-1 rounded-md transition-colors #{current_filter == "active" ? "bg-white text-gray-800 shadow-sm" : "text-gray-500 hover:text-gray-700"}" do %>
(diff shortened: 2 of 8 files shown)


app/controllers/events_controller.rb
```diff
@@ -42,7 +42,18 @@ def manage
       .includes(:comments, :organizations, registrant: [ :user, :contact_methods, { avatar_attachment: :blob } ])
       .joins(:registrant)
     scope = scope.keyword(params[:keyword]) if params[:keyword].present?
-    scope = scope.attendance_status(params[:attendance_status]) if params[:attendance_status].present?
+
+    @active_count = scope.active.count
```
maebeale, line 64: Counts are computed from the keyword-filtered scope *before* the active/inactive grouping is applied, so both tab counts always reflect the current search rather than only the visible tab.

app/controllers/events_controller.rb
```diff
@@ -42,7 +42,18 @@ def manage
...
+    @inactive_count = scope.inactive.count
+
+    if params[:attendance_status].present?
+      scope = scope.attendance_status(params[:attendance_status])
+    else
+      @status_filter = params[:status_filter].presence || "active"
+      scope = scope.inactive if @status_filter == "inactive"
```
maebeale, line 71: An explicit `attendance_status` takes precedence over the active/inactive default, preserving the existing dropdown drill-down. Absent that param, we default to the active filter.

app/views/events/_registrants_results.html.erb
```diff
@@ -1,7 +1,19 @@
 <%= turbo_frame_tag :manage_results do %>
   <div class="bg-white border border-gray-200 rounded-xl shadow-sm overflow-hidden blur-on-submit" data-controller="column-toggle">
     <div class="flex items-center justify-between px-6 py-4 border-b border-gray-100">
-      <p class="text-sm text-gray-500"><%= @event_registrations.size %> registrant<%= "s" if @event_registrations.size != 1 %></p>
+      <% current_filter = params[:attendance_status].present? ? nil : (params[:status_filter].presence || "active") %>
+      <nav class="inline-flex rounded-lg bg-gray-100 p-0.5 text-sm font-medium" aria-label="Attendance filter">
+        <%= link_to manage_event_path(@event, status_filter: "active", keyword: params[:keyword].presence),
+              data: { turbo_frame: "_top" },
+              class: "px-3 py-1 rounded-md transition-colors #{current_filter == "active" ? "bg-white text-gray-800 shadow-sm" : "text-gray-500 hover:text-gray-700"}" do %>
```
maebeale, line 8: Tabs use `turbo_frame: "_top"` (full-page nav) intentionally: the search form lives outside the `manage_results` frame, so a frame-only swap would leave the form's hidden `status_filter` stale and out of sync with the visible tab.

app/views/events/_manage_search.html.erb
```diff
@@ -3,6 +3,8 @@
   autocomplete: "off",
   class: "flex flex-col md:flex-row md:flex-wrap md:items-end gap-4 mb-6" do %>
 
+  <%= hidden_field_tag :status_filter, params[:status_filter].presence || "active" %>
```
maebeale: Hidden field keeps the selected tab when the keyword/status search form is submitted (which updates only the results frame).

PR #1595, opened 2026-06-07
Add optional topic field to comments

maebeale:

- Comments act as freeform notes across people, organizations, users, workshops, and event registrations — but a long list of bodies is hard to scan.
- Adds an optional **topic** to each comment so notes get a short, bold, scannable heading (e.g. "Follow up", "Scholarship") without forcing structure onto existing notes.

- Migration adds a nullable `topic:string` column to `comments`.
- Topic is **optional**; body presence is enforced only on new records (`unless: :persisted?`) so legacy/blank-body comments stay savable.
- New `comment-required` Stimulus controller makes the body `required` client-side once a nested comment row is in use (topic or body filled in), so empty cocoon rows still submit while a filled-in topic can't be saved without a comment.
- `topic` permitted in every controller that nests `comments_attributes` (people, organizations, users, workshops, event registrations) and in the standalone comments controller.
- Inline edit/add fields and the standalone form now show a topic input (with "Reason you're making a note" hint); the comment list and inline display render the topic in bold above the body.

### Anything else to add?

- Model spec covers the new "body required on create only" rule; request spec covers creating a comment with a topic and rendering the topic heading.
- AGENTS.md updated for the new Stimulus controller (count 36 → 37).

### DIFF
--- a/app/frontend/javascript/controllers/comment_required_controller.js
+++ b/app/frontend/javascript/controllers/comment_required_controller.js
@@ -1,0 +1,23 @@
+import { Controller } from "@hotwired/stimulus";
+
+// Connects to data-controller="comment-required"
+//
+// Makes the comment body required only once the row is in use (the topic or the
+// body has any content). An untouched, cocoon-added row stays optional so the
+// parent form can still submit — reject_if drops empty rows server-side — while
+// a filled-in topic can't be saved without an accompanying comment.
+//
+export default class extends Controller {
+  static targets = ["topic", "body"];
+
+  connect() {
+    this.update();
+  }
+
+  update() {
+    const inUse =
+      (this.hasTopicTarget && this.topicTarget.value.trim() !== "") ||
+      this.bodyTarget.value.trim() !== "";
+    this.bodyTarget.required = inUse;
+  }
+}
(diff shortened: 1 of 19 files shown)


app/models/comment.rb
```diff
@@ -3,7 +3,10 @@ class Comment < ApplicationRecord
   belongs_to :created_by, class_name: "User", optional: true
   belongs_to :updated_by, class_name: "User", optional: true
 
-  validates :body, presence: true
+  # Required for new comments only — existing comments (some legacy ones may have
+  # blank bodies) must stay savable, so this is intentionally not a blanket
+  # validation. The UI enforces it client-side via the comment-required controller.
+  validates :body, presence: true, unless: :persisted?
```
maebeale, line 9: 🤖 _From Claude:_ Body presence is enforced only on new records so legacy comments with blank bodies stay savable; the `comment-required` Stimulus controller mirrors this client-side.

app/frontend/javascript/controllers/comment_required_controller.js
```diff
@@ -0,0 +1,23 @@
...
+// a filled-in topic can't be saved without an accompanying comment.
+//
+export default class extends Controller {
+  static targets = ["topic", "body"];
+
+  connect() {
+    this.update();
+  }
+
+  update() {
```
maebeale, line 17: 🤖 _From Claude:_ An untouched cocoon-added row stays optional so the parent form still submits (reject_if drops empty rows server-side); a filled-in topic can't be saved without a body.

maebeale:
Adds "Topic" field to shared comments partial.

Screenshot 2026-06-07 at 9 51 48 PM

PR #1594, opened 2026-06-07
Add Grant model with scholarships, funder display, and budget guards

maebeale:
### What is the goal of this PR and why is this important?

Introduces a **Grant** model — money donated by an Organization or Person — as the parent of **Scholarships**, so program staff can track a funding source's eligibility criteria, required tasks, deadlines, and remaining budget, and award scholarships against it without overspending.

### How did you approach the change?

**Grant model & CRUD**
- New `Grant` with a polymorphic `donor` (Organization or Person), `amount_cents`, `application_deadline`, `funds_received_on`, and newline-separated `eligibility_criteria` / `tasks` lists.
- Full admin-only CRUD (`GrantsController`, `GrantPolicy`, `GrantDecorator`), its own `grants` domain theme color, and an index/show/new/edit set of views.

**Grant → Scholarship relationship**
- `Scholarship` gains an optional `grant`; a validation rejects any scholarship whose amount would push the grant's awarded total past its donation amount.
- Grant **show and edit** pages list their scholarships (server-paginated, mirroring the allocations list) with an **"Add scholarship"** jump-link.
- Scholarships can now be created **directly from a grant** (no event registration): the recipient is chosen via person search, and on save you're returned to the grant page you came from (show or edit).

**Funder display**
- When a scholarship is grant-funded, the form shows the funder as a themed icon + name — a person icon in the people color, or a building icon in the organizations color (no `Person:`/`Organization:` label).

**Safety**
- A grant with associated scholarships **cannot be deleted** (`restrict_with_error`); the Delete button is disabled with a tooltip and the controller handles the blocked deletion gracefully.

### UI Testing Checklist
- [ ] Grants index / show / new / edit render and CRUD works
- [ ] Add a scholarship from a grant's show and edit pages; confirm it returns to the right page after save
- [ ] Funder shows the correct themed icon + name for person vs organization donors
- [ ] Awarding more than the grant's remaining funds is rejected
- [ ] A grant with scholarships cannot be deleted
- [ ] _Screenshots to be added_ (UI changes; not capturable in this headless run)

### Anything else to add?
- Grant-funded scholarships are standalone (recipient chosen directly), while the existing event-registration scholarship flow is unchanged.
- Eligibility criteria and tasks are modeled as newline-separated text rather than child records — happy to promote them to first-class records if preferred.
- Rebased on top of the registration-suite (#1591); ~235 specs pass and RuboCop is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1593, opened 2026-06-07
Redesign event registration edit and management UI

maebeale:
### What is the goal of this PR and why is this important?

Overhauls the **event registration edit** experience and the surrounding allocations / scholarship / payment / notification flows so staff can manage a registrant from one cohesive, scannable page instead of hopping between screens.

### How did you approach the change?

The work is split into focused commits:

- **Add check memo line and lighten payment form inputs** — a Memo field for check payments (new `payments.memo` column) and white inputs so the fields are legible on the green allocation form.
- **Restyle allocations pages to match the redesign** — the allocations index and per-registration focused view now follow the scholarship-edit / manage-registrants design, with allocatable link/label helpers.
- **Add inline manual notification logging** — admins can log a communication (email/phone/text/video) with a sender against a registration; backed by new `notifications.channel` and `notifications.sender_id` columns. `autoemail` is reserved for system-sent email.
- **Show scholarship application answers and rework requested flag** — the scholarship edit page surfaces the registrant's scholarship form answers plus a "View full submission" jump link. The "Requested" toggle is now a plain flag (no longer auto-creates an award); an empty, unrequested scholarship is cleaned up on save.
- **Redesign the event registration edit page** — compact registrant + status card (with a color-coded user-account square and an "Unsaved" status hint), inline organizations, scholarship/CE boxes, a Registration allocations summary, and a Registration communications panel. Deferred-save fields show amber "unsaved" cues that turn green/normal once saved.
- **Wire up shared header, helpers, controller registry, and schema** — shared event page header partial, form-answer helper, Stimulus registry, AGENTS notes, and the schema for the new columns.

New columns via reversible migrations: `payments.memo`, `notifications.channel`, `notifications.sender_id`, `event_registrations.ce_credit_requested`.

### UI Testing Checklist

- [ ] Registration edit: status dropdown shows "Unsaved" until saved; org chips/scholarship toggle amber→green on save
- [ ] Scholarship edit: scholarship answers render with a working "View full submission" link
- [ ] Allocations index + focused view render and paginate correctly
- [ ] Check payment form: Memo field saves and shows on the payment page
- [ ] Notifications panel: inline "Add notification" logs an entry; list is read-only and paginated

### Anything else to add?

- Request, model, and system specs added/updated; each spec file passes in isolation. Note: this repo has known intermittent flakiness on request/system specs under full-suite ordering — watch CI for phantom failures rather than real regressions.
- Screenshots to be added for the UI changes.

[...]

PR #1596, opened 2026-06-07
Rename event "manage" route/views to "registrants"

maebeale:

- The event registrant-management page is labelled **"Registrants"** in the UI (tab, heading, back-links), but the route, action, views, and path helper were still named `manage`.
- This renames everything to match so the URL reads `/events/:id/registrants` instead of `/events/:id/manage` — aligning the code with the language users actually see.

- **Route:** `get :manage` → `get :registrants` (helper `registrants_event_path`)
- **Action:** `EventsController#manage` → `#registrants`
- **Views (renamed via `git mv`, history preserved):**
  - `events/manage.html.erb` → `events/registrants.html.erb`
  - `events/_manage_search.html.erb` → `events/_registrants_search.html.erb`
  - `events/_manage_results.html.erb` → `events/_registrants_results.html.erb`
  - Renamed the `manage_results` Turbo frame and its `render` references to `registrants_results`
- **Path helper:** every `manage_event_path` → `registrants_event_path` (controllers, views, specs)
- **`return_to` token:** the `"manage"` value used to route post-action redirects back to this page → `"registrants"` (producers, consumers, specs)
- **Remaining stale labels/wording** normalized to "Registrants" (e.g. "← Back to manage", "Manage registrants", a partial comment, a spec description)
- **Subnav:** moved the **Registrants** tab ahead of Background now that it's the primary view

### Deliberate decision — `EventPolicy#manage?` is preserved

- The `manage?` authorization rule is **not** renamed. It overrides `ApplicationPolicy`'s `default_rule :manage?` and gates ~6 other event views (`_form`, `_card`, `show`, `staff`, `_subnav`, `_registration_section`) as the general "can administer this event" check — it is not specific to this page.
- Renaming it would have silently fallen those calls back to the admin-only default and changed authorization behavior. The `registrants` action still authorizes with `to: :manage?`.

### Anything else to add?

- Rebased onto `main` past #1593 (event registration UI redesign) and #1594 (Grant model / Scholarship CRUD), both of which touched these files; the rename absorbed all the `manage_event_path` references they introduced (incl. `scholarships/_form.html.erb` and `spec/requests/scholarships_spec.rb`).
- Verified: `ai/routes -g registrants` shows the new route; `ai/lint` clean; **322 examples, 0 failures** across the event, event-registration, scholarship, policy, and view specs.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -2,7 +2,7 @@ class EventsController < ApplicationController
   include AhoyTracking, TagAssignable
   skip_before_action :authenticate_user!, only: [ :index, :show, :staff ]
   skip_before_action :verify_authenticity_token, only: [ :preview ]
-  before_action :set_event, only: %i[ show edit update destroy preview dashboard background manage staff recipients preview_reminder send_reminder copy_registration_form ]
+  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants staff recipients preview_reminder send_reminder copy_registration_form ]
 
   def index
     authorize!
@@ -47,7 +47,7 @@ class EventsController < ApplicationController
     @dashboard = EventDashboard.new(@event)
   end
 
-  def manage
+  def registrants
     authorize! @event, to: :manage?
     @event = @event.decorate
     scope = @event.event_registrations
@@ -129,7 +129,7 @@ class EventsController < ApplicationController
       EventMailer.event_registration_reminder(event_registration, days_until_event: days_until).deliver_later
     end
 
-    redirect_to manage_event_path(@event), notice: "Reminder emails are being sent to #{registrations.size} registrant#{'s' if registrations.size != 1}."
+    redirect_to registrants_event_path(@event), notice: "Reminder emails are being sent to #{registrations.size} registrant#{'s' if registrations.size != 1}."
   end
 
   def create
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -48,7 +48,7 @@ class EventsController < ApplicationController
   end
 
   def registrants
-    authorize! @event, to: :manage?
+    authorize! @event, to: :registrants?
     @event = @event.decorate
     scope = @event.event_registrations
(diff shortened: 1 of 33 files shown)


app/controllers/events_controller.rb
```diff
@@ -47,7 +47,7 @@ def background
     @dashboard = EventDashboard.new(@event)
   end
 
-  def manage
+  def registrants
```
maebeale, line 50: 🤖 _From Claude:_ Action renamed to `registrants` but authorization deliberately stays `to: :manage?` — `EventPolicy#manage?` is the shared "can administer this event" rule used by ~6 other event views, not specific to this page. Renaming it would silently fall those calls back to the admin-only default.

app/policies/event_policy.rb
```diff
@@ -46,6 +46,13 @@ def manage?
     admin? || owner?
   end
 
+  # Who can view/manage an event's registrants. Mirrors manage? for now, but kept
+  # as its own rule so registrant access can be tightened later without affecting
+  # other manage? actions.
+  def registrants?
```
maebeale, line 52: 🤖 _From Claude:_ Delegates to `manage?` rather than duplicating `admin? || owner?`, matching the `recipients?`/`staff?` pattern. Gives registrant access its own seam to tighten later without touching other `manage?` actions.

spec/requests/scholarships_spec.rb
```diff
@@ -51,7 +51,10 @@
       # Event title links to the event, with the training date alongside.
       expect(response.body).to include(event.title)
       expect(response.body).to include(event_path(event))
-      expect(response.body).to include(event.start_date.in_time_zone(Time.zone).strftime("%b %-d"))
+      # The header renders event times in the viewer's zone (Pacific by default),
+      # so assert the date in that same zone rather than the test's UTC default.
+      display_zone = "Pacific Time (US & Canada)"
+      expect(response.body).to include(event.start_date.in_time_zone(display_zone).strftime("%b %-d"))
```
maebeale, line 57: 🤖 _From Claude:_ The view renders event times in the viewer's zone (Pacific via `set_time_zone_from_user`); asserting in UTC made the day off-by-one when `start_date` straddled midnight. Now asserts in the same zone the header uses.

PR #1598, opened 2026-06-08
Add form/payment quick links and form usage navigation to event dashboard

maebeale:

- Admins managing an event had no fast path from the dashboard to the live registration/scholarship forms — they had to dig through the registrants page menu.
- There was also no way to navigate between a form and the events that use it.
- This stitches those journeys together and tidies up the form-editing entry points so the labels and destinations are consistent.

**Event dashboard**
- Added a quick-links row under the title: **Public registration form**, **Scholarship form**, and a **Bulk payment** placeholder ("Soon" — feature not built yet).
- Each form link is gated by its event setting: registration link needs an associated registration form; scholarship link needs that form **and** the "Enable scholarship application" checkbox (the scholarship link targets the public registration page, which dead-ends without a registration form).

**Registrants "Form actions" menu**
- Renamed **"Change base form" → "Change form settings"**.
- Added the same **Bulk payment** placeholder ("Soon").

**Public registration page (admin bar)**
- Simplified to **Edit form** + **Dashboard** (removed "Registrants"); "Edit form" is gated on `edit?` since form editing is admin-only.

**Form editing entry points**
- Pointed the "edit" links on the forms index, form show page, and the public-registration "Edit form" button at the **sections editor** (`edit_sections`).
- Renamed the field-editor heading from "Edit:" to **"Edit questions:"**.

**Form ↔ events navigation**
- The forms index **Events** count is now a button linking to a **form-filtered events index** (new `Event.using_form` scope + `form_id` param in `search_by_params`), with a "Showing events using form … / Clear filter" banner on the events index.
- Added an **events-connected** count (links to the filtered index) and a **submissions** count (display-only — no submissions index exists) under the form edit / edit-sections titles.

### UI Testing Checklist
- [ ] Dashboard quick links appear/hide per the event's form settings and open the correct forms in a new tab
- [ ] Bulk payment placeholder shows as a disabled "Soon" item on both the dashboard and the Form actions menu
- [ ] Public registration admin bar shows "Edit form" + "Dashboard" only; "Edit form" goes to the sections editor
- [ ] Forms index event count links to the filtered events index; banner shows and clears correctly
- [ ] Events-connected / submissions counts render under the form edit titles

### Anything else to add?
- **Bulk payment** is a placeholder in two spots — it needs a real destination once that feature exists.
- **Submissions** count is intentionally not a link: there is no submissions controller/route/index in the app.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -49,6 +49,7 @@ class Event < ApplicationRecord
   scope :featured, -> { registerable.where(published: true, featured: true) } # overrides Publishable
   scope :publicly_featured, -> { where(published: true, publicly_visible: true, publicly_featured: true) } # overrides Featureable
   scope :registerable, -> { where("registration_close_date IS NULL OR registration_close_date >= ?", Time.current) }
+  scope :using_form, ->(form_id) { joins(:event_forms).where(event_forms: { form_id: form_id }).distinct }
 
   def self.search_by_params(params)
     stories = is_a?(ActiveRecord::Relation) ? self : all
@@ -56,6 +57,7 @@ class Event < ApplicationRecord
     stories = stories.sector_names_all(params[:sector_names_all]) if params[:sector_names_all].present?
     stories = stories.category_names_all(params[:category_names_all]) if params[:category_names_all].present?
     stories = stories.windows_type_name(params[:windows_type_name]) if params[:windows_type_name].present?
+    stories = stories.using_form(params[:form_id]) if params[:form_id].present?
     stories
   end
 
--- a/app/views/events/dashboard.html.erb
+++ b/app/views/events/dashboard.html.erb
@@ -20,6 +20,39 @@
     <% end %>
   </div>
 
+  <%# Quick links: registration forms and bulk payment. Each form link is gated by
+      its event setting — registration on an associated registration form, scholarship
+      on that form plus the "Enable scholarship application" checkbox (which is what
+      creates the scholarship event_form). The scholarship link points at the public
+      registration page, so it also needs the registration form or it dead-ends there.
+      Links open in a new tab. Bulk payment is a placeholder until that feature is built. %>
+  <% link_base = "inline-flex items-center gap-2 rounded-lg border bg-white px-3 py-2 text-sm font-medium shadow-sm transition-colors" %>
+  <div class="flex flex-wrap items-center justify-center gap-2 sm:gap-3 mb-8">
+    <% if @event.event_forms.registration.exists? %>
+      <%= link_to new_event_public_registration_path(@event, as_visitor: true), target: "_blank", rel: "noopener noreferrer",
+            class: "#{link_base} #{DomainTheme.border_class_for(:events, intensity: 200)} text-gray-700 hover:#{DomainTheme.border_class_for(:events, intensity: 300)} hover:bg-gray-50" do %>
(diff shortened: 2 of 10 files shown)


app/models/event.rb
```diff
@@ -49,13 +49,15 @@ class Event < ApplicationRecord
   scope :featured, -> { registerable.where(published: true, featured: true) } # overrides Publishable
   scope :publicly_featured, -> { where(published: true, publicly_visible: true, publicly_featured: true) } # overrides Featureable
   scope :registerable, -> { where("registration_close_date IS NULL OR registration_close_date >= ?", Time.current) }
+  scope :using_form, ->(form_id) { joins(:event_forms).where(event_forms: { form_id: form_id }).distinct }
```
maebeale, line 52: 🤖 _From Claude:_ `.distinct` because the `event_forms` join can return duplicate event rows when an event uses the same form in multiple roles.

app/views/events/dashboard.html.erb
```diff
@@ -20,6 +20,39 @@
...
+    <% if @event.event_forms.registration.exists? %>
+      <%= link_to new_event_public_registration_path(@event, as_visitor: true), target: "_blank", rel: "noopener noreferrer",
+            class: "#{link_base} #{DomainTheme.border_class_for(:events, intensity: 200)} text-gray-700 hover:#{DomainTheme.border_class_for(:events, intensity: 300)} hover:bg-gray-50" do %>
+        <i class="fa-solid fa-clipboard-list <%= DomainTheme.text_class_for(:events, intensity: 600) %>"></i>
+        Public registration form
+        <i class="fa-solid fa-arrow-up-right-from-square text-xs text-gray-400"></i>
+      <% end %>
+    <% end %>
+
+    <% if @event.event_forms.registration.exists? && @event.scholarship_form %>
```
maebeale, line 40: 🤖 _From Claude:_ Scholarship link requires a registration form too — its destination is the public registration page, which redirects with "Registration form is not available" when none exists, so without this gate it would dead-end.

app/views/events/public_registrations/new.html.erb
```diff
@@ -7,8 +7,10 @@
     <% end %>
     <% if allowed_to?(:dashboard?, @event) %>
       <div class="ml-auto flex flex-wrap items-center gap-1">
+        <% if allowed_to?(:edit?, @form) %>
```
maebeale, line 10: 🤖 _From Claude:_ Gated on `edit?` (admin-only via FormPolicy) rather than the surrounding `dashboard?` check, since event owners can reach the dashboard but can't edit forms.

app/views/forms/_stats.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <% else %>
+    <span class="inline-flex items-center gap-1.5 rounded-full bg-gray-100 px-3 py-1 text-sm text-gray-500">
+      <i class="fa-solid fa-calendar-day text-gray-400"></i>
+      No events connected
+    </span>
+  <% end %>
+
+  <span class="inline-flex items-center gap-1.5 rounded-full bg-gray-100 px-3 py-1 text-sm text-gray-600">
+    <i class="fa-solid fa-inbox text-gray-400"></i>
+    <%= pluralize(form.form_submissions.size, "submission") %>
```
maebeale, line 21: 🤖 _From Claude:_ Submissions is intentionally display-only — there is no submissions controller/route/index in the app to link to.

PR #1601, opened 2026-06-08
Show sender on notifications index

maebeale:
- Surfaces who sent each system notification on the index, so staff can tell portal-generated emails apart from staff-sent ones without opening each record.

- Renamed the **Recipient** column to **People** and split the cell into two lines: `To: [recipient email]` and a smaller, lighter `From: [sender name]`.
- `From` falls back to **AWBW portal** when a notification has no sender (the association is optional).
- Eager-loaded `sender: :person` in the controller so the new `From` line doesn't trigger N+1 queries.

### UI Testing Checklist
- [ ] System notifications index shows a **People** column with To/From lines
- [ ] Notifications with a sender show the sender's name; sender-less notifications show **AWBW portal**

### Anything else to add?
- The `From` line uses `sender.full_name`, which falls back to the sender's email when their person/name is blank.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -6,7 +6,7 @@ class NotificationsController < ApplicationController
 
     if turbo_frame_request?
       per_page = params[:number_of_items_per_page].presence || 25
-      base_scope = authorized_scope(Notification.includes(:noticeable))
+      base_scope = authorized_scope(Notification.includes(:noticeable, sender: :person))
       filtered = base_scope.search_by_params(params.to_unsafe_h)
       @notifications = filtered.order(created_at: :desc)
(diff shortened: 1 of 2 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
 
     if turbo_frame_request?
       per_page = params[:number_of_items_per_page].presence || 25
-      base_scope = authorized_scope(Notification.includes(:noticeable))
+      base_scope = authorized_scope(Notification.includes(:noticeable, sender: :person))
```
maebeale, line 9: 🤖 _From Claude:_ Added `sender: :person` to the eager-load so the new `From:` line (which reads through `sender.full_name` → `sender.person`) doesn't trigger N+1 queries on the index.

PR #1606, opened 2026-06-09
Add per-field width, min-word, and field-option controls to form builder

maebeale:

- Give form admins real layout control over the form builder instead of relying on hardcoded, identifier-based field groupings
- Let questions sit at **full / half / third / quarter** width so multi-column forms can be composed for any field, not just the predefined name/email/address rows
- Support long, multi-sentence question prompts that were previously hitting the `varchar(255)` limit and raising `ActiveRecord::ValueTooLong`
- Add per-field **minimum word count** and **maximum character count** validation, plus an expandable options editor, to round out the field-configuration UI

- Added a `width` enum to `FormField` (`full/half/third/quarter`, prefixed to avoid colliding with AR finder methods like `.third`) plus a `grid_span_class` helper mapping each width to a Tailwind 12-column span
- Replaced the brittle `row_groups` machinery (duplicated across the preview, live, and read-only registration views) with a single `md:grid-cols-12` grid driven by each field's width; group headers and full-width fields take a whole row
- Backfilled `width` on existing fields in the migration (name pairs → half, city/state/zip → thirds, …) and set sensible defaults in `FormBuilderService` so existing forms keep their current appearance
- Widened `form_fields.name` to `text`; added nullable `min_words` and `max_characters` columns
- Min/max are enforced server-side in both the public-registration and bulk-payment flows via `min_words_error` / `max_characters_error`, surfaced as `maxlength` + hint text on the rendered fields, and edited in the form builder's collapsible "More options" panel (width / min words / max chars)
- Added `field_options` and `expand_all` Stimulus controllers for the editor UI
- Updated the Brakeman ignore fingerprint for the (admin-only, pre-existing) `forms_controller` permit warning, since the permit list changed

### Anything else to add?

- Min words / max chars apply only to free-form text answer types (one-line and paragraph); they're left to the presence check when blank
- One trade-off: the old layout had a couple of 2/3-width special cases (e.g. `secondary_email`); those now map to the nearest of the four supported widths (half), adjustable per-field via the new control
- New Tailwind classes (`md:grid-cols-12`, `md:col-span-{12,6,4,3}`) required a Vite dev rebuild
- Tests: model enum + `grid_span_class`, `min_words_error` / `max_characters_error` + their validations, service width defaults, controller enforcement in both flows, and the existing form/registration/bulk-payment suites — all green

- Known follow-up (deferred): the new `expand_all` / `field_options` Stimulus controllers duplicate the existing `expandable_cards` / `expandable_card` pattern. Kept as-is for now (the new pair uses the Outlets API, which our conventions prefer over `expandable_cards`' `@window` event broadcasting); cross-referencing `TODO(dedupe)` notes are in both controllers. To consolidate later.

- Scholarship fields now render in the same `md:grid-cols-12` grid (respecting per-field width) and have min-word / max-char rules enforced on submit; the shared per-field checks were extracted into `collect_field_errors` so registration and scholarship forms validate identically

### DIFF
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -332,6 +332,8 @@ module Events
           errors[field.id] = "must be a valid email address"
         elsif (word_error = field.min_words_error(value))
           errors[field.id] = word_error
+        elsif (char_error = field.max_characters_error(value))
+          errors[field.id] = char_error
         end
       end
 
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -330,6 +330,8 @@ module Events
           errors[field.id] = "must be a whole number"
         elsif field.field_identifier&.match?(/email(?!_type|_confirmation)/) && value.to_s !~ /\A[^@\s]+@[^@\s]+\z/
           errors[field.id] = "must be a valid email address"
+        elsif (word_error = field.min_words_error(value))
+          errors[field.id] = word_error
         end
       end
 
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -40,6 +40,10 @@ module Events
       registration_params, scholarship_params = split_form_params(all_params)
 
       @field_errors = validate_required_fields(registration_params)
+      if @scholarship_form
+        scholarship_fields = @scholarship_form.form_fields.where.not(answer_type: :group_header).to_a
+        @field_errors = @field_errors.merge(collect_field_errors(scholarship_fields, scholarship_params))
+      end
       if @field_errors.any?
         @form_fields = visible_form_fields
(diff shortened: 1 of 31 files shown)


app/controllers/events/public_registrations_controller.rb
```diff
@@ -310,11 +314,30 @@ def ensure_registerable
...
+        end
+      end
+
+      errors
+    end
+
+    # Per-field validation shared by every form role (registration, scholarship):
+    # presence, number/email format, and the per-field min-words / max-characters
+    # settings. Keyed by field id so results from multiple forms can be merged.
+    def collect_field_errors(fields, form_params)
```
maebeale, line 337: 🤖 _From Claude:_ Extracted the per-field presence/format/min-word/max-char checks here so the scholarship form reuses the exact same validation as registration via merged `field_errors`. The `confirm_email` cross-field check stays in `validate_required_fields` since it's registration-specific.

PR #1607, opened 2026-06-09
Align field-option controls under the name field in form builder

maebeale:
- In the form builder, the collapsible per-field options row (width / min words / max chars) lined up under the **visibility chip** instead of under the **field name**, which read as a misaligned, ragged second row.
- This is a follow-up polish to #1606 (per-field width / min-word / field-option controls), which has merged.

- Pulled the visibility chip out of the `field-options` container and made it a sibling of it, so the container holding both rows now begins at the field name. The options row therefore starts at the name field's left edge.
- Gave the chip `self-start mt-1` so it stays parallel to the name field whether the options panel is collapsed or expanded.
- Added request specs covering rich (unescaped) HTML rendering of header/field labels and the per-field options controls.

### UI Testing Checklist
- [ ] Open a form in the builder; confirm the "Full width / Min words / Max chars" row starts under the field name, not the visibility chip
- [ ] Confirm the visibility chip stays aligned with the name field when the options panel is expanded and collapsed
- [ ] Confirm group-header rows are unaffected

### Anything else to add?
- View-only layout change plus tests; no model or controller changes.

### DIFF
--- a/app/views/forms/_form_field_fields.html.erb
+++ b/app/views/forms/_form_field_fields.html.erb
@@ -41,12 +41,12 @@
       </div>
     </div>
   <% else %>
-    <div class="flex-1" data-controller="field-options">
+    <%= f.select :visibility, visibility_options,
+          {},
+          class: "self-start mt-1 text-xs rounded-full border px-2 py-0.5 shrink-0 cursor-pointer",
+          data: { controller: "chip-select", chip_select_styles_value: visibility_styles, action: "chip-select#update" } %>
+    <div class="flex-1 min-w-0" data-controller="field-options">
       <div class="flex flex-wrap items-center gap-2">
-        <%= f.select :visibility, visibility_options,
-              {},
-              class: "text-xs rounded-full border px-2 py-0.5 shrink-0 cursor-pointer",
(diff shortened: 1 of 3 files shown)


app/views/forms/_form_field_fields.html.erb
```diff
@@ -25,49 +25,74 @@
...
-        <%= f.check_box :required, class: "rounded border-gray-300 text-blue-600" %>
-        Required
-      </label>
-      <label class="flex items-center gap-1 text-sm text-gray-600 shrink-0" title="Hide if already answered on any form (not just this event)">
-        <%= f.check_box :one_time, class: "rounded border-gray-300 text-amber-600" %>
-        One-time
-      </label>
-      <%= link_to_remove_association "Remove", f,
-            class: "text-sm text-gray-400 hover:text-red-600 underline shrink-0 ml-auto" %>
+    <%= f.select :visibility, visibility_options,
```
maebeale, line 44: 🤖 _From Claude:_ Moved the visibility chip out of the `field-options` container so the options panel (width/min/max) aligns under the field name instead of the chip. `self-start mt-1` keeps the chip parallel to the name field in both collapsed and expanded states.

PR #1608, opened 2026-06-09
Order new form sections by page order and warn on toggle

maebeale:

- On the edit-sections page, newly checked sections were always appended at the **end** of the form, regardless of where they sit in the checkbox list — so adding a section "in the middle" left fields out of order relative to the page.
- Saving the edit_sections form with no actual changes to section lists still tore down and rebuilt fields, which is wasteful and risks churn on field positions.
- Unchecking a saved section is destructive (removes its questions), but nothing prompted the user about that before submitting.

- **Page-order reordering:** after adding fields for newly checked sections, `FormBuilderService.reorder_to_page_order!` renumbers all fields to the canonical `SECTIONS` order while preserving each field's order within its own section. A new `SECTION_GROUPS` map bridges section keys to the `section` column values their builders assign (some differ, e.g. `bulk_payment` → `group_payment`).
- **No-op guard:** `FormsController#update_sections` redirects with "No section changes." when the submitted section set matches the form's current sections — skipping the rebuild entirely.
- **Inline toggle feedback:** new `section_toggle_controller.js` Stimulus controller shows a red warning when a saved section is unchecked and a green note when a new section is checked, staying silent when the checkbox matches its saved state. The edit-sections list is now a numbered `<ol>` reflecting page order.

### UI Testing Checklist

- [ ] Add a section that sorts before an existing one → its fields appear in page order, not at the end
- [ ] Save with no checkbox changes → "No section changes." notice, field positions untouched
- [ ] Uncheck a saved section → red removal warning appears inline
- [ ] Check a new section → green additive note appears inline
- [ ] Toggle a checkbox back to its saved state → message clears

### Anything else to add?

- Specs added: page-order insertion + sequential positions (service spec), unchanged-set no-op (request spec). Full `forms_spec` + `form_builder_service_spec` green (37 examples).


app/services/form_builder_service.rb
```diff
@@ -66,6 +66,22 @@ def call
...
+    person_identifier: %w[person_identifier],
+    person_contact_info: %w[person_contact_info],
+    person_background: %w[background],
+    professional_info: %w[professional],
+    marketing: %w[marketing],
+    scholarship: %w[scholarship],
+    payment: %w[payment],
+    consent: %w[consent],
+    post_event_feedback: %w[post_event_feedback],
+    bulk_payment: %w[bulk_payment group_payment]
```
maebeale, line 82: 🤖 _From Claude:_ `SECTION_GROUPS` exists because several builders write a `section` column value that differs from the section key (e.g. `bulk_payment` fields are tagged `group_payment`). Without this map, those fields would not group back to their section when renumbering.

app/services/form_builder_service.rb
```diff
@@ -85,20 +101,40 @@ def self.update_sections!(form, new_sections)
...
+  # (the order of SECTIONS / the edit-sections checkboxes), preserving each
+  # field's existing order within its own section.
+  def self.reorder_to_page_order!(form, section_order)
+    section_for_group = SECTION_GROUPS.flat_map { |key, groups| groups.map { |group| [ group, key ] } }.to_h
+    rank = SECTIONS.keys.select { |key| section_order.include?(key) }.each_with_index.to_h
+
+    ordered = form.form_fields.reorder(:position).to_a.each_with_index.sort_by do |field, index|
+      section_key = section_for_group[field.section]
+      [ rank.fetch(section_key, rank.size), index ]
+    end
```
maebeale, line 130: 🤖 _From Claude:_ Sort is stable on the original `position` index, so each field keeps its order within its own section; only the cross-section ordering changes. `update_column` skips the no-op writes to avoid touching unchanged rows.

app/controllers/forms_controller.rb
```diff
@@ -76,6 +76,12 @@ def update_sections
       return
     end
 
+    current_sections = (@form.sections || []).map(&:to_sym)
+    if sections.to_set == current_sections.to_set
```
maebeale, line 80: 🤖 _From Claude:_ Set comparison so checkbox order does not count as a change — only the section membership matters here.

PR #1609, opened 2026-06-09
Show all fields in forms preview; apply conditional logic only on registration

maebeale:
- The **forms preview** (`/forms/:id`) was simulating conditional visibility with toggle controls, duplicating logic that already runs on the registration and public registration routes.
- An admin viewing a form wants to see **every** field, with a clear signal for which ones are conditionally shown — not a simulation of one runtime scenario at a time.
- The conditional filtering itself stays where it belongs: the live registration / public registration flows.

- `forms#show` now loads **all** fields ordered by position; removed the `preview_form_fields` param-driven filtering.
- The preview renders every field and highlights any that aren't *always ask* with a colored left-border marker and a visibility badge (Logged out only / Answers on file / Scholarship only), matching the editor's chip colors. Group headers with conditional visibility get a badge too.
- Removed the old "Preview mode" toggle UI; replaced with a short note explaining the preview shows all fields and the conditional logic runs on the live form.
- Added `FormField#visibility_label` and `#conditional_visibility?` (anything other than `always_ask`).
- Styled the preview as an admin-only page via `content_for(:page_bg_class, "admin-only bg-blue-100")` and registered it in the `page_bg_class` alignment spec (the `FormPolicy` is admin-only).

### UI Testing Checklist
- [ ] View a form with only always-ask fields — all fields shown, no highlights, note reads "All fields are always asked."
- [ ] View a form with logged-out-only / answers-on-file fields — every field renders, conditional ones show a badge + marker.
- [ ] Confirm the registration and public registration forms still apply the conditional hiding as before.
- [ ] Confirm the admin background color applies on the forms preview page.

### Anything else to add?
- `scholarship_only` isn't selectable in the field editor (only `always_ask` / `logged_out_only` / `answers_on_file` are) but exists in the enum — the preview styles it (blue) so it degrades gracefully if any field carries that value.

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -8,7 +8,10 @@ class FormsController < ApplicationController
 
   def show
     authorize! @form
-    @form_fields = preview_form_fields
+    # The forms preview shows every field, highlighting any with conditional
+    # visibility. The conditional logic itself only runs on the live
+    # registration and public registration forms.
+    @form_fields = @form.form_fields.reorder(position: :asc)
   end
 
   def new
@@ -100,20 +103,6 @@ class FormsController < ApplicationController
 
   def set_form
     @form = Form.find(params[:id])
-  end
-
-  def preview_form_fields
-    scope = @form.form_fields
-
-    if params[:preview_logged_in].present?
-      scope = scope.where.not(visibility: :logged_out_only)
-    end
-
-    if params[:preview_answered].present?
-      scope = scope.where.not(visibility: :answers_on_file)
-    end
-
-    scope.reorder(position: :asc)
   end
 
   def form_params
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -42,6 +42,14 @@ class FormField < ApplicationRecord
   # Prefixed because :third/:first/etc. collide with Active Record finder methods
   enum :width, [ :full, :half, :third, :quarter ], prefix: true
 
+  # Human-friendly labels for visibility, mirroring the form editor's options
+  VISIBILITY_LABELS = {
+    "always_ask" => "Always ask",
+    "scholarship_only" => "Scholarship only",
+    "logged_out_only" => "Logged out only",
+    "answers_on_file" => "Answers on file"
+  }.freeze
+
   # Human-friendly labels for answer types (radio is "single choice" because only one option can be picked)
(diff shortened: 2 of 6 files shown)


app/controllers/forms_controller.rb
```diff
@@ -8,7 +8,10 @@ def index
 
   def show
     authorize! @form
-    @form_fields = preview_form_fields
+    # The forms preview shows every field, highlighting any with conditional
+    # visibility. The conditional logic itself only runs on the live
+    # registration and public registration forms.
+    @form_fields = @form.form_fields.reorder(position: :asc)
```
maebeale, line 14: 🤖 _From Claude:_ Core behavior change: the preview now loads every field unconditionally. The `preview_*` param filtering moved out — conditional visibility is applied only on the live registration/public-registration routes.

app/models/form_field.rb
```diff
@@ -83,6 +91,16 @@ def answer_type_label
     ANSWER_TYPE_LABELS.fetch(answer_type, answer_type&.titleize)
   end
 
+  def visibility_label
+    VISIBILITY_LABELS.fetch(visibility, visibility&.humanize)
+  end
+
+  # True when the field is only asked under certain conditions, i.e. anything
+  # other than "always ask". The forms preview highlights these.
+  def conditional_visibility?
```
maebeale, line 100: 🤖 _From Claude:_ Defined as `!always_ask?` so any non-default visibility (incl. the not-editor-selectable `scholarship_only`) counts as conditional and gets highlighted.

app/views/forms/show.html.erb
```diff
@@ -80,18 +29,54 @@
...
         end
+
+        # Mirror the editor's visibility chip colors so the same field reads
+        # the same in both places.
+        visibility_badge_styles = {
+          "scholarship_only" => "bg-blue-100 text-blue-700 border-blue-200",
+          "logged_out_only" => "bg-emerald-100 text-emerald-700 border-emerald-200",
+          "answers_on_file" => "bg-amber-100 text-amber-700 border-amber-200"
+        }
+        visibility_marker_styles = {
```
maebeale, line 40: 🤖 _From Claude:_ Badge/marker colors intentionally mirror the editor chip styles in `_form_field_fields.html.erb` so a field reads the same in both the editor and the preview.

PR #1612, opened 2026-06-10
Edit multiple-choice answer options in the form builder

maebeale:

- Form admins can now **edit a multiple-choice field's answer options** directly in the form builder — previously there was no UI to add, rename, or remove choices.
- Makes the form-builder field rows readable at a glance and gives admins a **true picture of a form's sections** when deciding what to include.

**Answer options editing**

- Added an **"Answer options" toggle** inside the expanded field panel (only for radio/checkbox fields). It lists each option as an editable input with add/remove, persisted via nested attributes.
- **Avoided cross-field rename leakage:** options are stored as shared `AnswerOption` records (built with `find_or_create_by!`), so editing text **re-points the join at a matching `AnswerOption`** instead of renaming the shared record. Changing one field's "Work" → "Home" leaves every other field that reuses "Work" untouched.
- `FormField` nested attributes now `allow_destroy` and reject blank new rows; controller permits `form_field_answer_options_attributes: [:id, :option_name, :_destroy]`.
- New `answer-options` Stimulus controller mirrors the existing `field-options` show/hide pattern.
- Refreshed the existing Brakeman ignore fingerprint for the (pre-existing, intentional) `:role` mass-assignment entry, since the permit line changed.

**Field-row layout & section drag**

- **Right-aligned** each field's options (width / min words / max chars) next to **Remove**, and moved the **Answer options** toggle onto that same row, so the expanded panel mirrors the collapsed row's columns instead of drifting left.
- Moved the **section-header (informational) row's Remove link into the same "More options" dropdown** the other field types use, so every row follows one pattern.
- Dropped the **Required / Only-ask-once column headers** (the inline labels next to each checkbox already name them) and re-added those columns as **invisible spacers** so the header's flex-grow math matches the rows and Question / Type / Visibility sit directly above their inputs.
- Added a **distinct section drag handle** on group-header rows (4-way move icon on an indigo pill) to signal that dragging a header moves the **whole section** — behavior the `form-fields-sortable` controller already implements; this just makes it discoverable.
- Collapsed the conditional-visibility legend into a **details/summary column header** so it reads as a key for the right-aligned per-field visibility chips rather than as a standalone block.

**Editable sections overview**

- The edit-sections page now lists **custom (admin-added) sections alongside the built-in ones**, ordered by their real position on the form, via new `FormBuilderService.editable_sections`. Removal/reorder decisions are made against what the form actually contains.

### UI Testing Checklist

- [ ] Expand a multiple-choice field → "Answer options" toggle appears (on the controls row) with the current options below
- [ ] Edit an option's text and save → updates only this field, not other fields sharing the same option
- [ ] Add a new option and save → option persists and renders on the form
- [ ] Remove an option and save → option is gone
- [ ] Blank new option rows are ignored on save
- [ ] Field options row (width / min / max / Remove) is right-aligned and columns line up with the collapsed row
- [ ] Section-header row's Remove sits inside its "More options" dropdown like other rows
- [ ] Column headers (Question / Type / Visibility) line up with their inputs; no Required/Only-ask-once header text shows
- [ ] Group-header rows show the indigo move handle; dragging one reorders its whole section
- [ ] Edit-sections page shows custom sections interleaved in form order (no per-question bullet list)

### Anything else to add?

- Options are not drag-reorderable here; `AnswerOption.position` is global/shared today, so per-field ordering isn't cleanly modeled yet — left as a follow-up.
- Specs: model specs for `option_name` (get/set, dedup, re-point-not-rename, whitespace), request specs (render, rename, add, blank-ignore, remove), and service specs for `editable_sections`. Full suite green.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -69,7 +69,8 @@ class FormField < ApplicationRecord
   }.freeze
 
   # Nested attributes
-  accepts_nested_attributes_for :form_field_answer_options
+  accepts_nested_attributes_for :form_field_answer_options, allow_destroy: true,
+    reject_if: ->(attrs) { attrs[:option_name].blank? }
 
   default_scope { order(position: :desc) }
   scope :published, -> { where(status: "active") }
--- a/app/models/form_field_answer_option.rb
+++ b/app/models/form_field_answer_option.rb
@@ -5,4 +5,20 @@ class FormFieldAnswerOption < ApplicationRecord
   def name
     answer_option.name if answer_option
   end
+
+  # Virtual attribute used by the form builder to edit an option's text.
+  alias_method :option_name, :name
+
+  # Editing the text re-points this join at a matching AnswerOption (creating
+  # one if needed) rather than renaming the existing record, which is shared
+  # across fields (built via AnswerOption.find_or_create_by!) — renaming it
+  # would change every other field that uses the same option.
+  def option_name=(value)
+    value = value.to_s.strip
+    self.answer_option = if value.present?
+      AnswerOption.find_or_create_by!(name: value) do |option|
+        option.position = (AnswerOption.maximum(:position) || 0) + 1
+      end
+    end
+  end
 end
--- a/app/services/form_builder_service.rb
+++ b/app/services/form_builder_service.rb
@@ -82,10 +82,83 @@ class FormBuilderService
     bulk_payment: %w[bulk_payment]
   }.freeze
 
-  # Update sections on an existing form: add new sections, remove unchecked ones
-  def self.update_sections!(form, new_sections)
+  # Built-in section keys that apply to a form with the given role. Scholarship
+  # and bulk-payment forms show only their own section; every other role shows
+  # all sections except those two.
+  def self.applicable_section_keys(role)
+    SECTIONS.keys.select do |key|
+      case role
+      when "scholarship" then key == :scholarship
+      when "bulk_payment" then key == :bulk_payment
+      else key != :scholarship && key != :bulk_payment
+      end
+    end
+  end
+
+  # Ordered list of sections for the edit-sections page. Each entry is a hash
+  # with a :kind of :builtin or :custom. Built-in sections carry their :label,
+  # section :key, and :included state; custom (user-added) sections carry the
+  # group_header :label and the :questions that follow it on the form.
+  #
+  # Sections present on the form are ordered by their position there, so a
+  # custom section slots into the list wherever it sits on the form. Built-in
+  # sections not currently on the form follow their canonical predecessor.
+  def self.editable_sections(form)
+    group_to_key = SECTION_GROUPS.flat_map { |key, groups| groups.map { |group| [ group, key ] } }.to_h
+
+    positions = {}
+    header_names = {}
+    customs = []
+    current = nil
+
+    # Rank by iteration order rather than the stored position, which is nullable
+    # and can collide, so the sort keys below are always comparable integers.
+    form.form_fields.reorder(position: :asc).each_with_index do |field, order|
+      key = group_to_key[field.section]
+      if key
+        current = nil
+        positions[key] ||= order
+        header_names[key] ||= field.name if field.group_header?
+      elsif field.group_header?
+        current = { kind: :custom, header: field, questions: [] }
+        customs << [ order, current ]
+      elsif current
+        current[:questions] << field
+      end
+    end
+
+    entries = []
+    last_position = 0
+    applicable_section_keys(form.role).each_with_index do |key, rank|
+      position = positions[key]
+      last_position = position if position
+      default_header = SECTION_HEADERS.fetch(key).first
+      header_name = header_names[key]
+      entries << {
+        sort: [ position || last_position, position ? 0 : 1, rank ],
+        kind: :builtin,
+        key: key,
+        label: SECTIONS.fetch(key)[:label],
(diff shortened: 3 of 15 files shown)


app/models/form_field_answer_option.rb
```diff
@@ -5,4 +5,20 @@ class FormFieldAnswerOption < ApplicationRecord
...
+  # Editing the text re-points this join at a matching AnswerOption (creating
+  # one if needed) rather than renaming the existing record, which is shared
+  # across fields (built via AnswerOption.find_or_create_by!) — renaming it
+  # would change every other field that uses the same option.
+  def option_name=(value)
+    value = value.to_s.strip
+    self.answer_option = if value.present?
+      AnswerOption.find_or_create_by!(name: value) do |option|
+        option.position = (AnswerOption.maximum(:position) || 0) + 1
+      end
```
maebeale, line 21: 🤖 _From Claude:_ Re-points the join at a found/created `AnswerOption` rather than renaming the existing one, because options are shared across fields (`find_or_create_by!`) — an in-place rename would leak to every other field using that option.

app/models/form_field.rb
```diff
@@ -69,7 +69,8 @@ class FormField < ApplicationRecord
   }.freeze
 
   # Nested attributes
-  accepts_nested_attributes_for :form_field_answer_options
+  accepts_nested_attributes_for :form_field_answer_options, allow_destroy: true,
+    reject_if: ->(attrs) { attrs[:option_name].blank? }
```
maebeale, line 73: 🤖 _From Claude:_ `reject_if` on blank `option_name` drops empty "+ Add option" rows; an existing row keeps its value when removed (cocoon sets `_destroy`), so removals still go through.

config/brakeman.ignore
```diff
@@ -117,13 +117,13 @@
     {
       "warning_type": "Mass Assignment",
       "warning_code": 105,
-      "fingerprint": "5473b25a3e7836314952560daf8b227d719838ca52c1fd7388a03b60415dd015",
+      "fingerprint": "409787bd055764cbde1429558af080ceadef1d12b5a5b082746e6bec159f15b2",
```
maebeale, line 120: 🤖 _From Claude:_ Only refreshed the fingerprint/code for the pre-existing, intentional `:role` mass-assignment ignore — the permit line changed, which re-fingerprinted it. No new suppression added.

app/services/form_builder_service.rb
```diff
@@ -82,11 +82,84 @@ def call
...
+
+    entries = []
+    last_position = 0
+    applicable_section_keys(form.role).each_with_index do |key, rank|
+      position = positions[key]
+      last_position = position if position
+      default_header = SECTION_HEADERS.fetch(key).first
+      header_name = header_names[key]
+      entries << {
+        sort: [ position || last_position, position ? 0 : 1, rank ],
```
maebeale, line 138: 🤖 _From Claude:_ The 3-element sort key resolves ordering deterministically: real form position first, then `position ? 0 : 1` keeps absent built-in sections after the last present one rather than at position 0, and `rank` breaks ties by canonical section order. Without the middle element, an unincluded section would sort to the top.

app/services/form_builder_service.rb
```diff
@@ -82,11 +82,84 @@ def call
...
+  # sections not currently on the form follow their canonical predecessor.
+  def self.editable_sections(form)
+    group_to_key = SECTION_GROUPS.flat_map { |key, groups| groups.map { |group| [ group, key ] } }.to_h
+
+    positions = {}
+    header_names = {}
+    customs = []
+    current = nil
+
+    # Rank by iteration order rather than the stored position, which is nullable
```
maebeale, line 114: 🤖 _From Claude:_ Using `each_with_index` order instead of the stored `position` matters here because `position` is nullable and can collide — the sort keys downstream need comparable integers, and iteration order gives a stable total order.

app/views/forms/_form_field_fields.html.erb
```diff
@@ -71,27 +88,48 @@
         </button>
       </div>
```
maebeale, line 89: 🤖 _From Claude:_ The `answer-options` controller moved up to the panel wrapper (conditionally, only for multiple-choice) so its `icon` target (in the toggle, now on the controls row) and `panel` target (the option values, a row below) stay within one controller scope.

app/views/forms/edit.html.erb
```diff
@@ -67,23 +67,52 @@
...
-            <span class="text-xs text-gray-500">Always shown to all users</span>
+      <details class="border-b border-gray-100">
+        <summary class="cursor-pointer select-none list-none [&::-webkit-details-marker]:hidden hover:bg-gray-50">
+          <div class="flex items-center gap-3 px-6 py-2 text-xs font-medium text-gray-500">
+            <span class="w-3 shrink-0" aria-hidden="true"></span>
+            <div class="flex-1 min-w-0">
+              <div class="flex items-center gap-2">
+                <span class="min-w-0 flex-[3_1_10rem]">Question</span>
+                <span class="flex-[2_1_9rem]">Type</span>
+                <span class="flex items-center gap-1 text-sm shrink-0 invisible" aria-hidden="true">
```
maebeale, line 78: 🤖 _From Claude:_ These two columns are `invisible` spacers, not visible headers — the rows grow Question/Type via `flex-grow`, so the header needs the same column footprint or Type drifts right of its select. The labels stay only on the inline checkboxes below.

PR #1620, opened 2026-06-10
Add event-aware headers to public registration forms

maebeale:
- A standalone registration form is shared across many events, so its intro copy can't hard-code dates, fees, location, or platform — it goes stale the moment the form is reused for the next event.
- This lets an admin write the header once with placeholders that fill from whichever event the form is shown for.
- The scholarship form's header was also being dropped entirely when the form is appended to an event registration, so admin-authored scholarship copy never reached applicants.

- Added a token layer to `form_header_html`: `{{event_title}}`, `{{event_dates}}`, `{{event_times}}`, `{{event_fee}}`, `{{event_platform}}`, `{{event_location}}`, `{{event_month_year}}`, `{{registration_close}}` — each filled from the event, mirroring the event show page's date/time/fee formatting, with neutral fallbacks on the standalone preview (no event).
- Surfaced the available placeholders as a reference panel beside the header field in the form editor so they're discoverable. The header field now grows to fill available width/height while the placeholder panel stays a fixed width pinned right.
- On the standalone form preview (no event in scope), added a notice that any placeholders render as example text and fill from the event on the live page — so the fallbacks don't mislead.
- Passed the event into the `forms/_header` partial on both public registration pages, and drove the virtual-platform badge from event videoconference settings.
- Rendered the scholarship form's header inline above the scholarship questions when it's appended to an event registration (previously dropped).
- Seeded warm, scannable example headers (AWBW green + purple call-to-action panels): one on the registration form, and a scholarship-specific one scoped to eligibility and recipient obligations — deliberately not repeating registration-header content since both render on the same page. Set the flagship training to run on Zoom.

### UI Testing Checklist
- [ ] Public registration page renders the header with tokens filled from the event (dates, times, fee, platform, location, registration close).
- [ ] Scholarship application section renders the scholarship form's header above its questions.
- [ ] Standalone form preview (no event) renders neutral fallback phrasing and shows the "this is a preview" token notice.
- [ ] Form editor shows the placeholder reference panel next to the header field, sized to the right.
- [ ] Header HTML stays sanitized — admin styling renders, scripts/unsafe attrs stripped.

### Anything else to add?
- Header content is still sanitized through `form_label_html`; the seeds use only inline styles that survive the allowlist.
- 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,6 +25,7 @@ module ApplicationHelper
     "{{event_times}}" => [ "Event start-end time", "9 am - 4:30 pm PST" ],
     "{{event_fee}}" => [ "Registration fee", "$1,500" ],
     "{{event_platform}}" => [ "Virtual platform", "Zoom" ],
+    "{{event_location}}" => [ "In-person location", "Los Angeles, CA" ],
     "{{event_month_year}}" => [ "Event month and year", "July 2026" ],
     "{{registration_close}}" => [ "Registration close date", "June 20, 2026 9:00 am PST" ]
   }.freeze
@@ -40,9 +41,17 @@ module ApplicationHelper
       .gsub("{{event_times}}", event_times_label(event) || "the scheduled time")
       .gsub("{{event_fee}}", event_fee_label(event) || "the registration fee")
       .gsub("{{event_platform}}", event_platform_label(event) || "online")
+      .gsub("{{event_location}}", event_location_label(event) || "the event location")
       .gsub("{{event_month_year}}", event&.start_date&.strftime("%B %Y") || "upcoming")
       .gsub("{{registration_close}}", event_registration_close_label(event) || "soon")
     form_label_html(text)
+  end
+
+  # True when the header contains any {{token}} placeholder, so previews (which
+  # have no event in scope) can flag that the live values come from the event.
+  def form_header_uses_tokens?(form)
+    header = form&.header.to_s
+    FORM_HEADER_TOKENS.keys.any? { |token| header.include?(token) }
   end
 
   # Event date or date range as plain text (e.g. "July 23-24, 2026"), mirroring the
@@ -88,6 +97,12 @@ module ApplicationHelper
   def event_platform_label(event)
     return unless event&.videoconference_url.present?
     event.videoconference_label.presence
+  end
+
+  # In-person location name (e.g. "Los Angeles, CA"), or nil when the event has no
+  # physical location set.
+  def event_location_label(event)
+    event&.location&.name.presence
   end
 
   # Registration close date formatted like the event show page (e.g.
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -14,6 +14,106 @@ def form_label_html(text)
...
+    "{{event_location}}" => [ "In-person location", "Los Angeles, CA" ],
+    "{{event_month_year}}" => [ "Event month and year", "July 2026" ],
+    "{{registration_close}}" => [ "Registration close date", "June 20, 2026 9:00 am PST" ]
+  }.freeze
+
+  # Render a form header, filling event-driven tokens (see FORM_HEADER_TOKENS) from
+  # the event the form is being shown for. The public registration pages pass
+  # `event:`; the standalone form preview has none, so tokens fall back to a neutral
+  # phrase. HTML is sanitized via form_label_html.
+  def form_header_html(form, event: nil)
```
maebeale, line 37: 🤖 _From Claude:_ Each token falls back to a neutral phrase rather than the raw `{{token}}` so the standalone preview (no event) and any event missing a field still read as a sentence — the public page passes the real event, the preview shows the fallback.

app/views/forms/show.html.erb
```diff
@@ -20,6 +21,12 @@
     </div>
 
     <div class="px-6 py-6">
+      <% if form_header_uses_tokens?(@form) %>
```
maebeale, line 24: 🤖 _From Claude:_ Without an event in scope here the tokens render their fallbacks, which could read as the real values — this notice tells the admin the live page fills them from the event.

db/seeds/dev/events_management.rb
```diff
@@ -34,6 +34,27 @@
...
+# the public registration page) onto the base registration form, mirroring the
+# real AWBW Facilitator Training copy. Training dates, time, fee, and platform are
+# intentionally omitted — the public page already renders those from the event
+# (the date/time block, the "Cost:" badge, and the videoconference/platform badge),
+# and the form itself already notes that fields marked * are required. The form is
+# shared across events, so event-specific dates use tokens filled at render time
+# (see form_header_html): {{event_month_year}} from the event's start date and
+# {{registration_close}} from its registration close date, rather than hard-coded
+# values. Only set when blank so admin edits survive a re-seed.
+if registration_form.header.blank?
```
maebeale, line 47: 🤖 _From Claude:_ Guarded on `blank?` so a re-seed never clobbers an admin-edited header. The copy uses only inline styles that survive `form_label_html` sanitization (verified color, font-size, background-color, border, padding, border-radius all pass).

app/views/events/public_registrations/new.html.erb
```diff
@@ -121,6 +126,7 @@
             <h3 class="flex items-center gap-2.5 text-lg font-bold text-blue-900 mb-5">
               <i class="fa-solid fa-hand-holding-heart text-accent"></i>Scholarship application
             </h3>
+            <%= render "forms/header", form: @scholarship_form, event: @event %>
```
maebeale, line 129: 🤖 _From Claude:_ This header was previously only rendered for the registration form (`@form`), so any scholarship-form intro copy was silently dropped on appended registrations. Passing `event:` keeps the `{{...}}` tokens working here too.

PR #1621, opened 2026-06-10
Handle stripe links

jmilljr24:

Stripe payments can occur inside outside of the portal. This taps into the stripe webhooks and processes external payments and refunds.

### DIFF
--- a/app/models/external_processor_payment.rb
+++ b/app/models/external_processor_payment.rb
@@ -4,10 +4,6 @@ class ExternalProcessorPayment < Payment
 
   validate :pay_charge_presence, on: :create, unless: :skip_pay_charge_validation
 
-  scope :with_metadata_key, ->(key, value) {
-    where("JSON_EXTRACT(metadata, ?) = ?", "$.#{key}", value.as_json)
-  }
-
   private
 
   def pay_charge_presence
--- a/app/models/external_processor_payment.rb
+++ b/app/models/external_processor_payment.rb
@@ -2,7 +2,7 @@ class ExternalProcessorPayment < Payment
   belongs_to :pay_charge, class_name: "Pay::Charge", optional: true
   attr_accessor :skip_pay_charge_validation
 
-  validate :pay_charge_presence, unless: :skip_pay_charge_validation
+  validate :pay_charge_presence, on: :create, unless: :skip_pay_charge_validation
 
   scope :with_metadata_key, ->(key, value) {
     where("JSON_EXTRACT(metadata, ?) = ?", "$.#{key}", value.as_json)
--- a/app/models/external_processor_payment.rb
+++ b/app/models/external_processor_payment.rb
@@ -4,6 +4,10 @@ class ExternalProcessorPayment < Payment
 
   validate :pay_charge_presence, unless: :skip_pay_charge_validation
 
+  scope :with_metadata_key, ->(key, value) {
+    where("JSON_EXTRACT(metadata, ?) = ?", "$.#{key}", value.as_json)
+  }
+
   private
 
   def pay_charge_presence
--- a/app/models/external_processor_payment.rb
+++ b/app/models/external_processor_payment.rb
@@ -1,4 +1,12 @@
 class ExternalProcessorPayment < Payment
-  belongs_to :pay_charge, class_name: "Pay::Charge"
-  # validates :pay_charge_id, presence: true
+  belongs_to :pay_charge, class_name: "Pay::Charge", optional: true
+  attr_accessor :skip_pay_charge_validation
+
+  validate :pay_charge_presence, unless: :skip_pay_charge_validation
+
+  private
+
+  def pay_charge_presence
+    errors.add(:pay_charge_id, "must exist") if pay_charge_id.blank?
+  end
 end
--- a/app/views/payments/show.html.erb
+++ b/app/views/payments/show.html.erb
@@ -44,6 +44,18 @@
         <dd class="mt-1 text-sm text-gray-900"><%= @payment.pay_charge_id %></dd>
       </div>
     <% end %>
+    <% receipt_url = @payment.metadata.dig("stripe_charge", "receipt_url") %>
+    <% if receipt_url.present? %>
+      <div>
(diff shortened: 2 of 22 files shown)


app/webhooks/stripe_charge_succeeded_processor.rb
```diff
@@ -0,0 +1,65 @@
...
+  def call(event)
+    stripe_charge = event.data.object
+    return unless stripe_charge.paid
+
+    metadata = stripe_charge.metadata || {}
+
+    return if (metadata.keys & APP_METADATA_KEYS).any?
+
+    pay_charge = Pay::Charge.find_by(processor_id: stripe_charge.id)
+    return if pay_charge && ExternalProcessorPayment.exists?(pay_charge_id: pay_charge.id)
```
jmilljr24: Pay gem runs before this processor. If the pay gem creates records for us, we return else create our Payment record.

One of the main reasons the pay-gem would not create a local record is if there is no Pay::Customer record locally associated with the stripe transaction. Any payments created within the portal will find or create a Pay::Customer but this does not happen with stripe transactions initiated outside of the portal.

The pay-gem does have an auto-reconciliation for pay customers but the documents warn against using it so I think that is a bigger discussion and bigger project that needed right now.

app/views/payments/show.html.erb
```diff
@@ -5,7 +5,14 @@
     <div class="flex gap-2">
       <% if @payment.amount_cents_remaining > 0 %>
         <%= link_to "Allocate", new_allocation_path(source_sgid: @payment.to_sgid.to_s), class: "btn btn-primary" %>
-        <%= link_to "Refund", new_refund_path(payment_sgid: @payment.to_sgid.to_s), class: "btn btn-warning" %>
+        <% if @payment.is_a?(ExternalProcessorPayment) && @payment.pay_charge_id.blank? %>
```
maebeale, line 8: why would we end up with no `pay_charge_id` sometimes?

app/views/payments/show.html.erb
```diff
@@ -44,6 +51,18 @@
         <dd class="mt-1 text-sm text-gray-900"><%= @payment.pay_charge_id %></dd>
       </div>
     <% end %>
+    <% receipt_url = @payment.metadata.dig("stripe_charge", "receipt_url") %>
+    <% if receipt_url.present? %>
+      <div>
+        <dt class="text-sm font-medium text-gray-500">Stripe Receipt</dt>
```
maebeale, line 57: oooo! nice we don't have to do receipts!

does stripe also have a one-button invoicing?

app/views/payments/show.html.erb
maebeale: nvm i think i got it

app/models/external_processor_payment.rb
```diff
@@ -1,4 +1,16 @@
 class ExternalProcessorPayment < Payment
-  belongs_to :pay_charge, class_name: "Pay::Charge"
-  # validates :pay_charge_id, presence: true
+  belongs_to :pay_charge, class_name: "Pay::Charge", optional: true
+  attr_accessor :skip_pay_charge_validation
+
+  validate :pay_charge_presence, on: :create, unless: :skip_pay_charge_validation
+
+  scope :with_metadata_key, ->(key, value) {
+    where("JSON_EXTRACT(metadata, ?) = ?", "$.#{key}", value.as_json)
```
maebeale: i think we maybe need to add a db index if we're searching json. idk if that's a thing in mysql tho.

app/webhooks/stripe_charge_succeeded_processor.rb
```diff
@@ -0,0 +1,65 @@
...
+    return unless customer_id
+
+    customer = Stripe::Customer.retrieve(customer_id)
+    customer&.email
+  rescue Stripe::StripeError
+    nil
+  end
+
+  def external_payer
+    Person.find_or_create_by!(first_name: "External", last_name: "Payer")
```
maebeale, line 57: i think this should maybe be more obvious it's not a name? 

maybe: (External to the Portal) (Payer)

eh. that's not good either. i'm ok w fake person, just want names to have parens and make sense to an admin.

app/webhooks/stripe_charge_refunded_processor.rb
```diff
@@ -0,0 +1,29 @@
...
+    (charge.refunds&.data || []).each do |stripe_refund|
+      next unless stripe_refund.status == "succeeded"
+      next if Refund.exists?(stripe_refund_id: stripe_refund.id)
+
+      Refund.create!(
+        refundable: external_payment,
+        recipient: external_payment.payer,
+        amount_cents: stripe_refund.amount,
+        method: "stripe",
+        stripe_refund_id: stripe_refund.id
```
maebeale, line 23: do we have or need unique index in the DB on stripe_refund_id ? prob good for idempotency and race conditions w db txns.

app/webhooks/stripe_charge_succeeded_processor.rb
maebeale: do we need failover things if worker isn't available or sync didn't work? 

we maybe want to have a rake tasks that we run manually after first month just to see that we've got parity between portal and stripe?

maebeale (approved):
awesome work here, @jmilljr24 !!!

app/views/payments/show.html.erb
jmilljr24: From the stripe dashboard there is invoicing and this PR would capture the payment. We would have to discuss of the stripe invoice is sufficient for what they need at this time.

app/models/external_processor_payment.rb
jmilljr24: Ill take a look.

jmilljr24: I ended up removing this as there isn't a good way with mysql. I added the stripe_payment_id directly to the Payment model. The downside is that if a new payment processor is added we will need to keep adding new processor specific id columns. All is Payment since is STI but only used by ExternalProcessorPayment. Not a huge deal and better that json full table search.

PR #1626, opened 2026-06-11
HOLD UNTIL REQUESTED: Let admins register attendees without sending them a confirmation email

maebeale:

- Staff frequently register people *on their behalf*, where the automatic "you're registered" email to the attendee is unwanted noise (and sometimes goes to someone who never asked to be emailed).
- This adds an admin-only "on behalf" mode to the public registration form that captures the full registration but suppresses the registrant-facing confirmation — while still sending the staff FYI notification.

- `PublicRegistration` gained a `send_confirmation:` flag; the controller only honors `on_behalf` for users who can manage the event, finds/creates the registrant from the form (not the admin), and keeps the admin FYI.
- Because the on-behalf attendee isn't logged in, the form now renders the full logged-out field set for admins and reveals the identity fields it normally hides for a logged-in user via a small `on-behalf` Stimulus controller, with "Register on behalf" entry points on the registrants roster and the event Form actions menu.

### UI Testing Checklist
- [ ] As an admin, open an event's registrants page and click "Register on behalf" — the toggle is pre-checked and the name/email identity fields are visible.
- [ ] Submitting in on-behalf mode creates the registration but sends **no** confirmation email to the attendee (staff FYI still arrives).
- [ ] Unchecking "Registering on someone's behalf" re-hides the identity fields; re-checking reveals them.
- [ ] As a logged-out visitor, the on-behalf toggle is not present and registration still emails a confirmation.

### Anything else to add?
- The existing admin "New registration" → confirm-page flow (opt-in email) is unchanged; this complements it for the full-form case. Covered by new request and service specs.

### DIFF
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -18,7 +18,7 @@ module Events
         return
       end
 
-      @form_fields = visible_form_fields
+      set_field_variables
       @scholarship = scholarship_mode?
       @scholarship_form = @event.scholarship_form if @scholarship
       @event = @event.decorate
@@ -45,7 +45,7 @@ module Events
         @field_errors = @field_errors.merge(FormAnswerValidator.call(scholarship_fields, scholarship_params))
       end
       if @field_errors.any?
-        @form_fields = visible_form_fields
+        set_field_variables
         @event = @event.decorate
         render :new, status: :unprocessable_content
         return
@@ -58,7 +58,10 @@ module Events
         form: @form,
         form_params: registration_params,
         scholarship_requested: @scholarship,
-        person: current_user&.person
+        # On-behalf mode finds/creates the registrant from the form rather than
+        # tying the registration to the admin filling it out.
+        person: registration_person,
+        send_confirmation: !on_behalf?
       )
 
       if result.success?
@@ -72,7 +75,7 @@ module Events
                       notice: "You have been successfully registered!"
         end
       else
-        @form_fields = visible_form_fields
+        set_field_variables
         @event = @event.decorate
         flash.now[:alert] = result.errors.join(", ")
         render :new, status: :unprocessable_content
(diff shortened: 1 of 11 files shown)


app/controllers/events/public_registrations_controller.rb
```diff
@@ -112,6 +115,36 @@ def show
 
     private
 
+    # An admin (or event owner) registering someone else: only honored for users
+    # who can manage the event, so a guest can't suppress their own confirmation.
+    def on_behalf?
```
maebeale, line 120: 🤖 _From Claude:_ Gated on `allowed_to?(:dashboard?, @event)` so a logged-out visitor can never set `on_behalf` to suppress their own confirmation — the flag is ignored unless the requester can manage the event.

app/controllers/events/public_registrations_controller.rb
```diff
@@ -112,6 +115,36 @@ def show
...
+    # logged-in registrant, so we treat it as a fresh (logged-out) submission.
+    def registration_person
+      on_behalf? ? nil : current_user&.person
+    end
+
+    # Builds @form_fields for the registration form. Admins always get the full
+    # logged-out form so they can register a not-logged-in person; the fields
+    # that are normally hidden for the logged-in admin are flagged as
+    # "behalf-only" and revealed client-side when "on behalf" is checked.
+    def set_field_variables
```
maebeale, line 136: 🤖 _From Claude:_ Admins always get the full logged-out field set because the person being registered isn't logged in. The diff against the logged-in set (`@behalf_only_field_ids`) is what the Stimulus controller reveals/hides as the toggle changes.

maebeale:
@jmilljr24 this offers an alternative to `as_visitor`. i reuse this controller in subsequent pr to offer a similar option for the bulk payment form. lmk what you think!

maebeale:
🤖 _From Claude:_ Closing — we've decided not to allow skipping the registrant confirmation email or admin "register on behalf" registration, which is this PR's entire purpose. The only change here unrelated to that feature (dropping the dead `as_visitor` param from the dashboard registration links) has since landed on `main` independently, so nothing needs to be salvaged into a separate PR.

PR #1627, opened 2026-06-11
HOLD: Add admin-only option to skip the payer confirmation email (bulk payments)

maebeale:

> **Stacked on #1647** (bulk payment confirmation + FYI emails). Review/merge that first; this PR's diff is only the admin skip-email layer. Once #1647 merges, this PR will be retargeted to `main`.

- Staff frequently submit bulk payments **on someone's behalf**, where the automatic confirmation email to the payer is unwanted noise (and sometimes reaches someone who never asked to be emailed).
- This adds an admin-only option to suppress the payer confirmation for those submissions, while still sending the staff FYI.

- `BulkPayment` gained a `send_confirmation:` flag; the controller only honors `on_behalf` for users who can manage the event (so a guest can't suppress their own confirmation) and finds/creates the payer from the form rather than the admin.
- An admin-only checkbox on the bulk payment form toggles this, and a small `on-behalf` Stimulus controller reveals the payer identity fields that are normally hidden for a logged-in admin. Added a "Bulk payment on behalf" entry to the event Form actions menu.

### UI Testing Checklist
- [ ] As an admin, open a bulk payment form — the blue "Submitting on someone's behalf / Admin only" toggle appears (hidden from logged-out visitors).
- [ ] Checking the toggle reveals the payer name/email fields; unchecking re-hides them.
- [ ] Submitting on-behalf sends **no** payer confirmation (staff FYI still arrives); a normal submission emails both.
- [ ] "Bulk payment on behalf" entry in the Form actions menu pre-checks the toggle.

### Anything else to add?
- Overlaps with PR #1626, which introduces the same `on_behalf_controller.js`; this branch carries an identical copy.
- rubocop clean; on-behalf request + service specs pass.

### DIFF
--- a/app/controllers/events/bulk_payments_controller.rb
+++ b/app/controllers/events/bulk_payments_controller.rb
@@ -12,10 +12,8 @@ module Events
     def new
       authorize! :bulk_payment, to: :new?
 
-      @form_fields = visible_form_fields
+      set_field_variables
       @event = @event.decorate
-
-      @attendee_field = @form.form_fields.find_by(field_identifier: "bulk_payment_attendees")
     end
 
     def create
@@ -25,9 +23,8 @@ module Events
 
       @field_errors = validate_required_fields
       if @field_errors.any?
-        @form_fields = visible_form_fields
+        set_field_variables
         @event = @event.decorate
-        @attendee_field = @form.form_fields.find_by(field_identifier: "bulk_payment_attendees")
         render :new, status: :unprocessable_content
         return
       end
@@ -36,7 +33,10 @@ module Events
         event: @event,
         form: @form,
         form_params: @form_params,
-        person: current_user&.person
+        # On-behalf mode finds/creates the payer from the form rather than tying
+        # the submission to the admin filling it out.
+        person: registration_person,
+        send_confirmation: !on_behalf?
       )
 
       if result.success?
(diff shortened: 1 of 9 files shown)


app/controllers/events/bulk_payments_controller.rb
```diff
@@ -48,9 +48,8 @@ def create
                       notice: "Your bulk payment information has been submitted."
         end
```
maebeale, line 49: 🤖 _From Claude:_ `on_behalf?` is gated by `allowed_to?(:dashboard?, @event)` so a logged-out visitor can never set `on_behalf=1` to suppress the payer confirmation — the param alone is not enough.

app/services/event_registration_services/bulk_payment.rb
```diff
@@ -27,6 +29,31 @@ def call
 
     private
 
+    # Emails the payer a confirmation (unless an admin suppressed it via the
+    # on-behalf option) and always notifies staff with an FYI.
+    def send_notifications(submission, person)
+      if @send_confirmation
+        payer_email = person.preferred_email.presence || field_value("payer_email")&.strip
+        if payer_email.present?
+          NotificationServices::CreateNotification.call(
```
maebeale, line 38: 🤖 _From Claude:_ Only the payer confirmation is gated by `@send_confirmation`; the staff FYI always sends. This matches PR #1626 — on-behalf submissions still notify staff, they just do not email the payer.

app/models/form_submission.rb
```diff
@@ -5,4 +5,32 @@ class FormSubmission < ApplicationRecord
   has_many :payments
 
   accepts_nested_attributes_for :form_answers
+
+  # The event this submission belongs to, resolved through the join role that
+  # matches the submission's own role (e.g. a "bulk_payment" submission maps to
+  # the event_form with role "bulk_payment").
+  def event
+    form.events.find_by(event_forms: { role: role })
```
maebeale: 🤖 _From Claude:_ Resolves the event by matching the submission `role` against the `event_forms` join role (e.g. `bulk_payment`), since `FormSubmission` has no direct event association. The mailers rely on this to render the event block.

PR #1628, opened 2026-06-11
Add license number to person profile

maebeale:
- Adds `license_number` and `license_type` fields to `Person` so admins can record a professional credential (the number and what kind it is) alongside other contact details — there was previously nowhere to store this.

- One reversible migration adds two nullable string columns (`license_number`, `license_type`) to `people`; both are permitted in `person_params`.
- Reworked the birthday/timezone row of the person edit form into two cards: birthday + timezone on the left, license number + license type on the right, so related fields read as a unit. Added a hint to the birthday field (month/day only) to balance the row.

### UI Testing Checklist
- [ ] On the person edit form, two cards appear in one row: Birthday/Time zone on the left, License number/License type on the right.
- [ ] Entering license number and type, saving, and reopening the form shows the saved values.
- [ ] For a person with no associated user, the Time zone field is absent and the left card still lays out cleanly.

### Anything else to add?
- `license_number` and `license_type` are optional free-form strings (no format/uniqueness validation, `license_type` is not a constrained enum) and are not yet shown on the profile/show page — easy to add if desired.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -364,6 +364,7 @@ class PeopleController < ApplicationController
       :best_time_to_call,
       :date_of_birth,
       :license_number,
+      :license_type,
       :bio, :notes,
       :display_name_preference,
       :pronouns,
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -363,6 +363,7 @@ class PeopleController < ApplicationController
       :street_address, :city, :state, :zip, :country, :mailing_address_type,
       :best_time_to_call,
       :date_of_birth,
+      :license_number,
       :bio, :notes,
       :display_name_preference,
       :pronouns,
--- a/app/views/people/_form.html.erb
+++ b/app/views/people/_form.html.erb
@@ -280,7 +280,7 @@
               hint: "Maximum 250 characters" %>
 
   <div class="flex flex-col md:flex-row gap-6 items-start">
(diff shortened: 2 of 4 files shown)


app/views/people/_form.html.erb
```diff
@@ -279,7 +279,7 @@
               },
               hint: "Maximum 250 characters" %>
 
-  <div class="flex flex-col md:flex-row gap-6 items-end">
+  <div class="flex flex-col md:flex-row gap-6 items-start">
```
maebeale: 🤖 _From Claude:_ Switched this row from `items-end` to `items-start`. With `items-end` the Time zone hint made that block taller, so Birthday's selects dropped to line up with the bottom — now all three fields align along the top.

db/migrate/20260610040000_add_license_number_to_people.rb
```diff
@@ -0,0 +1,9 @@
+class AddLicenseNumberToPeople < ActiveRecord::Migration[8.1]
+  def up
+    add_column :people, :license_number, :string
```
maebeale, line 3: 🤖 _From Claude:_ Free-form optional string, no format/uniqueness validation — let me know if license numbers should be constrained.

app/controllers/people_controller.rb
```diff
@@ -363,6 +363,8 @@ def person_params
       :street_address, :city, :state, :zip, :country, :mailing_address_type,
       :best_time_to_call,
       :date_of_birth,
+      :license_number,
```
jmilljr24, line 366: @maebeale Do people have more than one license?

PR #1629, opened 2026-06-11
Add follow-up flag to comments

maebeale:
- Admins need a lightweight way to mark a comment as needing follow-up so important notes don't get lost.
- A `flagged` boolean on comments makes those notes visually stand out and filterable (a `Comment.flagged` scope is added).

- Added a reversible migration plus the `flagged` column, model scope, and `:flagged` permit across all commentable controllers' nested attributes.
- Surfaced flagging in three places: a flag checkbox when adding a comment, an inline Turbo-powered toggle on comment lists (new `update` action + `flag_toggle` partial + turbo stream), and via nested attributes on parent forms — all gated behind the existing admin `manage?` policy.

### UI Testing Checklist
- [ ] Add a comment with the flag checked and confirm it persists as flagged.
- [ ] Toggle the inline flag on a comment list and confirm it updates without a full page reload.
- [ ] Edit a parent record's comments and confirm flag state saves via nested attributes.

### Anything else to add?
- Specs cover the model scope, create/update requests, and nested-attribute flagging.

### DIFF
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -1,5 +1,5 @@
 class CommentPolicy < ApplicationPolicy
-  alias_rule :index?, :create?, to: :manage?
+  alias_rule :index?, :create?, :update?, to: :manage?
 
   def manage?
     admin?
--- a/app/views/comments/_flag_toggle.html.erb
+++ b/app/views/comments/_flag_toggle.html.erb
@@ -1,0 +1,8 @@
+<%= button_to polymorphic_path([ commentable, comment ]),
+              method: :patch,
+              params: { comment: { flagged: !comment.flagged } },
+              form: { id: dom_id(comment, :flag) },
+              class: "leading-none",
+              title: comment.flagged? ? "Flagged for follow-up — click to clear" : "Flag for follow-up" do %>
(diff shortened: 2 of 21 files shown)


app/policies/comment_policy.rb
```diff
@@ -1,5 +1,5 @@
 class CommentPolicy < ApplicationPolicy
-  alias_rule :index?, :create?, to: :manage?
+  alias_rule :index?, :create?, :update?, to: :manage?
```
maebeale, line 2: 🤖 _From Claude:_ Flag toggling is gated behind the existing admin `manage?` rule by aliasing `update?` to it — non-admins can't flag.

app/views/comments/_flag_toggle.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= button_to polymorphic_path([ commentable, comment ]),
```
maebeale, line 1: 🤖 _From Claude:_ `button_to` with `polymorphic_path([commentable, comment])` hits the nested comment `update` route; the stable `dom_id(comment, :flag)` form id lets the turbo stream replace just this toggle.

PR #1631, opened 2026-06-11
Restore one-click registration for signed-in members on public events

maebeale:
Since the custom-forms refactor, every publicly-registerable event is required to carry a registration form (logged-out visitors need it), which forced signed-in members through the full form and silently removed their one-click "Register" — the reported regression. This restores one-click for members without taking the form away from guests.

Added a per-event `signed_in_one_click_enabled` flag and an `Event#one_click_for_signed_in?` decision (one-click when no registration form is linked, or when the flag opts members out); the event card and registration section now branch on it. While in the Registration Form settings card, I also clarified its layout (grouped "Registration options", an explicit "Change" affordance for the public-registration status) and added preview links for the scholarship and bulk-payment add-ons to match the registration form's existing preview link.

### UI Testing Checklist
- [ ] Signed-in user on a public event with a registration form: "Register" goes to the form
- [ ] Enable "Signed-in users register in one click": signed-in "Register" now one-clicks; logged-out visitors still get the form
- [ ] Event with no registration form: signed-in "Register" one-clicks
- [ ] Registration Form settings card: scholarship/bulk-payment preview links appear only when each add-on is enabled and open the right page

### Anything else to add?
Covered by new model specs (`Event#one_click_for_signed_in?`) and request specs (register button rendering + add-on preview links). The toggle lives in the Registration Form card next to the scholarship/bulk-payment options.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -71,6 +71,14 @@ class Event < ApplicationRecord
     forms.find_by(event_forms: { role: "registration" })
   end
 
+  # Whether a signed-in user should register in one click rather than being
+  # routed to the registration form. True when no registration form is linked,
+  # or when an admin has explicitly opted members out of the form. A linked form
+  # (even one without fields yet) routes signed-in users to the form.
+  def one_click_for_signed_in?
+    signed_in_one_click_enabled? || registration_form.nil?
+  end
+
   def scholarship_form
     forms.find_by(event_forms: { role: "scholarship" })
   end
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -335,31 +335,6 @@
         <div id="registration_form_section" class="border border-gray-300 p-4 rounded-b-md" data-reveal-section-target="content">
           <div class="grid grid-cols-1 md:grid-cols-2 gap-6">
             <div>
-              <p class="text-sm mb-2">
-                Public registration:
-                <% if @event.public_registration_enabled? %>
-                  <a href="#public_registration_setting"
-                     class="text-green-700 font-medium hover:underline"
-                     data-controller="anchor-highlight"
-                     data-action="click->anchor-highlight#flash"
-                     data-anchor-highlight-target-param="public_registration_setting">Enabled</a>
-                <% else %>
-                  <a href="#public_registration_setting"
-                     class="text-gray-500 font-medium hover:underline"
-                     data-controller="anchor-highlight"
-                     data-action="click->anchor-highlight#flash"
-                     data-anchor-highlight-target-param="public_registration_setting">Disabled</a>
-                <% end %>
-              </p>
-              <% if @event.event_forms.registration.exists? %>
-                <%= link_to "View public registration page",
-                            new_event_public_registration_path(@event),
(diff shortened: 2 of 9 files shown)


app/models/event.rb
```diff
@@ -71,6 +71,14 @@ def registration_form
     forms.find_by(event_forms: { role: "registration" })
   end
 
+  # Whether a signed-in user should register in one click rather than being
+  # routed to the registration form. True when no registration form is linked,
+  # or when an admin has explicitly opted members out of the form. A linked form
+  # (even one without fields yet) routes signed-in users to the form.
+  def one_click_for_signed_in?
+    signed_in_one_click_enabled? || registration_form.nil?
+  end
```
maebeale, line 80: 🤖 _From Claude:_ Keys on whether a registration form is *linked*, not its field count — an empty linked form still routes signed-in users to the form. The `signed_in_one_click_enabled` flag is the admin override that lets members skip a form that exists for logged-out visitors.

maebeale, line 80: 🤖 _From Claude:_ Keys on whether a registration form is *linked*, not its field count — an empty linked form still routes signed-in users to the form. The signed_in_one_click_enabled flag is the admin override that lets members skip a form that exists for logged-out visitors.

app/views/events/_form.html.erb
```diff
@@ -384,19 +359,74 @@
...
+              </div>
+              <div>
+                <label class="flex items-center gap-2 cursor-pointer">
+                  <%= check_box_tag "event[scholarship_enabled]", "1", @event.scholarship_form.present? %>
+                  <span class="text-sm font-medium text-gray-700">Enable scholarship application</span>
+                </label>
+                <p class="text-xs text-gray-500 mt-1">Add a scholarship form so registrants can request financial assistance.</p>
+                <% if @event.scholarship_form.present? %>
+                  <%= link_to new_event_public_registration_path(@event, scholarship_requested: true),
+                              class: "inline-flex items-center gap-1 text-xs text-blue-700 hover:text-blue-900 underline mt-1",
```
maebeale, line 409: 🤖 _From Claude:_ Scholarship has no standalone page — it renders into the public registration page via the scholarship_requested=true param (the "combo" form), so the preview points there. Bulk payment has its own page.

maebeale:
Screenshot 2026-06-11 at 7 59 21 PM

maebeale:
@jmilljr24 this should reinstate the functionality for one-click for users and form for visitors (if public). you good w this?

jmilljr24 (approved):
I think it's a fine option for now.

I know we had some discussion on if/how we were to keep this once we added more registration flows but they seem to like it so it should stay in some form.

In my head (not sure if we talked about it) I thought that broad idea was that when you click to registered, the system would check the user against the form for any question we don't have on file for you yet. For the short for of just name etc, a logged in used would just be a one-click. If its a longer form but all questions are one time only, again it could be a one-click for a user who has already done all those questions. 

Having this explicit flag just muddies the water a bit but can certainly be refactored in the future.

maebeale:
@jmilljr24 yes, great points, and agree. my thinking was launch this now so current compass event can have one-click super clearly, and then we decouple one-click from on-file stuff later. TYY for raising this!

PR #1633, opened 2026-06-12
Refine registration close date display and default

maebeale:
- Make the registration deadline read naturally in form copy and faster for staff to set.

- Removed the filler "it only takes a minute." line from the public registration card header.
- Reformatted the `{{registration_close}}` form-header token to `July 20th at 9am PST` — ordinal day, no year, compact time (minutes shown only when not on the hour) — via `event_registration_close_label`.
- Defaulted the event-edit registration close datetime field to **9am on the Monday before the event's start date** (`event_registration_close_default`), falling back to two days out at 9am for new events that have no start date yet. The field is a `datetime-local`, so both date and time stay editable.
- Updated the token reference example shown on the form editor and added helper specs (compact-time, not-on-the-hour, and the Monday-before default including the Monday-start edge case).

### Anything else to add?
- The interpolated label shows the **actual** chosen close time (year omitted); it is not pinned to 9am — only the edit field *defaults* to 9am, so an admin's chosen time is always what registrants see.
- "Monday before" resolves to the prior Monday when the event itself starts on a Monday.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -27,7 +27,7 @@ module ApplicationHelper
     "{{event_platform}}" => [ "Virtual platform", "Zoom" ],
     "{{event_location}}" => [ "In-person location", "Los Angeles, CA" ],
     "{{event_month_year}}" => [ "Event month and year", "July 2026" ],
-    "{{registration_close}}" => [ "Registration close date", "June 20, 2026 9:00 am PST" ]
+    "{{registration_close}}" => [ "Registration close date", "July 20th at 9am PST" ]
   }.freeze
 
   # Render a form header, filling event-driven tokens (see FORM_HEADER_TOKENS) from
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -105,13 +105,17 @@ def event_location_label(event)
...
 
-  # Registration close date formatted like the event show page (e.g.
-  # "June 20, 2026 9:00 am PST"), or nil when there's no close date.
+  # Registration close date for form-header interpolation, e.g.
+  # "July 20th at 9am PST": ordinal day, no year, compact time (minutes only
+  # when not on the hour). Nil when there's no close date.
   def event_registration_close_label(event)
     close = event&.registration_close_date
     return unless close
     local = close.in_time_zone(Time.zone)
```
maebeale, line 114: 🤖 _From Claude:_ Builds the compact "9am"/"2:30pm" form: minutes are appended only when not on the hour, mirroring the existing `event_times_label` convention but with no space before am/pm.

app/views/events/_form.html.erb
```diff
@@ -117,11 +117,12 @@
                 "Registration close time",
                 class: "block font-medium mb-1" %>
 
+          <% registration_close_default = @event.registration_close_date || 2.days.from_now.change(hour: 9, min: 0) %>
```
maebeale: 🤖 _From Claude:_ Default is 9am two days out (matching the autoshow label's existing 2-days-from-now fallback). The field stays a fully-editable datetime, so this only prefills an empty value.

app/helpers/application_helper.rb
```diff
@@ -105,13 +105,26 @@ def event_location_label(event)
...
+    time += local.strftime("%P")
+    "#{local.strftime("%B")} #{local.day.ordinalize} at #{time} #{local.strftime("%Z")}"
+  end
+
+  # Default registration close datetime suggested on the event form: 9am on the
+  # Monday before the event's start date. New events without a start date yet
+  # fall back to two days out at 9am.
+  def event_registration_close_default(event)
+    start = event&.start_date
+    base = start ? (start.in_time_zone(Time.zone) - 1.day).beginning_of_week(:monday) : 2.days.from_now
```
maebeale, line 126: 🤖 _From Claude:_ `(start - 1.day).beginning_of_week(:monday)` yields the same-week Monday for most events, and the *prior* Monday when the event starts on a Monday — so the default is always strictly before the start date.

PR #1635, opened 2026-06-12
Make field subtitles editable in the form editor

maebeale:
- Admins had no way to set a field/section **subtitle**. The `hint_text` column existed and already rendered on public forms, but it wasn't editable.

- Added a "Subtitle" input to the field editor — inline under the input-type dropdown for fields (aligned beneath it) and in the section-header options panel.
- Section-header subtitles now render under the heading on the public registration form and the preview; regular-field subtitles already rendered.
- No migration or controller change — `hint_text` was already permitted. Added request specs for the editor input, saving, and rendering.

### UI Testing Checklist
- [ ] Expand a field's options — the Subtitle box sits under the input-type dropdown.
- [ ] Section header — subtitle editable in its options panel; renders under the heading on the public form and preview.
- [ ] Set subtitles, save, confirm they persist and render.

### Anything else to add?
- This is the first of three stacked PRs split by topic. The next two build on it: **form preview** and **form builder UI**.
- Subtitles are plain text (not rich HTML like section names) — say the word if HTML/links should be supported.

### DIFF
--- a/app/views/forms/_form_field_fields.html.erb
+++ b/app/views/forms/_form_field_fields.html.erb
@@ -52,7 +52,10 @@
         </button>
       </div>
 
-      <div class="hidden mt-2 flex items-center" data-field-options-target="panel">
+      <div class="hidden mt-2 flex flex-col gap-2" data-field-options-target="panel">
+        <%= f.text_area :hint_text, rows: 2,
+              class: "w-full text-sm text-gray-600 rounded border-gray-300 shadow-sm px-2 py-1",
+              placeholder: "Subtitle" %>
         <%= link_to_remove_association "Remove", f,
               class: "text-sm text-gray-400 hover:text-red-600 underline shrink-0 ml-auto" %>
       </div>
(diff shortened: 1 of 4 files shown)


app/views/forms/_form_field_fields.html.erb
```diff
@@ -89,15 +92,17 @@
...
           <% if field.multiple_choice? %>
             <button type="button"
-                    class="flex items-center gap-1.5 text-sm font-medium text-gray-600 hover:text-gray-800 cursor-pointer"
+                    class="min-w-0 flex-[3_1_10rem] flex items-center gap-1.5 text-sm font-medium text-gray-600 hover:text-gray-800 cursor-pointer"
                     data-action="answer-options#toggle">
               <i class="fa-solid fa-chevron-right text-xs transition-transform" data-answer-options-target="icon"></i>
               Answer options
               <span class="text-xs text-gray-400">(<%= field.form_field_answer_options.size %>)</span>
             </button>
+          <% else %>
```
maebeale: 🤖 _From Claude:_ The empty spacer + matching `flex-[3_1_10rem]`/`flex-[2_1_9rem]` bases (and `gap-2`) are what align the Subtitle box directly under the input-type dropdown in the row above. For multiple-choice fields the "Answer options" button doubles as that spacer.

app/frontend/javascript/controllers/field_options_controller.js
```diff
@@ -13,5 +13,7 @@ export default class extends Controller {
   expandedValueChanged() {
     this.panelTarget.classList.toggle("hidden", !this.expandedValue)
     this.iconTarget.classList.toggle("rotate-180", this.expandedValue)
+    // Tint the whole row while open so it's clear which field you're editing.
+    this.element.closest(".nested-fields")?.classList.toggle("bg-amber-50", this.expandedValue)
```
maebeale: 🤖 _From Claude:_ Tints the whole `.nested-fields` row (the controller element is the inner content div), so the active field stands out from the list. The row is outside the controller scope, so `closest` is the idiomatic reach here.

app/helpers/forms_helper.rb
```diff
@@ -0,0 +1,20 @@
+module FormsHelper
+  # The form editor's top-right links. The standalone "Preview form" is always
+  # available. When we also know which event the admin came from, we add a
+  # "View form" link to the real public-facing registration page — the preview
+  # can't fill header tokens or run conditional logic without an event.
+  def form_editor_view_links(form, event)
+    link_class = "text-sm text-gray-500 hover:text-gray-700 px-2 py-1"
+    preview = link_to "Preview", form_path(form), class: link_class
+    return preview unless event
```
maebeale: 🤖 _From Claude:_ "Preview" (standalone) is always available; "View" only appears when we know the originating event, since the live registration page needs an event to fill header tokens and run conditional logic.

PR #1613, opened 2026-06-10
Add optional form header and restyle form builder

maebeale:
Closes [link an issue or remove this line]

### What is the goal of this PR and why is this important?
- Adds an optional **form header**: an admin-authored HTML intro shown directly under the title on the form preview and the public registration pages, so organizers can give context or instructions without code changes.
- Restyles the **form builder (edit) page** so it reads at a glance instead of as a stack of disconnected white cards on a blank page.

### How did you approach the change?
- **Header feature**
  - New `forms.header` text column (`AddHeaderToForms` migration, reversible with guards).
  - Permit `:header` in `form_params`; updated the Brakeman ignore fingerprint for the pre-existing "admin only" `:role` mass-assignment entry (the permit line changed).
  - Shared `forms/_header` partial renders the intro via `form_label_html` when present; wired into the form preview (`forms/show`) and both public registration views (`new`/`show`).
- **Form builder restyle** (`forms/edit`)
  - Nested all sections inside a single full-width dark-blue banner; widened it to fill the blue page background.
  - Pulled the events/submissions stats up into the banner header (top-right) and removed the distracting white strip.
  - Unified the section header color (Form settings / Form header) to one blue.
  - Larger, bold **Form name** input.
  - Top nav links restyled to the standard admin gray; **Delete** moved out of the top text-links into the bottom actions row as a `btn-danger-outline`, inside a white actions card alongside Save/Cancel.

### UI Testing Checklist
- [ ] Form preview (`/forms/:id`) shows the header intro under the title when set, nothing when blank
- [ ] Public registration `new` and `show` pages render the header intro
- [ ] Editing/saving a form persists the header field
- [ ] Form builder banner spans full width; stats appear in the header; Delete works from the bottom row (only when no event forms are connected)

### Anything else to add?
- The `_stats` partial is intentionally left untouched (still used on the light-background edit-sections page); the dark header variant is inlined in `forms/edit`.
- Affected request specs pass (`forms_spec`, `public_registrations_spec`), RuboCop and Brakeman are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1636, opened 2026-06-12
Use the rhino editor for the form header field

maebeale:

- The optional form header was a plain-HTML `<textarea>` (added in #1613), forcing admins to hand-write markup like `<strong>…</strong>`.
- This switches it to the **rhino rich-text editor** — the same one already used for Event headers — so editing the form intro matches the rest of the app and admins never touch raw HTML.

- Followed the existing Event convention exactly:
  - `Form` now declares `has_rich_text :rhino_header`.
  - The form builder edit view renders `rhino_editor(f, :header)` in place of the textarea.
  - `forms_controller` permits `:rhino_header` instead of `:header`.
  - The `forms/_header` partial (used on the preview, public registration `new`/`show`) renders `form.rhino_header` as `prose` rich text.
- Migration `SwitchFormHeaderToRichText` copies any existing `header` HTML into the `rhino_header` ActionText record, then drops the now-unused `forms.header` column. The column was one day old (added in #1613) with no seed/factory data, so it's removed rather than left orphaned; the migration is reversible.
- Refreshed the Brakeman ignore baseline: the pre-existing `PermitAttributes` warning on `form_params` is fingerprinted on the code string, so renaming the permitted attribute required re-pointing the existing ignore entry.

### UI Testing Checklist

- [ ] In the form builder, the **Form header** card shows the rich-text editor (not a textarea); bold/italic/links/lists work.
- [ ] Saving a header persists and the editor re-opens with the saved content.
- [ ] The header renders correctly under the title on the form **preview**, and the **public registration** new/show pages.
- [ ] A form with no header renders nothing in that slot.

### Anything else to add?

- Left the `form_label_html` helper in place — it's still used for form-field and section-header labels elsewhere.
- Specs in `spec/requests/forms_spec.rb` updated to exercise `rhino_header` / the rich-text editor (36 examples, 0 failures); `ai/lint` clean.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -117,13 +117,13 @@
     {
       "warning_type": "Mass Assignment",
       "warning_code": 105,
-      "fingerprint": "ed621305ecb26c24fe0cfb0b55955439f0d2a7d0e7ff19e5911657f47bcb3ff1",
+      "fingerprint": "0ad5e5e590fdaeea9c44b9733922828022e23817520e96117d4828ff29825768",
       "check_name": "PermitAttributes",
       "message": "Potentially dangerous key allowed for mass assignment",
       "file": "app/controllers/forms_controller.rb",
       "line": 121,
       "link": "https://brakemanscanner.org/docs/warning_types/mass_assignment/",
-      "code": "params.require(:form).permit(:name, :role, :header, :hide_answered_person_questions, :hide_answered_form_questions, :form_fields_attributes => ([:id, :name, :answer_type, :required, :hint_text, :field_identifier, :section, :position, :visibility, :one_time, :width, :min_words, :max_characters, :_destroy, { :form_field_answer_options_attributes => ([:id, :option_name, :_destroy]) }]))",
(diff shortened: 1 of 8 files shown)


db/migrate/20260609180000_switch_form_header_to_rich_text.rb
```diff
@@ -0,0 +1,28 @@
+class SwitchFormHeaderToRichText < ActiveRecord::Migration[8.1]
+  # The optional form header moved from a plain-HTML `header` text column to an
+  # ActionText rich-text field (`rhino_header`), edited with the rhino editor.
+  def up
+    if column_exists?(:forms, :header)
+      execute(<<~SQL)
+        INSERT INTO action_text_rich_texts (name, body, record_type, record_id, created_at, updated_at)
+        SELECT 'rhino_header', header, 'Form', id, NOW(), NOW()
```
maebeale, line 8: 🤖 _From Claude:_ Copies existing plain-HTML headers into the `rhino_header` ActionText record before dropping the column. The column was one day old (#1613) with no seed/factory data, so this is belt-and-suspenders for any local data; `down` restores it.

config/brakeman.ignore
```diff
@@ -117,13 +117,13 @@
     {
       "warning_type": "Mass Assignment",
       "warning_code": 105,
-      "fingerprint": "ed621305ecb26c24fe0cfb0b55955439f0d2a7d0e7ff19e5911657f47bcb3ff1",
+      "fingerprint": "0ad5e5e590fdaeea9c44b9733922828022e23817520e96117d4828ff29825768",
```
maebeale, line 120: 🤖 _From Claude:_ Re-pointed (not added) this existing ignore entry — the `PermitAttributes` warning on `:role` is fingerprinted on the permit code string, so renaming `:header`→`:rhino_header` changed its fingerprint.

PR #1637, opened 2026-06-12
Show and toggle scholarship status on event recipients roster

maebeale:

- Event organizers reviewing the recipients roster (`events/recipients`) had no visibility into scholarship status — who was awarded, for how much, who funded it, or whether their tasks were complete
- They also had no way to finalize or unwind an award without leaving the page to edit the scholarship record
- This surfaces that status inline and lets organizers flip a recipient's tasks-completed state directly from the roster

- **Page-wide show/hide switch** — a Stimulus controller (`scholarship-status-toggle`) toggles every recipient card's scholarship status panel at once, keeping the switch label, track, knob, and `aria-checked` in sync. Status stays visible regardless of a card's expand/collapse state
- **Status panel** — shows the awarded amount, the funding donor's name when the scholarship is drawn from a grant (omitted when grant-free), and a tasks-completed badge
- **Inline tasks toggle** — for admins the badge is a `button_to` hitting a new `PATCH /scholarships/:id/toggle_tasks`; the existing model `after_update` allocates the award when complete and de-allocates it when outstanding, so one toggle finalizes or unwinds the award. Responds with a Turbo Stream that swaps just the status pill in place; falls back to a redirect for non-Turbo requests. Non-admins see a static badge
- **N+1 fix** — `EventDashboard#scholarship_by_recipient` now eager-loads `grant: :donor` for the donor name
- **Seeds** — reworked dev scholarship seeds so most event-allocated awards draw from a parent grant (mix of org and individual polymorphic donors), exercising both the donor and no-donor cases on the recipients page

### Anything else to add?

- New request specs cover the status panel, donor line presence/absence, the outstanding flag, authorization, and both the Turbo-stream and redirect responses of `toggle_tasks`

### DIFF
--- a/app/controllers/scholarships_controller.rb
+++ b/app/controllers/scholarships_controller.rb
@@ -1,5 +1,5 @@
 class ScholarshipsController < ApplicationController
-  before_action :set_scholarship, only: [ :show, :edit, :update, :destroy ]
+  before_action :set_scholarship, only: [ :show, :edit, :update, :destroy, :toggle_tasks ]
   before_action :set_grant, only: [ :new, :create ]
 
   def show
@@ -86,10 +86,30 @@ class ScholarshipsController < ApplicationController
     redirect_to scholarship_return_path(grant), notice: "Scholarship removed."
   end
 
+  # Flip a recipient's tasks-completed state from the event recipients roster.
+  # The model's after_update allocates the award when complete and de-allocates
+  # it when outstanding, so this single toggle finalizes or unwinds the award.
+  def toggle_tasks
+    authorize! @scholarship, to: :update?
+    @scholarship.update!(tasks_completed: !@scholarship.tasks_completed?)
+
+    respond_to do |format|
+      format.turbo_stream
+      format.html { redirect_back fallback_location: recipients_fallback_path, notice: "Scholarship updated." }
+    end
+  end
+
   private
 
   def set_scholarship
     @scholarship = Scholarship.find(params[:id])
+  end
+
+  # The recipients roster the tasks toggle is operated from — used as the
+  # fallback when there's no referer to return to.
+  def recipients_fallback_path
+    event = @scholarship.allocation&.allocatable.try(:event)
+    event ? recipients_event_path(event) : root_path
   end
 
   def set_grant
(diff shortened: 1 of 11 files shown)


app/controllers/scholarships_controller.rb
```diff
@@ -86,12 +86,32 @@ def destroy
     redirect_to scholarship_return_path(grant), notice: "Scholarship removed."
   end
 
+  # Flip a recipient's tasks-completed state from the event recipients roster.
+  # The model's after_update allocates the award when complete and de-allocates
+  # it when outstanding, so this single toggle finalizes or unwinds the award.
+  def toggle_tasks
+    authorize! @scholarship, to: :update?
```
maebeale, line 93: 🤖 _From Claude:_ One toggle both flips `tasks_completed` and (via the model's `after_update`) allocates or de-allocates the award — no separate allocation step needed here.

app/services/event_dashboard.rb
```diff
@@ -100,7 +100,7 @@ def scholarship_shoutouts
   # decide whether to flag a registrant as a scholarship recipient. First
   # scholarship wins if a person has several.
   def scholarship_by_recipient
-    @scholarship_by_recipient ||= scholarships.group_by(&:recipient_id).transform_values(&:first)
+    @scholarship_by_recipient ||= scholarships.includes(grant: :donor).group_by(&:recipient_id).transform_values(&:first)
```
maebeale, line 103: 🤖 _From Claude:_ Eager-loads `grant: :donor` to avoid an N+1 when the recipients page renders each scholarship's funding donor name.

app/views/scholarships/_tasks_status.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    everyone else it's a static badge. Wrapped in a stable id so the toggle can
+    swap just this pill in place over Turbo. %>
+<% completed = scholarship.tasks_completed? %>
+<% pill_classes = class_names(
+     "inline-flex items-center gap-1.5 rounded-full px-2.5 py-0.5 text-xs font-medium",
+     "bg-green-50 text-green-700" => completed,
+     "bg-amber-50 text-amber-700" => !completed
+   ) %>
+<span id="<%= dom_id(scholarship, :tasks_status) %>" class="inline-block">
+  <% if allowed_to?(:update?, scholarship) %>
```
maebeale, line 13: 🤖 _From Claude:_ Stable `dom_id` wrapper so the Turbo Stream can replace just this pill in place after a toggle.

PR #1640, opened 2026-06-12
Distinguish partial payments in manage registrants status column

maebeale:
- On the **manage registrants** page, a registrant who made a *partial* payment looked identical to one who had paid nothing — both showed only `$X due`.
- The amount already paid was hidden in a hover `title` tooltip, which is invisible on touch devices and easy to miss.
- Admins had no at-a-glance way to tell who had started paying versus who owed the full amount.

- Added an `EventRegistration#partially_paid?` predicate (non-zero allocations that don't yet cover the cost).
- Gave the status badge a distinct third state:
  - **Paid in full** — green, `fa-circle-check`, "Paid"
  - **Partial** — blue, `fa-circle-half-stroke`, "Partial · $X due"
  - **Nothing paid** — amber, `fa-circle-exclamation`, "$X due"

### UI Testing Checklist
- [ ] Registrant with no payment shows amber "$X due"
- [ ] Registrant with a partial payment shows blue "Partial · $X due"
- [ ] Fully-paid registrant shows green "Paid"
- [ ] Free events (cost 0) show no payment badge

### Anything else to add?
- Covered by request specs (all three badge states render through the real manage page) and model specs for `partially_paid?`.
- Considered showing both amounts inline ("$6 due ($4 paid)") but went with the more compact "Partial · $X due" label.

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -208,6 +208,10 @@ class EventRegistration < ApplicationRecord
     allocations_sum >= event.cost_cents.to_i
   end
 
+  def partially_paid?
+    !paid_in_full? && allocations_sum.to_i.positive?
+  end
+
   def joinable?
     active? && paid?
   end
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -208,8 +208,12 @@ class EventRegistration < ApplicationRecord
     allocations_sum >= event.cost_cents.to_i
   end
 
+  def payments_sum
+    allocations.where(source_type: Payment.polymorphic_name).sum(:amount)
+  end
+
   def partially_paid?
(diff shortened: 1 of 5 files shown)


app/models/event_registration.rb
```diff
@@ -115,6 +115,10 @@ def paid_in_full?
     allocations_sum >= event.cost_cents.to_i
   end
 
+  def partially_paid?
+    !paid_in_full? && allocations_sum.to_i.positive?
```
maebeale: 🤖 _From Claude:_ `positive?` here is what excludes the zero-paid case — without it, an unpaid registration (sum 0) would read as partial. Free events short-circuit via `paid_in_full?` returning true.

PR #1641, opened 2026-06-12
Add editable descriptions for workshop setting options

maebeale:

- The workshop setting options (Clinical setting, Outreach program, Residential program, …) are short labels whose meaning isn't always obvious to registrants.
- This gives each setting an optional **description** — the clarifying examples — shown as muted subtext under the option on the public registration form.
- Keeps the option *label* a clean, clickable phrase while the examples live in their own editable field, rather than being baked into the option name.

- Added a nullable `description` column to `categories` (reversible migration).
- Extended the shared `dynamic_form_field_options` helper to return `[ label, value, description ]` tuples for category-backed fields; the public form's radio and checkbox branches render the description as subtext when present.
- Made `description` editable per category from the Categories admin (form field + show page) and permitted in strong params.
- Seeded the canonical 14 workshop settings with clean names + descriptions, reconciling existing data non-destructively (renumber positions safely; unpublish — never delete — settings no longer offered, so historical taggings survive).
- Updated the `FormBuilderService` hard-coded option names to match.

### Layout

- Each option renders as a checkbox/name row (vertically centered, so the box no longer floats above the label) with the description stacked below, indented to line up under the name.
- Abbreviated examples drop the trailing period after "etc".
- Widened the registration and confirmation containers from `max-w-2xl` to `max-w-3xl` for a bit more breathing room.

### Relationship to #1634

- #1634 landed the same canonical 14 settings but with the examples **baked into the option name** (e.g. `"Clinical setting (community mental health, outpatient, etc.)"`).
- This PR builds on that refactor and moves the parentheticals out of the names into the new `description` column, so labels stay clean and the examples are separately editable.

### Anything else to add?

- Descriptions also render for category-backed fields switched to single-choice (radio), for parity.
- New request spec covers the description rendering; main's dynamic-option radio/checkbox specs and the helper specs pass.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -141,8 +141,9 @@ module ApplicationHelper
     "primary_age_group" => "AgeRange"
   }.freeze
 
-  # Returns the selectable options for a form field as [ label, value ] pairs,
-  # or nil when the field has no dynamic source (callers fall back to the
+  # Returns the selectable options for a form field as [ label, value, description ]
+  # tuples (description is nil for sectors and present only for categories that have
+  # one), or nil when the field has no dynamic source (callers fall back to the
   # field's own stored answer options). Shared by the public form's radio and
   # checkbox rendering so a dynamic field renders the same options regardless
   # of which single/multiple choice type it was set to.
@@ -152,7 +153,7 @@ module ApplicationHelper
       Sector.published.order(:name).map { |sector| [ sector.name, sector.id.to_s ] }
     when *DYNAMIC_FIELD_CATEGORY_TYPES.keys
       type = CategoryType.find_by(name: DYNAMIC_FIELD_CATEGORY_TYPES[field.field_identifier])
-      (type&.categories&.published&.order(:position, :name) || []).map { |category| [ category.name, category.id.to_s ] }
+      (type&.categories&.published&.order(:position, :name) || []).map { |category| [ category.name, category.id.to_s, category.description ] }
     end
   end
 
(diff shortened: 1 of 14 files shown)


db/seeds.rb
```diff
@@ -304,27 +304,43 @@ def find_or_create_by_name!(klass, name, **attrs, &block)
...
 end
+
+workshop_settings.each_with_index do |(name, description), index|
+  cat = workshop_settings_type.categories.where("LOWER(name) = LOWER(?)", name).first
+  cat ||= workshop_settings_type.categories.create!(name: name)
+  cat.update!(name: name, published: true, description: description)
+  cat.update_columns(position: index + 1)
+end
+
+# Hide settings that are no longer offered without destroying historical taggings.
```
maebeale, line 347: 🤖 _From Claude:_ Reconcile is intentionally non-destructive: settings dropped from the canonical list are unpublished, never deleted, so existing registrant taggings (CategorizableItems) survive. Positions are parked above the range first to avoid tripping the unique [category_type_id, position] index mid-renumber.

app/helpers/application_helper.rb
```diff
@@ -152,7 +153,7 @@ def dynamic_form_field_options(field)
       Sector.published.order(:name).map { |sector| [ sector.name, sector.id.to_s ] }
     when *DYNAMIC_FIELD_CATEGORY_TYPES.keys
       type = CategoryType.find_by(name: DYNAMIC_FIELD_CATEGORY_TYPES[field.field_identifier])
-      (type&.categories&.published&.order(:position, :name) || []).map { |category| [ category.name, category.id.to_s ] }
+      (type&.categories&.published&.order(:position, :name) || []).map { |category| [ category.name, category.id.to_s, category.description ] }
```
maebeale, line 156: 🤖 _From Claude:_ Added a third tuple element (description) here rather than a separate lookup so the public form gets it for free in both the radio and checkbox branches. Sectors return 2-tuples, so the destructured `option_description` is simply nil for them.

PR #1643, opened 2026-06-12
Extract FormAnswerValidator and drop FormField position default_scope

maebeale:

Follow-up to the custom-forms refactor. Two fragile spots left behind:

- **Duplicated answer validation** — the public-registration and bulk-payment controllers each carried their own inline copy of the same presence / integer / email / min-words / max-characters checks, free to drift apart.
- **Implicit `default_scope`** — `FormField` carried `default_scope { order(position: :desc) }`, which silently reversed field order anywhere it wasn't `unscoped`, leaked into unrelated queries, and made ordering a property of the model rather than the call site that needs it.

- Add `FormAnswerValidator` service (`app/services/form_answer_validator.rb`): presence (required), integer/email format, and per-field min-words/max-characters. Returns `{ field_id => error_message }` so results from multiple forms (registration + scholarship) merge cleanly. Group headers are skipped; other exclusions (e.g. bulk payment's nested attendees) are the caller's responsibility.
- Both controllers now delegate to it — bulk payments rejects the `bulk_payment_attendees` field before calling, since that nested field is validated separately.
- Add `FormField#email_field?` backed by an explicit identifier allowlist (`EMAIL_FIELD_IDENTIFIERS`), so email format-checking is opt-in by exact match — a `*_type` selector or unrelated field can't trigger it by accident.
- Drop the `position: :desc` `default_scope`; order **explicitly (`:asc`)** at each call site that relied on it — report/workshop log fields, the report-builder controllers, and the workshop decorator. Removed a now-unneeded `.unscoped` in the form-builder spec.

### Anything else to add?

- The signed-in one-click registration path (`Events::RegistrationsController#create`) deliberately does **not** use the validator — it submits no form fields (one-click only), so there's nothing to validate.
- Specs: new `form_answer_validator_spec.rb`, plus the touched model/controller/request specs all green (130 + 111 examples). Rubecop clean; Brakeman 0 warnings.

### DIFF
--- a/app/controllers/events/bulk_payments_controller.rb
+++ b/app/controllers/events/bulk_payments_controller.rb
@@ -96,34 +96,9 @@ module Events
     end
 
     def validate_required_fields
-      errors = {}
-      visible_fields = visible_form_fields
-
-      visible_fields.find_each do |field|
-        next if field.group_header?
-        next if field.field_identifier == "bulk_payment_attendees"
-
-        value = @form_params[field.id.to_s]
-
-        if field.required && (value.blank? || (value.is_a?(Array) && value.reject(&:blank?).empty?))
-          errors[field.id] = "can't be blank"
-          next
-        end
-
-        next if value.blank?
-
-        if field.number_integer? && value.to_s !~ /\A\d+\z/
-          errors[field.id] = "must be a whole number"
-        elsif field.field_identifier == "payer_email" && value.to_s !~ /\A[^@\s]+@[^@\s]+\z/
-          errors[field.id] = "must be a valid email address"
-        elsif (word_error = field.min_words_error(value))
-          errors[field.id] = word_error
-        elsif (char_error = field.max_characters_error(value))
-          errors[field.id] = char_error
-        end
-      end
-
-      errors
+      # The nested attendees field is validated separately, so exclude it here.
+      fields = visible_form_fields.reject { |field| field.field_identifier == "bulk_payment_attendees" }
+      FormAnswerValidator.call(fields, @form_params)
     end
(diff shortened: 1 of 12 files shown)


app/models/form_field.rb
```diff
@@ -106,6 +105,16 @@ def free_form_text?
     FREE_FORM_TEXT_TYPES.include?(answer_type)
   end
 
+  # Field identifiers (system-assigned by FormBuilderService) that collect an
+  # email address, so a submitted value should be format-checked. The "*_type"
+  # selector fields are deliberately excluded — this is an exact allowlist, not
```
maebeale, line 110: 🤖 _From Claude:_ Exact allowlist (not a `*email*` name match) is deliberate — it stops `primary_email_type` selector fields and any unrelated field from accidentally opting into email format validation.

app/controllers/events/bulk_payments_controller.rb
```diff
@@ -96,34 +96,9 @@ def set_form
...
-        elsif (word_error = field.min_words_error(value))
-          errors[field.id] = word_error
-        elsif (char_error = field.max_characters_error(value))
-          errors[field.id] = char_error
-        end
-      end
-
-      errors
+      # The nested attendees field is validated separately, so exclude it here.
+      fields = visible_form_fields.reject { |field| field.field_identifier == "bulk_payment_attendees" }
```
maebeale, line 100: 🤖 _From Claude:_ `bulk_payment_attendees` is excluded here because the nested attendee rows are validated on their own path; passing it to the shared validator would flag the whole nested field as blank.

PR #1644, opened 2026-06-12
Single-select dropdown answer type, enum cleanup, and primary/additional sector editor

maebeale:

Three related strands around how participants pick service areas and how those sectors are managed:

- **Single-select dropdown answer type** — registrants can pick **one** primary service area from a dropdown (`primary_service_area_single`), while the existing checkbox field becomes "Additional service areas".
- **Answer-type enum cleanup** — the old `multiple_choice_radio`/`multiple_choice_dropdown` names were misleading (both pick exactly one). Renamed to say what they mean.
- **Person sector editor redesign** — the old editor crammed a sector name plus two text checkboxes into each pill (hard to read), and let an admin mark several sectors "primary" when only one is allowed.

**Dropdown answer type**
- Added a `single_select_dropdown` answer type with a render branch in `_form_field.html.erb` that reuses the shared `dynamic_form_field_options` helper (so dynamic Sector/Category options work for dropdowns too).
- `PublicRegistration` recognizes `primary_service_area_single`; registration collects sector ids from both the single dropdown and the additional-areas checkbox.
- Event background page charts the dropdown answers as a new **"Primary service area"** breakdown and renames the combined-tags chart to **"All service areas"**, each with per-row drill-in links.

**Answer-type enum rename (no migration — enum is integer-backed, order preserved)**
- `multiple_choice_radio` → `single_select_radio`, `multiple_choice_dropdown` → `single_select_dropdown`, `multiple_choice_checkbox` → `multi_select_checkbox`.
- Predicate `multiple_choice?` → `selectable?`, backed by an explicit `SELECTABLE_ANSWER_TYPES` list instead of a fragile `"multiple_choice"` substring match.
- The JS editor controller had the **same** substring check (`value.includes("multiple_choice")`) with no test covering it — the rename would have silently broken the "show answer options" toggle. Fixed to `includes("select")`.

**Person sector editor redesign**
- Each sector is now a chip with two Font Awesome toggles: a **star** (primary, single-select — starring one clears the others) and a ghosted **crown** (leader, multi-select, pure CSS). The starred chip darkens and jumps to the front.
- New `primary_sector` Stimulus controller enforces the single primary and the move-to-front.
- New `SectorsTaggable` concern (Person + Organization) validates **at most one primary sector** (checking the in-memory set, so nested-attribute saves are caught), and exposes `sectorable_items_primary_first` so the editor lists the primary first, then the rest alphabetically.
- Renamed the partial's misleading `show_is_leader` local to `show_admin_flags`.

### UI Testing Checklist

- [ ] On the registration form, the primary service area renders as a single-select dropdown and the additional areas as checkboxes.
- [ ] On the person edit form, starring a sector clears any other star, darkens the chip, and moves it to the front; the crown toggles leader independently.
- [ ] Saving a person persists one primary sector; the editor re-renders primary-first then alphabetical.

### Anything else to add?

- Profile-flag spec marker updated for #1645's split `Primary sector` / `Additional sectors` headings.
- Specs: `form_field_spec`, `form_builder_service_spec`, `event_dashboard_spec`, `person_spec` (validation + ordering), and request specs; full sets green, RuboCop clean.
- **Not yet done:** the Organization form renders the shared partial without `show_admin_flags`, so org-side primary editing is still pending.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,7 +51,7 @@ This codebase (Rails 8.1)
 | `app/models/` | ActiveRecord models | ~75 files |
 | `app/services/` | Service objects for complex logic | ~23 files |
 | `app/jobs/` | SolidQueue background jobs | 3 files |
-| `app/models/concerns/` | Shared model modules | 13 concerns |
+| `app/models/concerns/` | Shared model modules | 14 concerns |
 
 ### Presentation
 
@@ -71,7 +71,7 @@ This codebase (Rails 8.1)
 | Directory | Purpose |
 |---|---|
 | `app/frontend/entrypoints/` | Vite entry points (application.js, application.css) |
-| `app/frontend/javascript/controllers/` | Stimulus controllers (60) |
+| `app/frontend/javascript/controllers/` | Stimulus controllers (61) |
 | `app/frontend/javascript/rhino/` | Rich text editor customizations (mentions, grid) |
 | `app/frontend/stylesheets/` | Tailwind CSS and component styles |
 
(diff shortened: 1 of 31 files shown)


maebeale:
Screenshot 2026-06-12 at 10 10 30 PM

PR #1645, opened 2026-06-13
Split participant sectors into primary and additional service areas

maebeale:
- The participant profile showed one flat "Sectors" list, so the sector a person picked as their **primary service area** during registration was indistinguishable from sectors they were tagged with incidentally.
- This surfaces what someone primarily does versus where they also have a footprint, making profiles more informative for staff.

- Added an `is_primary` flag to `sectorable_items` (migration + schema) and set it from the `primary_service_area` registration answer in `PublicRegistration`, promoting an already-tagged sector to primary if later selected.
- Split the profile "Sectors" block into two side-by-side columns — **Primary sector** and **Additional sectors** — and added a "Primary" checkbox (alongside "Leader") so admins can adjust it manually; permitted the new param.

### UI Testing Checklist
- [ ] Register for a public event selecting a primary service area → that sector appears under **Primary sector** on the registrant's profile.
- [ ] A sector tagged manually (not via registration) appears under **Additional sectors**.
- [ ] Each column shows "None selected." when empty and the two columns stack on mobile.
- [ ] On the person edit form, the new **Primary** checkbox saves correctly.

### Anything else to add?
- `primary_service_area` is a multi-select, so a person can have more than one primary sector; all are rendered in that column.
- Scoped to the profile (show) page — the people listing table's sector column is unchanged.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -218,7 +218,7 @@ module EventRegistrationServices
 
       if sector_ids.any?
         sectors = Sector.where(id: sector_ids)
-        person.sectors = (person.sectors + sectors).uniq
+        assign_primary_sectors(person, sectors)
         organization.sectors = (organization.sectors + sectors).uniq if organization
       end
 
@@ -226,6 +226,13 @@ module EventRegistrationServices
         categories = Category.where(id: category_ids)
         person.categories = (person.categories + categories).uniq
         organization.categories = (organization.categories + categories).uniq if organization
+      end
+    end
+
+    def assign_primary_sectors(person, sectors)
+      sectors.each do |sector|
+        item = person.sectorable_items.find_or_initialize_by(sector: sector)
+        item.update!(is_primary: true)
       end
     end
 
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -118,24 +118,42 @@
           <% end %>
           <!-- Sectors -->
           <% if @person.profile_show_sectors? %>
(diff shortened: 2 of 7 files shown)


app/views/people/show.html.erb
```diff
@@ -118,24 +118,42 @@
...
-                         .each do |si| %>
-                    <%= render "sectors/tagging_label",
-                                 sector: si.sector,
-                                 display_leader: true,
-                                 is_leader: si.is_leader %>
+              <div class="grid grid-cols-1 sm:grid-cols-2 gap-4">
+                <div>
+                  <h2 class="text-lg font-semibold text-gray-800 mb-3">Primary sector</h2>
+                  <% if primary_sector_items.any? %>
+                    <div class="flex flex-wrap gap-2">
```
maebeale, line 129: 🤖 _From Claude:_ `primary_service_area` is a multi-select checkbox, so this column can render multiple primary sectors. Flagging in case we want to constrain to a single primary later.

app/services/event_registration_services/public_registration.rb
```diff
@@ -229,6 +229,13 @@ def assign_tags(person, organization)
       end
     end
 
+    def assign_primary_sectors(person, sectors)
+      sectors.each do |sector|
+        item = person.sectorable_items.find_or_initialize_by(sector: sector)
```
maebeale, line 234: 🤖 _From Claude:_ `find_or_initialize_by` means a sector already tagged as "additional" gets promoted to primary here rather than duplicated — the unique index on (sector, sectorable) guarantees one row per sector, so primary/additional is mutually exclusive.

PR #1652, opened 2026-06-13
Fix duplicate questions/sections when re-editing form sections after a header rename

maebeale:

- Editing a form's sections could **duplicate section headers and every question under them** — a real risk as admins navigate events, their registration forms, and the edit-sections page.
- The edit-sections page had two disagreeing sources of truth for "which sections does this form have":
  - The checkboxes (`editable_sections`) derive *included* from the **fields actually on the form** (each field's `section` group).
  - `update_sections!` derived its add/remove diff from the stored **`form.sections` column**.
- These drift in one concrete case: **rename a section header, then remove the section.** Removal matched headers by their *default name*, so the renamed header survived as an orphan — keeping the section "present" in the fields but absent from the column. Re-saving then treated it as newly added and rebuilt a **second header plus a second copy of every question.**

- Made the field set the single source of truth via a new `FormBuilderService.present_section_keys` (uses `pluck` so repeated calls within one update see deletions made between them).
- `update_sections!` now diffs against `present_section_keys` instead of the `sections` column, so a section is only added/removed when its checkbox is **newly checked or newly unchecked**.
- Removal now deletes by **section group** (`where(section: …)`) rather than canonical identifier + default header name, so renamed headers (and any questions added under a section) are removed too — nothing is left orphaned to resurface as a duplicate.
- The controller's "No section changes" short-circuit uses the same basis for consistency.
- Verified the event form-link paths (`build_public_registration_form`, `copy_registration_form`, `assign_event_forms`) are already idempotent, so navigating events doesn't double-link forms.

### Anything else to add?

- Added failing-first specs: renamed header is removed, the rename → remove → re-add round-trip yields exactly one header and one question, re-submitting an already-present section adds nothing despite column drift, plus `present_section_keys` coverage.
- One existing custom-section test was updated: its hand-built `background` header now legitimately counts as a present section under the new truth.
- Tests: `form_builder_service_spec` + `forms_spec` (90), `events_spec` + `event_spec` (146) all pass; lint clean.

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -83,7 +83,7 @@ class FormsController < ApplicationController
     end
 
     removed_custom_ids = removed_custom_section_ids
-    current_sections = (@form.sections || []).map(&:to_sym)
+    current_sections = FormBuilderService.present_section_keys(@form)
     if sections.to_set == current_sections.to_set && removed_custom_ids.empty?
       redirect_to edit_form_path(@form, event_id: params[:event_id]), notice: "No section changes."
       return
--- a/app/services/form_builder_service.rb
+++ b/app/services/form_builder_service.rb
@@ -101,6 +101,17 @@ class FormBuilderService
     end
   end
 
+  # Built-in section keys that currently have fields on the form. This — not the
+  # stored `sections` column — is the source of truth for which sections are
+  # present, so add/remove stays consistent with the edit-sections checkboxes
+  # (which are rendered from the fields) even if the column drifts.
+  def self.present_section_keys(form)
+    group_to_key = SECTION_GROUPS.flat_map { |key, groups| groups.map { |group| [ group, key ] } }.to_h
+    # pluck runs a fresh query rather than reading (and caching) the association,
+    # so repeated calls within a single update see deletions made between them.
+    form.form_fields.pluck(:section).filter_map { |section| group_to_key[section] }.uniq
+  end
+
   # Ordered list of sections for the edit-sections page. Each entry is a hash
   # with a :kind of :builtin or :custom. Built-in sections carry their :label,
   # section :key, and :included state; custom (user-added) sections carry the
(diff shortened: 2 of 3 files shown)


app/services/form_builder_service.rb
```diff
@@ -162,22 +173,19 @@ def self.editable_sections(form)
...
 
     remove_custom_sections!(form, remove_custom_section_ids)
 
     added = new_sections - old_sections
     removed = old_sections - new_sections
 
-    # Remove fields and headers belonging to removed sections
+    # Remove every field belonging to a removed section by its section group.
+    # Keying off the group (not the canonical field_identifier or default header
+    # name) deletes renamed headers too, so nothing is left orphaned to resurface
```
maebeale, line 185: 🤖 _From Claude:_ Deleting by section group (not by canonical `field_identifier` + default header name) is the core fix: it also removes renamed headers and any questions an admin added under the section, so nothing is left orphaned to resurface as a duplicate on re-add.

app/controllers/forms_controller.rb
```diff
@@ -83,7 +83,7 @@ def update_sections
     end
 
     removed_custom_ids = removed_custom_section_ids
-    current_sections = (@form.sections || []).map(&:to_sym)
+    current_sections = FormBuilderService.present_section_keys(@form)
```
maebeale, line 86: 🤖 _From Claude:_ Uses the same fields-based truth as `update_sections!` so the "No section changes" short-circuit stays consistent with the diff the service actually computes.

app/services/form_builder_service.rb
```diff
@@ -101,6 +101,17 @@ def self.applicable_section_keys(role)
...
 
+  # Built-in section keys that currently have fields on the form. This — not the
+  # stored `sections` column — is the source of truth for which sections are
+  # present, so add/remove stays consistent with the edit-sections checkboxes
+  # (which are rendered from the fields) even if the column drifts.
+  def self.present_section_keys(form)
+    group_to_key = SECTION_GROUPS.flat_map { |key, groups| groups.map { |group| [ group, key ] } }.to_h
+    # pluck runs a fresh query rather than reading (and caching) the association,
+    # so repeated calls within a single update see deletions made between them.
+    form.form_fields.pluck(:section).filter_map { |section| group_to_key[section] }.uniq
```
maebeale, line 112: 🤖 _From Claude:_ `pluck` (not iterating the association) is deliberate — `update_sections!` calls this before and after deleting fields in the same request, and a cached association would make the second call see stale, pre-deletion records.

PR #1656, opened 2026-06-13
Notifications index: link form submission noticeables to a viewable page

maebeale:

- The notifications index (`/notifications`) was throwing a production 500: `NoMethodError: undefined method 'form_submission_path'` (reported via Honeybadger).
- Root cause: #1647 began creating `Notification` records with a `FormSubmission` as their `noticeable` (bulk-payment confirmation + FYI emails). `FormSubmission` had **no route**, so the view's `polymorphic_path(notification.noticeable)` resolved to a nonexistent path and raised — breaking the entire index for staff whenever a bulk-payment notification appeared.
- Beyond stopping the crash, a `FormSubmission` noticeable should actually be **viewable** — staff need to click through to see the submission.

- **Stop the 500 (first commit):** `routable_path` returns a link or `nil`, and the notifications view links only when routable — defensively covering any routeless noticeable.
- **Make it viewable (this work):** `routable_path` now gives a `FormSubmission` a real destination via `form_submission_link_path`:
  - **Registration submissions** (submitter has an `EventRegistration`) → the public **registration details** page, keyed by the registration slug — the same "ticket"/details view the public already sees.
  - **Bulk-payment / event-less submissions** (no registration) → a new RESTful **`form_submissions#show`**, admin-only via `FormSubmissionPolicy`.
- **One display, two pages:** `form_submissions#show` and the public `bulk_payments#show` now render a shared `form_submissions/_submission` partial (fields, submitter, attendees). Removed the now-dead `@form_fields`/`@responses`/`@attendees` setup and `parse_attendees` from `BulkPaymentsController`.

### Anything else to add?

- Test-first throughout: request specs for `form_submissions#show` (admin 200 / non-admin redirect), an updated notifications spec asserting the row now links, a `FormSubmissionPolicy` spec, and helper specs covering both `routable_path` branches.
- No data migration needed — view/routing only.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -260,6 +260,15 @@ module ApplicationHelper
     end
   end
 
+  # polymorphic_path for a record, or nil when the record's model has no route
+  # (e.g. FormSubmission, which is referenced as a notification's noticeable but
+  # has no controller/show page). Lets views link only when a route exists.
+  def routable_path(record)
+    polymorphic_path(record)
+  rescue NoMethodError
+    nil
+  end
+
   def search_page(params)
     params[:search] ? params[:search][:page] : 1
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -260,13 +260,26 @@ module ApplicationHelper
     end
   end
 
-  # polymorphic_path for a record, or nil when the record's model has no route
-  # (e.g. FormSubmission, which is referenced as a notification's noticeable but
-  # has no controller/show page). Lets views link only when a route exists.
+  # Best viewable path for a record (e.g. a notification's noticeable), or nil
+  # when its model has no route. FormSubmissions have no polymorphic route, so
+  # they get a tailored destination via form_submission_link_path.
   def routable_path(record)
+    return form_submission_link_path(record) if record.is_a?(FormSubmission)
     polymorphic_path(record)
   rescue NoMethodError
     nil
+  end
+
+  # Where to view a form submission. When the submitter has a public event
+  # registration, send the viewer to their registration details page (the public
+  # "ticket"/details view, keyed by the registration slug). Otherwise — bulk
+  # payments and event-less submissions, which have no registration — fall back
+  # to the form submission show page.
+  def form_submission_link_path(submission)
+    event = submission.event
+    registration = event && submission.person.event_registrations.find_by(event: event)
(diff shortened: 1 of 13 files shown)


app/helpers/application_helper.rb
```diff
@@ -260,6 +260,15 @@ def index_button(collection, params: {}, path: nil, label: nil, icon: nil, hide_
     end
   end
 
+  # polymorphic_path for a record, or nil when the record's model has no route
+  # (e.g. FormSubmission, which is referenced as a notification's noticeable but
+  # has no controller/show page). Lets views link only when a route exists.
+  def routable_path(record)
+    polymorphic_path(record)
+  rescue NoMethodError
```
maebeale, line 269: 🤖 _From Claude:_ Rescuing `NoMethodError` (rather than enumerating routable types) keeps this generic — it catches any current or future noticeable whose model has no route, not just `FormSubmission`.

app/views/form_submissions/_submission.html.erb
```diff
@@ -0,0 +1,66 @@
+<%# Renders a form submission's answers — field responses, the submitter, and any
```
maebeale, line 1: 🤖 _From Claude:_ Extracted from `bulk_payments/show` into a shared partial so `form_submissions#show` and the public bulk-payment page render submissions identically. It's self-contained from `submission` (computes fields/responses/attendees itself), which let the controller drop `@form_fields`/`@responses`/`@attendees` and the `parse_attendees` helper.

app/helpers/application_helper.rb
```diff
@@ -260,6 +260,28 @@ def index_button(collection, params: {}, path: nil, label: nil, icon: nil, hide_
...
+  # registration, send the viewer to their registration details page (the public
+  # "ticket"/details view, keyed by the registration slug). Otherwise — bulk
+  # payments and event-less submissions, which have no registration — fall back
+  # to the form submission show page.
+  def form_submission_link_path(submission)
+    event = submission.event
+    registration = event && submission.person.event_registrations.find_by(event: event)
+    return event_public_registration_path(event, reg: registration.slug) if registration&.slug.present?
+    form_submission_path(submission)
+  end
```
maebeale, line 283: 🤖 _From Claude:_ The branching lives here so the notifications view stays dumb: registration submissions route to the existing public registration details page (by slug), while bulk-payment/event-less submissions — which have no EventRegistration — fall back to form_submissions#show. Keying on registration presence (not just event presence) is deliberate: bulk-payment submissions do resolve an event but have no registration, so they correctly take the fallback.

PR #1658, opened 2026-06-13
Preserve registration form choices on failed event save

maebeale:

- When an admin saves the event form and validation fails (e.g. `start_date`/`end_date` left blank), their **Registration form**, **Enable scholarship application**, and **Enable bulk payment** selections were silently reset on the re-rendered form.
- These controls aren't Event attributes — they're virtual params turned into `EventForm` join records by `assign_event_forms`, which only runs *after* a successful save. The checkboxes/select read their state from `@event.scholarship_form` / `registration_form` (DB lookups), so on a failed save nothing is written and the controls come back empty.
- Plain model attributes (title, dates, toggles, hints, rich text) and the `sector_ids`/`category_ids` checkboxes already survive via Rails' normal in-memory repopulation — only these post-save-derived controls were affected.

- In `events/_form.html.erb`, added a `resubmitted = params[:event].present?` guard inside the registration form section.
- On a re-render, the registration-form `<select>`, scholarship checkbox, and bulk-payment checkbox now read back from the submitted params; on a normal new/edit load they still read from the persisted event (unchanged default-form logic preserved).

### Tests

- `spec/requests/events_spec.rb` — new failing-first request spec confirming the registration form select + scholarship/bulk-payment checkboxes are retained after a validation failure.
- Added a companion spec asserting sector/category checkboxes are retained (they already were — locks in the behavior).
- Full `events_spec` passes (112 examples, 0 failures).

### Anything else to add?

- Image uploads (primary/gallery) are still cleared on any failed submit — that's a browser security limitation for file inputs, not something the server can repopulate.

### DIFF
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -235,11 +235,17 @@
         </button>
 
         <div id="registration_form_section" class="border border-gray-300 p-4 rounded-b-md" data-reveal-section-target="content">
+          <%# On a failed save the event_forms aren't persisted, so re-read the
+              admin's registration choices from the submitted params instead of
+              the (now-empty) associations. %>
+          <% resubmitted = params[:event].present? %>
           <div class="grid grid-cols-1 md:grid-cols-2 gap-6">
             <div>
               <label class="block text-sm font-medium text-gray-700 mb-1">Registration form</label>
               <% current_reg_form = @event.registration_form %>
-              <% default_form_id = if current_reg_form
+              <% default_form_id = if resubmitted
+                                     params.dig(:event, :registration_form_id).presence&.to_i
+                                   elsif current_reg_form
                                      current_reg_form.id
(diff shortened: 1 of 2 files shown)


app/views/events/_form.html.erb
```diff
@@ -235,11 +235,17 @@
         </button>
 
         <div id="registration_form_section" class="border border-gray-300 p-4 rounded-b-md" data-reveal-section-target="content">
+          <%# On a failed save the event_forms aren't persisted, so re-read the
+              admin's registration choices from the submitted params instead of
+              the (now-empty) associations. %>
+          <% resubmitted = params[:event].present? %>
```
maebeale, line 241: 🤖 _From Claude:_ `params[:event].present?` distinguishes a re-render after a failed save (params present) from a fresh new/edit GET (no params), so the fix only overrides the source-of-truth on resubmit and leaves normal loads reading from the persisted event.

PR #1661, opened 2026-06-13
Set ce_credit_requested from the seeded CE-interest question

maebeale:
- Public registrants seeking **continuing-education (CE) hours** had no way to indicate it on the registration form — admins had to toggle `ce_credit_requested` on each registration by hand.
- This wires a seeded Yes/No "magic question" on the registration form to the registration's CE flag, so the request is captured at submission time.

- Added `EventRegistrationServices::PublicRegistration::CE_CREDIT_INTEREST_IDENTIFIER` (`"ce_credit_interest"`) as the single source of truth shared by the seed, service, and specs.
- Seeded a standalone **Continuing education** section + a required-less `single_select_radio` Yes/No question onto the dev registration form, with its own `section` so the form builder's add/remove-section logic leaves it untouched. The seed is idempotent.
- `PublicRegistration` now sets `ce_credit_requested: true` on new registrations when the answer is "Yes" (case-insensitive), and also toggles it on for an existing registration that re-submits with "Yes" — mirroring the existing scholarship-flag pattern (only ever toggles on, never off).
- Added service specs covering Yes / No / unanswered and the existing-registration path.

### UI Testing Checklist
- [ ] On a public registration form, the new "Continuing education" question appears and answering **Yes** results in a registration with CE requested.

### Anything else to add?
- The flag only toggles **on**, never off, consistent with how `scholarship_requested` is handled in the same service.
- No new services/controllers/Stimulus/mailers/rake tasks, so no AGENTS.md changes.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -2,6 +2,11 @@ module EventRegistrationServices
   class PublicRegistration
     Result = Struct.new(:success?, :event_registration, :form_submission, :errors, keyword_init: true)
 
+    # Well-known field_identifier of the "magic" CE question seeded onto the
+    # registration form. Answering it "Yes" toggles the registration's
+    # ce_credit_requested flag. Kept here so the seed, service, and specs agree.
+    CE_CREDIT_INTEREST_IDENTIFIER = "ce_credit_interest".freeze
+
     def self.call(event:, form:, form_params:, scholarship_requested: false, person: nil)
       new(event:, form:, form_params:, scholarship_requested:, person:).call
     end
@@ -31,6 +36,7 @@ module EventRegistrationServices
         existing = @event.event_registrations.find_by(registrant: person)
         if existing
           existing.update!(scholarship_requested: true) if @scholarship_requested
+          existing.update!(ce_credit_requested: true) if ce_credit_requested?
           if existing.status == "cancelled"
             existing.update!(status: "registered")
             send_notifications(existing)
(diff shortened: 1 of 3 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -31,6 +36,7 @@ def call
         existing = @event.event_registrations.find_by(registrant: person)
         if existing
           existing.update!(scholarship_requested: true) if @scholarship_requested
+          existing.update!(ce_credit_requested: true) if ce_credit_requested?
```
maebeale, line 39: 🤖 _From Claude:_ Mirrors the scholarship line above — only ever toggles CE **on**, never off, so a re-submit can add the request but won't silently clear an admin's existing flag.

app/services/event_registration_services/public_registration.rb
```diff
@@ -248,10 +254,16 @@ def collect_ids_from_checkboxes(identifier)
...
       @event.event_registrations.create!(
         registrant: person,
-        scholarship_requested: @scholarship_requested
+        scholarship_requested: @scholarship_requested,
+        ce_credit_requested: ce_credit_requested?
       )
     end
 
+    # True when the registrant answered "Yes" to the seeded CE-interest question.
+    def ce_credit_requested?
```
maebeale, line 263: 🤖 _From Claude:_ `.to_s` guards a nil/unanswered field (→ false), and `casecmp?` makes the match case-insensitive since the radio value is the option label ("Yes").

db/seeds/dev/events_management.rb
```diff
@@ -94,6 +94,41 @@
...
 end
 
+# The CE-interest "magic question": a single Yes/No whose answer drives the
+# resulting registration's ce_credit_requested flag (see
+# EventRegistrationServices::PublicRegistration). Seeded straight onto the form
+# with its own section so the form builder's add/remove-section logic leaves it
+# alone, and carrying the well-known field_identifier the service keys off.
+ce_identifier = EventRegistrationServices::PublicRegistration::CE_CREDIT_INTEREST_IDENTIFIER
+if registration_form.form_fields.where(field_identifier: ce_identifier).none?
+  next_position = (registration_form.form_fields.maximum(:position) || 0) + 1
```
maebeale, line 104: 🤖 _From Claude:_ Custom `section: "continuing_education"` (not a FormBuilderService group) so `update_sections!` never destroys this field when an admin edits the built-in sections. Idempotent via the `none?` guard.

PR #1663, opened 2026-06-13
Notifications: link noticeable records to viewable pages with friendly names

maebeale:

- The notifications **Record** column (index) and **Noticeable** field (detail page) previously linked using the raw model class name and `polymorphic_path`, which:
  - **Crashed on the detail page** for `FormSubmission` noticeables (bulk payments) — they have no polymorphic route, so `polymorphic_path` raised.
  - Showed an unhelpful label (`EventRegistration`, `FormSubmission`) that said nothing about *which* record it was.
- This makes notification records traceable to the actual person/event/form they're about, and viewable from both the list and the detail page.

- **Detail page routing** — reused the existing `routable_path` helper (added by #1656 for the index) on `show.html.erb` so it links the same viewable destinations and degrades to plain text for routeless records, instead of calling `polymorphic_path` unconditionally.
- **Friendly labels** — added two helpers:
  - `noticeable_label` — the record's name (registrant · event, submitter · form, or the record's own `title`/`name`).
  - `noticeable_type_label` — plain-language type (`Registration`, `Bulk payment`, `Form submission`, else humanized model name).
- **Format decision** — rendered as a small uppercase type label over a text link (mirroring the existing **People** column), rather than a button, so it doesn't compete with the row's existing **View** action button.
- **Bulk payments** — no dedicated handling needed: #1656's `form_submission_path` + shared `_submission` partial already render bulk-payment submissions (including the attendee table); they route there automatically.

### UI Testing Checklist

- [ ] Notifications index: Record column shows a type label + record name, links to the right page
- [ ] Notification detail page: Noticeable field links correctly (no crash) for a bulk-payment notification
- [ ] Routeless noticeables fall back to plain text (no link, no error)

### Anything else to add?

- Helper specs cover both new label helpers across registration / bulk payment / submission / other model cases.
- Tests: helper specs 56/56, notifications request specs 28/28, lint clean.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -282,6 +282,34 @@ module ApplicationHelper
     form_submission_path(submission)
   end
 
+  # A friendly type name for a noticeable record, e.g. "Registration" or "Bulk
+  # payment" rather than the raw model name ("EventRegistration").
+  def noticeable_type_label(record)
+    return "Registration" if record.is_a?(EventRegistration)
+
+    if record.is_a?(FormSubmission)
+      return record.role == "bulk_payment" ? "Bulk payment" : "Form submission"
+    end
+
+    record.class.name.underscore.humanize
+  end
+
+  # A human-friendly name for a noticeable record (the registrant and event for a
+  # registration, the submitter and form for a submission, otherwise the record's
+  # own title/name), so links read as the thing rather than its model class.
+  def noticeable_label(record)
+    label = case record
+    when EventRegistration
+      [ record.registrant&.name, record.event&.title ].compact_blank.join(" · ")
+    when FormSubmission
+      [ record.person&.name, record.form&.name ].compact_blank.join(" · ")
+    else
+      record.try(:title) || record.try(:name) || record.try(:full_name)
+    end
+
+    label.presence || "##{record.id}"
+  end
+
   def search_page(params)
     params[:search] ? params[:search][:page] : 1
   end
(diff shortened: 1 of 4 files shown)


app/views/notifications/show.html.erb
```diff
@@ -144,9 +144,15 @@
         <dt class="w-28 text-gray-500">Noticeable</dt>
         <dd class="text-gray-900">
           <% if @notification.noticeable.present? %>
-            <%= link_to @notification.noticeable.class.name,
-                        polymorphic_path(@notification.noticeable),
-                        class: "text-blue-600 hover:underline" %>
+            <% record_path = routable_path(@notification.noticeable) %>
```
maebeale, line 147: 🤖 _From Claude:_ This is the actual bug fix — #1656 only updated the index, leaving the detail page calling `polymorphic_path` unconditionally, which raised for `FormSubmission` noticeables (bulk payments). Reusing `routable_path` here makes it consistent and crash-safe.

app/helpers/application_helper.rb
```diff
@@ -282,6 +282,34 @@ def form_submission_link_path(submission)
...
+    return "Registration" if record.is_a?(EventRegistration)
+
+    if record.is_a?(FormSubmission)
+      return record.role == "bulk_payment" ? "Bulk payment" : "Form submission"
+    end
+
+    record.class.name.underscore.humanize
+  end
+
+  # A human-friendly name for a noticeable record (the registrant and event for a
```
maebeale, line 297: 🤖 _From Claude:_ `noticeable_label` special-cases registrations/submissions (registrant · event, submitter · form) and otherwise falls back through `title`/`name`/`full_name`, so any future noticeable type gets a sensible label without changes here.

PR #1664, opened 2026-06-13
Add optional event-specific staff bio

maebeale:

- Lets event organizers tailor how a staff member is introduced on a **specific event's** public staff page, instead of being limited to the one profile bio that person carries everywhere.
- The per-event bio is **optional** and **overrides** the profile bio for that event only; left blank, the public staff page falls back to the person's profile bio.

- **Model/data:** added a `bio` column to `event_staffs`; `EventStaff#public_bio` resolves the event bio first, then the profile bio (only when the person allows it to be shown). Blank event bios normalize to `nil`.
- **Staff form (`/events/:id/staff/edit`):** each staff card now shows the editable **event bio** on the left beside the person's **read-only profile bio** (with an "Edit on profile" link) on the right, so organizers can see what they're overriding.
- **Dynamic profile bio:** for a freshly added row the person isn't known server-side, so the profile bio is fetched when a person is picked, via a new admin-gated `GET /people/:id/bio` endpoint (returns sanitized HTML) wired through the `event-staff-bio` Stimulus controller.
- **Duplicate guard:** adding the same person to a staff twice in one submission previously passed per-record validation and then hit the unique index → 500. Added an Event-level `validate :staff_members_are_unique, on: :update` so it re-renders the form with a clear error instead.
- **Seeds:** Amy (event-bio override) and Maria Johnson (profile-bio fallback) are staffed on the flagship training so both states are visible in dev.

### Anything else to add?

- The "Edit on profile" link targets the person's profile **edit** page (admin-only). Person search on this form is already admin-gated, so that's consistent — but worth noting if non-admin event owners are ever meant to use this page.
- Tests: new `people_bio_spec`, model specs for `public_bio` / duplicate guard, and request specs for saving/clearing the event bio. Migration is reversible; AGENTS.md updated for the new endpoint/controller.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -1,6 +1,6 @@
 class PeopleController < ApplicationController
   include AhoyTracking, TagAssignable
-  before_action :set_person, only: %i[ show edit update destroy workshop_logs checkout ]
+  before_action :set_person, only: %i[ show edit update destroy workshop_logs checkout bio ]
 
   def index
     authorize!
@@ -87,6 +87,21 @@ class PeopleController < ApplicationController
     elsif params[:external_title].present?
       @filtered_logs = all_logs.select { |log| log.external_workshop_title == params[:external_title] }
     end
+  end
+
+  # Read-only profile bio for the event-staff editor, fetched when a person is
+  # picked for a new staff row. Returns sanitized HTML so the client can render
+  # it directly. Person search is admin-gated, so show? (admin or owner) is the
+  # right authorization here.
+  def bio
+    authorize! @person, to: :show?
+    has_bio = @person.profile_show_bio? && @person.bio.present?
+    render json: {
+      show_bio: @person.profile_show_bio?,
+      has_bio: has_bio,
+      bio_html: (helpers.sanitize(@person.bio, tags: %w[p br strong em a ul ol li b i], attributes: %w[href]) if has_bio),
+      edit_path: (edit_person_path(@person) if allowed_to?(:edit?, @person))
+    }
   end
 
   def checkout
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -38,6 +38,7 @@ class Event < ApplicationRecord
   validates_inclusion_of :published, in: [ true, false ]
   validates_numericality_of :cost_cents, greater_than_or_equal_to: 0, allow_nil: true
   validate :registration_form_required_when_publicly_registerable, on: :update
+  validate :staff_members_are_unique, on: :update
 
   # Nested attributes
   accepts_nested_attributes_for :primary_asset, allow_destroy: true, reject_if: :all_blank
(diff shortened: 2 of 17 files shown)


app/models/event.rb
```diff
@@ -161,6 +162,16 @@ def registration_form_required_when_publicly_registerable
     errors.add(:public_registration_enabled, "requires a registration form to be selected")
   end
 
+  # The event_staffs unique index catches duplicates already persisted, but two
+  # new rows for the same person in one submission both pass the per-record
+  # uniqueness check (neither is saved yet) and would hit the index on save.
+  def staff_members_are_unique
```
maebeale, line 168: 🤖 _From Claude:_ Scoped to `on: :update` deliberately: staff is only edited via the update_staff PATCH path, and running this on create made the factory's `create(:event)` load (and cache) an empty `event_staffs` association on the event object, which broke unrelated specs.

app/models/event_staff.rb
```diff
@@ -4,7 +4,17 @@ class EventStaff < ApplicationRecord
...
+  normalizes :bio, with: ->(value) { value&.strip.presence }
+
   scope :ordered_by_name, -> {
     joins(:person).order(Arel.sql("people.first_name, people.last_name"))
   }
+
+  # Bio shown on the public staff page. A per-event bio overrides the person's
+  # profile bio; otherwise fall back to the profile bio when the person allows
+  # it to be shown.
+  def public_bio
```
maebeale, line 16: 🤖 _From Claude:_ Resolution order is event bio → profile bio, and the profile-bio fallback is gated on `profile_show_bio?` so we never surface a bio the person chose to hide.

app/controllers/people_controller.rb
```diff
@@ -89,6 +89,21 @@ def workshop_logs
     end
   end
 
+  # Read-only profile bio for the event-staff editor, fetched when a person is
+  # picked for a new staff row. Returns sanitized HTML so the client can render
+  # it directly. Person search is admin-gated, so show? (admin or owner) is the
+  # right authorization here.
+  def bio
+    authorize! @person, to: :show?
```
maebeale, line 97: 🤖 _From Claude:_ `show?` (admin or owner) is the right gate here because the person search that feeds this form is already admin-only — so only admins reach the new-row fetch. Returns pre-sanitized HTML so the client renders it directly.

app/views/events/_event_staff_fields.html.erb
```diff
@@ -1,43 +1,80 @@
...
+        <%= f.hidden_field :person_id %>
+      <% else %>
+        <%= f.input :person_id,
+          include_blank: true,
+          required: true,
+          input_html: {
+            data: {
+              controller: "remote-select",
+              remote_select_model_value: "person",
+              action: "change->event-staff-bio#personChanged"
```
maebeale, line 20: 🤖 _From Claude:_ The `change` action bubbles from the person `<select>` up to the `event-staff-bio` controller on the card, which fetches the picked person's profile bio (the URL template on line 3 has `__ID__` swapped in JS).

PR #1665, opened 2026-06-13
Reveal "please specify" text input for form "Other" options

maebeale:

- When a form field (radio or checkbox) has an option literally named **Other**, selecting it now reveals an inline "Please specify" text input inside that option's box so the form filler can describe their choice.
- Previously organizers had to add a separate free-text question to capture this, leaving the bare "Other" answer meaningless on submissions.

- Added an `other-option` Stimulus controller that shows/hides the text input as the Other control is toggled and folds the typed text into the control's submitted value as `Other: <text>` — no extra form params.
- The text input has **no `name`**, so it never submits on its own; only its folded value reaches the server. This means **no controller, service, or schema changes** — the answer stores and displays through the existing `submitted_answer` path.
- Updated `app/views/events/public_registrations/_form_field.html.erb` to render the controller wrapper (only when an Other option exists), the control target, and the revealed text input for both radio and checkbox fields.
- Added `EventHelper#other_option?`, `#other_option_selected?`, and `#other_option_text` to detect the Other option and parse a stored `Other: <text>` answer back into a checked state + prefilled text after a validation error re-render.

### UI Testing Checklist

- [ ] Radio field with an "Other" option: selecting Other reveals the text box; typing stores `Other: <text>`; selecting another option hides it.
- [ ] Checkbox field with an "Other" option: same reveal/hide behavior; submitted alongside other checked values.
- [ ] Submitting with validation errors re-checks Other and keeps the typed text.
- [ ] Fields without an "Other" option are unaffected (no controller attached).

### Anything else to add?

- Covered by helper specs (`spec/helpers/event_helper_spec.rb`) and request specs for render + error re-render. JS reveal behavior is driven by Stimulus; the request specs assert the markup/targets are present since the suite's public-registration specs run under `rack_test`.

### DIFF
--- a/app/frontend/javascript/controllers/other_option_controller.js
+++ b/app/frontend/javascript/controllers/other_option_controller.js
@@ -1,0 +1,33 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Reveals a free-text "please specify" input inside an "Other" option box for
+// radio/checkbox form fields. When Other is selected the text input appears,
+// and whatever the user types is folded into the option's submitted value as
+// "Other: <text>" — so the server stores it with no extra params or changes.
+// The text input has no name attribute, so it never submits on its own.
+export default class extends Controller {
+  static targets = ["control", "text"]
+
+  connect() {
+    this.update()
+  }
+
+  // Fires on any change within the field (a radio/checkbox toggling on or off).
+  update() {
+    const selected = this.controlTarget.checked
+    this.textTarget.classList.toggle("hidden", !selected)
+    if (selected) this.foldValue()
+  }
+
+  // Fires while the user types in the Other box — keep Other selected and sync.
+  typed() {
+    if (!this.controlTarget.checked) this.controlTarget.checked = true
+    this.textTarget.classList.remove("hidden")
+    this.foldValue()
+  }
+
+  foldValue() {
+    const text = this.textTarget.value.trim()
+    this.controlTarget.value = text ? `Other: ${text}` : "Other"
+  }
+}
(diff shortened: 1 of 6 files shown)


app/frontend/javascript/controllers/other_option_controller.js
```diff
@@ -0,0 +1,33 @@
...
+  // Fires while the user types in the Other box — keep Other selected and sync.
+  typed() {
+    if (!this.controlTarget.checked) this.controlTarget.checked = true
+    this.textTarget.classList.remove("hidden")
+    this.foldValue()
+  }
+
+  foldValue() {
+    const text = this.textTarget.value.trim()
+    this.controlTarget.value = text ? `Other: ${text}` : "Other"
```
maebeale, line 31: 🤖 _From Claude:_ The typed text folds into the option value as `Other: <text>`, so the existing `submitted_answer` path stores and displays it with no server changes. `EventHelper#other_option_selected?`/`#other_option_text` parse this format back for re-checking and prefilling after a validation error.

app/views/events/public_registrations/_form_field.html.erb
```diff
@@ -72,24 +72,35 @@
...
           </span>
           <% if option_description.present? %>
             <span class="pl-6 text-xs font-normal text-gray-500"><%= option_description %></span>
           <% end %>
+          <% if is_other %>
+            <input type="text"
+                   placeholder="Please specify"
+                   value="<%= other_option_text(value) %>"
+                   data-other-option-target="text"
+                   data-action="input->other-option#typed"
```
maebeale, line 98: 🤖 _From Claude:_ The text input intentionally has no `name`, so it never submits on its own — only its value, folded into the radio/checkbox value by the controller, reaches the server. That keeps this a view+JS-only change.

PR #1668, opened 2026-06-13
Surface "Other" registration answers on person profile and edit

maebeale:

- On the sector/category-backed registration questions (primary service area, workshop environments, etc.), a registrant can pick **"Other"** and type free text. That gets folded into the stored answer (e.g. `5, Other: Equine therapy`).
- Those free-text values can't be `Sector`/`Category` records, so they were **silently dropped** — they never appeared among the profile's sector tags or the edit page's workshop-setting checkboxes.
- Admins viewing a person had no way to see what the registrant actually entered, even though the value lives in `FormAnswer#submitted_answer`.
- This surfaces those values so the registrant's intent isn't lost.

- Added `OtherOption.texts` (`app/services/other_option.rb`) to pull the free-text portion(s) out of a stored answer like `5, 12, Other: Equine therapy`.
- Added `Person#other_service_area_responses` and `Person#other_workshop_setting_responses`, which read the relevant form answers by `field_identifier` and return the unique "Other" texts.
- Rendered them as **read-only chips** (`people/_other_responses`) next to the real tags:
  - **Profile** (`people/show`): beside the Primary sector tags.
  - **Person edit** (`people/_form`): in the Sectors section and the workshop-settings (categories) section.
- They're read-only by design — a free-text value can't be promoted to a real `Sector`/`Category`, so the chip flags it for an admin rather than pretending to be a tag.

### Anything else to add?

- The profile only displays sectors (not categories), so the category "Other" texts surface on the **edit** page only — that matches where workshop settings are shown.
- Scoped to a separate branch off `main` (not #1667) so #1667 can merge without retriggering CI.
- Known limitation inherited from the storage format: a free-text answer containing `", "` is split on the comma (answers are stored comma-joined), so such values render truncated.

### Demo data

- Seeded Amy User's Facilitator Training (event 1) submission with two free-text "Other" answers — `primary_service_area` = `"Other: Equine-assisted therapy"` and `workshop_environments` = `"Other: Mobile art van"` — so the chips are visible after `rake db:seed:dev`. Amy's profile shows the chip beside her real sector tags; her edit page shows both the service-area and workshop-setting chips.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -207,7 +207,33 @@ class Person < ApplicationRecord
     }
   end
 
+  # Field identifiers whose "Other" free text maps onto the profile's sectors.
+  OTHER_SERVICE_AREA_IDENTIFIERS = %w[primary_service_area primary_service_area_single].freeze
+  # Field identifiers whose "Other" free text maps onto the workshop-setting
+  # categories shown on the edit page.
+  OTHER_WORKSHOP_SETTING_IDENTIFIERS = %w[workshop_environments client_life_experiences primary_age_group].freeze
+
+  # Free-text "Other" service areas the person typed on registration forms.
+  # They can't be Sector records, so they're surfaced beside the sector tags.
+  def other_service_area_responses
+    other_form_responses(OTHER_SERVICE_AREA_IDENTIFIERS)
+  end
+
+  # Free-text "Other" workshop settings (category-backed fields) from forms.
+  def other_workshop_setting_responses
+    other_form_responses(OTHER_WORKSHOP_SETTING_IDENTIFIERS)
+  end
+
   private
+
+  def other_form_responses(identifiers)
+    form_submissions
+      .joins(form_answers: :form_field)
+      .where(form_fields: { field_identifier: identifiers })
+      .pluck("form_answers.submitted_answer")
+      .flat_map { |answer| OtherOption.texts(answer) }
+      .uniq
+  end
 
   def strip_whitespace
     self.first_name = first_name&.strip
--- a/app/services/other_option.rb
+++ b/app/services/other_option.rb
@@ -1,0 +1,22 @@
+# Extracts free-text "Other" answers out of a stored form answer.
+#
+# When a registrant picks the "Other" option on a form question, their answer is
+# folded into "Other: <text>" (see the other-option Stimulus controller). On the
+# sector/category-backed questions the chosen ids are joined with the free text,
+# e.g. "5, 12, Other: Equine therapy". Those free-text values can't be Sector or
+# Category records, so they never become tags — this pulls them out so they can
+# be surfaced alongside the real tags.
+module OtherOption
+  PREFIX = "Other"
+
+  # The free-text portions of a stored answer, e.g.
+  # "5, Other: Equine therapy" => [ "Equine therapy" ]. A bare "Other" with no
+  # accompanying text yields nothing.
+  def self.texts(submitted_answer)
+    submitted_answer.to_s.split(", ").filter_map do |token|
+      stripped = token.strip
+      next unless stripped.downcase.start_with?("#{PREFIX.downcase}:")
+      stripped[(PREFIX.length + 1)..].strip.presence
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


app/services/other_option.rb
```diff
@@ -0,0 +1,22 @@
...
+# sector/category-backed questions the chosen ids are joined with the free text,
+# e.g. "5, 12, Other: Equine therapy". Those free-text values can't be Sector or
+# Category records, so they never become tags — this pulls them out so they can
+# be surfaced alongside the real tags.
+module OtherOption
+  PREFIX = "Other"
+
+  # The free-text portions of a stored answer, e.g.
+  # "5, Other: Equine therapy" => [ "Equine therapy" ]. A bare "Other" with no
+  # accompanying text yields nothing.
```
maebeale, line 14: 🤖 _From Claude:_ Matches the prefix case-insensitively and skips a bare `Other` (no text) so empty chips never render. Free text containing `", "` will split because answers are stored comma-joined — a known limitation of the storage format, not introduced here.

app/models/person.rb
```diff
@@ -207,8 +207,34 @@ def remote_search_label
...
+    other_form_responses(OTHER_WORKSHOP_SETTING_IDENTIFIERS)
+  end
+
   private
 
+  def other_form_responses(identifiers)
+    form_submissions
+      .joins(form_answers: :form_field)
+      .where(form_fields: { field_identifier: identifiers })
+      .pluck("form_answers.submitted_answer")
```
maebeale, line 233: 🤖 _From Claude:_ Sourced from `form_answers` rather than a person attribute because the "Other" value has no `Sector`/`Category` to hang off of. `.uniq` collapses the same free text repeated across multiple registrations.

app/views/people/show.html.erb
```diff
@@ -134,9 +135,10 @@
                                      is_leader: si.is_leader %>
                       <% end %>
                     </div>
```
maebeale, line 138: 🤖 _From Claude:_ Suppress "None selected." when there are no real sector tags but there is an Other response, so the section never reads as empty while showing a chip.

app/views/people/_form.html.erb
```diff
@@ -131,11 +131,13 @@
                                               show_admin_flags: true } },
                                   class: "btn btn-secondary-outline" %>
     </div>
+    <%= render "people/other_responses", responses: @person.other_service_area_responses %>
   </div>
 
   <!-- Profile-specific Categories -->
   <% if @person_categories_grouped.present? %>
     <div class="form-group space-y-4">
```
maebeale, line 139: 🤖 _From Claude:_ Category "Other" chips live inside the existing `@person_categories_grouped.present?` block — the workshop-settings section only exists when profile-specific categories are configured, which is also where these belong.

db/seeds/dev/events_management.rb
```diff
@@ -559,6 +559,29 @@
...
+# field (Primary service area) and a category-backed field (Workshop Settings).
+# These free-text values can't be Sector/Category records, so they only surface
+# via Person#other_service_area_responses / #other_workshop_setting_responses.
+# Seeded before the professional-answer enrichment below so the primary_service_area
+# value survives its "skip if already answered" guard. Idempotent.
+if facilitator_training && amy_person
+  amy_submission = FormSubmission.find_by(person: amy_person, form: facilitator_training.registration_form)
+  if amy_submission
+    {
+      "primary_service_area" => "Other: Equine-assisted therapy",
```
maebeale, line 574: 🤖 _From Claude:_ Seeded before the professional-answer enrichment below so this `primary_service_area` value survives that block's "skip if already answered" guard. Amy still gets real sector tags from the enrichment, so her profile shows the real tags **and** this "Other" chip.

app/views/people/_other_responses.html.erb
```diff
@@ -0,0 +1,13 @@
+<%# Read-only chips for free-text "Other" answers pulled from registration form
+    responses. These aren't Sector/Category records, so they're shown for
+    reference rather than as editable tags. Renders bare chips — the caller
+    supplies the surrounding flex container so they sit at the end of the
+    relevant section (sectors box, workshop-settings card, etc.). %>
+<% responses ||= [] %>
+<% responses.each do |text| %>
```
maebeale, line 7: 🤖 _From Claude:_ Partial now emits bare chips (no wrapper) so each caller can drop them into its own flex container — that lets the chips sit as peers at the end of the green Sectors box / Primary sector row / Workshop Settings card instead of floating below as an orphaned block.

PR #1672, opened 2026-06-14
Add default max-characters safety net for free-form answers

maebeale:

- Free-form answer fields (`free_form_input_one_line`, `free_form_input_paragraph`) with no explicit `max_characters` previously accepted **unbounded** text into a TEXT column.
- A single pathological submission (megabyte answer) could bloat the DB and break admin list / email rendering — there was no client-side `maxlength` either.
- Separately, an admin could configure a `min_words`/`max_characters` pair that no applicant could satisfy.
- This adds a generous per-type **default cap as a silent safety net**, a validation against **contradictory min/max settings**, and documents the answer-escaping strategy — without taking away the ability to have long inputs.

**Default max-characters safety net**
- Added `FormField::DEFAULT_MAX_CHARACTERS` (1,000 for one-line, 10,000 for paragraph) and an `effective_max_characters` method: explicit `max_characters` when set, otherwise the per-type default, `nil` for non-free-form fields.
- Routed `max_characters_error` through `effective_max_characters` so server-side validation enforces the default when nothing is configured.
- Used `effective_max_characters` for the browser `maxlength` attribute in the registration and bulk-payment forms (silent cap).
- **An explicit per-field `max_characters` always wins, including when larger than the default** — so longer answers stay possible by opting in (e.g. set 50,000). The visible "Maximum of X characters" helper text remains tied to the *explicit* value, keeping the default invisible to normal use.

**Contradictory min/max validation**
- Added a `FormField` validation (fires on form-edit save) that blocks an unsatisfiable pair, e.g. min 250 words capped at 1,000 chars — applicants would hit `maxlength` before reaching the word minimum and could never submit.
- Requires `max_characters >= min_words * 6` (a realistic average word length incl. spacing). Checked against the **explicit** max only; the generous default never conflicts.

**Sanitization**
- Confirmed every place an answer is rendered uses plain `<%= %>` auto-escaping (no `raw`/`html_safe`/`simple_format`), so there is no active XSS exposure. Rather than lossy `strip_tags`-on-write (which would mangle legitimate answers like `revenue < $5k`), documented the escape-on-output strategy on `FormAnswer` as a guardrail against future regressions. Brakeman is clean.

### UI Testing Checklist

- [ ] Registration free-form fields with no configured max stop accepting input at the default cap and reject over-length on submit
- [ ] A field with an explicit `max_characters` larger than the default still allows up to that larger value
- [ ] The "Maximum of X characters" helper text only shows when an explicit max is set
- [ ] Saving a form field with min 250 words and max 1,000 chars is blocked with a clear error

### Anything else to add?

- No DB change — the column default stays `nil` so "blank = use default" admin semantics are preserved; the cap is applied at the validation/render layer.
- Tests: `spec/models/form_field_spec.rb` (effective cap, default error, min/max consistency) and `spec/services/form_answer_validator_spec.rb` (default surfaces on submission). Full suite for touched files green; lint and security clean.

### DIFF
--- a/app/models/form_answer.rb
+++ b/app/models/form_answer.rb
@@ -1,3 +1,12 @@
+# `submitted_answer` is stored verbatim — exactly what the applicant typed — so
+# their words are never silently altered (e.g. "revenue < $5k" or "C++ <T>"
+# survive intact). It is NOT HTML-sanitized on write.
+#
+# Safety is enforced at render time instead: every place that displays an answer
+# uses plain `<%= %>`, which auto-escapes, so any markup shows as inert text.
+# Do NOT render a submitted_answer through `raw`, `html_safe`, `<%==`, or
+# `simple_format` — that would reintroduce an XSS hole. Volume abuse (giant
+# answers) is bounded separately by FormField's effective max-characters cap.
 class FormAnswer < ApplicationRecord
   belongs_to :form_field, optional: true
   belongs_to :form_submission
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -11,6 +11,16 @@ class FormField < ApplicationRecord
   # Answer types that collect free-form text, where a minimum word count applies
   FREE_FORM_TEXT_TYPES = %w[free_form_input_one_line free_form_input_paragraph].freeze
 
+  # Fallback character ceilings applied when a free-form field has no explicit
+  # max_characters set. This is a safety net against pathological submissions
+  # (megabyte answers that bloat the DB and break admin/email rendering), not a
+  # UX limit — the values are far above any realistic answer, and an explicit
+  # per-field max_characters always wins, including when it is larger.
+  DEFAULT_MAX_CHARACTERS = {
+    "free_form_input_one_line" => 1_000,
+    "free_form_input_paragraph" => 10_000
+  }.freeze
+
   # Validations
   validates_presence_of :name
   # Keeps an over-long name as a friendly validation error instead of a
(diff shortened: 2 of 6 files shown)


app/models/form_field.rb
```diff
@@ -11,6 +11,16 @@ class FormField < ApplicationRecord
   # Answer types that collect free-form text, where a minimum word count applies
   FREE_FORM_TEXT_TYPES = %w[free_form_input_one_line free_form_input_paragraph].freeze
 
+  # Fallback character ceilings applied when a free-form field has no explicit
+  # max_characters set. This is a safety net against pathological submissions
+  # (megabyte answers that bloat the DB and break admin/email rendering), not a
+  # UX limit — the values are far above any realistic answer, and an explicit
+  # per-field max_characters always wins, including when it is larger.
+  DEFAULT_MAX_CHARACTERS = {
```
maebeale, line 19: 🤖 _From Claude:_ Defaults are a safety net, not a UX limit — set far above any realistic answer. An explicit per-field `max_characters` overrides this (including when larger), so long inputs stay possible by opting in.

app/models/form_field.rb
```diff
@@ -136,15 +146,24 @@ def min_words_error(value)
     "must be at least #{min_words} #{"word".pluralize(min_words)}"
   end
 
+  # The character ceiling actually enforced for this field: the explicit
+  # max_characters when set, otherwise the per-type default safety net. Returns
+  # nil for non-free-form fields (no cap applies).
+  def effective_max_characters
```
maebeale, line 172: 🤖 _From Claude:_ Single source of truth for the cap (explicit ⟶ default ⟶ nil). `max_characters_error` and the view `maxlength` attrs all route through this so server and browser enforcement agree.

app/models/form_answer.rb
```diff
@@ -1,3 +1,12 @@
+# `submitted_answer` is stored verbatim — exactly what the applicant typed — so
+# their words are never silently altered (e.g. "revenue < $5k" or "C++ <T>"
+# survive intact). It is NOT HTML-sanitized on write.
+#
+# Safety is enforced at render time instead: every place that displays an answer
```
maebeale, line 5: 🤖 _From Claude:_ Chose escape-on-output over `strip_tags`-on-write: stripping is lossy (mangles `revenue < $5k`, `C++ <T>`) and only hides a future `raw`/`html_safe` regression rather than preventing it. All current render sites already auto-escape; Brakeman is clean.

maebeale (commented):
🤖 _From Claude:_ Notes on the key decisions in this change.

app/models/form_field.rb
```diff
@@ -136,15 +152,38 @@ def min_words_error(value)
     "must be at least #{min_words} #{"word".pluralize(min_words)}"
   end
 
+  # Blocks an admin from saving a free-form field whose explicit max_characters
+  # is too small to ever satisfy its min_words minimum (e.g. 250 words capped at
+  # 1,000 characters) — otherwise the field could never be submitted. Compared
+  # against the explicit max only; the generous default never conflicts.
+  def max_characters_allows_min_words
```
maebeale, line 159: 🤖 _From Claude:_ Guards against an unsatisfiable field (e.g. min 250 words capped at 1000 chars — applicants hit `maxlength` before reaching the word minimum and can never submit). The 6 chars/word floor (`MIN_CHARS_PER_WORD`) is a realistic average; bump it for a stricter check. Validated against the explicit max only, since the default never conflicts.

PR #1673, opened 2026-06-14
Validate selectable answers against the field's offered options

maebeale:

- Follow-up to #1672. Answers to **selectable fields** (radio / dropdown / checkbox) were stored verbatim with **no server-side check** that the value was actually one of the options the form presented.
- A tampered/forged request could submit an arbitrary value — or a **valid-but-unoffered Sector/Category id** — which would be stored and then counted in dashboard analytics and shown to admins, polluting registrant data.
- Not an XSS or auth issue (output is auto-escaped; an attacker can only affect their own registration), but a real **data-integrity** gap. This closes it with a server-side allowlist check.

- Added `FormField#answer_inclusion_error(value)` (mirroring `min_words_error` / `max_characters_error`) and wired it into `FormAnswerValidator`'s error chain. It rejects any submitted selectable value that isn't one of the field's offered options.
- Handles **single and multi-select**, and still allows **"Other" / "Other: \<text>"** free-text answers when the field offers an Other option.
- **Single source of truth:** to guarantee validation matches exactly what the form *renders* (so it can never reject a legitimate submission), the option-source knowledge is now centralized on `FormField`:
  - Relocated `DYNAMIC_FIELD_CATEGORY_TYPES`, `SERVICE_AREA_FIELD_IDENTIFIERS`, and `OTHER_OPTION_PREFIX` onto the model.
  - Pointed the view helpers (`dynamic_form_field_options`, `form_field_option_source`, `EventHelper`) at the relocated constants.
  - `FormField#allowed_answer_values` derives the valid set the same way the form builds its options (stored option names, or published Sector/Category ids).
- **Strictness decision:** if an option is renamed/removed mid-session, the submission is rejected (strict — favors data correctness; the form re-renders current options on load). Easy to soften later if it proves user-facing.

### UI Testing Checklist

- [ ] Submitting a normal radio/dropdown/checkbox selection still works
- [ ] A forged value (not in the option list) is rejected with "has an invalid selection"
- [ ] An "Other: \<text>" answer still submits on fields that offer Other
- [ ] Dynamic fields (service area, age range, etc.) still render and accept valid selections

### Anything else to add?

- No DB or schema change; validation only.
- Tests: `spec/models/form_field_spec.rb` (`#answer_inclusion_error`, incl. dynamic Sector/Category cases and Other handling) and `spec/services/form_answer_validator_spec.rb` (inclusion surfaces on submission). Full registration request + system specs, helper specs, lint, and Brakeman all green.
- Based on `main` (which already includes #1672); independent of that PR.

### DIFF
--- a/app/helpers/event_helper.rb
+++ b/app/helpers/event_helper.rb
@@ -1,6 +1,7 @@
 module EventHelper
   # The special free-text option label that reveals a "please specify" input.
-  OTHER_OPTION_PREFIX = "Other"
+  # Canonical definition lives on FormField (shared with answer validation).
+  OTHER_OPTION_PREFIX = FormField::OTHER_OPTION_PREFIX
 
   # True when an option label is the special free-text "Other" choice.
   def other_option?(label)
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -11,6 +11,24 @@ class FormField < ApplicationRecord
   # Answer types that collect free-form text, where a minimum word count applies
   FREE_FORM_TEXT_TYPES = %w[free_form_input_one_line free_form_input_paragraph].freeze
 
+  # Field identifiers whose selectable options are sourced dynamically from
+  # Sector records rather than the field's own stored answer options. The
+  # submitted value for these is a Sector id (as a string).
+  SERVICE_AREA_FIELD_IDENTIFIERS = %w[primary_service_area primary_service_area_single].freeze
+
+  # Field identifiers whose selectable options are sourced dynamically from a
+  # CategoryType's published categories. The submitted value is a Category id
+  # (as a string). Maps the field identifier to its backing CategoryType name.
+  DYNAMIC_FIELD_CATEGORY_TYPES = {
+    "workshop_environments" => "WorkshopEnvironment",
+    "client_life_experiences" => "StoryPopulation",
+    "primary_age_group" => "AgeRange"
+  }.freeze
+
+  # The special free-text option label that lets a respondent supply their own
+  # value; a chosen "Other" answer is stored as "Other" or "Other: <text>".
+  OTHER_OPTION_PREFIX = "Other"
+
   # Fallback character ceilings applied when a free-form field has no explicit
   # max_characters set. This is a safety net against pathological submissions
   # (megabyte answers that bloat the DB and break admin/email rendering), not a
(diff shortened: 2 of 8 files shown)


app/models/form_field.rb
```diff
@@ -186,6 +204,63 @@ def max_characters_error(value)
...
+  def dynamic_options?
+    field_identifier.in?(SERVICE_AREA_FIELD_IDENTIFIERS) ||
+      DYNAMIC_FIELD_CATEGORY_TYPES.key?(field_identifier)
+  end
+
+  # The set of values a submission may legitimately contain for this selectable
+  # field — stored option names, or dynamic Sector/Category ids as strings —
+  # exactly mirroring what the public form renders. nil for non-selectable
+  # fields. This is the source of truth for both rendering and validation.
+  def allowed_answer_values
```
maebeale, line 218: 🤖 _From Claude:_ Deliberately built the same way the form renders options (stored names, or published Sector/Category ids) so the validator can never reject a value the form actually offered. Centralizing the option-source constants on the model is what makes that guarantee hold.

app/models/form_field.rb
```diff
@@ -186,6 +204,63 @@ def max_characters_error(value)
...
+
+    value == OTHER_OPTION_PREFIX || value.start_with?("#{OTHER_OPTION_PREFIX}:")
+  end
+
+  # Returns a validation error string when a submitted selectable value is not
+  # one of the field's offered options — guarding against tampered/forged
+  # submissions — or nil when it passes / does not apply. Handles single and
+  # multi-select, and allows "Other" / "Other: <text>" when the field offers it.
+  # Blank values are left to the presence (required) check.
+  def answer_inclusion_error(value)
```
maebeale, line 253: 🤖 _From Claude:_ Handles single + multi-select via `Array(value)`, leaves blanks to the required check, and permits `Other`/`Other: <text>` only when the field offers Other. Strict on renamed/removed options (rejects) — favors data correctness; the form re-renders current options on load.

app/helpers/event_helper.rb
```diff
@@ -1,6 +1,7 @@
 module EventHelper
   # The special free-text option label that reveals a "please specify" input.
-  OTHER_OPTION_PREFIX = "Other"
+  # Canonical definition lives on FormField (shared with answer validation).
+  OTHER_OPTION_PREFIX = FormField::OTHER_OPTION_PREFIX
```
maebeale, line 4: 🤖 _From Claude:_ Aliased to the model constant rather than removed, so existing `EventHelper::OTHER_OPTION_PREFIX` references keep working while the canonical value lives on FormField (shared with validation).

PR #1675, opened 2026-06-14
Jump-link icons on registrant chips + context-aware back links

maebeale:
- On the event registrants roster, the scholarship and payment chips jump to separate admin pages, but nothing signaled that — so this adds the standard jump-link icon (the same one used elsewhere) to all scholarship chips (Completed, Recipient, Requested, Create) and the payment chips.
- It also fixes a navigation papercut: those destination pages always sent the "back" link to the registration edit page, even when the user arrived from the registrants roster.

- Each registrants-roster chip now passes `return_to: "registrants"`, and the scholarship `new`/`edit` and allocations pages honor that to send the back link to the registrants roster, while arriving from a registration still returns to the registration.
- Added request specs covering both back-link contexts on the scholarship and allocations pages.

### UI Testing Checklist
- [ ] Scholarship and payment chips on `events/:id/registrants` show the jump-link icon.
- [ ] From the registrants roster, opening a scholarship or payment page shows a "← Registrants" back link.
- [ ] From a registration, opening a scholarship or payment page shows a "← Registration" back link.

### Anything else to add?
- The jump-link icon is on all payment states (including "Paid") for visual consistency within the column.

### DIFF
--- a/app/views/events/_registrants_results.html.erb
+++ b/app/views/events/_registrants_results.html.erb
@@ -136,34 +136,38 @@
                 <td class="px-4 py-2 text-sm text-center">
                   <% if (s = registration.scholarships.first) %>
                     <% if s.tasks_completed? %>
-                      <%= link_to edit_scholarship_path(s),
+                      <%= link_to edit_scholarship_path(s, return_to: "registrants"),
                                   class: "inline-flex items-center gap-1.5 rounded-full text-xs font-medium border px-5 py-0.5 bg-green-50 text-green-700 border-green-200 hover:opacity-80",
                                   data: { turbo_frame: "_top" } do %>
                         <i class="fas fa-graduation-cap"></i>
                         <span>Completed</span>
+                        <i class="fa-solid fa-arrow-up-right-from-square text-[0.6rem] opacity-70"></i>
                       <% end %>
                     <% else %>
-                      <%= link_to edit_scholarship_path(s),
+                      <%= link_to edit_scholarship_path(s, return_to: "registrants"),
                                   class: "inline-flex items-center gap-1.5 rounded-full text-xs font-medium border px-5 py-0.5 bg-blue-50 text-blue-700 border-blue-200 hover:opacity-80",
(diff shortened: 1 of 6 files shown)


app/views/scholarships/new.html.erb
```diff
@@ -8,10 +8,14 @@
       <%= link_to (params[:return_to] == "grant_edit" ? edit_grant_path(@grant) : grant_path(@grant)), class: "text-sm text-gray-500 hover:text-gray-700" do %>
         <i class="fa-solid fa-arrow-left text-xs"></i> Grant
       <% end %>
-    <% elsif @allocatable.respond_to?(:event) %>
+    <% elsif @allocatable.respond_to?(:event) && params[:return_to] == "registration" %>
       <%= link_to edit_event_registration_path(@allocatable), class: "text-sm text-gray-500 hover:text-gray-700" do %>
         <i class="fa-solid fa-arrow-left text-xs"></i> Registration
       <% end %>
+    <% elsif @allocatable.respond_to?(:event) %>
+      <%= link_to registrants_event_path(@allocatable.event), class: "text-sm text-gray-500 hover:text-gray-700" do %>
```
maebeale, line 16: 🤖 _From Claude:_ Previously this always pointed back to the registration edit page; now it branches on `return_to` so users arriving from the registrants roster (which passes `return_to: "registrants"`) land back there instead. Mirrors the logic already in `edit.html.erb`.

app/views/events/_registrants_results.html.erb
```diff
@@ -199,6 +203,7 @@
                       <% else %>
                         <span><%= "$%.2f due" % (due_cents / 100.0) %></span>
                       <% end %>
+                      <i class="fa-solid fa-arrow-up-right-from-square text-[0.6rem] opacity-70"></i>
```
maebeale, line 206: 🤖 _From Claude:_ Single icon after the conditional label so it renders on all three payment states (Paid / Partial / due) — they all jump to the same allocations page, so the icon stays consistent across the column.

PR #1676, opened 2026-06-14
Warn before adding, removing, or editing facilitator affiliations

maebeale:
A facilitator affiliation's title and start/end dates drive the organization's status with AWBW, which is recalculated automatically by the `Affiliation` model's save/destroy callbacks. Today an admin can add, remove, or edit one of these rows and silently flip the org's status without realizing it.

Added a `affiliation-facilitator-warning` Stimulus controller on the Organization and Person edit forms that snapshots the affiliation rows on connect and, on submit, shows a confirmation if a facilitator affiliation was added, removed, or edited (title/start/end changed) — gating at submit time since that's when the DB write and status recalculation actually happen. Cancelling aborts the save while preserving the unsaved-changes guard.

### UI Testing Checklist
- [ ] Editing a facilitator affiliation's dates and saving prompts a confirmation; confirming saves, cancelling stays on the form
- [ ] Removing a facilitator affiliation and saving prompts a confirmation
- [ ] Adding a new affiliation (defaults to "Facilitator") and saving prompts a confirmation
- [ ] Editing only a non-facilitator affiliation saves with no prompt

### Anything else to add?
Covered by `spec/system/organization_facilitator_warning_spec.rb`. The same controller is wired to both the organization and person forms since affiliations are editable from either side.

### DIFF
--- a/app/frontend/javascript/controllers/affiliation_facilitator_warning_controller.js
+++ b/app/frontend/javascript/controllers/affiliation_facilitator_warning_controller.js
@@ -1,0 +1,100 @@
+import { Controller } from "@hotwired/stimulus";
+
+// Connects to data-controller="affiliation-facilitator-warning"
+//
+// Warns before saving when a facilitator affiliation (one with "facilitator"
+// in its title) is added, removed, or edited. Those affiliations — together
+// with their start and end dates — determine the organization's status with
+// AWBW, which is calculated dynamically, so the user is asked to confirm
+// the change before the form submits.
+//
+// Attach to the <form> element. Snapshots the affiliation rows on connect and
+// compares them against the current state on submit; if a facilitator-relevant
+// change is detected the user must confirm or the submission is cancelled.
+export default class extends Controller {
+  connect() {
+    this.snapshots = this.captureSnapshots();
+    this.handleSubmit = (event) => this.guardSubmit(event);
+    // Capture phase so this runs before other submit listeners (e.g. dirty-form).
+    this.element.addEventListener("submit", this.handleSubmit, true);
+  }
+
+  disconnect() {
+    this.element.removeEventListener("submit", this.handleSubmit, true);
+  }
+
+  guardSubmit(event) {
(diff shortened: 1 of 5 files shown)


app/frontend/javascript/controllers/affiliation_facilitator_warning_controller.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const message =
+      "You're adding, removing, or editing a facilitator affiliation. " +
+      "This affects the organization's status with AWBW, which is calculated " +
+      "from facilitator affiliations and their start and end dates.\n\n" +
+      "Are you sure you want to save this change?";
+
+    if (!window.confirm(message)) {
+      event.preventDefault();
+      event.stopImmediatePropagation();
```
maebeale, line 37: 🤖 _From Claude:_ Listening in the capture phase (line 18) plus `stopImmediatePropagation` here means a cancelled submit never reaches the form's `dirty-form` handler, so the unsaved-changes guard stays armed.

app/frontend/javascript/controllers/affiliation_facilitator_warning_controller.js
```diff
@@ -0,0 +1,100 @@
...
+      const previous = key ? this.snapshots[key] : null;
+
+      // New row added in this session — relevant if it's a non-removed facilitator.
+      if (!previous) return current.facilitator && !current.destroyed;
+
+      // Removal — relevant only if the row was a facilitator before.
+      if (current.destroyed) return previous.facilitator && !previous.destroyed;
+
+      // Edit — relevant if the row is (or was) a facilitator and any field changed.
+      const involvesFacilitator = previous.facilitator || current.facilitator;
```
maebeale, line 91: 🤖 _From Claude:_ `previous.facilitator || current.facilitator` is intentional — it catches a title edited *into* or *out of* facilitator status, not just edits to an existing facilitator row.

jmilljr24:
Great idea!!

PR #1677, opened 2026-06-14
Keep event datetime fields readable and rebalance the details form layout

maebeale:

- The event details form had two layout problems:
  - The **Start / End / Registration-close** datetime row used a fixed 3-column grid, so at narrower viewports each `datetime-local` input was squeezed until its value truncated (e.g. `08/13/202`), **hiding the time** and making it look like the time couldn't be edited (the registration-close field is a full `datetime`).
  - Conversely, when only two fields fit, they stretched far wider than their value, leaving a large gap between the value and the calendar icon.
  - The surrounding fields (pre-date text, cost, location, videoconference) were laid out in unbalanced rows.

- **Datetime row** — replaced the fixed grid with an auto-fit grid + per-field minimum (`grid-cols-[repeat(auto-fit,minmax(min(100%,13rem),1fr))] gap-3`):
  - Three fields now pack onto one row inside the 2/3-width left column; `1fr` sizes them evenly so there's no leftover whitespace.
  - 13rem is still wide enough to show the full `MM/DD/YYYY, HH:MM AM` value plus the calendar icon.
  - The row reflows 3 → 2 → 1 columns as the viewport narrows instead of truncating.
  - `min(100%, …)` prevents overflow on very narrow screens.
- **Field layout** — reorganized the left column into balanced 50/50 rows:
  - **Event cost** (left) paired with **Pre-date text** (right).
  - **Videoconference** (left) paired with **Location** (right), moved below the cost row.

### UI Testing Checklist

- [ ] Wide viewport: all three datetime fields sit on one row with no large gap; cost+pre-date and videoconference+location each form a 50/50 row
- [ ] Narrowing the window reflows the datetime fields to 2, then 1 per row (no truncation of the time)
- [ ] The time portion of each datetime field is visible and editable at every width

### Anything else to add?

- No model, migration, or behavior changes — purely a layout fix. The registration-close field was already a `datetime` column with a `datetime-local` input.

### DIFF
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -67,7 +67,7 @@
       <div class="grid grid-cols-1 md:grid-cols-3 gap-6">
       <%# LEFT: Date/time, cost + pre-date text, videoconference + location %>
       <div class="md:col-span-2 space-y-6">
-      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,15rem),1fr))] gap-3">
+      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,13rem),1fr))] gap-3">
         <div>
           <%= f.label :start_date,
                   "Start time",
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -76,7 +76,7 @@
                 } %>
       </div>
 
-      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,16rem),1fr))] gap-4">
+      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,15rem),1fr))] gap-3">
         <div>
           <%= f.label :start_date,
                   "Start time",
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -65,17 +65,8 @@
 
     <div id="event_details" class="bg-gray-50 border border-gray-300 rounded-b-md p-6">
       <div class="grid grid-cols-1 md:grid-cols-3 gap-6">
-      <%# LEFT: Date/time, location, videoconference, cost %>
+      <%# LEFT: Date/time, cost + pre-date text, videoconference + location %>
       <div class="md:col-span-2 space-y-6">
-      <div>
-        <%= f.input :pre_date_text,
-                label: "Pre-date text",
-                input_html: {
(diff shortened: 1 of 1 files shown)


app/views/events/_form.html.erb
```diff
@@ -76,7 +76,7 @@
                 } %>
       </div>
 
-      <div class="grid grid-cols-1 md:grid-cols-3 gap-4">
+      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,16rem),1fr))] gap-4">
```
maebeale: 🤖 _From Claude:_ `min(100%, 16rem)` is intentional — the bare `minmax(16rem, 1fr)` would overflow horizontally on viewports narrower than 256px; wrapping the track minimum in `min()` lets it fall back to `100%` there. 16rem is the smallest width that still shows the full `MM/DD/YYYY, HH:MM AM` value plus the calendar icon.

app/views/events/_form.html.erb
```diff
@@ -65,18 +65,9 @@
...
-        <%= f.input :pre_date_text,
-                label: "Pre-date text",
-                input_html: {
-                  class: "w-full rounded border-gray-300 shadow-sm px-3 py-2 focus:ring-blue-500 focus:border-blue-500",
-                  placeholder: "e.g. Upcoming Workshop"
-                } %>
-      </div>
-
-      <div class="grid grid-cols-1 md:grid-cols-3 gap-4">
+      <div class="grid grid-cols-[repeat(auto-fit,minmax(min(100%,15rem),1fr))] gap-3">
```
maebeale: 🤖 _From Claude:_ Final sizing is 15rem min + `gap-3` (was 16rem/`gap-4`): three fields need 3×15rem+2×0.75rem ≈ 744px, which fits the form on common laptop windows (~760–790px) so it stays 3-up there, while 15rem is still wide enough to show the full `MM/DD/YYYY, HH:MM AM` value + icon. `min(100%, …)` keeps it from overflowing below 240px.

jmilljr24:
I noticed Christy useing Firefox ( as well as Rudy in past meeting) which both have the issue of no native datetime picker. When the field is datetime the browser only has a date picker. You cannot submit the time.

There are some convoluted options with js that are brittle to browser changes.I think the least intrusive option is to have to separate form inputs for date and time, then we can merge them vs needing to make true separate db columns.

PR #1679, opened 2026-06-15
Bulk payment allocation UI and styling

maebeale:
- Admins need to see and manage where a bulk payment's money is allocated, directly from the bulk payments page.
- The recently added allocation views (#1678) didn't match the app's established component patterns; this brings them in line and fills in the missing workflow.

- **Bulk payment cards** — expandable per-submission cards that fuzzy-match submitted attendees to active registrants, with columns for Profile / Registration / Allocations / **Action**. The Action column holds the inline allocate form (amount + Allocate); there's also a searchable "New allocation" form for non-matched registrants.
- **Responsive submitted-info rows** — the per-attendee rows use a fixed-track grid so columns align row-to-row, gated on a **container query** (`@4xl`) rather than a viewport breakpoint: when the card is wide the aligned grid shows; when it's narrow (e.g. sidebar open) each attendee falls back to a stacked, wrapping layout instead of being crushed. Rows have a hover state.
- **Cross-card Turbo broadcast** — allocating updates the affected registration's due/allocated totals (and clears the inline allocate box once fully paid) on **every** card that lists the same registrant, since one payer can appear across multiple submissions.
- **Performance** — preload allocation sums once per page via a single grouped query (`allocated_cents_by_registration`), instead of querying per registration per card.
- **Dashboard equation** — added **unallocated bulk payments** as a clickable addend in the grand-total equation, and made each addend link to its area. New `bulk_payments` domain theme color (amber).
- **Confirmation page** — added admin Dashboard / Bulk payments links and a read-only allocations section to the post-checkout bulk payment page (admin-gated).
- **Navigation** — context-aware back links (`return_to` / `expand`) so links that open in a new tab return the user to exactly where they came from. Returning expands that submission's card server-side (the view omits the `hidden` class based on `params[:expand]`) and the matching `#payment-card-<id>` anchor scrolls to it; the dropdown controller's `open` value is synced from the same flag so its state matches the rendered card. Centralized in `EventHelper#bulk_payments_return_path`.

### Rebase onto main (#1680 + #1681)
- **#1680 (event on form submissions):** the list reads submissions via `Event#form_submissions` (stored `event_id`); the dev seed + request specs set the event on bulk-payment submissions to match the production `BulkPayment` service.
- **#1681 (allocation refactor):** integrated main's additions on top of this branch's styled UI —
  - Adopted `FormSubmissionDecorator#matched_attendees` / `answers_by_identifier` (matching logic moved out of the card partial).
  - Kept main's new **record-payment** feature (`create_bulk_payment` action + `_bulk_payment_form` + `conditional_fields` Stimulus controller + route); wired it to re-render this branch's card (expanded, with preloaded totals).
  - Standardized the shared local on `event_registrations`. (The card root id is `payment-card-<id>` so it matches the return anchor; the turbo_stream replace targets use the same id.)
  - **Decision:** kept this branch's searchable `remote-select` registrant picker for "New allocation" (main had swapped it for a plain select); the record-payment form still uses `remote-select` for person/organization, so the controller remains in use.

### UI Testing Checklist
- [ ] Bulk payments page: cards expand, attendee matches render, inline + searchable allocate forms work
- [ ] Allocating updates the badge, allocations table, and remaining balance via Turbo — including other cards that list the same registrant
- [ ] Recording a cash/check payment on a no-payment submission works and re-renders the card expanded
- [ ] Dashboard: unallocated bulk payments shows as an equation addend and each addend links to the right place
- [ ] Submitted-info rows: columns align when the card is wide, and stack/wrap cleanly (no character-by-character email wrap) when the card is narrow

### Anything else to add?
- Known limitation: the Pending→Paid transition for card payments depends on the Stripe `charge.succeeded` webhook creating the `Payment`. There's also a latent conflict between two Stripe-charge → payment creation paths (`PayChargeExtensions#create_bulk_payment` vs the generic `StripeChargeSucceededProcessor` from #1621) — not addressed here, flagged for follow-up.


app/controllers/events/bulk_payments_controller.rb
```diff
@@ -60,6 +60,7 @@ def show
       authorize! :bulk_payment, to: :show?
 
       @submission = FormSubmission.find(params[:submission_id])
+      @payment = @submission.payments.first
```
maebeale: @jmilljr24 are there ever multiple payments created by the webhook? like if the first one created failed doesn't it create another one that would also be associated w this form? or does it just update this one payment to success state?

jmilljr24: A submission should only have one payment, Model should be updated to reflect that. `has_one` payment.

If they choose `credit card (now)` the form submission is created. If the checkout succeeds, then the payment is created and associated to the form_submission. If the checkout fails we keep the form_submission but no Payment.

app/views/events/_bulk_payment_card.html.erb
```diff
@@ -1,102 +1,186 @@
...
+            <h5 class="text-xs font-medium text-gray-400 uppercase tracking-wide mb-2">New allocation</h5>
             <%= form_tag allocate_bulk_payment_event_path(@event),
                          class: "flex items-center gap-2" do %>
               <%= hidden_field_tag :payment_id, payment.id %>
-              <%= select_tag :event_registration_id,
-                   options_from_collection_for_select(event_registrations, :id, ->(er) { "#{er.registrant.name} — #{er.registrant.preferred_email}" }),
-                   prompt: "Select registrant...",
-                   class: "w-64 rounded-lg border border-gray-300 bg-white px-2 py-1.5 text-sm shadow-sm" %>
+              <div class="w-64 shrink-0 rounded-lg border border-gray-300 bg-white px-3 py-2 shadow-sm focus-within:ring focus-within:ring-blue-200 focus-within:border-blue-500">
+                <%= select_tag :event_registration_id, nil,
```
jmilljr24: I forgot to mention this. I went back to a simple select because the remote-select would need to be updated to scope the event_registrations to just this event. It seemed overly complex when the list of registrants is relatively small.

PR #1680, opened 2026-06-15
Record the event each form submission belongs to

maebeale:
Form submissions had no direct link to their event — the event was inferred through the form's join role, which returned `nil` for role-less registration submissions, so we couldn't reliably tell (or filter by) which event a form was submitted for. This adds an explicit `event_id` and switches event-scoped queries to use it.

- Added a nullable, indexed `event_id` FK to `form_submissions` and a `belongs_to :event, optional: true`; replaced the indirect `#event` method. Every creation site (bulk payment, registration create/update, scholarship) now sets the event.
- Added `Event#form_submissions` (`dependent: :restrict_with_error`, to avoid orphaning payments) and moved event-scoped queries onto the column: the bulk-payment list, the public-registration "answers on file" hide check, and the registration-form icon on the registrations index.
- These read `event_id` only — pre-existing/genuinely event-less submissions are intentionally **not** backfilled, so the column drives new/forward filtering. A helper (`form_submission_link_path`) still infers the event for older rows when building view links.
- Factory is event-less by default with a `:with_event` trait; seed cohort submissions are linked to their event.

### UI Testing Checklist
- [ ] Event bulk-payments page lists that event's bulk-payment submissions.
- [ ] On the registrations index, registrants who submitted their event's form show the green "completed form" icon.

### Anything else to add?
`event_id` is kept nullable since some submissions legitimately have no event. If we later confirm production data, a follow-up could backfill historical rows so they're filterable too.

### DIFF
--- a/app/models/form_submission.rb
+++ b/app/models/form_submission.rb
@@ -1,17 +1,11 @@
 class FormSubmission < ApplicationRecord
   belongs_to :person
   belongs_to :form
+  belongs_to :event, optional: true
   has_many :form_answers, dependent: :destroy
   has_many :payments
 
   accepts_nested_attributes_for :form_answers
-
-  # The event this submission belongs to, resolved through the join role that
-  # matches the submission's own role (e.g. a "bulk_payment" submission maps to
-  # the event_form with role "bulk_payment").
-  def event
-    form.events.find_by(event_forms: { role: role })
-  end
 
   # Answers keyed by their field's identifier. Bulk payment (and similar) forms
   # address fields by identifier rather than position.
--- a/db/migrate/20260615115052_add_event_to_form_submissions.rb
+++ b/db/migrate/20260615115052_add_event_to_form_submissions.rb
@@ -1,19 +1,6 @@
 class AddEventToFormSubmissions < ActiveRecord::Migration[8.0]
   def up
-    add_reference :form_submissions, :event, null: true, index: true
-    add_foreign_key :form_submissions, :events unless foreign_key_exists?(:form_submissions, :events)
-
-    # Backfill the event for existing submissions by matching the join role.
-    # Registration submissions carry no role, so they resolve through the
-    # "registration" event_form.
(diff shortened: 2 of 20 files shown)


db/migrate/20260615115052_add_event_to_form_submissions.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    # Backfill the event for existing submissions by matching the join role.
+    # Registration submissions carry no role, so they resolve through the
+    # "registration" event_form.
+    execute(<<~SQL.squish)
+      UPDATE form_submissions fs
+      JOIN event_forms ef
+        ON ef.form_id = fs.form_id
+        AND ef.role = COALESCE(fs.role, 'registration')
+      SET fs.event_id = ef.event_id
```
maebeale: 🤖 _From Claude:_ Backfill resolves the event through the form's join role, mapping role-less registration submissions to the `registration` event_form (the old indirect lookup returned nil for these). Kept nullable + best-effort so any unresolvable legacy row doesn't block the migration.

app/models/form_submission.rb
```diff
@@ -1,18 +1,12 @@
 class FormSubmission < ApplicationRecord
   belongs_to :person
   belongs_to :form
+  belongs_to :event, optional: true
```
maebeale, line 4: 🤖 _From Claude:_ `optional: true` because `event_id` is nullable for backfill safety; every creation path now sets the event, so new rows are always populated. Can tighten to required in a follow-up once prod backfill is confirmed null-free.

db/migrate/20260615115052_add_event_to_form_submissions.rb
maebeale: removed data changes from migration

app/models/form_submission.rb
maebeale: it's not just for backfill. some form_submissions later won't have an event.

PR #1684, opened 2026-06-15
Omit the "Other" sector from the primary service-area dropdown

maebeale:

- A respondent's single **primary** service area must be a concrete sector, so offering the catch-all **Other** choice there is misleading.
- The multi-select **additional** service-areas field still needs **Other**, so the two fields must diverge.

- Added `Sector.excluding_other` scope and an `OTHER_SECTOR_NAME` constant as the single source of truth for the catch-all sector.
- Centralized the offered sectors in `FormField#service_area_sectors`: the single-select `primary_service_area_single` field excludes **Other**, the multi-select `primary_service_area` field includes it.
- Routed both the public form's dropdown rendering (`dynamic_form_field_options`) and submission validation (`answer_inclusion_error`) through that method so they stay in agreement — a value rejected in the UI is also rejected server-side.

### Anything else to add?

- No migration: this only changes which existing published sectors are offered/accepted for the primary field.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -16,6 +16,11 @@ class FormField < ApplicationRecord
   # submitted value for these is a Sector id (as a string).
   SERVICE_AREA_FIELD_IDENTIFIERS = %w[primary_service_area primary_service_area_single].freeze
 
+  # The single-select "primary" service-area field. Unlike the multi-select
+  # "additional" field, it omits the catch-all "Other" sector — a respondent's
+  # primary service area must be a concrete sector.
+  PRIMARY_SERVICE_AREA_FIELD_IDENTIFIER = "primary_service_area_single"
+
   # Field identifiers whose selectable options are sourced dynamically from a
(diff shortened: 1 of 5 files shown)


app/models/form_field.rb
```diff
@@ -230,6 +235,15 @@ def allowed_answer_values
     values.to_set
   end
 
+  # The published Sector records a service-area field offers, in name order. The
+  # single-select "primary" field omits the catch-all "Other" sector; the
+  # multi-select "additional" field includes it. Source of truth shared by the
+  # public form's rendering and submission validation.
+  def service_area_sectors
+    scope = Sector.published.order(:name)
+    field_identifier == PRIMARY_SERVICE_AREA_FIELD_IDENTIFIER ? scope.excluding_other : scope
```
maebeale, line 244: 🤖 _From Claude:_ This method is the single source of truth shared by `dynamic_form_field_options` (rendering) and `answer_inclusion_error` (validation), so the dropdown options and the accepted submission values can never drift apart.

PR #1686, opened 2026-06-15
Add the ability to archive a person and their user account

maebeale:

- AWBW needed a **reversible** way to take a person out of circulation without losing their data.
- This adds soft-delete (**archiving**) for people and their user accounts.

> **Scope note:** Hard deletion of people is handled in a separate PR. This PR is archive-only.

**Soft-delete via the `discard` gem** (not `paranoia`):
- `discard` adds `discarded_at` + `kept`/`discarded` scopes without overriding `destroy` or installing a `default_scope`, so the app's existing `dependent: :destroy` cascades keep working unchanged. `paranoia`'s own README marks it "not recommended for new projects."

**Behavior:**
- **Archive** discards the person **and** their user together. A discarded user is blocked from authenticating (`active_for_authentication?`) and hidden from default people/user listings. Fully reversible via **Restore**.
- Admins get **Archive / Restore** controls + an "Archived" badge on the profile, and a **View archived** toggle on the people index.

**Details:**
- `Discard::Model` on `Person` and `User`; `User#active_for_authentication?` + `inactive_message` block discarded users; `has_access` scoped `.kept`.
- New `PersonArchivalService` (`archive!` / `restore!`), wrapped in a transaction so person + user stay in sync.
- `PersonPolicy#archive?` (admin); `relation_scope` filters `.kept` so archived people drop out of non-admin scopes.
- Controller `archive` / `unarchive` member routes. People & users indexes hide archived by default with a `?archived=true` toggle.
- Migration adds indexed `discarded_at` to `people` and `users` (reversible up/down).

### UI Testing Checklist

- [ ] As admin, archive a person → redirected to profile, "Archived" badge shows, person gone from default People list
- [ ] Archived person appears under **View archived**; **Restore** brings them back
- [ ] Archived user cannot log in

### Anything else to add?

- Tests: service, policy, request, and model specs (archive/restore, login block, archived filter). All green; RuboCop clean.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -1,6 +1,6 @@
 class PeopleController < ApplicationController
   include AhoyTracking, TagAssignable
-  before_action :set_person, only: %i[ show edit update destroy workshop_logs checkout bio ]
+  before_action :set_person, only: %i[ show edit update destroy archive unarchive workshop_logs checkout bio ]
 
   def index
     authorize!
@@ -14,6 +14,7 @@ class PeopleController < ApplicationController
         affiliations: :organization,
         categorizable_items: { category: :category_type }
       ).references(:user))
+      base_scope = ActiveModel::Type::Boolean.new.cast(params[:archived]) ? base_scope.discarded : base_scope.kept
       filtered = base_scope.search_by_params(params.to_unsafe_h)
                            .order(:first_name, :last_name)
       @count_display = filtered.count
@@ -224,6 +225,24 @@ class PeopleController < ApplicationController
 
     respond_to do |format|
       format.html { redirect_to people_path, status: :see_other, notice: "Person was successfully destroyed." }
+    end
+  end
+
+  def archive
+    authorize! @person, to: :archive?
+    PersonArchivalService.new(@person).archive!
+
+    respond_to do |format|
+      format.html { redirect_to @person, status: :see_other, notice: "Person was archived." }
+    end
+  end
+
+  def unarchive
+    authorize! @person, to: :archive?
+    PersonArchivalService.new(@person).restore!
+
+    respond_to do |format|
+      format.html { redirect_to @person, status: :see_other, notice: "Person was restored." }
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  include Discard::Model
+
   # Include default devise modules. Others available are:
   # :confirmable, :timeoutable and :omniauthable
   devise :database_authenticatable, :recoverable, :confirmable,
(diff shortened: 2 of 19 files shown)


app/services/person_deletion_service.rb
```diff
@@ -0,0 +1,48 @@
...
+    purge_ahoy_records(user)
+    user.destroy!
+  end
+
+  # Ahoy visits/events reference the user by id without an association, so remove
+  # them explicitly. Events are deleted both by user_id and by their parent visit
+  # so none are left orphaned.
+  def purge_ahoy_records(user)
+    visit_ids = Ahoy::Visit.where(user_id: user.id).pluck(:id)
+    Ahoy::Event.where(user_id: user.id).or(Ahoy::Event.where(visit_id: visit_ids)).delete_all
```
maebeale: 🤖 _From Claude:_ Ahoy events are deleted by both `user_id` and parent `visit_id` so events on the user's visits that were never attributed to a user (e.g. pre-login) are not left orphaned.

app/policies/person_policy.rb
```diff
@@ -48,9 +61,13 @@ def owner?
     record.user == user
   end
 
-  def has_associated_data?
-    record.user.present? ||
-      record.affiliations.exists? ||
-      record.stories_as_spotlighted_facilitator.exists?
+  def has_significant_associated_data?
```
maebeale: 🤖 _From Claude:_ A bare user account is intentionally **not** "significant" here — that's exactly the "person only has a user" case we want plain Delete to allow. Affiliations/registrations/scholarships/etc. still block it.

app/models/user.rb
```diff
@@ -118,7 +120,11 @@ def remote_search_label
   end
 
   def active_for_authentication?
```
maebeale, line 122: 🤖 _From Claude:_ Discarded (archived) users are blocked from authenticating here in addition to the existing inactive check, so archiving an account immediately revokes login.

app/controllers/people_controller.rb
```diff
@@ -217,11 +218,36 @@ def update
   end
 
   def destroy
-    authorize! @person
-    @person.destroy
+    if ActiveModel::Type::Boolean.new.cast(params[:admin])
```
maebeale: 🤖 _From Claude:_ `?admin=true` selects the full-purge path (person + user + all associated data) and authorizes against `full_destroy?`; the UI gates it behind a browser confirmation.

app/models/user.rb
maebeale, line 122: 🤖 _From Claude:_ Discarded (archived) users are blocked from authenticating here alongside the existing inactive check, so archiving immediately revokes login.

app/services/person_archival_service.rb
```diff
@@ -0,0 +1,22 @@
+class PersonArchivalService
+  def initialize(person)
+    @person = person
+  end
+
+  # Soft-delete (archive) the person and their user together. A discarded user
+  # is blocked from authenticating and hidden from default listings.
+  def archive!
+    ActiveRecord::Base.transaction do
+      @person.user&.discard!
```
maebeale, line 10: 🤖 _From Claude:_ Person and user are discarded together in a transaction so they never drift out of sync; `&.` covers people with no user account.

app/policies/person_policy.rb
```diff
@@ -38,7 +43,7 @@ def search?
 
   relation_scope do |relation|
     next relation if admin?
-    relation.searchable.with_active_affiliations.where_user_not_locked
+    relation.kept.searchable.with_active_affiliations.where_user_not_locked
```
maebeale, line 46: 🤖 _From Claude:_ `.kept` drops archived people out of the non-admin scope so they disappear from public/searchable listings while staying restorable by admins.

PR #1687, opened 2026-06-15
HOLD: Distinguish new-tab links from same-tab jumps

maebeale:

- The box-arrow icon (`fa-arrow-up-right-from-square`) was used for **two different behaviors**: links that open in a new tab *and* links that just navigate in place (Turbo `_top` break-outs, drill-ins). Users couldn't tell what a link would do before clicking.
- New-tab links also lacked the screen-reader warning [WCAG G201](https://www.w3.org/WAI/WCAG21/Techniques/general/G201) expects, and the icon's size was set ad-hoc per call site (`text-[0.55rem]`, `text-[10px]`, `w-3 h-3`, …).
- Establishes one clear standard so the affordance always matches the behavior, grounded in [NN/G](https://www.nngroup.com/articles/new-browser-windows-and-tabs/) and current accessibility guidance.

- **New `new_tab_link` helper** (`app/helpers/application_helper.rb`) is now the *only* origin of the box-arrow icon. It sets `target="_blank"` + `rel="noopener"` and appends screen-reader-only `(opens in new tab)` text. Named for the behavior (opens a new tab), not destination — these are internal links, not external sites.
- **Migrated every genuine new-tab link** (payments, forms, event registrations, scholarships, events) to the helper — adding the missing a11y text and collapsing the size drift into one default.
- **Same-tab navigation** now uses a directional arrow (`fa-arrow-right`) or no icon; the box-arrow no longer leaks onto in-place links. Also fixes the lone `fa-external-link-alt` outlier in `taggings/index`.
- **Documented the standard** in `.impeccable.md` so future design work (and the impeccable skills) follow it.

### How was it tested?

- New helper specs cover target/rel merging, sr-only text, block form, and `icon_class` sizing.
- Request specs that render the migrated views pass (payments, event_registrations, scholarships, grants, allocations, taggings, events — 350+ examples).
- Rubocop clean.

### Anything else to add?

Same-tab links were kept same-tab (the principled fix: prefer same-tab, just correct the icon). If any of those should actually open a new tab instead, that's a per-link product call — happy to flip specific ones.


app/helpers/application_helper.rb
```diff
@@ -260,6 +260,41 @@ def index_button(collection, params: {}, path: nil, label: nil, icon: nil, hide_
...
+  # "opens in a new tab" affordance appended after its content: the box-arrow
+  # icon (decorative) plus screen-reader-only warning text (WCAG G201), and sets
+  # target/rel for you. This is the ONLY place the box-arrow icon should
+  # originate — never reuse `fa-arrow-up-right-from-square` for same-tab or
+  # in-page navigation, where it misleads; use a directional arrow there instead.
+  # Pass `icon_class:` to size/color the trailing icon to the surrounding text.
+  #
+  #   <%= new_tab_link "View receipt", receipt_url, class: "underline" %>
+  #   <%= new_tab_link grants_path, class: "..." do %>View all grants<% end %>
+  def new_tab_link(name = nil, options = nil, html_options = nil, &block)
```
maebeale, line 315: 🤖 _From Claude:_ This helper is intentionally the **only** place `fa-arrow-up-right-from-square` is emitted — that keeps the box-arrow meaning exactly "opens in a new tab" and prevents it from drifting back onto same-tab links. `rel` merges with any caller value (so `rel: "noreferrer"` becomes `noreferrer noopener`).

app/views/allocations/_allocated_to.html.erb
```diff
@@ -7,7 +7,7 @@
     <%= link_to edit_event_registration_path(allocatable),
                 class: "relative block rounded-lg border border-gray-300 bg-white px-3 py-2 pr-8 transition-colors hover:border-gray-400 hover:bg-gray-50",
                 data: { turbo_frame: "_top" } do %>
-      <i class="fa-solid fa-arrow-up-right-from-square absolute right-2 top-2 text-[0.6rem] text-gray-400"></i>
+      <i class="fa-solid fa-arrow-right absolute right-2 top-2 text-[0.6rem] text-gray-400"></i>
```
maebeale, line 10: 🤖 _From Claude:_ This link breaks out of the Turbo frame in the **same tab** (`turbo_frame: "_top"`), so it now uses a directional arrow instead of the box-arrow, which had implied a new window.

PR #1688, opened 2026-06-15
HOLD: Add page-header design reference (styleguide)

maebeale:

- Page headings across the app have drifted into ~5 inconsistent variants (h1 vs h2, semibold/bold, text-2xl/3xl, centered vs left, ad-hoc back-links and breadcrumb rows) spread over 500+ templates.
- Before building shared header partials and sweeping the site, we need to **agree on the target look-and-feel**. This PR adds an internal design-reference page so those decisions can be made visually, against real design tokens, rather than in the abstract.

- Added `GET /styleguide/headers` (`StyleguideController#headers`) — logged-in only, no model data, authorization skipped.
- Built `app/views/styleguide/headers.html.erb` plus two demo partials (`_demo`, `_utility_row`) that render every candidate treatment with sample content using the real `DomainTheme` color tokens.
- The page is organized as a decision surface:
  - **Anatomy + rules** — one shared utility row (back-link left → defaults to `admin_path` on admin pages; subnav right), actions in the title row.
  - **Families** — list (A), topper (B), centered hero (C), and the proposed **charcoal base + thin model-color stripe** (J) that unifies admin chrome while marking individual models.
  - **Questions** — alignment/centered-vs-left (E), event registration admin-vs-public (F), band-everywhere combo (G), more subnavs per-record & per-area (H), and rich gradients for public contribution forms (I).
- Organizing principle surfaced by the exercise: **rich = public-facing, quiet = admin** (gradients/centered heroes signal outward pages; charcoal/flat signal internal tools).

### Anything else to add?

- This is a **reference/spec only** — no existing pages are changed yet. Once the variants are chosen, the follow-up builds `shared/_page_header` + `shared/_subnav` and converts pages in batches.
- Per-model stripe/gradient colors are dynamic, so they'll need adding to the Tailwind safelist alongside the existing `DomainTheme` colors when the real partials land.
- Open decisions captured on the page: background alignment, Style A for the admin registration form, charcoal+stripe vs light band, subnav standard (D1), subnav expansion, and gradient approach for public forms.

### DIFF
--- a/app/controllers/styleguide_controller.rb
+++ b/app/controllers/styleguide_controller.rb
@@ -1,0 +1,8 @@
+class StyleguideController < ApplicationController
+  # Internal design reference for comparing page-header treatments.
+  # Logged-in only; no model data required, so authorization is skipped.
+  skip_verify_authorized
+
+  def headers
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+  # Internal design reference for comparing page-header treatments
+  get "styleguide/headers", to: "styleguide#headers"
+
   # temporary direct routes to images for migration audit
   resources :attachments, only: [ :show ]
   resources :media_files, only: [ :show ]
--- a/app/views/styleguide/_demo.html.erb
+++ b/app/views/styleguide/_demo.html.erb
@@ -1,0 +1,17 @@
+<%# locals: id, name, usage, recommended: false %>
+<% recommended = local_assigns.fetch(:recommended, false) %>
+<div class="rounded-xl border <%= recommended ? "border-green-300 ring-1 ring-green-200" : "border-gray-200" %> overflow-hidden">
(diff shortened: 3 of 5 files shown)


app/controllers/styleguide_controller.rb
```diff
@@ -0,0 +1,8 @@
+class StyleguideController < ApplicationController
+  # Internal design reference for comparing page-header treatments.
+  # Logged-in only; no model data required, so authorization is skipped.
+  skip_verify_authorized
```
maebeale, line 4: 🤖 _From Claude:_ Auth is skipped (not `authorize!`) because this is a static, data-free internal reference with no records to authorize against. Still behind `authenticate_user!`, so logged-in only.

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+  # Internal design reference for comparing page-header treatments
+  get "styleguide/headers", to: "styleguide#headers"
```
maebeale, line 3: 🤖 _From Claude:_ Reference/spec route only — meant to be removed (or moved behind an admin constraint) once the header system is chosen and the real shared partials land.

PR #1689, opened 2026-06-15
Show scholarship form's headers and name in registration section

maebeale:

- The public registration **scholarship section** ignored the scholarship form's own data:
  - The section **title was hard-coded** ("Scholarship application") instead of using the form's name.
  - Every **group header in the scholarship form was skipped** (`next if field.group_header?`), so admin-defined section headers never appeared.
- This made the scholarship section inconsistent with the rest of the form and unable to reflect how an admin configured the scholarship form.

- **Title is now the form's name** — the heart icon is shown beside `@scholarship_form.display_name` rather than a hard-coded string.
- **Section headers render from the form** — group headers in the scholarship form now display using the shared section-header partial (rainbow bar, no heart).
- **No duplicate title** — the default standalone scholarship form (built by `FormBuilderService`) has both a form name *and* a leading "Scholarship Application" group header. A leading header whose text matches the form name is skipped so the title isn't shown twice; genuine sub-section headers still render.
- **Extracted `_group_header.html.erb`** — the section-header markup is now a partial reused by both the main registration loop and the scholarship loop (DRY).
- The heart icon appears **only** on the section title, not on group headers.

### Anything else to add?

- Heads-up: the shared scholarship form's name is currently "Scholarship Application" (title case). The title now renders the form's name verbatim, so to match sentence-case UI conventions it can be renamed under `/forms` (data change, not code).
- Tests in `spec/system/public_registration_new_spec.rb` cover: title shows the form name + heart, the form's intro header renders, the form's section headers render, and a header matching the form name is not repeated.

### DIFF
--- a/app/views/events/public_registrations/new.html.erb
+++ b/app/views/events/public_registrations/new.html.erb
@@ -118,14 +118,7 @@
         <div class="grid grid-cols-1 md:grid-cols-12 gap-x-4">
           <% @form_fields.each do |field| %>
             <% if field.group_header? %>
-              <div class="md:col-span-12 pt-7 pb-4">
-                <h3 class="rich-label flex items-center gap-2.5 text-lg font-bold text-purple-900">
-                  <span class="<%= form_section_bar_class %>"></span><%= form_label_html(field.name) %>
-                </h3>
-                <% if field.hint_text.present? %>
-                  <p class="mt-1.5 text-sm text-gray-500"><%= field.hint_text %></p>
-                <% end %>
-              </div>
+              <%= render "events/public_registrations/group_header", field: field %>
             <% else %>
               <% submitted_value = params.dig(:public_registration, :form_fields, field.id.to_s) %>
               <% if submitted_value.blank? && field.field_identifier == "payment_method" %>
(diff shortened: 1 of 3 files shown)


app/views/events/public_registrations/new.html.erb
```diff
@@ -141,16 +134,21 @@
...
             <div class="grid grid-cols-1 md:grid-cols-12 gap-x-4">
               <% @scholarship_form.form_fields.reorder(position: :asc).each do |field| %>
-                <% next if field.group_header? %>
-                <% submitted_value = params.dig(:public_registration, :form_fields, field.id.to_s) %>
-                <div class="<%= field.grid_span_class %>">
-                  <%= render "events/public_registrations/form_field", field: field, value: submitted_value %>
-                </div>
+                <% if field.group_header? %>
+                  <%# Skip a section header that just repeats the form name shown as the title above %>
+                  <% next if field.name.to_s.strip.casecmp?(@scholarship_form.display_name.to_s.strip) %>
```
maebeale, line 144: 🤖 _From Claude:_ The default standalone scholarship form has both a `name` and a leading "Scholarship Application" group header (both created by `FormBuilderService`). Since the form name is now the section title, this skips a leading header that just repeats it — avoiding a duplicate heading while still rendering any genuine sub-section headers.

PR #1690, opened 2026-06-15
Flag scholarship_requested when awarding an event scholarship

maebeale:
- When a scholarship is awarded against an event registration, the registration's `scholarship_requested` flag should be set to `true` if it wasn't already.
- This keeps the registrant roster and the scholarship-status filters consistent: an awarded scholarship implies one was requested, even when the request flag wasn't previously checked.

- Added an `after_create_commit` callback on `Scholarship` that flips the connected `EventRegistration`'s `scholarship_requested` to `true` when it's currently `false`.
- Guarded the callback so it only runs when the scholarship's allocation points at an `EventRegistration` (grant-only scholarships are unaffected).
- Deliberately did **not** touch the flag on destroy — removing an awarded scholarship doesn't undo the fact that one was requested, so the flag stays on.

### Anything else to add?
- TDD: added model specs covering the set-on-create, leave-on-destroy, and not-connected cases (failing first, then green).

### DIFF
--- a/app/models/scholarship.rb
+++ b/app/models/scholarship.rb
@@ -13,6 +13,7 @@ class Scholarship < ApplicationRecord
   validate :within_grant_budget, if: :grant
 
   after_update :sync_allocation_amount, if: -> { saved_change_to_amount_cents? }
+  after_create_commit :flag_event_registration_scholarship_requested
 
   scope :completed, -> { where(tasks_completed: true) }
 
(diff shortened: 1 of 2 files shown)


app/models/scholarship.rb
```diff
@@ -48,4 +49,15 @@ def sync_allocation_amount
...
+  # When a scholarship is awarded against an event registration, the registration
+  # should reflect that a scholarship was requested. We never clear this on delete:
+  # removing an awarded scholarship doesn't undo the fact that one was requested.
+  def flag_event_registration_scholarship_requested
+    registration = allocation&.allocatable
+    return unless registration.is_a?(EventRegistration)
+    return if registration.scholarship_requested?
+
+    registration.update!(scholarship_requested: true)
+  end
```
maebeale, line 62: 🤖 _From Claude:_ Intentionally one-directional: we set the flag on create but never clear it on destroy, since removing an awarded scholarship doesn't undo that one was requested. The `is_a?(EventRegistration)` guard keeps grant-only scholarships (no allocation) from triggering this.

PR #1692, opened 2026-06-16
Gate videoconference link to a 30-min window around the event

maebeale:
- The Zoom/videoconference link was shown to any paid registrant at all times — days before the event and indefinitely after — even though the link is only useful right around the event.
- This restricts the link to a join window so attendees see it when they need it and not before/after.

- Added `Event#videoconference_window_open?` plus a `VIDEOCONFERENCE_JOIN_BUFFER = 30.minutes` constant: the window runs from 30 min before `start_date` through 30 min after `end_date`.
- Wired it into `EventRegistration#joinable?` (`active? && paid? && event.videoconference_window_open?`), so every view that gates on `joinable?` (event show page and registration ticket) hides the link outside the window. Payment-in-full gating is unchanged.

### UI Testing Checklist
- [ ] As a paid registrant, the join link is hidden more than 30 minutes before start.
- [ ] The join link appears from 30 minutes before start through 30 minutes after end.
- [ ] The join link is hidden more than 30 minutes after end.
- [ ] Unpaid / partially-paid / cancelled registrants still never see the link.

### Anything else to add?
- Covered by model specs (window boundaries + `joinable?`) and a system spec asserting a fully-paid registrant is blocked until the window opens. Followed red→green TDD.

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -210,7 +210,7 @@ class EventRegistration < ApplicationRecord
   end
 
   def joinable?
-    active? && paid?
+    active? && paid? && event.videoconference_window_open?
   end
 
   def attendance_status_label
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -3,6 +3,10 @@ class Event < ApplicationRecord
   include ActionText::Attachable
   include RemoteSearchable
 
+  # Grace window on either side of the event during which the videoconference
+  # link is available to paid registrants.
+  VIDEOCONFERENCE_JOIN_BUFFER = 30.minutes
+
   has_rich_text :rhino_header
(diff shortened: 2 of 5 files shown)


app/models/event_registration.rb
```diff
@@ -210,7 +210,7 @@ def partially_paid?
   end
 
   def joinable?
-    active? && paid?
+    active? && paid? && event.videoconference_window_open?
```
maebeale, line 213: 🤖 _From Claude:_ Note this makes link visibility time-dependent — `joinable?` flips with the clock, no DB change. The 30-min buffer lives in `Event::VIDEOCONFERENCE_JOIN_BUFFER`.

PR #1693, opened 2026-06-16
Hide "Other" option on dropdown form fields

maebeale:
- Dropdown form fields can't capture free text, so a selected "Other" would submit a bare `"Other"` with no way to clarify it — unlike radio/checkbox fields, which reveal a "please specify" input.
- This hides "Other" from dropdowns on the public registration form and warns form authors so they can switch to radio/checkboxes if they need an "Other" choice.

- In the public form field renderer, the `single_select_dropdown` branch now rejects any "Other" option (detected by label via the existing `other_option?` helper); other field types are untouched.
- Added a shared `_dropdown_other_warning` partial, rendered server-side (no Stimulus, shown on page load) in both the form editor and the form preview whenever a dropdown carries an "Other" option.
- The warning uses a `dropdown_hides_other?` helper that checks the field's *effective* options — dynamic sources (sectors/categories) as well as author-managed stored options — so dynamic dropdowns whose source list includes an "Other" entry warn too.

### UI Testing Checklist
- [ ] On the public registration form, a dropdown field with an "Other" option no longer lists "Other"; its other options still appear.
- [ ] In the form editor and preview, a dropdown field with an "Other" option shows the amber warning.
- [ ] A dynamic dropdown (e.g. service area / sector or a category-backed field) whose source list contains an "Other" entry also shows the warning.
- [ ] No warning shows for dropdowns without an "Other" option, or for radio/checkbox fields (which still reveal the "please specify" input).

### Anything else to add?
- The warning reflects saved state on page load only; switching a field to "dropdown" in the editor surfaces it after a reload (matches the no-Stimulus requirement).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -182,6 +182,19 @@ module ApplicationHelper
     end
   end
 
+  # True when a dropdown field carries an "Other" option that the public form
+  # strips (dropdowns have no free-text input). Checks the field's effective
+  # options — dynamic sources (sectors/categories) as well as author-managed
+  # stored options — so the editor and preview warn wherever "Other" would
+  # silently disappear.
+  def dropdown_hides_other?(field)
+    return false unless field.answer_type == "single_select_dropdown"
+
+    options = dynamic_form_field_options(field) ||
+      field.form_field_answer_options.includes(:answer_option).map { |ffao| [ ffao.answer_option.name ] }
+    options.any? { |label, _| other_option?(label) }
+  end
+
   # Describes where a dynamic-option field's choices come from, for the form
   # editor badge: a sentence-case label and a link to the filtered admin list
   # that manages those options. Returns nil for fields with stored options.
--- a/app/views/events/public_registrations/_form_field.html.erb
+++ b/app/views/events/public_registrations/_form_field.html.erb
@@ -105,8 +105,10 @@
   <% when "single_select_dropdown" %>
     <%# Dynamic fields (e.g. primary_service_area_single) source options from
         Sector/Category data; everything else uses the field's own stored options. %>
-    <% dropdown_options = dynamic_form_field_options(field) ||
-         field.form_field_answer_options.includes(:answer_option).map { |ffao| [ ffao.answer_option.name, ffao.answer_option.name ] } %>
+    <%# Dropdowns have no free-text input, so the "Other" option is omitted here. %>
+    <% dropdown_options = (dynamic_form_field_options(field) ||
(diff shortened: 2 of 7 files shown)


app/views/events/public_registrations/_form_field.html.erb
```diff
@@ -105,8 +105,10 @@
   <% when "single_select_dropdown" %>
     <%# Dynamic fields (e.g. primary_service_area_single) source options from
         Sector/Category data; everything else uses the field's own stored options. %>
-    <% dropdown_options = dynamic_form_field_options(field) ||
-         field.form_field_answer_options.includes(:answer_option).map { |ffao| [ ffao.answer_option.name, ffao.answer_option.name ] } %>
+    <%# Dropdowns have no free-text input, so the "Other" option is omitted here. %>
+    <% dropdown_options = (dynamic_form_field_options(field) ||
+         field.form_field_answer_options.includes(:answer_option).map { |ffao| [ ffao.answer_option.name, ffao.answer_option.name ] })
+         .reject { |option_label, _| other_option?(option_label) } %>
```
maebeale, line 111: 🤖 _From Claude:_ Detection is by option **label** (`other_option?`), not value — this matters for category-backed dynamic fields where the value is a numeric id, so a value-based filter would miss "Other".

app/views/forms/_dropdown_other_warning.html.erb
```diff
@@ -0,0 +1,8 @@
+<%# Dropdowns can't collect free text, so the public form strips the "Other" option.
+    Warn the author whenever a dropdown still carries an "Other" option. %>
+<% if field.answer_type == "single_select_dropdown" && field.other_option? %>
```
maebeale: 🤖 _From Claude:_ Server-rendered on page load (no Stimulus, per request), so the warning reflects saved state — switching a field to dropdown surfaces it after a reload. `field.other_option?` already returns false for dynamic-option fields, so those never trigger it.

app/helpers/application_helper.rb
```diff
@@ -182,6 +182,19 @@ def dynamic_form_field_options(field)
     end
   end
 
+  # True when a dropdown field carries an "Other" option that the public form
+  # strips (dropdowns have no free-text input). Checks the field's effective
+  # options — dynamic sources (sectors/categories) as well as author-managed
+  # stored options — so the editor and preview warn wherever "Other" would
+  # silently disappear.
+  def dropdown_hides_other?(field)
```
maebeale, line 190: 🤖 _From Claude:_ Checks **effective** options (dynamic sources + stored) rather than `FormField#other_option?`, which short-circuits to false for dynamic fields — that gap is exactly why a sector/category-backed dropdown with an "Other" entry warns now.

PR #1694, opened 2026-06-16
Prevent display-only form fields from being marked required

maebeale:
- Informational-only and section-header form fields collect no answer, so marking them "required" is meaningless — it produced a stray asterisk and, for informational-only, an unsatisfiable presence check that could block submissions.
- This makes "required" unavailable for those display-only types across the model, the form builder, and the public/report views.

- Added a `FormField` validation (`non_input_types_cannot_be_required`) plus a `collects_input?` helper and `NON_INPUT_ANSWER_TYPES` constant as the server-side backstop.
- In the form builder, section headers get a hidden `required: false`, while the shared field row hides + unchecks the Required toggle via the `answer-options` Stimulus controller when the type is display-only (the type can be switched live there, so a static hidden field would be wrong).
- Stopped rendering the required marker for display-only fields on the public form and report views, and updated the factory/specs so display-only fields default to not-required.

### UI Testing Checklist
- [ ] In the form builder, select "Informational-only" or "Section header" as a field type and confirm the Required checkbox disappears.
- [ ] Switch the field back to an input type and confirm the Required checkbox reappears.
- [ ] Confirm no asterisk renders next to display-only fields on the public form and report views.

### Anything else to add?
The model validation guarantees correctness even with JavaScript disabled; the Stimulus controller and hidden field are UX/data conveniences layered on top.

### DIFF
--- a/app/frontend/javascript/controllers/answer_options_controller.js
+++ b/app/frontend/javascript/controllers/answer_options_controller.js
@@ -7,8 +7,13 @@ import { Controller } from "@hotwired/stimulus"
 // "Single select radio" would save with no options and render nothing on the
 // public form. The chevron toggle expands/collapses the option list itself.
 export default class extends Controller {
-  static targets = ["type", "section", "list", "icon"]
+  static targets = ["type", "section", "list", "icon", "required", "requiredLabel"]
   static values = { expanded: Boolean }
+
+  // Display-only answer types (informational text, section headers) never
+  // collect a response, so they can't be required — mirrors FormField's
+  // NON_INPUT_ANSWER_TYPES on the server.
+  nonInputTypes = ["no_user_input", "group_header"]
 
   connect() {
     this.render()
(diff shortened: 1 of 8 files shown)


app/views/forms/_form_field_fields.html.erb
```diff
@@ -33,6 +33,8 @@
   <%= f.hidden_field :position, value: field.position %>
 
   <% if field.group_header? %>
+    <%# Section headers never collect an answer, so they're never required. %>
+    <%= f.hidden_field :required, value: false %>
```
maebeale, line 37: 🤖 _From Claude:_ Section headers render in this fixed branch with no Required toggle, so a static hidden `false` is safe. The shared field row (info-only) intentionally does NOT use this — its type can switch live, so a static value would clobber the real checkbox.

app/frontend/javascript/controllers/answer_options_controller.js
```diff
@@ -32,11 +37,27 @@ export default class extends Controller {
...
+  }
+
+  // Display-only fields (informational text, section headers) can't be
+  // required, so hide the Required box entirely — like section headers, but
+  // without their heading font. Unchecking leaves the check_box helper's
+  // hidden "0" field to record the field as not required.
+  renderRequired() {
+    if (this.hasRequiredTarget && this.nonInput) this.requiredTarget.checked = false
+    if (this.hasRequiredLabelTarget) {
+      this.requiredLabelTarget.classList.toggle("hidden", this.nonInput)
```
maebeale, line 60: 🤖 _From Claude:_ Hides the Required box (rather than disabling) for display-only types, matching section headers but without the heading font. Unchecking lets the `check_box` hidden `"0"` record it as not-required.

spec/factories/form_fields.rb
```diff
@@ -9,6 +9,8 @@
     input_type { :text_alphanumeric } # Default datatype
     sequence(:position) { |n| n }
     parent_id { nil }
+    # Display-only types (informational text, section headers) can't be required.
+    required { FormField::NON_INPUT_ANSWER_TYPES.exclude?(answer_type.to_s) }
```
maebeale, line 13: 🤖 _From Claude:_ A required section header was an invalid fixture under the new validation; this derives `required` from the answer type so default factory builds stay valid.

PR #1695, opened 2026-06-16
Surface event "Before you attend" details on the registration ticket

maebeale:

- Registrants routinely miss the materials, supplies, and policy info that used to live in a long confirmation email.
- This gives events an editable "details" field shown on its own ticket-linked page behind a prominent call-out, so the info is unmissable but the ticket essentials aren't buried.

- Added an `event_details` field (basic-HTML textarea, edited and sanitized the same way as the existing form-header intro) plus an editable heading (`event_details_label`, default "Before you attend") on events.
- Surfaced it on the registrant ticket as a prominent amber call-out linking to a new public details page (`details_event_path`); both are hidden when no content is set.
- Renamed the registrant-facing "Registration details" to "Your form responses" so it's clear that page shows their submitted form, and seeded the flagship training with the real art-supply copy.

### UI Testing Checklist
- [ ] As an admin, add "Details content" (and optionally change the label) on an event and save.
- [ ] On the registration ticket, confirm the amber "Before you attend" call-out appears and links to the details page, which renders the HTML.
- [ ] Clear the details content and confirm the call-out and page are hidden (the page redirects to the event).
- [ ] Confirm the ticket link now reads "View your form responses" and opens the form-responses page titled "Your form responses".

### Anything else to add?
- Email injection was intentionally left out of scope; the details currently live on the ticket + page only. Covered by new request and system specs.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -96,7 +96,7 @@ class EventsController < ApplicationController
   def details
     authorize! @event, to: :details?
 
-    if @event.rhino_event_details.blank?
+    if @event.event_details.blank?
       redirect_to event_path(@event, reg: params[:reg].presence)
       return
     end
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,8 +1,8 @@
 class EventsController < ApplicationController
   include AhoyTracking, TagAssignable
-  skip_before_action :authenticate_user!, only: [ :index, :show, :staff ]
+  skip_before_action :authenticate_user!, only: [ :index, :show, :staff, :details ]
   skip_before_action :verify_authenticity_token, only: [ :preview ]
-  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants staff edit_staff update_staff recipients bulk_payments preview_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
+  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants details staff edit_staff update_staff recipients bulk_payments preview_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
 
   def index
     authorize!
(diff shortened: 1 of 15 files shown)


app/views/events/details.html.erb
```diff
@@ -0,0 +1,33 @@
...
+          <h1 class="text-xl font-semibold tracking-wide leading-tight"><%= @event.event_details_label %></h1>
+          <p class="text-sm text-blue-200"><%= @event.title %></p>
+        </div>
+      </div>
+      <div class="absolute inset-x-0 bottom-0 h-1 bg-gradient-to-r from-accent via-amber-400 to-accent"></div>
+    </div>
+
+    <div class="px-6 sm:px-8 py-7">
+      <div class="rich-label prose max-w-none text-gray-700 leading-relaxed prose-headings:text-gray-800 prose-a:text-blue-700">
+        <%= form_label_html(@event.event_details) %>
```
maebeale, line 29: 🤖 _From Claude:_ Reusing `form_label_html` so the details HTML goes through the same sanitizing allowlist as form headers/labels — admin-authored markup can render but can't inject XSS.

app/models/event.rb
```diff
@@ -139,6 +139,12 @@ def name
     title
   end
 
+  # Heading shown on the ticket call-out and the details page. Falls back to the
+  # default even when an admin clears it, so the section never renders unlabelled.
+  def event_details_label
+    super.presence || "Before you attend"
```
maebeale, line 155: 🤖 _From Claude:_ Falls back to the default even when an admin clears the label, so the call-out/page heading is never empty while details are present.

app/controllers/events_controller.rb
```diff
@@ -90,6 +90,20 @@ def registrants
     end
   end
 
+  # Public "Before you attend" page (materials, supplies, policies). Linked from
+  # the registration ticket. When no details are set there is nothing to show, so
+  # fall back to the event page.
+  def details
+    authorize! @event, to: :details?
+
+    if @event.event_details.blank?
```
maebeale, line 99: 🤖 _From Claude:_ When no details are set there's nothing to show, so we redirect to the event rather than rendering an empty page — mirrors the "hidden when blank" behavior of the ticket call-out.

PR #1696, opened 2026-06-16
Add per-option type-in boxes to registration form options

maebeale:
On the public registration form, only the "Other" option revealed a free-text "please specify" box, but three "how did you hear about us" sources also need to capture a specific detail. This lets "Work(ed) at an agency…", "Word of Mouth", and "Foundation/Funder" each reveal their own type-in box (e.g. "Please list the name of the person") so we collect that context at sign-up.

Rather than special-casing each option, I generalized the existing single-purpose "Other" reveal into a reusable specify-option mechanism keyed off the option label, where each label maps to its own placeholder (config lives on `FormField`); the answer is still stored inline as `"<label>: <typed text>"` so the server needs no new params or columns. The Stimulus controller (renamed `other-option` → `specify-option`) now handles multiple control/text pairs per field so one radio/checkbox group can offer several type-in boxes, and validation accepts any offered specify option's free-text form.

### UI Testing Checklist
- [ ] On a registration form with these options, selecting "Work(ed) at an agency…", "Word of Mouth", or "Foundation/Funder" reveals a type-in box with the correct placeholder
- [ ] Typing in a box and submitting stores the value as `"<option>: <text>"`; re-rendering after a validation error re-checks the option and prefills the text
- [ ] Selecting a non-specify option (or a different radio) hides the boxes; "Other" still works as before
- [ ] Multi-select checkbox fields can show several boxes at once independently

### Anything else to add?
Matching is by option label (case/whitespace-insensitive), so the placeholders only appear when the stored `AnswerOption` names match the constants in `FormField::SPECIFY_OPTION_PLACEHOLDERS` exactly — notably the long "Work(ed) at an agency that has/had an AWBW program" label.


app/models/form_field.rb
```diff
@@ -25,10 +25,22 @@ class FormField < ApplicationRecord
...
+  # The generic free-text option label that lets a respondent supply their own
   # value; a chosen "Other" answer is stored as "Other" or "Other: <text>".
   OTHER_OPTION_PREFIX = "Other"
 
+  # Selectable option labels that reveal a free-text "please specify" box when
+  # chosen, each mapped to that box's placeholder. A chosen answer is stored as
+  # "<label>: <typed text>" (or the bare label when nothing is typed), so the
+  # server needs no extra params. "Other" is the generic catch-all; the rest
+  # capture a specific named source for "how did you hear about us" questions.
+  SPECIFY_OPTION_PLACEHOLDERS = {
```
maebeale, line 46: 🤖 _From Claude:_ This config is the single source of truth for both rendering (helper) and validation, keyed by option **label** — same approach as the prior "Other" reveal. Tradeoff: placeholders only appear if stored `AnswerOption` names match these strings exactly (case/whitespace-insensitive), so editing a label in the form builder requires updating the constant.

app/frontend/javascript/controllers/specify_option_controller.js
```diff
@@ -0,0 +1,46 @@
...
+      control.dataset.specifyBase = control.value
+    })
+    this.update()
+  }
+
+  // Fires on any change within the field — reveal each selected option's text
+  // input (folding its value), hide the rest.
+  update() {
+    this.controlTargets.forEach((control, i) => {
+      const text = this.textTargets[i]
```
maebeale, line 26: 🤖 _From Claude:_ Controls and text inputs are paired by DOM order (nth control ↔ nth text). This holds because every specify option renders exactly one control then one text, and non-specify options render neither — so the two target arrays stay index-aligned.

app/frontend/javascript/controllers/specify_option_controller.js
```diff
@@ -0,0 +1,46 @@
...
+// nth control is paired with the nth text input by their order in the DOM. The
+// text inputs have no name attribute, so they never submit on their own.
+export default class extends Controller {
+  static targets = ["control", "text"]
+
+  connect() {
+    // Capture each control's base value (its option label) before folding
+    // rewrites it, so we can rebuild "<option>: <text>" on every change.
+    this.controlTargets.forEach((control) => {
+      control.dataset.specifyBase = control.value
```
maebeale, line 17: 🤖 _From Claude:_ We capture the base value on connect (before any fold rewrites it to `"<label>: <text>"`) because the server always re-renders the control with its bare label as the value, so connect is the one moment it is guaranteed clean.

app/views/events/public_registrations/_form_field.html.erb
```diff
@@ -72,31 +72,32 @@
...
       <% radio_options.each do |option_label, option_value, option_description| %>
-        <% is_other = other_option?(option_label) %>
+        <% placeholder = specify_placeholder(option_label) %>
+        <% is_specify = placeholder.present? %>
         <label class="flex flex-col gap-1 rounded-lg border border-gray-300 bg-white px-3.5 py-2 text-sm text-gray-700 cursor-pointer transition hover:border-blue-400 hover:bg-blue-50 has-[:checked]:border-blue-500 has-[:checked]:bg-blue-50 has-[:checked]:text-blue-900 has-[:checked]:font-medium">
           <span class="flex items-center gap-2">
             <input type="radio"
                    name="<%= field_name %>"
-                   value="<%= is_other ? "Other" : option_value %>"
+                   value="<%= is_specify ? option_label : option_value %>"
```
maebeale, line 84: 🤖 _From Claude:_ Specify options submit their **label**, not `option_value`. This matters for dynamic (Sector/Category-backed) fields whose `option_value` is a record id — the stored answer must be the literal label (e.g. `"Other"`), so it is detected and folded correctly.

PR #1697, opened 2026-06-16
Hide "Mixed-age groups" from the registration age-group field

maebeale:
- The registration form's "Primary Age Group(s) Served" question asks respondents to name the concrete age groups they serve, so the catch-all "Mixed-age groups" AgeRange category shouldn't be selectable there.
- It still needs to stay available everywhere else age ranges are tagged (workshops, windows types, reporting), so it can't simply be unpublished.

- Added a `Category.excluding_mixed_age` scope and a `FormField#dynamic_categories` method, mirroring the existing `Sector.excluding_other` / `service_area_sectors` pattern that omits "Other" from the primary service-area field.
- Routed both the public form rendering (and editor preview) and submission validation through this one source of truth, so a forged "Mixed-age groups" answer is rejected just as it's hidden from the UI.

### UI Testing Checklist
- [ ] On the registration form, the "Primary Age Group(s) Served" field no longer lists "Mixed-age groups".
- [ ] "Mixed-age groups" still appears wherever age ranges are tagged elsewhere (workshops, windows types).

### Anything else to add?
- This is a hardcoded exclusion (matching the "Other" sector precedent) rather than a per-category admin toggle.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -34,6 +34,11 @@ class FormField < ApplicationRecord
     "primary_age_group" => "AgeRange"
   }.freeze
 
+  # The "primary age group(s) served" field. Backed by AgeRange categories but
+  # omits the catch-all "Mixed-age groups" category — a respondent names the
+  # concrete age groups they serve, not the mixed bucket.
+  PRIMARY_AGE_GROUP_FIELD_IDENTIFIER = "primary_age_group"
+
   # The generic free-text option label that lets a respondent supply their own
   # value; a chosen "Other" answer is stored as "Other" or "Other: <text>".
   OTHER_OPTION_PREFIX = "Other"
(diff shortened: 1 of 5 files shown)


app/models/form_field.rb
```diff
@@ -276,6 +280,19 @@ def service_area_sectors
     field_identifier == PRIMARY_SERVICE_AREA_FIELD_IDENTIFIER ? scope.excluding_other : scope
   end
 
+  # The published Category records a category-backed dynamic field offers, in
+  # position/name order. The "primary age group" field omits the catch-all
+  # "Mixed-age groups" AgeRange category — a respondent names the concrete age
+  # groups they serve. Empty when the backing CategoryType is missing. Source of
+  # truth shared by the public form's rendering and submission validation.
+  def dynamic_categories
```
maebeale, line 288: 🤖 _From Claude:_ Routing both form rendering and `allowed_answer_values` through this one method keeps the rendered options and the submission-validation allowlist from drifting — a value hidden from the UI is also rejected if forged.

maebeale:
@jmilljr24 this isn't a great solution, but we need something for tonight so doing this for now.

jmilljr24:
> @jmilljr24 this isn't a great solution, but we need something for tonight so doing this for now.

I don't have time to review but if you are comfortable, have at it.

You have better organizational skills than I do so if you have a good place to remember clean up actions that would be a great place to add this.

PR #1698, opened 2026-06-16
Show the primary service area consistently with a starred chip

maebeale:
- A person's **primary service area** — the one they singled out on their registration form — was displayed inconsistently and often indistinguishable from their other sectors.
- This gives the primary sector one consistent treatment everywhere it appears: a darker-green chip led by a gold star (matching the chip editor, where star = primary) with the standard hover state.

- **Shared chip** (`sectors/_tagging_label`): added an `is_primary` local (defaults false). When set, the chip renders with a gold star (`fa-star`) + darker-green fill (`bg-lime-200`) and a "Primary service area" tooltip. All other callers are unaffected.
- **Scholarship recipients page**: sectors render primary-first (`sectorable_items_primary_first`), with the primary starred — read-only.
- **Person profile**: collapsed the two-column "Primary sector / Additional sectors" layout into one primary-first "Sectors" list; primary leads as the starred chip, followed by additional sectors and any free-text "Other" responses.
- **People index/search**: shows **only** the primary sector as the starred chip, keeping the table scannable (the full set lives on the profile).
- **Event staff cards**: replaced the plain compact sector badges with the same shared chip — primary first and starred.
- **Seed** (`record_professional_answers`): the first submitted service area becomes each registrant's single primary `SectorableItem` (demoting any other first), so demo data matches what they selected on the form.

### Rebase note
Rebased onto main, which now includes #1700 ("Show 'Other' service-area answers as additional sectors"). That's **not** a duplicate — it tweaked the old two-column profile layout to move the "Other" chip into the Additional column. Our single combined list already achieves that intent (the "Other" free-text chip renders after all sectors, never under/with the primary), so we kept our layout and realigned #1700's request spec to assert the Other chip renders as a free-text `(other)` chip rather than keying off the now-removed "Additional sectors" heading. #1700's seed change (`excluding_other`) is complementary and preserved.

### Out of scope (follow-up PRs)
- Event **Background/Dashboard** primary-sector charts — left as-is per request.
- **Organization** sector views — left as-is per request.

### UI Testing Checklist
- [ ] Recipients page: primary shows first as a darker-green starred chip; others light-lime
- [ ] Person profile: single "Sectors" list, primary first and starred, "Other" responses still shown
- [ ] People index: only the primary sector renders, as the starred chip
- [ ] Event staff cards: primary sector first and starred
- [ ] Other sector listings (workshops, logs, taggings) unchanged

### Anything else to add?
- Branch name says "crown" — the marker is a **star** (star = primary, crown = leader in this codebase); renaming post-creation would close the PR, so it's left as-is.
- Added view specs for the default and primary (starred) chip.

### DIFF
--- a/app/views/sectors/_tagging_label.html.erb
+++ b/app/views/sectors/_tagging_label.html.erb
@@ -4,11 +4,18 @@
 
 <% display_leader ||= false %>
 
-<% bg_color ||= DomainTheme.bg_class_for(:sectors) %>
+<% is_primary ||= false %>
+
+<%# Primary service areas read as a darker-green chip with a star so they stand
+    out from the additional sectors, matching the star used in the chip editor;
+    everyone else keeps the light-lime default. %>
+<% bg_color ||= is_primary ? "bg-lime-200" : DomainTheme.bg_class_for(:sectors) %>
 
 <%# bg_hover_color ||= DomainTheme.bg_class_for(:sectors, hover: true) %>
 
-<% bg_hover_color ||= "bg-lime-200" %>
+<% bg_hover_color ||= is_primary ? "bg-lime-300" : "bg-lime-200" %>
+
+<% text_color ||= is_primary ? "text-lime-800" : "text-gray-500" %>
 
 <% if sector %>
   <%= link_to taggings_path(sector_names_all: sector.name),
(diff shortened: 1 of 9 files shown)


db/seeds/dev/events_management.rb
```diff
@@ -671,6 +671,15 @@
   end
   # Service area chart reads SectorableItem tags, mirroring assign_tags.
   sectors.each { |sector| SectorableItem.find_or_create_by!(sector: sector, sectorable: person) }
+  # Make the first submitted service area the person's single primary sector, so the
+  # recipients page + profile crown a primary that matches what they selected on the
+  # registration form. Demote any other primary first to keep exactly one (a person
+  # registered for several events is enriched once per event). Idempotent.
+  primary_sector = sectors.first
+  if primary_sector
```
maebeale, line 682: 🤖 _From Claude:_ Demoting any other primary before setting this one keeps exactly one primary even when a person is enriched once per event they registered for — otherwise a multi-event registrant could end up with two starred sectors.

app/views/sectors/_tagging_label.html.erb
```diff
@@ -4,11 +4,18 @@
 
 <% display_leader ||= false %>
 
-<% bg_color ||= DomainTheme.bg_class_for(:sectors) %>
+<% is_primary ||= false %>
```
maebeale, line 7: 🤖 _From Claude:_ `is_primary` defaults to false, so the ~10 existing callers of this partial render unchanged; only the recipients page opts in.

maebeale:
Screenshot 2026-06-16 at 5 44 22 PM
Screenshot 2026-06-16 at 5 45 15 PM
Screenshot 2026-06-16 at 5 46 30 PM

PR #1699, opened 2026-06-16
Collapse submitted content section on participant profile by default

maebeale:
- The "Submitted content" card on a participant profile (workshop ideas, variation ideas, story ideas, workshop logs) is only visible to the profile owner/admin and takes up significant vertical space.
- Collapsing it by default keeps the profile compact while the content stays one click away.

- Reused the existing `expandable-card` Stimulus controller (already used in `events/recipients.html.erb`) rather than adding new JS.
- Turned the section header into a full-width toggle button with a `fa-chevron-down` icon that rotates when expanded, and set `expanded-value="false"` so it starts collapsed.
- Marked the content as the controller's `body` target with `hidden` for the initial collapsed state; the controller syncs `hidden`, chevron rotation, and `aria-expanded` on connect and on each toggle.

### Anything else to add?
- No new files or JS — purely a markup change in `app/views/people/show.html.erb`.

### DIFF
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -246,17 +246,25 @@
     <!-- SUBMITTED SECTION -->
     <% if allowed_to?(:show?, @person) %>
       <div class="user-only <%= DomainTheme.bg_class_for(:user_only) %>
-            border <%= DomainTheme.border_class_for(:user_only) %> rounded-lg shadow-sm">
+            border <%= DomainTheme.border_class_for(:user_only) %> rounded-lg shadow-sm"
+           data-controller="expandable-card"
+           data-expandable-card-expanded-value="false">
         <!-- Header -->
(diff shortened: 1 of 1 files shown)


app/views/people/show.html.erb
```diff
@@ -246,17 +246,25 @@
     <!-- SUBMITTED SECTION -->
     <% if allowed_to?(:show?, @person) %>
       <div class="user-only <%= DomainTheme.bg_class_for(:user_only) %>
-            border <%= DomainTheme.border_class_for(:user_only) %> rounded-lg shadow-sm">
+            border <%= DomainTheme.border_class_for(:user_only) %> rounded-lg shadow-sm"
+           data-controller="expandable-card"
+           data-expandable-card-expanded-value="false">
```
maebeale, line 251: 🤖 _From Claude:_ Starts collapsed via `expanded-value="false"` plus `hidden` on the body target, so there's no expand-flash on load — the controller's `expandedValueChanged` also fires on connect to keep state, chevron, and `aria-expanded` in sync.

PR #1701, opened 2026-06-16
Show captured question wording for historical form answers

maebeale:
When a question's name is edited later, historical submission views were retroactively relabeling already-submitted answers, misrepresenting what each respondent was actually asked. This PR makes every submission view render the wording snapshotted at submission time so editing a question no longer rewrites history.

Added a `display_question_label(field, response)` helper that returns `question_name_when_answered` (the wording captured when the answer was saved), falling back to the field's current name only when no snapshot exists (e.g. a question added after this person submitted). Wired it into the public registration show, recipients, form submission, and scholarship submission views so the logic is centralized and consistent.

### UI Testing Checklist
- [ ] On a form submission view, reword a question after a response exists and confirm the original wording still shows next to that answer.
- [ ] Add a new question after a submission and confirm it falls back to the current field name.

### Anything else to add?
Note: answers created before the `question_name_when_answered` column existed have a blank snapshot, so they still fall back to the current field name (no backfill). Covered by new helper and request specs.

### DIFF
--- a/app/helpers/event_helper.rb
+++ b/app/helpers/event_helper.rb
@@ -113,6 +113,14 @@ module EventHelper
     end
   end
 
+  # The question label to show next to a stored answer: the wording captured at
+  # submission time (so editing a question later doesn't rewrite history), falling
+  # back to the field's current name when there's no answer on file — e.g. a
+  # question added after this person submitted.
+  def display_question_label(field, response)
+    response&.question_name_when_answered.presence || field&.name
+  end
+
   def display_response_text(field, response)
     text = resolve_answer_text(field, response&.submitted_answer)
     return tag.span("—", class: "text-gray-400") if text.blank?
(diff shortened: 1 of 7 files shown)


app/helpers/event_helper.rb
```diff
@@ -97,6 +97,14 @@ def event_profile_button(event, truncate_at: nil, subtitle: nil, data: {}, path:
     end
   end
 
+  # The question label to show next to a stored answer: the wording captured at
+  # submission time (so editing a question later doesn't rewrite history), falling
+  # back to the field's current name when there's no answer on file — e.g. a
+  # question added after this person submitted.
+  def display_question_label(field, response)
+    response&.question_name_when_answered.presence || field&.name
```
maebeale, line 121: 🤖 _From Claude:_ Fallback to the live `field.name` is intentional but means answers predating the `question_name_when_answered` column (blank snapshot) still show the *current* wording — there's no backfill, only forward capture.

PR #1705, opened 2026-06-17
filemaker initial-structure

jmilljr24:


maebeale (approved):
😍

app/models/resource.rb
```diff
@@ -12,7 +12,7 @@ def self.mentionable_rich_text_fields
   end
 
   PUBLISHED_KINDS = [ "Handout", "Template", "Toolkit", "Form" ]
-  KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme", "Scholarship" ]
+  KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme", "Scholarship", "FM Archive" ]
```
jmilljr24, line 15: Temp added for easy upload of fm copies via ui

app/renderers/fm_archive_pagination_renderer.rb
```diff
@@ -0,0 +1,11 @@
+class FmArchivePaginationRenderer < TailwindPaginationRenderer
```
jmilljr24, line 1: I didn't want to hack into the main pagination to make this work. It would be relatively small but the archive is somewhat self contained so thought to keep is separate.

app/views/organizations/edit.html.erb
```diff
@@ -8,6 +8,9 @@
           </div>
         <% end %>
         <div class="flex flex-wrap justify-end gap-2 mb-4">
+          <% if @organization.filemaker_code.present? %>
```
jmilljr24: This doesn't have to live here. Just want to surface a link for now.

app/controllers/fm_archives_controller.rb
```diff
@@ -0,0 +1,77 @@
+class FmArchivesController < ApplicationController
+  TABLES = {
```
maebeale, line 2: is this all the tables in there?

app/views/organizations/edit.html.erb
maebeale: fm code is already in this form. i think if we have a match on code we make the button there? fine w this option instead and/or in addition.

fm_data_structure.md
```diff
@@ -0,0 +1,424 @@
+# FileMaker Data Structure via ODBC API
```
maebeale, line 1: 👏

jmilljr24: Theres more to this story lol. We will still need to do a manual export but the odbc did help me discover all the tables.

app/views/organizations/edit.html.erb
jmilljr24: Whatever you think! I'm guessing when reviewing a person and they get the itch to look at the filemaker data, I suspect their first inclination is NOT to look for the edit field of the filemaker code. Where the best spot is, I don't know.

app/controllers/fm_archives_controller.rb
jmilljr24: Correct as far as could figure out.

app/views/organizations/edit.html.erb
maebeale: i like your idea of button at the top. let's also add link in hint of the fm code field "View filemaker archive", if there's a fm code present and it matches the archive.

PR #1706, opened 2026-06-17
Remove "Other" payment method from registration forms

maebeale:
- The "Other" payment method added noise to the payment choice without a clear meaning for registrants, so it is being removed from the registration and bulk payment forms.

- Dropped "Other" from `FormBuilderService::PAYMENT_METHOD_OPTIONS`, so newly built registration and bulk payment forms no longer offer it. Existing forms are left as-is (no data migration).
- Scholarship applicants previously had "Other" preselected as their payment method; they now see the same default as every other registrant and are still never charged.

### UI Testing Checklist
- [ ] On a paid event's registration form, the payment method choices show Credit card (now/later) and Check, with no "Other".
- [ ] On the bulk payment form, "Other" no longer appears among the payment method choices.
- [ ] A scholarship registration still shows the payment method section (without "Other").

### Anything else to add?
- Added request specs covering the removed option and the scholarship behavior.

### DIFF
--- a/app/services/form_builder_service.rb
+++ b/app/services/form_builder_service.rb
@@ -3,12 +3,7 @@ class FormBuilderService
   # charge, so controllers match against PAYMENT_METHOD_PAY_NOW rather than
   # repeating its label. Keep this the single source of truth for the label.
   PAYMENT_METHOD_PAY_NOW = "Credit card (now)".freeze
-  # "Other" is retained as a payment method in code and form data — the scholarship
-  # flow stores it as the payment method — but it is intentionally hidden from the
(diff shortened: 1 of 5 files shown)


jmilljr24 (approved):
If you think this is the direction it needs to go that is fine.

My understand from the conversation was to remove "other" completely. They will get user feedback on any confusion regarding scholarships and payments.

One down side to this is we now have another option of a hardcoded value with extra logic around it. I don't love having the stripe checkout dependent on a string (we already saw the issue of checkouts not working with this). My vote it so just remove other. One less thing to deal with right now.

maebeale:
ok updated to just remove

PR #1707, opened 2026-06-17
Add W-9 download to the registration ticket via an "Additional forms" question

maebeale:

- Registrants who need **AWBW's W-9** (or an invoice) had no way to request it during registration, and staff had no signal to act on.
- This adds an **"Additional forms"** question to the registration form and surfaces the requested document(s) on the registrant's digital ticket.
- Scoped to the **W-9 download** for now; the invoice flag is wired up but its print/page lives in a follow-up PR.

- Followed the existing `ce_credit_requested` **"magic question"** pattern so the seed, service, and ticket agree on one well-known `field_identifier`.
- **Migration:** added `w9_requested` and `invoice_requested` booleans to `event_registrations` (default `false`, not null).
- **Service** (`EventRegistrationServices::PublicRegistration`): new `ADDITIONAL_FORMS_IDENTIFIER` constant + `w9_requested?` / `invoice_requested?` that read the multi-select answer, setting both flags on create **and** on re-registration.
- **Seed:** added an "Additional forms" group header + `multi_select_checkbox` question (options **Invoice**, **W-9**) in its own section, idempotent on re-seed.
- **Ticket:** conditional **Download W-9** card (shown only when `w9_requested?`), styled to match the existing "Before you attend" callout.
- **W-9 file:** committed AWBW's W-9 to `public/documents/awbw-w9.pdf` — served at a stable, undigested `/documents/awbw-w9.pdf` URL (not the asset pipeline, which would fingerprint the filename).

### UI Testing Checklist

- [ ] Register for an event and check **W-9** under "Additional forms" → the ticket shows a **Download W-9** link that opens `public/documents/awbw-w9.pdf`.
- [ ] Register without checking W-9 → no download link on the ticket.
- [ ] Check **Invoice** → `invoice_requested` is set (no ticket link yet — follow-up PR).

### Anything else to add?

- **Decision:** the question is `required: false`. "Do you need either of the following?" must allow "neither," matching the CE magic-question precedent — despite the red `*` in the original mock.
- **Follow-up PR (deferred):** the invoice print page + its conditional ticket link, consuming the already-present `invoice_requested` flag.
- Tests: service specs (both / one / none / re-registration) + request specs (W-9 link shows iff `w9_requested`). All green; lint clean; seed verified idempotent.


app/views/event_registrations/_ticket.html.erb
```diff
@@ -109,6 +109,22 @@
         <% end %>
       <% end %>
 
+      <!-- Additional forms the registrant asked for during registration -->
+      <% if event_registration.w9_requested? %>
+        <%= link_to "/documents/awbw-w9.pdf",
```
maebeale, line 114: 🤖 _From Claude:_ Links to a stable, undigested public path on purpose — the W-9 lives at `public/documents/awbw-w9.pdf`, not the asset pipeline (which would fingerprint the filename and break this URL). Shown only when `w9_requested?`.

app/services/event_registration_services/public_registration.rb
```diff
@@ -264,6 +276,20 @@ def ce_credit_requested?
       field_value(CE_CREDIT_INTEREST_IDENTIFIER).to_s.strip.casecmp?("yes")
     end
 
+    # The "Additional forms" question is a multi-select, so its submitted value is
+    # an array of the checked option labels (e.g. [ "Invoice", "W-9" ]).
+    def additional_forms_selections
```
maebeale, line 281: 🤖 _From Claude:_ "Additional forms" is a multi-select, so its submitted value is an array of checked labels (e.g. `["Invoice", "W-9"]`). `Array(...)` normalizes the nil/single/array cases before the per-flag `casecmp?` checks below.

db/seeds/dev/events_management.rb
```diff
@@ -129,6 +129,46 @@
...
+    name: "Additional forms",
+    answer_type: :group_header,
+    status: :active,
+    position: next_position,
+    required: false,
+    section: "additional_forms",
+    visibility: :always_ask
+  )
+  additional_forms_field = registration_form.form_fields.create!(
+    name: "Do you need either of the following?",
```
maebeale, line 152: 🤖 _From Claude:_ Deliberately `required: false` (set on the field below): "Do you need either of the following?" must allow "neither," matching the CE magic-question precedent — despite the red `*` in the original mock. Flag if you actually want it required.

PR #1710, opened 2026-06-17
Add a CE hours detail section to the registration ticket

maebeale:
- Continuing-education requirements (CAMFT approval, license-number and payment rules, sign-in instructions, post-training evaluation) used to be re-sent in a long confirmation email for every CE-eligible event.
- This surfaces that info per event so admins keep the embedded links in one place instead of re-emailing.

- Mirrored the existing "Before you attend" feature: two new `events` columns (`ce_hours_details` HTML + `ce_hours_details_label`, default "CE hours"), an admin collapsible editor, a `GET /events/:id/ce_hours` sanitized-HTML details page, and a ticket call-out (indigo + graduation-cap, distinct from the amber "Before you attend" call-out).
- The call-out and page only render when content is present; blank details redirect to the event. Added request specs covering the page and the admin form fields.

### UI Testing Checklist
- [ ] Edit an event → open the **CE hours** section, set a label and HTML content (with links), and save.
- [ ] View a registration ticket for that event → the indigo **CE hours** call-out appears and links to the details page; embedded links work.
- [ ] Clear the content → call-out and page disappear (visiting the URL redirects to the event).

### Anything else to add?
- Content is admin-entered per event (deadline/cost vary), sanitized to the same allowed tags as "Before you attend". No content is seeded.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -156,6 +156,13 @@ class Event < ApplicationRecord
     super.presence || "Before you attend"
   end
 
+  # Heading shown on the CE hours ticket call-out and its details page. Falls
+  # back to the default even when an admin clears it, so the section never
+  # renders unlabelled.
+  def ce_hours_details_label
+    super.presence || "CE hours"
+  end
+
   # Virtual attributes for date/time inputs (Firefox datetime-local compat)
   attr_writer :start_date_date, :start_date_time,
               :end_date_date, :end_date_time,
--- a/app/views/event_registrations/_ticket.html.erb
+++ b/app/views/event_registrations/_ticket.html.erb
@@ -109,6 +109,21 @@
         <% end %>
       <% end %>
 
+      <!-- CE hours (continuing education requirements, payment, sign-in rules) -->
+      <% if event_registration.event.ce_hours_details.present? %>
+        <%= link_to ce_hours_event_path(event_registration.event, reg: event_registration.slug),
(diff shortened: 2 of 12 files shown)


app/models/event.rb
```diff
@@ -156,6 +156,13 @@ def event_details_label
     super.presence || "Before you attend"
```
maebeale, line 156: 🤖 _From Claude:_ Mirrors `event_details_label` — falls back to the default even when an admin clears the field, so the section never renders unlabelled.

app/views/event_registrations/_ticket.html.erb
```diff
@@ -109,6 +109,21 @@
         <% end %>
       <% end %>
 
+      <!-- CE hours (continuing education requirements, payment, sign-in rules) -->
+      <% if event_registration.event.ce_hours_details.present? %>
+        <%= link_to ce_hours_event_path(event_registration.event, reg: event_registration.slug),
+                      class: "flex items-center gap-3 rounded-xl border-2 border-indigo-300 bg-indigo-50 px-4 py-3 hover:bg-indigo-100 transition-colors" do %>
```
maebeale, line 115: 🤖 _From Claude:_ Intentionally indigo + graduation-cap to distinguish this call-out from the amber "Before you attend" one directly above, so two stacked call-outs on a ticket read as different things.

PR #1711, opened 2026-06-17
Add public slug-based bulk payment ticket page

maebeale:
- Bulk payment payers have no account, yet need a durable, shareable view of what they paid for.
- This gives bulk payments the **same two-page shape as registrations**: a styled *ticket* plus a public *submitted-form* page, both reachable by an unguessable slug.

- A bulk payment is a `FormSubmission` with `role: "bulk_payment"`, so I added a `slug` column (generated for bulk payment submissions only) and two public pages:
  - **Ticket** — `/bulk-payment/:slug` — event details, the **list of registrants covered**, payment status, invoice, and a "View submission" link. Unlike the registration ticket it shows *many* names (not one registrant) and omits attendee-only content (videoconference label/join link, "before you attend", per-person cancel/pay actions).
  - **Submission** — `/events/:id/bulk_payment?reg=:slug` — the submitted form responses, mirroring `public_registration#show`; backs to the ticket (or the dashboard when an admin arrives via `return_to=bulk_payments`).
- The payer confirmation email links to the ticket (like the registration confirmation, guarded by event presence); the staff FYI email and the admin dashboard's "Submission" button link to the submission page. The id-based `show` was replaced by the slug-based `?reg=` page.

### UI Testing Checklist
- [ ] Submit a bulk payment → redirected to `/bulk-payment/:slug` showing event, registrants list, and payment status
- [ ] Confirm the ticket omits the videoconference link/label, "before you attend", and per-person actions
- [ ] Click "View submission" on the ticket → lands on the submission page → "Back to ticket" returns
- [ ] From the admin bulk-payments dashboard, click "Submission" → submission page → "Back to bulk payments" returns to the expanded row
- [ ] Click "View invoice" on the ticket → back link returns to the ticket
- [ ] Payer confirmation email "View bulk payment" opens the ticket; staff FYI "View submission" opens the submission page
- [ ] As an admin, the ticket still shows the allocations section

### Anything else to add?
- There are no existing bulk payments, so no public backward compatibility was needed for the old id-based confirmation page.


app/models/form_submission.rb
```diff
@@ -7,6 +7,25 @@ class FormSubmission < ApplicationRecord
...
 
+  scope :bulk_payment, -> { where(role: "bulk_payment") }
+
+  validates :slug, uniqueness: true, allow_nil: true
+
+  # Bulk payment submissions are reachable by their payer (who has no account)
+  # through a public, slug-based ticket URL. Other roles are reached by id.
+  before_create :generate_slug, if: :bulk_payment?
+
+  def self.generate_unique_slug
```
maebeale, line 18: 🤖 _From Claude:_ Slugs are generated only for bulk payment submissions (other roles stay id-addressed). I deliberately did **not** override `to_param`, so the admin `form_submission_path` and `FormSubmission.find` keep working by id.

app/controllers/events/bulk_payments_controller.rb
```diff
@@ -64,6 +64,17 @@ def show
       @event = @event.decorate
     end
 
+    # Public, slug-based ticket for the payer. Shows the event details, the
+    # registrants they paid for, and the submitted form — but none of the
+    # per-person admin actions found on the bulk payments dashboard.
+    def ticket
```
maebeale, line 72: 🤖 _From Claude:_ The old id-based `show` is kept for back-compat with confirmation emails already in inboxes; new flows (create redirect, Stripe return, payer email) point at this slug ticket instead.

app/views/events/bulk_payments/ticket.html.erb
```diff
@@ -0,0 +1,209 @@
...
+          <i class="fa-solid fa-circle-info text-amber-500 text-xl shrink-0"></i>
+
+          <div class="flex-1 min-w-0 text-left">
+            <p class="font-semibold text-amber-900"><%= @event.event_details_label %></p>
+            <p class="text-sm text-amber-700">Important info for this event — please read</p>
+          </div>
+
+          <i class="fa-solid fa-arrow-right text-amber-500 shrink-0"></i>
+        <% end %>
+      <% end %>
```
maebeale: 🤖 _From Claude:_ `joinable: false` is passed explicitly so the videoconference *label* shows but the join link doesn't — a bulk payer isn't a registrant, so per-person join access doesn't apply here.

PR #1714, opened 2026-06-17
Hide select resources from the portal while keeping direct-link access

maebeale:

Some resources need to stay publicly accessible by their direct URL while being kept out of the user-facing portal (search results, listings, and the public homepage) so they can be shared by link without surfacing to everyone browsing.

Added a `hidden_from_search` boolean on `Resource` plus a `searchable` scope, and excluded hidden resources from every non-admin listing — the `/resources` search index and tag search (via `ResourcePolicy`'s relation scope) and the public homepage — while leaving `show?`/direct-link access untouched so the resources remain publicly reachable. Admins still find them, marked with an amber "Hidden from search" badge, and the lone direct `Resource.story` query was routed through `authorized_scope`. Also seeded two example hidden PDFs and tidied the resource form so Kind, Windows audience, and a shortened Display order field sit cleanly below the title.

### UI Testing Checklist

- [ ] As an admin, a `hidden_from_search` resource appears in `/resources` with a "Hidden from search" badge.
- [ ] As a regular user and as a guest, that resource does NOT appear in `/resources` search/listings or on the homepage.
- [ ] The resource's direct link (`/resources/:id`) still loads for guests.
- [ ] On resource edit, the "Hidden from search" flag toggles, and Kind / Windows audience / Display order render below the title (audience + display order on one row).

### Anything else to add?

`ResourcesController#stories` is currently unrouted (the `stories` routes map to a separate `StoriesController`); it was switched to `authorized_scope` for consistency but is slated for removal as dead code.


app/policies/resource_policy.rb
```diff
@@ -27,10 +27,7 @@ def filter_published?
 
   relation_scope do |relation|
     next relation if admin?
-    if authenticated?
-      relation.published
-    else
-      relation.publicly_visible
-    end
+    visible = authenticated? ? relation.published : relation.publicly_visible
+    visible.searchable
```
maebeale, line 31: 🤖 _From Claude:_ This is the central gate: hidden resources drop out of every non-admin `authorized_scope(Resource)` listing (the search index and tag search). Admins are exempt via the `next relation if admin?` above, so they still find them. Deliberately not a `default_scope` — that would hide them from admins and from direct-link `show?` too.

app/models/resource.rb
```diff
@@ -89,6 +90,9 @@ def self.mentionable_rich_text_fields
     result = value ? published_kinds.where(published: true) : where(published: false)
   end
   scope :recent, -> { published.by_created }
+  # Resources flagged hidden_from_search stay publicly accessible by direct link
+  # but are excluded from non-admin portal searches and listings.
+  scope :searchable, -> { where(hidden_from_search: false) }
```
maebeale, line 95: 🤖 _From Claude:_ `hidden_from_search` resources stay reachable by direct link (`show?` is untouched); this scope only removes them from search/listings.

app/controllers/home/resources_controller.rb
```diff
@@ -6,6 +6,7 @@ def index
       authorize! :home
       @resources = authorized_scope(Resource.includes(primary_asset: { file_attachment: :blob }, downloadable_asset: { file_attachment: :blob }, gallery_assets: { file_attachment: :blob })
                            .published
+                           .searchable
```
maebeale, line 9: 🤖 _From Claude:_ The homepage uses `HomePolicy` (not `ResourcePolicy`), which filters by featured/publicly_featured only — so the policy gate doesn't reach it and a hidden+featured resource would otherwise leak onto the public landing. `.searchable` can't live in `HomePolicy` since it's shared with events/workshops/etc. that have no such scope, so it's applied here.

app/controllers/resources_controller.rb
```diff
@@ -29,7 +29,7 @@ def index
 
   def stories
     authorize!
-    @stories = Resource.story.paginate(page: params[:page], per_page: 6).decorate
+    @stories = authorized_scope(Resource.story).paginate(page: params[:page], per_page: 6).decorate
```
maebeale, line 32: 🤖 _From Claude:_ Switched from a direct `Resource.story` call to `authorized_scope`. Note this action is currently unrouted (the `stories` routes map to a separate `StoriesController`), and `published` restricts to PUBLISHED_KINDS — so consider removing it as dead code in a follow-up.

app/helpers/title_display_helper.rb
```diff
@@ -43,6 +43,18 @@ def title_with_badges(record, font_size: "text-lg", record_title: nil,
       fragments << content_tag(:span, safe_join(visibility_badges, " "), class: "inline-flex items-center gap-2 flex-nowrap")
     end
 
+    # --- Hidden from search badge (publicly accessible, but excluded from portal searches) ---
+    if current_user&.super_user? && !home_page &&
```
maebeale, line 47: 🤖 _From Claude:_ Badge is gated to admins (`super_user?`) since the hidden/not-public status is internal info — regular users and guests who reach the resource by direct link don't see it.

PR #1716, opened 2026-06-18
Append newly added form fields to the bottom of the list

maebeale:
- In the form editor, a newly added field jumped to the **top** of the question list instead of staying where the admin added it.
- The cause: a new `FormField` saved without a `position` (the column is nullable), and the list is ordered by `position: :asc`, so `nil` sorted above every real field.

- Added a `before_validation` on `FormField` that, when a field is created without a position, assigns it `max(position) + 1` for its form — placing it at the bottom.
- Dragged fields already arrive with an explicit position (written by the sortable controller) and are left untouched, so they stay wherever they were dropped.

### UI Testing Checklist
- [ ] In the form editor, click "+ Add field" — the new field appears at the bottom of the list after saving, not the top.
- [ ] Add a field, drag it to a new spot, and save — it stays where it was dropped.

### Anything else to add?
- Covered by a new request spec asserting a positionless new field lands after existing fields. No JS changes were needed.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -91,6 +91,12 @@ class FormField < ApplicationRecord
   validate :max_characters_allows_min_words
   validate :non_input_types_cannot_be_required
 
+  # New fields submitted without a position (e.g. added in the form editor and
+  # not dragged) would otherwise save with a nil position and sort to the top of
+  # the list. Append them to the bottom instead. Dragged fields arrive with an
+  # explicit position and are left untouched.
+  before_validation :append_to_bottom, on: :create, if: -> { position.blank? }
+
   # Enum
   enum :status, [ :inactive, :active ]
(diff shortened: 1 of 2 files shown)


app/models/form_field.rb
```diff
@@ -91,6 +91,12 @@ class FormField < ApplicationRecord
   validate :max_characters_allows_min_words
   validate :non_input_types_cannot_be_required
 
+  # New fields submitted without a position (e.g. added in the form editor and
+  # not dragged) would otherwise save with a nil position and sort to the top of
+  # the list. Append them to the bottom instead. Dragged fields arrive with an
```
maebeale, line 96: 🤖 _From Claude:_ Guarded on `position.blank?` and `on: :create` so dragged fields (which submit an explicit position via the sortable controller) and `FormBuilderService` (which sets positions itself) are left untouched — only positionless new fields get appended.

PR #1717, opened 2026-06-18
Populate Additional Ages Served from age range categories

maebeale:
- Registration forms offered a category-driven **"Primary Age Group(s) Served"** question, but the parallel **"Additional"** age question — mirroring the existing primary/additional **sectors** pattern — was never seeded as a field.
- This adds an **"Additional Age Group(s) Served"** question so respondents can name the other age groups they serve, populated exactly like the primary question.

- Added the `additional_age_group` field to the professional-info section in `FormBuilderService` (and to `SECTION_FIELDS` / `SECTION_FIELD_NAMES`), modeled on the "Additional service areas" naming/hint.
- Wired it to the same `AgeRange` categories and the "Mixed-age groups" exclusion via a new `AGE_GROUP_FIELD_IDENTIFIERS` constant, so its options match Primary Age exactly.
- Made answer rendering (`EventHelper`) and "Other" free-text mapping (`Person`) treat it identically to the other category-backed professional fields; added a spec asserting its options equal the primary field's.

### Anything else to add?
- Forms are stored as `FormField` records, so this seeds the field into newly built forms only — existing event forms aren't backfilled. The event dashboard's "Primary age group" breakdown was intentionally left reading `primary_age_group` only.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -31,13 +31,16 @@ class FormField < ApplicationRecord
   DYNAMIC_FIELD_CATEGORY_TYPES = {
     "workshop_environments" => "WorkshopEnvironment",
     "client_life_experiences" => "StoryPopulation",
-    "primary_age_group" => "AgeRange"
+    "primary_age_group" => "AgeRange",
+    "additional_age_group" => "AgeRange"
   }.freeze
 
-  # The "primary age group(s) served" field. Backed by AgeRange categories but
-  # omits the catch-all "Mixed-age groups" category — a respondent names the
-  # concrete age groups they serve, not the mixed bucket.
+  # The "primary" and "additional" age group fields. Both are backed by AgeRange
+  # categories but omit the catch-all "Mixed-age groups" category — a respondent
+  # names the concrete age groups they serve, not the mixed bucket.
   PRIMARY_AGE_GROUP_FIELD_IDENTIFIER = "primary_age_group"
+  ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER = "additional_age_group"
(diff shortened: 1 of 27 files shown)


app/models/form_field.rb
```diff
@@ -292,16 +295,16 @@ def service_area_sectors
...
+  # groups" AgeRange category — a respondent names the concrete age groups they
+  # serve. Empty when the backing CategoryType is missing. Source of truth shared
+  # by the public form's rendering and submission validation.
   def dynamic_categories
     type = CategoryType.find_by(name: DYNAMIC_FIELD_CATEGORY_TYPES[field_identifier])
     return Category.none unless type
 
     scope = type.categories.published.order(:position, :name)
-    field_identifier == PRIMARY_AGE_GROUP_FIELD_IDENTIFIER ? scope.excluding_mixed_age : scope
+    AGE_GROUP_FIELD_IDENTIFIERS.include?(field_identifier) ? scope.excluding_mixed_age : scope
```
maebeale, line 319: 🤖 _From Claude:_ Generalized the single-field check to `AGE_GROUP_FIELD_IDENTIFIERS` so the additional field shares Primary Age's exact options, including the `excluding_mixed_age` filter that hides the catch-all bucket.

app/services/form_builder_service.rb
```diff
@@ -452,6 +452,9 @@ def build_professional_info_fields(form, position)
     position = add_field(form, position, "Primary Age Group(s) Served", :multi_select_checkbox,
                          key: "primary_age_group", group: "professional", required: false,
                          hint: "Select all age groups you primarily serve.")
+    position = add_field(form, position, "Additional Age Group(s) Served", :multi_select_checkbox,
```
maebeale, line 455: 🤖 _From Claude:_ Followed the primary/additional **sectors** naming pattern for the label and hint; key is `additional_age_group`, registered in `DYNAMIC_FIELD_CATEGORY_TYPES` above so it's AgeRange-backed.

PR #1718, opened 2026-06-18
Render form field subtitle as HTML and add a separate hint

maebeale:
- A form field's **subtitle** (the text shown under the field label) now renders sanitized HTML, just like the field title/header already do — so admins can include links, bold, and emphasis instead of seeing escaped tags.
- The historically-misnamed `hint_text` column has always acted as that subtitle, so it is renamed to `subtitle`, and a brand-new `hint_text` field is added for SimpleForm-style hint text shown **below the input**.

- Migration renames `form_fields.hint_text` → `subtitle` and adds a fresh `hint_text` text column (reversible; verified rollback + re-migrate).
- Updated every display/edit/service/seed reference: modern views render `subtitle` (and the new `hint_text`) via `form_label_html` sanitization; the form editor gains a "Hint" textarea alongside "Subtitle"; legacy report views were repointed to `subtitle` to preserve their existing behavior unchanged.
- The new hint stacks below the input independently of the auto-generated min-words / max-characters notices (which never used this column).

### UI Testing Checklist
- [ ] Subtitle with HTML (e.g. `<strong>`, `<a>`) renders formatted on the public registration form, the admin form preview, and the bulk-payment form.
- [ ] The new "Hint" field saves from the form editor and renders (as HTML) directly below the field input.
- [ ] A free-form field shows hint, min-words, and max-characters notices stacked in that order without overlap.
- [ ] Legacy monthly-report / workshop-log forms still display their subtext as before.

### Anything else to add?
- Both subtitle and the new hint are sanitized through the same `FORM_LABEL_TAGS`/`FORM_LABEL_ATTRIBUTES` allowlist as the title. Refreshed the Brakeman mass-assignment ignore fingerprint after the permitted-params list changed. Specs, RuboCop, and the security scan all pass.

### DIFF
--- a/app/views/events/public_registrations/_form_field.html.erb
+++ b/app/views/events/public_registrations/_form_field.html.erb
@@ -18,8 +18,8 @@
     <% end %>
   </div>
 
-  <% if field.hint_text.present? %>
-    <p class="text-xs text-gray-500 mb-1.5"><%= field.hint_text %></p>
+  <% if field.subtitle.present? %>
+    <p class="rich-label text-xs text-gray-500 mb-1.5"><%= form_label_html(field.subtitle) %></p>
   <% end %>
 
   <% field_name = "public_registration[form_fields][#{field.id}]" %>
@@ -164,6 +164,10 @@
       <% end %>
     <% end %>
 
+  <% end %>
+
+  <% if field.hint_text.present? %>
+    <p class="rich-label text-xs text-gray-500 mt-1.5"><%= form_label_html(field.hint_text) %></p>
   <% end %>
 
   <% if field.free_form_text? && field.min_words.to_i.positive? %>
--- a/app/views/shared/_form_field.html.erb
+++ b/app/views/shared/_form_field.html.erb
@@ -3,13 +3,13 @@
   <div class="label-plus-comment">
     <%# Display-only field — never collects input, so no required marker. %>
     <label><%= field.name %></label>
-    <div class="custom-smaller"> <%= field.hint_text %> </div>
+    <div class="custom-smaller"> <%= field.subtitle %> </div>
   </div>
 
 <% when :group_header %>
   <div class="label-plus-comment">
     <label><%= field.name %> </label>
-    <div class="custom-smaller"><%= field.hint_text %></div>
+    <div class="custom-smaller"><%= field.subtitle %></div>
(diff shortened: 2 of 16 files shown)


db/migrate/20260618014554_rename_form_field_hint_text_to_subtitle_and_add_hint_text.rb
```diff
@@ -0,0 +1,14 @@
+class RenameFormFieldHintTextToSubtitleAndAddHintText < ActiveRecord::Migration[8.0]
+  # The column historically called `hint_text` has always rendered as a subtitle
+  # under the field label, so it is renamed to `subtitle`. A fresh `hint_text`
+  # column is then added for SimpleForm-style hint text shown below the input.
```
maebeale, line 4: 🤖 _From Claude:_ Order matters: `up` renames `hint_text`→`subtitle` first, then adds a *new* `hint_text`; `down` reverses it (drop new column, rename back). Guarded with `column_exists?`/`if_exists` so reruns are idempotent — verified rollback + re-migrate.

app/views/events/public_registrations/_form_field.html.erb
```diff
@@ -166,6 +166,10 @@
 
   <% end %>
 
+  <% if field.hint_text.present? %>
+    <p class="rich-label text-xs text-gray-500 mt-1.5"><%= form_label_html(field.hint_text) %></p>
```
maebeale, line 170: 🤖 _From Claude:_ New hint renders through the same `form_label_html` sanitizer as the title/subtitle. It stacks independently above the min-words/max-chars notices, which never read this column.

app/views/shared/_form_field.html.erb
```diff
@@ -3,13 +3,13 @@
   <div class="label-plus-comment">
     <%# Display-only field — never collects input, so no required marker. %>
     <label><%= field.name %></label>
-    <div class="custom-smaller"> <%= field.hint_text %> </div>
+    <div class="custom-smaller"> <%= field.subtitle %> </div>
```
maebeale, line 6: 🤖 _From Claude:_ Legacy report views repointed `hint_text`→`subtitle` only to keep working — display behavior is intentionally unchanged (still escaped, no HTML), since their titles render escaped too.

PR #1719, opened 2026-06-18
Add intends-to-pay flag to grant ticket access before payment clears

maebeale:
- Some registrants commit to paying after the deadline but can't pay in time; previously they were locked out of their ticket's training materials and videoconference join link, since access keys off paid status.
- This adds an admin-settable `intends_to_pay` flag that grants that access while keeping the registration's reported payment status as "due" everywhere money is reported.

- Added an `intends_to_pay` boolean column plus `access_granted?` (`paid? || intends_to_pay?`) on `EventRegistration`, and routed `joinable?` (the only payment-gated access path) through it — so reporting surfaces (roster badge, both CSV exports, dashboard metrics) still reflect the true balance owed.
- Surfaced an "Intends to pay" toggle in the registration's payments/allocations card, an access-granted badge on the ticket, an "Intends to pay" roster marker and payment-status filter option, and a dedicated "Intends to pay" column in both registration CSV exports; seeded Aisha on the Facilitator Training (no payments, flagged) to demo the state alongside Amy.

### UI Testing Checklist
- [ ] Toggle "Intends to pay" on an unpaid registration; confirm the ticket shows the join link/materials and an "Access granted — payment pending" badge while status stays "Due".
- [ ] Confirm the roster badge, payment filter, and CSV exports still report the registrant as owing, and that the new "Intends to pay" CSV column reads Yes/No.

### Anything else to add?
- The `intends_to_pay` naming was kept intentionally; `access_granted?` is the method any future payment-gated resource (e.g. recordings) should use rather than `paid?` — now documented on the method.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -262,6 +262,7 @@ class EventRegistrationsController < ApplicationController
     params.require(:event_registration).permit(
       :event_id, :registrant_id, :status,
       :scholarship_requested,
+      :intends_to_pay,
       :ce_credit_requested,
       :ce_hours_requested,
       :ce_license_number,
@@ -296,7 +297,7 @@ class EventRegistrationsController < ApplicationController
           er.attendance_status_label,
           er.scholarships.any? ? "Yes" : "No",
           er.scholarships.completed.any? ? "Yes" : "No",
-          cost_required ? (er.paid_in_full? ? "Paid" : "Due") : "",
+          cost_required ? er.payment_status_label : "",
           total_cents.positive? ? format("%.2f", total_cents / 100.0) : ""
         ]
       end
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -282,7 +282,7 @@ class EventRegistrationsController < ApplicationController
 
   def csv_export(registrations)
     CSV.generate(headers: true) do |csv|
-      csv << [ "First name", "Last name", "Email", "Phone", "Event", "Status", "Scholarship", "Scholarship completed", "Payment status", "Payment total" ]
+      csv << [ "First name", "Last name", "Email", "Phone", "Event", "Status", "Scholarship", "Scholarship completed", "Payment status", "Intends to pay", "Payment total" ]
       registrations.find_each do |er|
         r = er.registrant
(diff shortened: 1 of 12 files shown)


app/models/event_registration.rb
```diff
@@ -184,6 +185,23 @@ def paid?
...
   end
 
+  # True when the registrant should be granted access to ticket materials
+  # (training links, etc.) even though they haven't paid in full yet. Admins
+  # flip the `intends_to_pay` flag when someone commits to paying after the
+  # deadline so they aren't locked out in the meantime. This does NOT mark the
+  # registration as paid — payment status still shows as due.
+  def access_granted?
+    paid? || intends_to_pay?
+  end
```
maebeale, line 201: 🤖 _From Claude:_ `access_granted?` is the single seam for "may this registrant reach paid content?" — `joinable?` (the only payment-gated path, the videoconference link) routes through it. Any future paid-gated resource (e.g. recordings) should use this, not `paid?`.

app/models/event_registration.rb
```diff
@@ -184,6 +185,23 @@ def paid?
...
+
+  # Human-readable payment status for rosters and CSV exports. Assumes the event
+  # has a cost — callers show nothing for free events.
+  def payment_status_label
```
maebeale, line 205: 🤖 _From Claude:_ Deliberately keyed off `paid_in_full?`, not `access_granted?` — reporting surfaces (rosters, CSV, dashboard metrics) must still show an intends-to-pay registrant as owing.

app/controllers/event_registrations_controller.rb
```diff
@@ -296,7 +297,8 @@ def csv_export(registrations)
           er.attendance_status_label,
           er.scholarships.any? ? "Yes" : "No",
           er.scholarships.completed.any? ? "Yes" : "No",
-          cost_required ? (er.paid_in_full? ? "Paid" : "Due") : "",
+          cost_required ? er.payment_status_label : "",
+          er.intends_to_pay? ? "Yes" : "No",
```
maebeale, line 301: 🤖 _From Claude:_ Dedicated `Intends to pay` Yes/No column added alongside the existing payment-status text, so finance can filter/sort on the raw flag rather than parsing the status label. Mirrored in the events-controller export.

PR #1720, opened 2026-06-18
Remove workshop settings and client life experiences registration questions

maebeale:
- The "Workshop Settings" and "Client Life Experiences" professional-info questions are no longer collected, so they are removed from the registration form builder and the code that consumed their submitted answers.

- Dropped both fields from `FormBuilderService` (builder + section registries) and from `FormField`'s dynamic category-type mapping, and stopped `PublicRegistration` from tagging registrants from them.
- Narrowed `Person`/`EventHelper` handling to the remaining `primary_age_group` field, removed the dev seed's "Other" workshop-setting demo answer, and updated the affected specs.

### Anything else to add?
- The event Background dashboard's "Settings" and "Life experiences" charts are **retained** for historical and admin-applied tags — the `WorkshopEnvironment`/`StoryPopulation` category types and their seeds stay; only the self-service registration questions were removed, so future registrations simply no longer write these tags.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -209,9 +209,9 @@ class Person < ApplicationRecord
 
   # Field identifiers whose "Other" free text maps onto the profile's sectors.
   OTHER_SERVICE_AREA_IDENTIFIERS = %w[primary_service_area primary_service_area_single].freeze
-  # Field identifiers whose "Other" free text maps onto the workshop-setting
-  # categories shown on the edit page.
-  OTHER_WORKSHOP_SETTING_IDENTIFIERS = %w[workshop_environments client_life_experiences primary_age_group additional_age_group].freeze
+  # Field identifiers whose "Other" free text maps onto the category-backed
+  # profile fields shown on the edit page.
+  OTHER_WORKSHOP_SETTING_IDENTIFIERS = %w[primary_age_group additional_age_group].freeze
 
   # Free-text "Other" service areas the person typed on registration forms.
   # They can't be Sector records, so they're surfaced beside the sector tags.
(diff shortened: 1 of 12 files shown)


app/models/person.rb
```diff
@@ -209,9 +209,9 @@ def remote_search_label
 
   # Field identifiers whose "Other" free text maps onto the profile's sectors.
   OTHER_SERVICE_AREA_IDENTIFIERS = %w[primary_service_area primary_service_area_single].freeze
-  # Field identifiers whose "Other" free text maps onto the workshop-setting
-  # categories shown on the edit page.
-  OTHER_WORKSHOP_SETTING_IDENTIFIERS = %w[workshop_environments client_life_experiences primary_age_group].freeze
+  # Field identifiers whose "Other" free text maps onto the category-backed
+  # profile fields shown on the edit page.
+  OTHER_WORKSHOP_SETTING_IDENTIFIERS = %w[primary_age_group].freeze
```
maebeale: 🤖 _From Claude:_ Kept `primary_age_group` as the sole identifier — it is still a category-backed field whose "Other" free text surfaces on the profile/edit pages; only the two removed questions left this list.

db/seeds/dev/events_management.rb
```diff
@@ -801,11 +801,11 @@
 
 puts "Tagging registrants with life experiences and workshop settings…"
```
maebeale, line 802: 🤖 _From Claude:_ Kept this tagging block (and the WorkshopEnvironment/StoryPopulation seeds) so the "Settings" and "Life experiences" background charts are retained for historical and admin-applied tags — only the self-service registration questions were removed, so public registration no longer writes these.

PR #1722, opened 2026-06-18
Compound payer & additional designation pickers on allocation payment form

maebeale:
The check/cash "add payment" subform on the allocations index made staff declare a payer type and then pick from separate person and organization dropdowns. This replaces those three fields with two compound pickers — **Payer** (hint: "Whose account is the money coming from?") and **Additional designation** — that each search people and organizations together, matching how staff actually think about a payment.

A new combined `person_or_organization` search endpoint returns both models as SGID-valued options (suffixed with the kind to disambiguate), and the Payment model gains virtual `payer_sgid` / `additional_designation_sgid` accessors that resolve the selection back into `payer_type` + `person_id` / `organization_id` on submit. The database schema is unchanged: the kind picked in **Payer** drives `payer_type`, and **Additional designation** fills the other person/org slot (a same-kind designation is dropped, since the table holds only one of each). Other payment forms that still post `payer_type` directly are untouched.

### UI Testing Checklist
- [ ] On the allocations index, "Add cash payment" / "Add check payment" show **Payer** and **Additional designation** pickers, each searching people and organizations.
- [ ] Opening the form from a registration preselects the registrant/organization sensibly.
- [ ] Submitting saves the correct `payer_type` and person/organization, and allocation succeeds.

### Anything else to add?
Note: `to_sgid` embeds an expiry timestamp and varies per call, so the computed SGID is memoized to keep the preselected option's value matching the field — otherwise prefill silently breaks. Full flexibility (e.g. two people) would require a schema migration and was intentionally left out.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,7 +1,13 @@
 class SearchController < ApplicationController
   skip_before_action :preload_current_user_associations, raise: false
 
+  # Virtual "model" that searches people and organizations together, used by the
+  # compound payer / additional-designation pickers on the payment form.
+  COMPOUND_PAYER_MODEL = "person_or_organization".freeze
+
   def index
+    return render json: compound_payer_results if params[:model] == COMPOUND_PAYER_MODEL
+
     model_class = allowed_model(params[:model])
     unless model_class
       skip_verify_authorized!
@@ -32,6 +38,19 @@ class SearchController < ApplicationController
   end
 
   private
+
+  def compound_payer_results
+    authorize! Person, to: :search?
+    authorize! Organization, to: :search?
+
+    query = params[:q].to_s.strip
+    return [] if query.blank?
+
+    records = authorized_scope(Person.remote_search(query)).limit(25).to_a +
+      authorized_scope(Organization.remote_search(query)).limit(25).to_a
+
+    records.map(&:compound_search_label)
+  end
 
   def allowed_model(model_param)
     {
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -16,7 +16,15 @@ class Payment < ApplicationRecord
 
   before_create :set_external_origin
   before_validation :set_amount_cents_remaining, if: :new_record?
+  before_validation :assign_parties_from_sgids, if: :sgid_assigned?
   before_validation :auto_set_payer_type
+
+  # Virtual attributes backing the compound "Payer" and "Additional designation"
+  # dropdowns, which each search across both people and organizations. The
+  # selected value is a signed global id; on submit we resolve it back to a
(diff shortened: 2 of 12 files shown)


app/models/payment.rb
```diff
@@ -47,6 +55,32 @@ def payer
...
+  def form_additional_designation
+    [ organization, person ].compact.find { |record| record != form_payer }
+  end
+
+  # Memoize the computed default so repeated reads in a single render return the
+  # same string — to_sgid embeds an expiry timestamp and varies per call, which
+  # would otherwise stop the preselected option from matching the field value.
+  def payer_sgid
+    return @payer_sgid if defined?(@payer_sgid)
+    @computed_payer_sgid ||= form_payer&.to_sgid&.to_s
```
maebeale: 🤖 _From Claude:_ Memoized because `to_sgid` embeds an expiry timestamp and returns a different string per call — without this, the preselected option value and the field's selected value never match and prefill silently breaks.

app/models/payment.rb
```diff
@@ -89,6 +123,50 @@ def auto_set_payer_type
...
+  # Fill whichever of person_id / organization_id the payer didn't already take.
+  # If the additional designation is the same kind as the payer, the single
+  # column for that kind is already used, so it's dropped (the schema only holds
+  # one person and one organization per payment).
+  def fill_party_slot(record)
+    case record
+    when Organization
+      self.organization_id ||= record.id
+    when Person
+      self.person_id ||= record.id
```
maebeale, line 155: 🤖 _From Claude:_ `||=` intentionally drops a same-kind additional designation (e.g. payer + designation both people): the schema holds only one person and one organization per payment, so the payer's slot wins. Full flexibility would need a migration.

app/controllers/search_controller.rb
```diff
@@ -33,6 +39,19 @@ def index
 
   private
 
+  def compound_payer_results
+    authorize! Person, to: :search?
+    authorize! Organization, to: :search?
```
maebeale, line 44: 🤖 _From Claude:_ Combined people + organizations search for the compound pickers. Both `search?` rules resolve to `admin?`, so authorizing each keeps parity with the single-model search path.

app/models/payment.rb
jmilljr24: I don't think this is necessary.  'to_sgid' is deterministic for the same inputs, both reads return the same string when this renders.

We should only need
```
def payer_sgid
  return @payer_sgid if defined?(@payer_sgid)
  form_payer&.to_sgid&.to_s
end
```

jmilljr24:
This is a nice way to handle this. Thanks

maebeale:
Screenshot 2026-06-19 at 11 15 58 AM

PR #1721, opened 2026-06-18
Add admin-configurable registration ticket callouts

maebeale:
### What is the goal of this PR and why is this important?
- Events had accreted one-off registration-ticket call-outs (CE hours, "Before you attend"), each requiring its own column, edit toggle, and detail page.
- This generalizes them into a `RegistrationTicketCallout` so admins can add any number of titled call-outs to an event's registration ticket from a single editor, each linking to its own public detail page.

### How did you approach the change?
- Added a `RegistrationTicketCallout` model (belongs_to `:event`) with title, subtitle, HTML body, `callout_type` (action/reference), Font Awesome icon, colour theme, a `show_if_paid` visibility flag, and a positioning-gem `position`.
- Callouts are managed in one collapsible editor in the event form (cocoon add/remove); drag-reordering reuses the existing categories `sortable` Stimulus controller via a per-row PUT, so no new JS controller was added.
- Each callout renders as a colour-themed call-out on the ticket linking to a detail page and honours `show_if_paid`; seeded on the art workshop, with model/request/system specs.

### UI Testing Checklist
- [ ] In an event's edit form, expand "Registration ticket callouts", add one (title, subtitle, type, colour, icon, body) and save.
- [ ] Confirm it appears on the registration ticket with the chosen icon/colour and links to its detail page.
- [ ] Drag to reorder saved callouts and confirm the order persists after reload.
- [ ] Set a callout to "only show once the registration is paid" and confirm it stays hidden until paid in full.

### Anything else to add?
- Renamed from `RegistrationTicketSection` → `RegistrationTicketCallout` (including the `callout_type` column) to match the codebase's existing "call-out" vocabulary.
- Also adds an AI-files rule (CLAUDE.md + copilot-instructions.md) to reuse or lightly adapt an existing JS controller before adding a new one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1723, opened 2026-06-18
Resolve duplicate migration version for intends-to-pay

maebeale:
- PR #1719 (intends-to-pay) and PR #1721 (registration ticket callouts) both merged a migration at version `20260618020000`, leaving two migrations with the same version on main.
- A fresh `rails db:migrate` now raises `DuplicateMigrationVersionError`, so this restores a clean migration sequence.

- Renamed `20260618020000_add_intends_to_pay_to_event_registrations.rb` to `20260618030000_…` (unique, later than the callouts migration) and bumped `db/schema.rb`'s version line to match.
- The migration body is idempotent (`add_column … unless column_exists?`), so databases that already applied the old version simply no-op on the renamed one — no data risk.

### UI Testing Checklist
- [ ] No UI change. Verify `rails db:migrate` runs cleanly from scratch and both `intends_to_pay` and `registration_ticket_callouts` exist.

### Anything else to add?
- Verified a from-scratch `db:drop/create/migrate` with no duplicate-version error; smoke specs for both features pass.


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_06_18_020000) do
+ActiveRecord::Schema[8.1].define(version: 2026_06_18_030000) do
```
maebeale, line 13: 🤖 _From Claude:_ Version bumped to match the renamed migration (`20260618030000`), now the newest. The renamed migration is idempotent, so DBs that already ran the old `20260618020000` re-run it as a no-op.

PR #1725, opened 2026-06-18
Live-preview ticket callout colour and icon in the event editor

maebeale:
The registration ticket callout editor cards gave no sense of how a call-out would look on the actual ticket — admins had to save and reload to check the colour and icon. This previews both inline so they can configure call-outs confidently in one pass.

A small `callout-preview` Stimulus controller tints each nested editor card with its selected theme (custom colour, or the per-type default) and, mirroring the public ticket row, echoes the call-out's Font Awesome icon on the left and an arrow on the right — all updating live as the type, colour, or icon fields change. The default **action** colour also moves from blue to orange so admin-added actions stand apart from the built-in blue download/invoice rows, which required adding an `orange` entry to `COLOR_THEMES`.

### UI Testing Checklist
- [ ] Editing an event: each callout card is tinted by its chosen colour, falling back to the type default (orange for action, indigo for reference).
- [ ] Changing the Type/Colour selects re-tints the card and arrow live; clearing Colour reverts to the type default.
- [ ] Typing a Font Awesome class in Icon shows it on the left of the card; clearing it hides the icon again.
- [ ] Adding a new callout row and reordering saved rows still works.

### Anything else to add?
Theme classes are already in the Tailwind safelist, so no CSS rebuild is needed. The existing blue theme stays selectable as a custom override — only the action *default* changed.

### DIFF
--- a/app/frontend/javascript/controllers/callout_preview_controller.js
+++ b/app/frontend/javascript/controllers/callout_preview_controller.js
@@ -32,14 +32,14 @@ export default class extends Controller {
     this.updateLeadingIcon(theme)
   }
 
-  // Mirror the call-out's leading icon on the left of the card — shown only when
-  // the admin has entered a Font Awesome class — tinted to match the theme.
+  // Mirror the call-out's leading icon in the card's left gutter — drawn only
+  // when the admin has entered a Font Awesome class, tinted to match the theme.
+  // The gutter keeps its width either way so the fields below stay aligned.
   updateLeadingIcon(theme) {
     if (!this.hasLeadingIconTarget || !this.hasIconInputTarget) return
     const glyph = this.iconInputTarget.value.trim()
-    this.leadingIconTarget.className = glyph
-      ? `${glyph} ${theme.icon} shrink-0 text-xl`
-      : "hidden"
+    const base = "w-6 shrink-0 text-center text-xl"
+    this.leadingIconTarget.className = glyph ? `${glyph} ${theme.icon} ${base}` : base
   }
 
   themeKey() {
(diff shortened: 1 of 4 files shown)


app/models/registration_ticket_callout.rb
```diff
@@ -20,7 +21,7 @@ class RegistrationTicketCallout < ApplicationRecord
   }.freeze
 
   DEFAULT_ICONS = { "action" => "fa-solid fa-arrow-right", "reference" => "fa-solid fa-circle-info" }.freeze
-  DEFAULT_COLORS = { "action" => "blue", "reference" => "indigo" }.freeze
+  DEFAULT_COLORS = { "action" => "orange", "reference" => "indigo" }.freeze
```
maebeale, line 24: 🤖 _From Claude:_ Switched the action default from blue to orange so admin-added action call-outs don't blend into the built-in blue download/invoice rows already on the ticket. Blue stays available as a custom override — only the default moved.

app/frontend/javascript/controllers/callout_preview_controller.js
```diff
@@ -0,0 +1,52 @@
...
+
+  connect() {
+    this.element.classList.remove("bg-white", "border-gray-200")
+    if (this.hasArrowTarget) this.arrowTarget.classList.remove("text-gray-300")
+    this.applied = []
+    this.update()
+  }
+
+  update() {
+    const theme = this.themesValue[this.themeKey()] || this.themesValue["indigo"]
```
maebeale, line 20: 🤖 _From Claude:_ The controller swaps Tailwind colour classes at runtime; these are all already in the `@source inline(...)` safelist (used by the public ticket), so JIT picks them up without a CSS rebuild. It reads the theme map straight from the model via data values so JS and Ruby never drift.

PR #1728, opened 2026-06-18
Fix upside-down chevrons on the event dashboard cards

maebeale:
- The collapsible cards on the event dashboard (money cards + headcount cards) showed chevrons that flipped upside down when a card was expanded.
- The money cards contain nested `<details>` rows, so opening an outer card incorrectly rotated the collapsed inner-row chevrons too.

- The chevrons used `[[open]_&]:rotate-180`, which Tailwind compiles to `[open] .…rotate-180` — matching on *any* ancestor with `[open]`, not the chevron's own `<details>`.
- Replaced it with named Tailwind groups: each card `<details>` is `group/card` (chevron `group-open/card:rotate-180`) and each money row `<details>` is `group/row` (chevron `group-open/row:rotate-180`), so every chevron only rotates when its own section is open.
- Verified by compiling the Tailwind v4 CSS and confirming the generated selectors are scoped to each `.group/card` / `.group/row` `[open]` element.

### UI Testing Checklist
- [ ] On the event dashboard, each card's chevron points down when collapsed and up when expanded.
- [ ] Expanding a money card (Registration fees / Scholarships / CE fees) does NOT flip its still-collapsed Paid/Due/Scholarship row chevrons.
- [ ] Expanding an inner money row rotates only that row's chevron.

### Anything else to add?
- The same `[[open]_&]` pattern exists in `forms/edit.html.erb` and `event_registrations/link_organization.html.erb`, but those use flat, non-nested `<details>`, so they don't hit this bug and are left unchanged.

### DIFF
--- a/app/views/events/_dashboard_money_row.html.erb
+++ b/app/views/events/_dashboard_money_row.html.erb
@@ -22,7 +22,7 @@
   amounts_by_registrant_id = local_assigns.fetch(:amounts_by_registrant_id, nil)
 %>
 <div class="flex items-start gap-1.5">
-  <details class="flex-1 min-w-0">
+  <details class="group/row flex-1 min-w-0">
     <summary class="flex items-center justify-between gap-2 rounded px-1 -mx-1 py-1 cursor-pointer select-none list-none [&::-webkit-details-marker]:hidden hover:bg-gray-50">
       <span class="flex items-center gap-1.5 text-xs font-medium text-gray-500">
         <% if icon_bg.present? %>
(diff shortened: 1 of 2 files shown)


app/views/events/_dashboard_money_row.html.erb
```diff
@@ -37,7 +37,7 @@
       </span>
       <span class="flex items-center gap-2">
         <span class="text-sm font-semibold tabular-nums <%= amount_color %>"><%= number_to_currency(amount_cents / 100.0) %></span>
-        <i class="fa-solid fa-chevron-down text-[10px] text-gray-300 transition-transform [[open]_&]:rotate-180"></i>
+        <i class="fa-solid fa-chevron-down text-[10px] text-gray-300 transition-transform group-open/row:rotate-180"></i>
```
maebeale, line 40: 🤖 _From Claude:_ Named groups (`group/card` vs `group/row`) are required here, not plain `group`: these `<details>` nest, and an unnamed group would let the outer card re-trigger the inner row chevrons — the exact bug this fixes.

PR #1731, opened 2026-06-18
Rename "System notifications" to "Notifications"

maebeale:
- The admin UI labeled these "System notifications," but the "System" qualifier added no meaning beyond the `Notification` model name.
- This relabels the UI to simply "Notifications," matching the model and reducing redundant copy.

- Updated the two hard-coded headings in the notifications index and show views.
- Replaced the hard-coded dashboard `custom_card("System notifications", notifications_path, …)` with `model_card(:notifications, …)`, so the card's title and path are derived from the model (styling is unchanged — it falls back to the same gray theme).
- Renamed a stale `it "creates a system notification"` test label to match.

### UI Testing Checklist
- [ ] Admin dashboard shows a "Notifications" card linking to `/notifications`.
- [ ] Notifications index heading reads "Notifications".
- [ ] A notification detail page heading reads "Notification".

### Anything else to add?
- No locale/en.yml entries existed for these strings, so nothing to remove there. `DomainTheme` has no `:notifications` key, so the card keeps its existing gray fallback styling.

### DIFF
--- a/app/helpers/admin_cards_helper.rb
+++ b/app/helpers/admin_cards_helper.rb
@@ -33,7 +33,7 @@ module AdminCardsHelper
       model_card(:quotes, icon: "💬", intensity: 100),
       model_card(:payments, icon: "💳"),
       custom_card("Scholarships", grants_path, icon: "🎓", color: :fuchsia),
-      custom_card("System notifications", notifications_path, icon: "🔔"),
+      model_card(:notifications, icon: "🔔"),
       model_card(:story_ideas, icon: "✍🏾️", intensity: 100),
       custom_card("Tags", tags_path, icon: "🏷️", color: :lime, intensity: 100),
       model_card(:workshop_ideas, icon: "💡", intensity: 100),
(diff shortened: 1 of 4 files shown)


app/helpers/admin_cards_helper.rb
```diff
@@ -29,7 +29,7 @@ def user_content_cards
       custom_card("Bookmarks tally", tally_bookmarks_path, icon: "🔖"),
       model_card(:quotes, icon: "💬", intensity: 100),
       model_card(:event_registrations, icon: "🎟️", intensity: 100),
-      custom_card("System notifications", notifications_path, icon: "🔔"),
+      model_card(:notifications, icon: "🔔"),
```
maebeale, line 36: 🤖 _From Claude:_ Switched from `custom_card` to `model_card(:notifications)` so the title and path derive from the model. Styling is unchanged — `DomainTheme` has no `:notifications` key, so it falls back to the same gray-50 the custom card used.

maebeale, line 36: 🤖 _From Claude:_ Switched from `custom_card` to `model_card(:notifications)` so the title and path derive from the model. Styling is unchanged — `DomainTheme` has no `:notifications` key, so it falls back to the same gray-50 the custom card used.

PR #1734, opened 2026-06-18
Filterable recipient picker and confirm step for bulk event reminders

maebeale:
- Admins sending bulk event reminders couldn't target a subset of registrants or review exactly who would be emailed before sending — they had to trust a checkbox list and a browser confirm dialog.
- This adds a filterable recipient picker and a confirmation interstitial so admins can precisely scope and verify a reminder send.

- Added a `ReminderRecipientFilter` service (matched in memory) that filters the recipient list by name, organization, grantor, comment, and payment/scholarship/CE/account status; filters keep everyone visible and only (un)check who matches, rendered into a turbo frame.
- Added a `confirm_reminder` interstitial that lists the selected recipients (comma-separated + full list) and shows the composed subject/body before the final send, with `send_reminder` and confirm sharing one `selected_reminder_registrations` helper.
- Showed each registrant's connected organizations (comma-separated), widened the page to fit the recipient table, and replaced the `ORGANIZATION_NAME` env indirection with the fixed "AWBW Portal" reminder subject.

### UI Testing Checklist
- [ ] Open an event's "Send reminder emails" page; confirm the recipient table shows connected organizations and the filters check/uncheck matching registrants.
- [ ] Apply each filter (name, organization, grantor, comment, payment/scholarship/CE/account status) and verify only matching rows stay checked.
- [ ] Proceed to the confirm step and verify the recipient list and rendered subject/body match the selection before sending.

### Anything else to add?
- The email-preview `raw` on the confirm page is whitelisted in `config/brakeman.ignore` (same accepted pattern as `preview_reminder`); the embedded custom message is sanitized via `reminder_message_html` and the subject is shown escaped.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -2,7 +2,7 @@ class EventsController < ApplicationController
   include AhoyTracking, TagAssignable
   skip_before_action :authenticate_user!, only: [ :index, :show, :staff, :details, :ce_hours ]
   skip_before_action :verify_authenticity_token, only: [ :preview ]
-  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
+  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder confirm_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
 
   def index
     authorize!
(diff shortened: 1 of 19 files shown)


app/services/reminder_recipient_filter.rb
```diff
@@ -0,0 +1,134 @@
...
+    account_status ce_status
+  ].freeze
+
+  def initialize(event_registrations, params)
+    @event_registrations = event_registrations
+    @params = params
+  end
+
+  def matched_ids
+    @event_registrations.select { |reg| matches?(reg) }.map(&:id).to_set
```
maebeale, line 20: 🤖 _From Claude:_ Design note: filters never remove rows — they return the matching ids and the view only (un)checks them, so an admin can hand-tweak the selection after filtering. Matching runs in memory against the already-loaded set, which is why the controller eager-loads orgs/comments/scholarships/grants here.

app/models/event_registration.rb
```diff
@@ -257,6 +257,19 @@ def joinable?
     active? && payment_access_granted? && event.videoconference_window_open?
   end
 
+  # Bucket the registrant's login account into one of four states for the bulk
+  # reminder filters. Precedence matters: a confirmed, unlocked, active account
+  # has access regardless of when it was invited; a still-pending account counts
+  # as "invited" only while it hasn't gained access. Returns one of
+  # "none", "has_access", "invited", "no_access".
+  def account_status
+    account = registrant&.user
```
maebeale, line 266: 🤖 _From Claude:_ Precedence is deliberate: a confirmed/unlocked/active account counts as "has access" regardless of when it was invited; "invited" only applies while it has not yet gained access.

app/controllers/events_controller.rb
```diff
@@ -416,6 +446,17 @@ def copy_registration_form
 
   private
 
+  # The registrations the admin checked on the recipient picker, narrowed to those
+  # we can actually email. Shared by the confirm interstitial and the send action
+  # so both operate on exactly the same set.
+  def selected_reminder_registrations
+    allowed_ids = Array(params[:registration_ids]).map(&:to_i).reject(&:zero?)
```
maebeale, line 453: 🤖 _From Claude:_ Extracted so the confirm interstitial and the actual send operate on exactly the same set (checked ids ∩ emailable), avoiding drift between what is previewed and what is sent.

PR #1736, opened 2026-06-18
Add intends-to-pay chip to reminder recipients and unify due-amount formatting

maebeale:
- On the bulk reminder-email recipient table, admins couldn't see whether an unpaid registrant intends to pay, and amounts always rendered trailing `.00` cents.
- They also had no quick way to inspect/fix a registration from that page without losing their in-progress message draft.

- Added an icon-only "intends to pay" chip (hover tooltip) beside the due amount on the reminders recipient table, moved email inline with the name, and added a registration button.
- The registration button opens in a **new tab** (`target=_blank`/`rel=noopener`) so the picker — including the subject/message draft, which lives only in the DOM and is too long to round-trip through URL params — stays fully intact. A `return_to: preview_reminder` eyebrow gives the new tab a sensible "Send reminders" back link.
- Unified all due-amount chips (reminders, registrants, ticket, bulk payments) to use `dollars_from_cents` so cents only show when non-zero, and standardized the registration-button icon to `fa-clipboard-list`.
- Seeded Maria and Anna on event 1 with a partial balance and `intends_to_pay` so the new chip has dev data.

### UI Testing Checklist
- [ ] Reminders recipient table shows the intends-to-pay chip after the due amount for unpaid registrants who intend to pay.
- [ ] Due amounts show whole dollars (e.g. `$100 due`) but keep cents when present (e.g. `$99.50 due`).
- [ ] The registration button opens the registration editor in a new tab; the picker tab keeps any typed subject/message.

### Anything else to add?
- Because the editor opens in a new tab, payment-status chips on the original picker tab won't reflect an edit until the list is re-filtered (it's a turbo frame). Happy to add a small "refresh" affordance if useful.

### DIFF
--- a/app/helpers/event_helper.rb
+++ b/app/helpers/event_helper.rb
@@ -71,21 +71,6 @@ module EventHelper
     )
   end
 
-  # Recipient filters the send-reminders page had active, sliced to the known
-  # filter keys so we only ever echo recognized params. Used to carry the filters
-  # into a registration edit (nested under `reminder_filters`) and to re-apply
-  # them when returning. See ReminderRecipientFilter and app/views/events/_reminder_recipients.html.erb.
-  def reminder_filter_params(source = params[:reminder_filters])
-    return {} if source.blank?
-    source.permit(*ReminderRecipientFilter::FILTER_KEYS).to_h.compact_blank
-  end
-
-  # Back link to the send-reminders page, re-applying whatever recipient filters
-  # were active when the admin opened a registration from it.
-  def preview_reminder_return_path(event)
-    preview_reminder_event_path(event, reminder_filter_params)
-  end
-
   def event_profile_button(event, truncate_at: nil, subtitle: nil, data: {}, path: nil)
     bg = DomainTheme.bg_class_for(:events, intensity: 100)
     hover_bg = DomainTheme.bg_class_for(:events, intensity: 100, hover: true)
--- a/app/helpers/event_helper.rb
+++ b/app/helpers/event_helper.rb
@@ -71,6 +71,21 @@ module EventHelper
     )
   end
 
+  # Recipient filters the send-reminders page had active, sliced to the known
+  # filter keys so we only ever echo recognized params. Used to carry the filters
+  # into a registration edit (nested under `reminder_filters`) and to re-apply
+  # them when returning. See ReminderRecipientFilter and app/views/events/_reminder_recipients.html.erb.
+  def reminder_filter_params(source = params[:reminder_filters])
+    return {} if source.blank?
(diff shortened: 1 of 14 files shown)


app/helpers/event_helper.rb
```diff
@@ -71,6 +71,21 @@ def bulk_payments_return_path(event)
     )
   end
 
+  # Recipient filters the send-reminders page had active, sliced to the known
+  # filter keys so we only ever echo recognized params. Used to carry the filters
+  # into a registration edit (nested under `reminder_filters`) and to re-apply
+  # them when returning. See ReminderRecipientFilter and app/views/events/_reminder_recipients.html.erb.
+  def reminder_filter_params(source = params[:reminder_filters])
```
maebeale: 🤖 _From Claude:_ Slicing to `ReminderRecipientFilter::FILTER_KEYS` (not echoing raw params) keeps this from reflecting arbitrary query params back into the redirect, and means the picker and the return path always agree on which keys count as filters.

app/views/event_registrations/_form.html.erb
```diff
@@ -1,5 +1,9 @@
 <%= simple_form_for(event_registration, data: { turbo: false }) do |f| %>
   <%= hidden_field_tag :return_to, params[:return_to] if params[:return_to].present? %>
+  <%# Carry the send-reminders recipient filters so the page re-applies them on return. %>
+  <% reminder_filter_params.each do |key, value| %>
+    <%= hidden_field_tag "reminder_filters[#{key}]", value %>
```
maebeale: 🤖 _From Claude:_ Filters ride through the edit as hidden fields so the post-save redirect (which only sees the submitted form) can re-apply them — same mechanism as the existing `return_to` hidden field just above.

app/views/events/_reminder_recipients.html.erb
```diff
@@ -26,24 +26,38 @@
...
                 <% end %>
               <% else %>
                 <span class="text-gray-400">—</span>
               <% end %>
             </td>
+            <td class="px-4 py-2 text-right">
+              <%# Opens in a new tab so the in-progress subject/message draft on this page isn't lost. %>
+              <%= link_to edit_event_registration_path(reg, return_to: "preview_reminder"),
+                          title: "View #{person.full_name}'s registration for #{@event.title} (opens in a new tab)",
+                          target: "_blank", rel: "noopener",
```
maebeale, line 56: 🤖 _From Claude:_ New tab is deliberate: the subject/message draft is DOM-only state and too long to round-trip via URL params, so a same-tab navigation would lose it. The tab keeps the picker intact; `return_to` just gives the new tab a sensible back link.

PR #1740, opened 2026-06-18
Add facilitator_training flag to events

maebeale:
- Events that train facilitators (the "TAC" a scholarship recipient attends) need a first-class boolean so downstream features can identify them reliably — without fragile title-string matching.
- The immediate consumer is the scholarship index's training column (built on a separate branch); this PR lands the field on its own so it can merge independently and that branch picks it up on rebase.

- Added `events.facilitator_training` boolean (default `false`, indexed) via a reversible migration, plus an `Event.facilitator_trainings` scope.
- Permitted the attribute in `EventPolicy` and exposed it as a **"This event is a facilitator training"** checkbox in the Event title box, under Pre-title — parallel to the "Show title on event page" checkbox under Title.
- Restructured the **Event title** box into a two-column grid (Title + "Show title on event page" left, Pre-title right) and trimmed the trailing margin under the checkbox (`mb-0!`) so the box height matches the Display-settings box.
- Defaulted the flag **on** for dev-seed events whose title matches `/training/i`, set on every re-seed for determinism.
- Added a request spec covering the permitted attribute.

### UI Testing Checklist
- [ ] Event title box and Display settings box are the same height
- [ ] "Show title on event page" still toggles `autoshow_title`
- [ ] New "This event is a facilitator training (TAC)" checkbox saves correctly
- [ ] Dev seed marks both Facilitator Trainings as trainings, nothing else

### Anything else to add?
- Extracted from the scholarships branch (biarritz). That branch keeps its consumers (`Person#completed_facilitator_trainings`, scholarship decorator/seeds/specs); they reference this column and will resolve once this merges and that branch rebases onto main.
- Padding fix uses the Tailwind v4 important suffix (`mb-0!`) because the simple_form boolean wrapper's base `mb-4` otherwise wins the cascade.

### DIFF
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -8,8 +8,10 @@
   <div class="grid grid-cols-1 md:grid-cols-2 gap-6 mb-12">
     <div>
       <label class="block font-medium mb-1 text-gray-700">Event title</label>
-      <div class="border border-gray-200 rounded-lg grid grid-cols-1 md:grid-cols-2 gap-4 p-4">
-        <div>
+      <div class="border border-gray-200 rounded-lg p-4">
+        <%# Inputs and checkboxes are split into two parallel grid rows so each %>
+        <%# checkbox lines up with its sibling regardless of the Pre-title hint height %>
+        <div class="grid grid-cols-1 md:grid-cols-2 gap-4">
           <%= f.input :title,
                   label: "Title",
                   required: true,
@@ -21,9 +23,6 @@
                       "w-full rounded border-gray-300 shadow-sm px-3 py-2
                                                 focus:ring-blue-500 focus:border-blue-500",
                   } %>
-          <%= f.input :autoshow_title, as: :boolean, label: "Show title on event page", wrapper_html: { class: "mb-0!" } %>
-        </div>
-        <div>
           <%= f.input :pre_title,
                   label: "Pre-title",
                   hint: "Displays above the title",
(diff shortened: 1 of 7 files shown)


app/views/events/_form.html.erb
```diff
@@ -21,8 +21,8 @@
                       "w-full rounded border-gray-300 shadow-sm px-3 py-2
                                                 focus:ring-blue-500 focus:border-blue-500",
                   } %>
+          <%= f.input :autoshow_title, as: :boolean, label: "Show title on event page", wrapper_html: { class: "mb-0!" } %>
```
maebeale: 🤖 _From Claude:_ `mb-0!` (Tailwind v4 important suffix) is deliberate — the simple_form boolean wrapper ships a base `mb-4`, and a plain `mb-0` loses the cascade (equal specificity, `mb-4` emitted later). The `!` is what trims the trailing space so this box matches the Display-settings box height.

db/seeds/dev/events_management.rb
```diff
@@ -223,7 +223,7 @@
   # find_or_create_by! only sets these on create, so without this an existing DB
   # keeps stale dates/cost and neither the index ordering, the multi-day span, nor
   # the registration fee would update.
-  event.update!(start_date: start_date, end_date: end_date, registration_close_date: registration_close, cost_cents: cost_cents)
+  event.update!(start_date: start_date, end_date: end_date, registration_close_date: registration_close, cost_cents: cost_cents, facilitator_training: title.match?(/training/i))
```
maebeale, line 226: 🤖 _From Claude:_ Set on the unconditional `update!` (not the `find_or_create_by!` block) so re-seeding an existing dev DB flips the flag on/off to match the current title — same pattern as the dates/cost refresh just above.

PR #1741, opened 2026-06-18
Tidy event form title/display area and rename page-content toggle

maebeale:
- Clarifies and tidies the event edit form's title/display area so admins can tell at a glance what each toggle controls.

- Renamed the "Page content" collapsible toggle to "Event show page" and moved the "Show title on event show page" boolean into it (next to header/description), since it governs the show page.
- Moved the new `facilitator_training` boolean directly under the Title field in the left column so it sits closer to the related inputs.
- Removed the flex stretch on the "Display settings" box so the blue box renders at its natural height rather than stretching to match the white box.

### UI Testing Checklist
- [ ] On the event edit form, confirm the "Show title on event show page" toggle now appears inside the expanded "Event show page" section.
- [ ] Confirm the "Facilitator training event" checkbox sits under the Title field, and the "Display settings" box height looks right.

### Anything else to add?
- View-only change to `app/views/events/_form.html.erb`; rebased on the `facilitator_training` flag from #1740.

### DIFF
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -9,20 +9,21 @@
     <div>
       <label class="block font-medium mb-1 text-gray-700">Event title</label>
       <div class="border border-gray-200 rounded-lg p-4">
-        <%# Inputs and checkboxes are split into two parallel grid rows so each %>
-        <%# checkbox lines up with its sibling regardless of the Pre-title hint height %>
         <div class="grid grid-cols-1 md:grid-cols-2 gap-4">
-          <%= f.input :title,
-                  label: "Title",
-                  required: true,
-                  as: :text,
-                  input_html: {
-                    rows: 1,
-                    value: f.object.title,
-                    class:
-                      "w-full rounded border-gray-300 shadow-sm px-3 py-2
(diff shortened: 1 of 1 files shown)


app/views/events/_form.html.erb
```diff
@@ -31,17 +32,13 @@
...
-        </div>
       </div>
     </div>
 
     <% if allowed_to?(:manage?, @event) %>
-      <div class="admin-only mt-6 md:mt-0 flex flex-col">
+      <div class="admin-only mt-6 md:mt-0">
         <label id="display_settings" class="block font-medium mb-1 text-gray-700">Display settings</label>
-        <div class="bg-blue-50 border border-gray-200 rounded-lg grid grid-cols-2 content-start gap-y-0 p-3 flex-1 [&_.form-group]:mb-1">
+        <div class="bg-blue-50 border border-gray-200 rounded-lg grid grid-cols-2 content-start gap-y-0 p-3 [&_.form-group]:mb-1">
```
maebeale, line 41: 🤖 _From Claude:_ Dropped the `flex flex-col`/`flex-1` stretch here so the Display settings box renders at its natural height instead of expanding to match the white box.

PR #1742, opened 2026-06-18
Stop marking additional service areas as primary on registration

maebeale:
- Registrants were ending up with **multiple primary service-area sectors**, which violates the "one primary" rule and crowns every sector on the person/recipients profile.
- Root cause: public registration fed both the single-select "Primary service area" dropdown (`primary_service_area_single`) **and** the multi-select "Additional service areas" checkboxes (`primary_service_area`) into one assignment that set `is_primary: true` on all of them.

- Only the dropdown selection is marked primary; the checkbox selections become additional (`is_primary: false`).
- A sector named in both is kept as a single primary row (no duplicate), and any previously-primary sector is demoted so exactly one primary survives.
- Renamed `assign_primary_sectors` → `assign_sectors` to reflect that it now assigns both primary and additional.

### UI Testing Checklist
- [ ] Register via the public form picking one primary + several additional service areas; confirm the person shows exactly one starred/primary sector and the rest as additional.
- [ ] Re-register naming a different primary; confirm the old primary demotes to additional.

### Anything else to add?
- The existing specs encoded the bug (submitted the *additional* field and asserted `is_primary: true`); rewritten to assert correct behavior, plus added cases for one-primary, no-duplicate, and demotion. Full service spec green (23 examples).
- Per-person uniqueness (sector-to-person, category-to-person) is already enforced by DB unique indexes, so a sector can't be both primary and additional for one person — this fix addresses the *primary-vs-additional* mislabeling that produced the duplicate primaries.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -227,15 +227,18 @@ module EventRegistrationServices
     end
 
     def assign_tags(person, organization)
-      sector_ids = collect_ids_from_checkboxes("primary_service_area_single") +
-                   collect_ids_from_checkboxes("primary_service_area")
+      # The single-select dropdown names the ONE primary service area; the
+      # checkbox field collects ADDITIONAL areas. A sector named in both is
+      # primary, never duplicated.
+      primary_sector_id = collect_ids_from_checkboxes("primary_service_area_single").first
+      sector_ids = (Array(primary_sector_id) + collect_ids_from_checkboxes("primary_service_area")).uniq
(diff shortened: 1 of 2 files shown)


maebeale:
🤖 _From Claude:_ Closing as superseded by #1834 (merged 2026-06-22), which fixes the same root cause more thoroughly: it removed `assign_primary_sectors` entirely, added `SectorsTaggable#tag_sectors` enforcing the single-primary rule (with the same demote-stale-primaries logic this PR introduced), plus a model-level `at_most_one_primary_sector` validation and disentangled field identifiers. Nothing here remains to merge.

PR #1747, opened 2026-06-18
Age ranges: single primary like sectors + visibility toggle

maebeale:
- Builds on #1717's primary/additional age groups, reshaping them to match **sectors** exactly and adding a per-profile **visibility toggle**.
- On the registration form a respondent now picks **one** primary age range from a dropdown and **many** additional ones from a multi-select — same shape as the primary/additional sector questions.

**Registration form**
- "Primary age range" is now a **single-select dropdown** (one primary); "Additional ages served" stays **multi-select**.
- A data migration converts existing forms' `primary_age_group` field to a dropdown and renames both questions.

**Single primary (sector parity)**
- `AgeGroupTaggable#tag_age_groups` and `#apply_primary_age_groups!` keep exactly one primary — choosing a new primary demotes the previous one — backed by an `at_most_one_primary_age_group` validation.
- A migration normalizes any existing data that marked several primaries down to one.
- Edit forms keep a single "Primary" toggle checked at a time via a new `single-select-checkbox` Stimulus controller (allows zero or one, like the star toggle on sectors).

**Visibility toggle**
- `profile_show_age_ranges` on people + organizations gates the "Age ranges served" section, mirroring `profile_show_sectors`. Added a "Show age ranges" checkbox to both edit forms.

**Reporting**
- The event dashboard age breakdown now reads **both** the primary and additional age-range answers (previously only the primary question), so the chart reflects every age range served.

### Anything else to add?
- This branch was previously a parallel implementation of #1717; it has been rebased onto main and reduced to just the deltas above (single-primary + visibility toggle + dashboard fix). Everything else from the old version was superseded by #1717.
- Specs: updated `AgeGroupTaggable` spec (single-primary, demote, validation), added a builder spec for the new age field types. Full affected suite green; rubocop clean; dev seed + Vite dev build run.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ This codebase (Rails 8.1)
 | Directory | Purpose |
 |---|---|
 | `app/frontend/entrypoints/` | Vite entry points (application.js, application.css) |
-| `app/frontend/javascript/controllers/` | Stimulus controllers (71) |
+| `app/frontend/javascript/controllers/` | Stimulus controllers (74) |
 | `app/frontend/javascript/rhino/` | Rich text editor customizations (mentions, grid) |
 | `app/frontend/stylesheets/` | Tailwind CSS and component styles |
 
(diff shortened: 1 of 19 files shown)


maebeale:
🤖 _From Claude:_ Rebased onto latest main (33 commits). Conflicts resolved: `form_builder_service.rb` (main renamed `hint:` → `subtitle:` in `add_field`, so the new age-range fields use `subtitle:`), `db/schema.rb` version line, and the AGENTS.md controller count.

Heads-up on the backfill migration (`20260617150000`): it's backdated relative to main's `hint_text` → `subtitle` rename (`20260618014554`), so it runs *before* the rename on a fresh in-order DB but *after* it on production (where the rename is already applied). It detects `column_exists?(:form_fields, :subtitle)` and writes to whichever column exists, so the descriptive text lands correctly in both orderings. On a fresh DB there are no forms yet, so the loop is a no-op anyway.

PR #1748, opened 2026-06-18
Flatten event location searchable-select to match sibling inputs

maebeale:

- The location searchable-select on the event edit form rendered as a **box-within-a-box** with a filled background, looking inconsistent with the neighboring text inputs and visually reading as a required/distinct field when it's optional.

- Root cause: Tom Select copies the original `<select>`'s utility classes onto its generated `.ts-wrapper`, then renders its own bordered, filled `.ts-control` inside it — two stacked boxes.
- Added a `ts-flat` marker class to the location select so the change is **scoped to the event form** (the other 7 searchable-selects keep their current look).
- Added a `.ts-flat` Tom Select variant in `application.tailwind.css` that makes the inner control transparent and borderless, leaving the single wrapper box; matched its content height to the sibling `px-3 py-2` text inputs and kept a blue focus ring for keyboard focus.

### UI Testing Checklist

- [ ] Event edit form: location field is a single, transparent box with no inner border
- [ ] Location field is the same height as the Videoconference URL input beside it
- [ ] Keyboard-focusing the location field shows a visible blue focus ring
- [ ] Search/select still works and the chosen location persists on save

### Anything else to add?

- Scoped to the event form via the `ts-flat` class; happy to make it the default for all searchable-selects if we want the flat look everywhere.

### DIFF
--- a/app/frontend/stylesheets/application.tailwind.css
+++ b/app/frontend/stylesheets/application.tailwind.css
@@ -94,6 +94,34 @@
   @apply bg-gray-100;
 }
 
+/* Tom Select "flat" variant: the wrapper inherits the field's bordered box,
+   so the inner control is transparent and borderless — no box-within-a-box.
+   Used for optional searchable-selects (e.g. the event location) so the field
+   reads as not-required. */
+.ts-wrapper.ts-flat {
+  @apply bg-transparent;
+}
+
+.ts-wrapper.ts-flat .ts-control,
+.ts-wrapper.ts-flat.focus .ts-control {
+  background: transparent;
+  border: 0;
+  box-shadow: none;
+  padding: 0;
+  min-height: 1.5rem;
+}
+
+/* Match the height of sibling text inputs (py-2 + 1.5rem line-height) */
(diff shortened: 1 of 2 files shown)


app/frontend/stylesheets/application.tailwind.css
```diff
@@ -94,6 +94,34 @@
...
+.ts-wrapper.ts-flat .ts-control,
+.ts-wrapper.ts-flat .ts-control input {
+  line-height: 1.5rem;
+  font-size: 1rem;
+}
+
+.ts-wrapper.ts-flat.focus {
+  @apply ring-1 ring-blue-500 border-blue-500;
+}
+
```
maebeale, line 124: 🤖 _From Claude:_ The field's `focus:ring/border` utilities sit on the original `<select>`, which Tom Select hides — so focus styling won't fire there. This rule reattaches a focus ring to the `.ts-wrapper.focus` class Tom Select toggles, preserving the keyboard focus indicator after we strip the inner control's border/shadow.

PR #1751, opened 2026-06-18
Refresh the registration ticket: thumbnail header, compact details, and status chips

maebeale:

- Tightens the registration ticket so registrants can scan the essentials and act, and consolidates payment/CE/scholarship status into the cards instead of one-off pills.
- Reclaims vertical space (the old centered banner image ate a full row) and gives the `MagicTicketCallouts` cards more prominence.

**Header**
- Event image → a small thumbnail inline with the title; title nudged to `text-2xl`.
- Date/times and `Cost · Location` moved up under the title as a compact three-line header (one-line date via the non-styled `times`); dropped the old centered Event Details block.
- `labelled_cost` now formats through `MoneyFormatter.dollars_from_cents`, so amounts carry a thousands separator (`$1,500`).

**Status chips (one `badge` / `badge_classes` on `MagicTicketCallouts::Card`, rendered in the subtitle row)**
- Each chip is themed to read against its own card, with amber reserved for "action needed", and carries a matching darker border:
  - **Payment** — `Make your payment` + amber `$X due` while due; `Payment` + blue `Paid` once settled.
  - **CE** — teal `$X due` once hours + license are on file; otherwise amber, naming what's missing: `$250 · License number needed`, `Hours needed`, or `Hours & license number needed` (amount prefixed when the hours/fee are known).
  - **Scholarship** — fuchsia award amount once tasks are met; amber `$X · Tasks outstanding` while tasks remain.
- All colour shades come from the existing `@source inline` Tailwind safelist (no rebuild needed).

**Payment page** (`events/callouts/payment.html.erb`)
- Added a **View invoice** card; tucked the pay-by-check details into a collapsed disclosure (keeping the grey card); chipped the amount-due total.

**Buttons**
- Fixed the brand-red pay/register buttons' hover: new `.btn-accent` variant carries its own border/bg so invert-on-hover works without an inline style overriding it. Applied on the ticket, the payment page, and the event-show register buttons.

### Testing

- `magic_ticket_callouts`, `event_registration_show` (system), `events/registrations` (request), and `event_decorator` specs all pass; Rubocop clean.

### Note

- Collaboration across parallel workspaces: the CE and scholarship amount chips landed separately and were unified onto the single `badge` mechanism here.
- **Open question:** the CE `Hours needed` badge can't show an amount (the fee derives from the hours, which are what's missing), so it has no `$` prefix — unlike `$250 · License number needed`.

### DIFF
--- a/app/views/event_registrations/_ticket.html.erb
+++ b/app/views/event_registrations/_ticket.html.erb
@@ -17,23 +17,21 @@
 
     <!-- Ticket Body -->
     <div class="bg-gray-50 px-6 py-6 space-y-5">
-      <!-- Event Image + Title -->
-      <div class="text-center space-y-4">
-        <div class="inline-block rounded-lg overflow-hidden border border-gray-200 shadow-sm">
-          <%= render "assets/display_image",
-                       resource: event_registration.event,
-                       width: 48, height: 32,
-                       variant: :index,
-                       link: event_path(event_registration.event, reg: event_registration.slug),
-                       file: event_registration.event.decorate.display_image %>
-        </div>
+      <!-- Event thumbnail + title -->
+      <div class="flex items-center justify-center gap-4">
+        <%= render "assets/display_image",
+                     resource: event_registration.event,
+                     width: 18, height: 18,
+                     variant: :index,
+                     link: event_path(event_registration.event, reg: event_registration.slug),
+                     file: event_registration.event.decorate.display_image %>
 
-        <div>
+        <div class="min-w-0 text-left">
           <% if event_registration.event.respond_to?(:pre_title) && event_registration.event.pre_title.present? %>
-            <p class="text-base font-semibold text-gray-700 mb-1" style="font-family: 'Telefon Bold', sans-serif"><%= event_registration.event.pre_title %></p>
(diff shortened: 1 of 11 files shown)


app/views/event_registrations/_ticket.html.erb
```diff
@@ -71,96 +69,114 @@
...
+                         variant: "reference",
+                         icon: "fa-file-pdf",
+                         arrow: "fa-download",
+                         title: "Download W-9",
+                         subtitle: "AWBW's W-9 tax form for your records" %>
+          <% end %>
+        </div>
       <% end %>
 
       <!-- Custom ticket callouts (admin-configured: actions & reference reading) -->
```
maebeale, line 110: 🤖 _From Claude:_ Rebase note — main's admin-configured ticket callouts (#1740-era) auto-merged here and now render as standalone cards alongside the new two-tier system. They use their own multi-color `callout.theme`, which is intentionally at odds with this PR's "two variants only, no third accent color" `_ticket_card` rule. Left as-is to preserve the feature, but worth deciding whether callouts should eventually route through `_ticket_card` (action/reference) or stay a separate styled system.

PR #1753, opened 2026-06-18
Rescue concurrent duplicate event registration inserts

maebeale:
- Fixes a production 500 (`ActiveRecord::RecordNotUnique`) when creating an event registration that collides with an existing one on `(registrant_id, event_id)`.
- The model's `uniqueness` validation runs as a separate `SELECT` before the `INSERT`, so a double-submit or two concurrent requests can both pass validation and have the second insert violate the DB unique index.

- Wrapped `EventRegistrationsController#create` in a `rescue ActiveRecord::RecordNotUnique` that treats the lost race exactly like a duplicate validation failure, redirecting back with a clear alert instead of surfacing a 500.
- Extracted the failure redirect into a `redirect_after_failed_create` helper so the validation and race paths share one `return_to`-aware redirect, and added a request spec that forces the race and asserts a graceful redirect.

### UI Testing Checklist
- [ ] Attempt to register the same person for the same event twice (or double-click submit) and confirm you see an alert instead of an error page.

### Anything else to add?
- No schema change; the existing unique index is what now gets handled gracefully at the controller layer.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -66,15 +66,14 @@ class EventRegistrationsController < ApplicationController
         }
       end
     else
-      respond_to do |format|
-        format.html {
-          case params[:return_to]
-          when "registrants" then redirect_to registrants_event_path(@event_registration.event), alert: @event_registration.errors.full_messages.to_sentence
-          else redirect_to event_registrations_path, alert: @event_registration.errors.full_messages.to_sentence
-          end
-        }
-      end
+      redirect_after_failed_create(@event_registration.errors.full_messages.to_sentence)
     end
+  rescue ActiveRecord::RecordNotUnique
+    # The uniqueness validation isn't atomic with the insert, so a concurrent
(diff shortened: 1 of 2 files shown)


app/controllers/event_registrations_controller.rb
```diff
@@ -66,15 +66,14 @@ def create
...
-        format.html {
-          case params[:return_to]
-          when "registrants" then redirect_to registrants_event_path(@event_registration.event), alert: @event_registration.errors.full_messages.to_sentence
-          else redirect_to event_registrations_path, alert: @event_registration.errors.full_messages.to_sentence
-          end
-        }
-      end
+      redirect_after_failed_create(@event_registration.errors.full_messages.to_sentence)
     end
+  rescue ActiveRecord::RecordNotUnique
```
maebeale, line 71: 🤖 _From Claude:_ The uniqueness validation isn't atomic with the insert, so this rescue is the real guard against duplicates under concurrency/double-submit — the validation just gives the common case a nicer message.

PR #1754, opened 2026-06-18
Fix UrlGenerationError on slugless bulk payment submissions

maebeale:
- Fixes a production `ActionController::UrlGenerationError` (`No route matches … slug: nil`) when viewing a bulk payment submission, and restores admin access to legacy submissions that have no slug (created before the `before_create :generate_slug` callback).

- `BulkPaymentsController#show` now mirrors `PublicRegistrations#show`: payers reach it publicly by slug (`?reg=`), while admins reach it by id (`?submission_id=`) — both render the same `form_submissions/submission` partial.
- The slug branch scopes with `where.not(slug: nil)` so a blank `reg` 404s instead of matching a slugless record and crashing; the id branch is gated by `FormSubmissionPolicy#show?` (admin only).
- The bulk-payments dashboard card links by slug when present, otherwise by `submission_id`; the view shows the "Back to ticket" eyebrow only when a slug exists.
- Added request specs covering the blank-reg 404, the admin id view of a slugless submission, and that signed-out viewers cannot reach a submission by id.

### Anything else to add?
- No schema or data migration; legacy slugless submissions stay as-is and are simply viewable again by admins.

### DIFF
--- a/app/controllers/events/bulk_payments_controller.rb
+++ b/app/controllers/events/bulk_payments_controller.rb
@@ -62,7 +62,10 @@ module Events
     def show
       authorize! :bulk_payment, to: :show?
 
-      @submission = FormSubmission.bulk_payment.find_by!(slug: params[:reg], event_id: @event.id)
+      # where.not(slug: nil) guards against a blank reg matching a legacy
+      # slugless bulk payment, which would then crash building its ticket link.
+      @submission = FormSubmission.bulk_payment.where.not(slug: nil)
+                                  .find_by!(slug: params[:reg], event_id: @event.id)
       @event = @event.decorate
     end
 
--- a/app/controllers/events/bulk_payments_controller.rb
+++ b/app/controllers/events/bulk_payments_controller.rb
@@ -56,16 +56,21 @@ module Events
       end
     end
 
-    # Public view of the submitted bulk payment form, reached by the payer (the
-    # "View submission" link in their confirmation email and on the ticket) and by
-    # admins from the dashboard. Backs to the ticket. Mirrors PublicRegistrations#show.
+    # View of the submitted bulk payment form, rendering the same partial either
+    # way. Mirrors PublicRegistrations#show: the payer reaches it publicly by slug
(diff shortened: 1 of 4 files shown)


app/controllers/events/bulk_payments_controller.rb
```diff
@@ -62,7 +62,10 @@ def create
     def show
       authorize! :bulk_payment, to: :show?
 
-      @submission = FormSubmission.bulk_payment.find_by!(slug: params[:reg], event_id: @event.id)
+      # where.not(slug: nil) guards against a blank reg matching a legacy
+      # slugless bulk payment, which would then crash building its ticket link.
```
maebeale: 🤖 _From Claude:_ The `where.not(slug: nil)` is what fixes the crash: without it, a blank `?reg=` matched a legacy slugless record via `find_by!(slug: nil)`, then the view's `bulk_payment_ticket_path(@submission.slug)` blew up on the nil slug. Now a blank reg 404s.

app/controllers/events/bulk_payments_controller.rb
```diff
@@ -56,13 +56,21 @@ def create
...
-      authorize! :bulk_payment, to: :show?
+      if params[:reg].present?
+        authorize! :bulk_payment, to: :show?
+        # where.not(slug: nil) keeps a blank reg from matching a slugless record.
+        @submission = FormSubmission.bulk_payment.where.not(slug: nil)
+                                    .find_by!(slug: params[:reg], event_id: @event.id)
+      else
+        @submission = FormSubmission.bulk_payment.find_by!(id: params[:submission_id], event_id: @event.id)
+        authorize! @submission, to: :show?
+      end
```
maebeale, line 72: 🤖 _From Claude:_ Mirrors `PublicRegistrations#show`: the public `?reg=` branch keeps the `where.not(slug: nil)` crash guard, while the `?submission_id=` branch is admin-only via `FormSubmissionPolicy#show?`. Both render the same submission partial, so legacy slugless submissions are viewable again without a data migration.

PR #1755, opened 2026-06-18
WAIT: Point incidental sector field-identifier references at the canonical names

maebeale:
🤖 PR, suggested 👤 review level: 👀 Skim — view-only: spec fixtures and comment copy, no logic or data changes

- #1834 made `additional_sectors` / `primary_sector_single` the canonical sector field identifiers, keeping the legacy `primary_sector` / `primary_service_area` names accepted via `FormField`'s `*_SECTOR_FIELD_IDENTIFIERS` constants.
- App code and seeds already moved to the canonical names; this cleans up the last incidental stragglers so nothing outside the legacy constants still reads as "service area".

- Repointed spec fixtures that just needed *a* sector field at the canonical identifiers, and fixed two stale "service areas" comments.
- Left the deliberate backward-compat tests (form_field_spec's parameterized legacy schemes, person_spec's `#other_sector_responses` block, event_dashboard_spec's legacy-resolution case) on the legacy names on purpose — they exist to prove the legacy entries in the constants still resolve.

### Anything else to add?
- No production logic or data changes (no migration). RuboCop clean; the affected specs (376 examples) pass.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -100,7 +100,7 @@ class FormField < ApplicationRecord
   # Keeps an over-long name as a friendly validation error instead of a
   # database ValueTooLong exception (the column is text, this is the UX cap).
   validates :name, length: { maximum: 1000 }
-  # A field_identifier wires a field to backend logic (Stripe, service areas,
+  # A field_identifier wires a field to backend logic (Stripe, sectors,
   # email checks, etc.), so the same identifier must not appear twice on one form
   # or that logic would target an ambiguous field. Ordinary fields carry no
   # identifier, so blanks are exempt — any number of them may coexist.
--- a/app/views/forms/_form_field_fields.html.erb
+++ b/app/views/forms/_form_field_fields.html.erb
@@ -151,7 +151,7 @@
         </div>
 
         <% if params[:admin] == "true" %>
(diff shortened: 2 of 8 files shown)


db/migrate/20260618191816_rename_service_area_field_identifiers_to_sector.rb
```diff
@@ -0,0 +1,22 @@
...
+  # about sectors (their options come from the Sectors admin), so the persisted
+  # FormField#field_identifier values are renamed to match the code constants.
+  # update_all is intended here: a one-shot bulk rename of existing records.
+  IDENTIFIER_RENAMES = {
+    "primary_service_area" => "primary_sector",
+    "primary_service_area_single" => "primary_sector_single"
+  }.freeze
+
+  def up
+    IDENTIFIER_RENAMES.each do |old_identifier, new_identifier|
```
maebeale: 🤖 _From Claude:_ `field_identifier` is a persisted, indexed column, so existing FormField records need this bulk rename — without it, forms already in the DB would keep the old identifier and silently lose their dynamic sector options. Reversible via `down`.

maebeale:
🤖 _From Claude:_ Closing — superseded on main; rebasing this branch onto main yields an empty diff. The sector field-identifier references now point at the canonical names via #2279/#2419.

PR #1703, opened 2026-06-17
Add sector descriptions and refresh the canonical sector list

maebeale:
### What is the goal of this PR and why is this important?
- Service areas (sectors) need a short clarification beyond their name, just like categories already have, so respondents understand options on the public registration form.
- The canonical sector list was also out of date and needed to match the current set of service-area tags.

### How did you approach the change?
- Added a `description` column to `Sector` (migration + admin form, show, and index, mirroring the category pattern) and surfaced it on the public registration form: as subtext under each checkbox in the additional service-areas list, and folded into a `Name (description)` label in the single primary service-area dropdown (which can't show subtext).
- Refreshed `SECTOR_TYPES` to the current tags, moved the parenthetical clarifications into seeded descriptions, and made seeding unpublish sectors no longer on the list (preserving historical taggings) while keeping the `Other` catch-all published.

### UI Testing Checklist
- [ ] Sector admin: create/edit shows and saves the Description field; show and index display it.
- [ ] Public registration form, additional service areas (checkboxes): descriptions render as subtext.
- [ ] Public registration form, primary service area (dropdown): options render as `Name (description)`.

### Anything else to add?
- Two apparent typos in the supplied list were corrected: tag `Systems/Policy Change` (was "Sytems") and the Climate/Environmental description `environmental trauma` (was "envirnomental"). Flagging in case either was intentional.

PR #1756, opened 2026-06-18
Migrate sector taggings onto the canonical sector list

maebeale:

- PR #1703 refreshed `Sector::SECTOR_TYPES` to the new canonical service-area list, seeded descriptions, and made seeds **create the new names while unpublishing (not destroying) the legacy ones**.
- It deliberately did **not** migrate taggings. So in production every existing workshop/story/video/etc. tagging is stranded on a now-**unpublished** legacy sector (e.g. `Child Abuse`, `Criminal/Legal`, `Student`), while the new published sectors (`Child Abuse/Neglect`, `Court/Legal System`, `Education`) are empty.
- This task closes that gap by re-homing the taggings, so the published canonical sectors carry the real data.

- Added `lib/tasks/consolidate_sectors.rake` (`data:consolidate_sectors`), **dry-run by default**; `DRY_RUN=false` to apply.
- **Rename when it's a 1:1 name change** (9 sectors, e.g. `Criminal/Legal → Court/Legal System`): in-place `update!`, which keeps the same row and all its taggings — nothing is removed.
- **Merge (via `ModelDeduper`) only when two rows must collapse into one:**
  - `Student` **and** `Education/Schools` both map to `Education` — a genuine 2:1 merge.
  - If the canonical name already exists (i.e. #1703 seeds ran first), a rename would collide, so it falls back to a merge onto the seeded published record.
  - The merge re-points all taggings onto the kept sector, then removes the redundant empty row. Taggings are never lost — this is the same merge the admin dedupe screen performs.
- **Idempotent:** a second run is a clean no-op.
- Verified end-to-end against the dev DB: 21 → 31 sectors, `Student`'s taggings folded into `Education` with no loss, second run skipped all steps. Rubocop clean.

### Anything else to add?

- No taggings are ever destroyed. Simple name changes keep their original `sectors` row (rename); only a genuine 2:1 merge removes a now-empty redundant row, which is the standard behavior of the existing `ModelDeduper` dedupe code.

### DIFF
--- a/lib/tasks/consolidate_sectors.rake
+++ b/lib/tasks/consolidate_sectors.rake
@@ -1,0 +1,150 @@
+# frozen_string_literal: true
+
+# Migrates sector TAGGINGS from the legacy service-area names onto the canonical
+# SECTOR_TYPES list. PR #1703 already refreshed SECTOR_TYPES, seeded descriptions,
+# and made seeds create the new names while *unpublishing* (not destroying) the
+# old ones — but it deliberately did not move taggings, so every existing
+# workshop/story/etc. tagging is stranded on a now-unpublished legacy sector while
+# the new published sectors are empty. This task closes that gap.
+#
+# It is deploy-order-agnostic by design:
+#   - Run before #1703 seeds (DB still on the old names): legacy names are renamed
+#     in place, preserving taggings, and Student is merged into Education.
+#   - Run after #1703 seeds (new names already exist): each rename hits the
+#     collision check and falls back to a merge, re-pointing taggings onto the
+#     seeded published record and destroying the empty legacy one.
+namespace :data do
+  desc "Migrate sector taggings from legacy names onto the canonical SECTOR_TYPES (rename in place or merge via ModelDeduper). DRY_RUN=false to apply."
+  task consolidate_sectors: :environment do
+    dry_run = ENV.fetch("DRY_RUN", "true") != "false"
+
+    # Current name => new name. Renames are done in place with update!, which
+    # preserves every tagging because sectorable_items reference sector_id, not
+    # the name string.
+    renames = {
+      "Child Abuse" => "Child Abuse/Neglect",
+      "Community Oppression/Violence" => "Community Violence",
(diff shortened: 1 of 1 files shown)


lib/tasks/consolidate_sectors.rake
```diff
@@ -0,0 +1,150 @@
...
+        skipped << "rename #{from.inspect} -> #{to.inspect}: source not found (already renamed?)"
+        next
+      end
+
+      existing_target = find_by_name.call(to)
+      if existing_target && existing_target.id != source.id
+        # The target name is already taken by a different record, so an in-place
+        # rename would violate the uniqueness validation. Fall back to a merge.
+        puts "  COLLISION #{from.inspect} -> #{to.inspect}: target ##{existing_target.id} exists, merging instead"
+        merges << [ from, to ]
```
maebeale, line 88: 🤖 _From Claude:_ This collision branch is what makes the task deploy-order-agnostic: once #1703 seeds have created the new canonical name, an in-place rename would violate the uniqueness validation, so we route it through a merge onto the already-seeded published record instead.

lib/tasks/consolidate_sectors.rake
```diff
@@ -0,0 +1,150 @@
...
+
+    # --- 2. Merges --------------------------------------------------------
+    puts "\nMerging sectors…"
+    deduper = ModelDeduper.new(model_class: Sector, dry_run: dry_run)
+    merges.each do |source_name, target_name|
+      source = find_by_name.call(source_name)
+      target = find_by_name.call(target_name)
+      # In a dry run the prerequisite rename hasn't been persisted, so resolve
+      # the target back through its pre-rename name to preview accurately.
+      target ||= find_by_name.call(renames.key(target_name))
```
maebeale, line 105: 🤖 _From Claude:_ Dry-run only: the prerequisite rename (e.g. Education/Schools → Education) is not persisted in a dry run, so the merge target wont be found by its new name. Resolving back through `renames.key` lets the dry run preview the Student → Education merge accurately.

PR #1760, opened 2026-06-18
Auto-scroll the registration error band into view

maebeale:

- A failed public event registration already renders a top-of-form error flash band (added in #1758), but after submitting from the bottom of a long form it lands off-screen — the registrant sees nothing change and assumes the page hung.

- Attach a small `form-errors` Stimulus controller to the **existing** error flash band so it scrolls itself into view on connect. It's on the band element, so it only connects when an alert is actually present.
- Works on both submit paths — Turbo (free events) and full-page (paid events).
- No new band: an earlier version of this PR added its own summary band, but #1758 landed the same generic band on `main` in the meantime, so this now just adds the scroll behaviour to that band (no duplication).

### UI Testing Checklist
- [ ] On a long registration form, submit with a required field missing → the page scrolls up to the red error band.
- [ ] Submit a valid registration → no band, normal success flow.

### Anything else to add?
- Bumped the Stimulus controller count in `AGENTS.md`.
- The success "You're registered!" banner (#1764) and over-length-answer handling (#1763, already merged) were split out of this branch earlier.

### DIFF
--- a/app/frontend/javascript/controllers/form_errors_controller.js
+++ b/app/frontend/javascript/controllers/form_errors_controller.js
@@ -1,0 +1,10 @@
+import { Controller } from "@hotwired/stimulus"
+
+// When the form re-renders with validation errors, scroll the error flash band
+// into view so the registrant immediately sees what needs fixing. Attached to
+// the band itself, so it only connects when an error message is present.
+export default class extends Controller {
+  connect() {
+    this.element.scrollIntoView({ behavior: "smooth", block: "center" })
+  }
+}
--- a/app/views/events/public_registrations/new.html.erb
+++ b/app/views/events/public_registrations/new.html.erb
@@ -85,8 +85,10 @@
       <% end %>
       <%= render "forms/header", form: @form, event: @event %>
 
+      <%# The error flash band carries the form-errors controller so it scrolls
+          itself into view on connect after a failed submit. %>
       <% if flash[:alert] %>
-        <div class="flex items-start gap-3 bg-red-50 border border-red-200 text-red-800 rounded-lg px-4 py-3 mb-6">
+        <div role="alert" class="scroll-mt-24 flex items-start gap-3 bg-red-50 border border-red-200 text-red-800 rounded-lg px-4 py-3 mb-6" data-controller="form-errors">
           <i class="fa-solid fa-circle-exclamation mt-0.5 text-red-500"></i>
           <span><%= flash[:alert] %></span>
         </div>
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ This codebase (Rails 8.1)
 | Directory | Purpose |
 |---|---|
 | `app/frontend/entrypoints/` | Vite entry points (application.js, application.css) |
-| `app/frontend/javascript/controllers/` | Stimulus controllers (73) |
+| `app/frontend/javascript/controllers/` | Stimulus controllers (74) |
 | `app/frontend/javascript/rhino/` | Rich text editor customizations (mentions, grid) |
 | `app/frontend/stylesheets/` | Tailwind CSS and component styles |
 
--- a/app/frontend/javascript/controllers/index.js
+++ b/app/frontend/javascript/controllers/index.js
@@ -84,6 +84,9 @@ application.register("file-preview", FilePreviewController)
 import FieldOptionsController from "./field_options_controller"
 application.register("field-options", FieldOptionsController)
 
+import FormErrorsController from "./form_errors_controller"
+application.register("form-errors", FormErrorsController)
+
 import FormFieldsSortableController from "./form_fields_sortable_controller"
 application.register("form-fields-sortable", FormFieldsSortableController)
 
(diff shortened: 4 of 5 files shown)


app/frontend/javascript/controllers/form_errors_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from "@hotwired/stimulus"
+
+// When the form re-renders with validation errors, scroll the error summary
+// banner into view so the registrant immediately sees what needs fixing.
+// Attached to the banner itself, so it only connects when errors are present.
+export default class extends Controller {
+  connect() {
```
maebeale, line 7: 🤖 _From Claude:_ Controller is attached to the error banner itself rather than the form, so it only ever connects when errors are present — no guard/target needed, and it scrolls its own element into view.

app/controllers/events/registrations_controller.rb
```diff
@@ -20,6 +20,8 @@ def show
       when "cancelled"
         flash.now[:alert] = "Payment was cancelled. You are registered for this event but payment may still be due."
       end
+
+      @just_registered = params[:registered].present? || params[:checkout] == "success"
```
maebeale: 🤖 _From Claude:_ Banner fires for a fresh registration (`?registered=true`) and for a successful Stripe return (`checkout=success`), but not on plain ticket re-visits.

jmilljr24:
@maebeale I think the standard flash message I added is sufficient in another PR. Rudy was happy with the change.

I would want further testing of this changes, no rushing

jmilljr24 (commented):
I'd like to wait on this.

app/services/event_registration_services/public_registration.rb
```diff
@@ -62,6 +62,8 @@ def call
 
         Result.new(success?: true, event_registration: event_registration, form_submission: submission, errors: [])
       end
+    rescue ActiveRecord::ValueTooLong => e
+      Result.new(success?: false, event_registration: nil, errors: [ too_long_message(e) ])
```
maebeale, line 81: 🤖 _From Claude:_ `ValueTooLong` is a `StatementInvalid`, not a `RecordInvalid`, so it needs its own rescue — otherwise an answer longer than its varchar(255) column 500s mid-registration instead of re-rendering the form.

maebeale:
@jmilljr24 what do you think about the one-time "You're registered" text? I could move that to another PR.

maebeale:
@jmilljr24 will your error also capture errors on scholarship fields?

maebeale:
@jmilljr24 also there's an extra validation re value is too long that wasn't being handled gracefully. i'd like to deploy that.

app/frontend/javascript/controllers/form_errors_controller.js
jmilljr24, line 1: I'd like to stick with standard rails flash and avoid extra js.

app/services/event_registration_services/public_registration.rb
```diff
@@ -62,6 +62,8 @@ def call
 
         Result.new(success?: true, event_registration: event_registration, form_submission: submission, errors: [])
       end
+    rescue ActiveRecord::ValueTooLong => e
```
jmilljr24, line 80: I haven't seen this but seems worthwhile.

app/views/events/public_registrations/new.html.erb
```diff
@@ -92,6 +92,22 @@
         </div>
       <% end %>
 
+      <% if @field_errors&.any? %>
```
jmilljr24: I'm not sure how this is different from the flash[:alert] on line 88

Image

app/views/events/registrations/show.html.erb
```diff
@@ -10,4 +10,15 @@
   <% end %>
   </div>
 </div>
+<% if @just_registered %>
```
jmilljr24: i think we can get away with sticking with a flash message and not adding any extra variables.

jmilljr24:
> @jmilljr24 what do you think about the one-time "You're registered" text? I could move that to another PR.

Nothing inherently wrong with it but in some ways is redundant to the flash message (standard flow) with additional conditional code. I don't feel its worth it.

app/views/events/public_registrations/new.html.erb
jmilljr24: Each section was already displaying the correct error for each field.

app/frontend/javascript/controllers/form_errors_controller.js
jmilljr24: On second look, I don't see where this is used.

app/views/events/public_registrations/new.html.erb
maebeale: bc people miss flash messages. there's too much text at top of form to see the field errors, and people aren't realizing the form didn't submit.

jmilljr24: Can you look at the screenshot and code again. In a previous PR you added the "flash"message directly in the view which is shown in the screenshot. (Not a dismissible flash).

The only thing missing was setting the actual information in the controller which has already been fixed and approved by the stakeholders

app/frontend/javascript/controllers/form_errors_controller.js
jmilljr24: Ah, I see it now. It seems excessive to have this. The section is at the top of the page so the scroll would be negligible anyways.

PR #1761, opened 2026-06-18
Add editable payment method to event registrations

maebeale:
> 📦 First of two PRs. This one adds the `payment_method` data + admin editing. The ticket-facing behavior for check payers is in the stacked follow-up **#1762**.

- Admins need to **see and change** a registrant's payment method (currently it only exists as a registration form answer, not on the registration record).
- This stores the choice on the registration and makes it editable, as groundwork for tailoring the ticket to check payers (#1762).

- Added a `payment_method` string column to `event_registrations`, populated from the registration form's `payment_method` answer (create and re-registration paths) — same options as the bulk payment form (`FormBuilderService::PAYMENT_METHOD_OPTIONS`), guarded by a light inclusion validation.
- Added an editable **icon dropdown** in the registration edit page's payments box (styled like the registration-status select, icon swaps per choice), laid out in two columns beside "Intends to pay", and permitted the param.
- Renamed the `attendance-status` Stimulus controller to `icon-select` since it now drives both the status and payment-method selects.

### UI Testing Checklist
- [ ] On a registration's edit page, change **Payment method** in the dropdown → icon updates, "Unsaved" hint shows, and saving persists the choice.
- [ ] Set it back to **Not specified** → clears the stored value.

### Anything else to add?
- No ticket/registrant-facing change here — that's #1762, which gates the credit-card button on `payment_method == "Check"`.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -60,6 +60,7 @@ module EventRegistrationServices
           existing.update!(ce_credit_requested: true, ce_hours_requested: ce_hours_requested) if ce_credit_requested?
           existing.update!(w9_requested: true) if w9_requested?
           existing.update!(invoice_requested: true) if invoice_requested?
+          existing.update!(payment_method: payment_method) if payment_method.present?
           if existing.status == "cancelled"
             existing.update!(status: "registered")
             send_notifications(existing)
@@ -311,7 +312,8 @@ module EventRegistrationServices
         ce_credit_requested: ce_credit_requested?,
         ce_hours_requested: ce_hours_requested,
         w9_requested: w9_requested?,
-        invoice_requested: invoice_requested?
+        invoice_requested: invoice_requested?,
+        payment_method: payment_method
       )
     end
 
@@ -348,6 +350,14 @@ module EventRegistrationServices
 
     def invoice_requested?
       additional_forms_selections.any? { |value| value.casecmp?(ADDITIONAL_FORMS_INVOICE) }
+    end
+
+    # The payment method the registrant chose (one of
+    # FormBuilderService::PAYMENT_METHOD_OPTIONS), persisted on the registration
+    # so admins can see/change it and the ticket can hide the credit-card button
+    # for check payers. nil when the form omits the question.
+    def payment_method
+      field_value("payment_method").presence
     end
 
     def create_form_submission(person)
--- a/app/views/event_registrations/_payment_history.html.erb
+++ b/app/views/event_registrations/_payment_history.html.erb
@@ -46,22 +46,73 @@
         </div>
       </dl>
 
-      <%# Intends to pay — grants ticket/material access while the balance above is
-          still owed, for registrants who've committed to paying after the deadline.
-          Does not change the amount due or mark the registration as paid. %>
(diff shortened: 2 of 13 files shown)


jmilljr24 (changes requested):
Pause.

The conversation was about changing text not a gated section. While possibly needed, they were going to discuss the text first.

app/views/event_registrations/_payment_history.html.erb
```diff
@@ -62,6 +62,26 @@
...
+          editable here. "Check" hides the "Pay with Credit Card" button on the
+          registrant's digital ticket. Options mirror the bulk payment form. %>
+      <fieldset class="mt-4 border-t border-gray-100 pt-4">
+        <legend class="text-sm font-medium text-gray-700">Payment method</legend>
+        <p class="mb-2 text-xs text-gray-400">Choosing "Check" hides the credit-card payment button on the registrant's ticket.</p>
+        <div class="flex flex-wrap gap-2.5">
+          <% [ [ "Not specified", "" ] ].concat(FormBuilderService::PAYMENT_METHOD_OPTIONS.map { |option| [ option, option ] }).each do |option_label, option_value| %>
+            <label class="flex items-center gap-2 rounded-lg border border-gray-300 bg-white px-3.5 py-2 text-sm text-gray-700 cursor-pointer transition hover:border-blue-400 hover:bg-blue-50 has-[:checked]:border-blue-500 has-[:checked]:bg-blue-50 has-[:checked]:text-blue-900 has-[:checked]:font-medium">
+              <input type="radio"
+                     name="event_registration[payment_method]"
```
maebeale: 🤖 _From Claude:_ Options mirror the bulk payment form (`PAYMENT_METHOD_OPTIONS`), prepended with a blank "Not specified" so an admin can clear a previously-set method back to nil.

app/services/event_registration_services/public_registration.rb
```diff
@@ -47,6 +47,7 @@ def call
           existing.update!(ce_credit_requested: true, ce_hours_requested: ce_hours_requested) if ce_credit_requested?
           existing.update!(w9_requested: true) if w9_requested?
           existing.update!(invoice_requested: true) if invoice_requested?
+          existing.update!(payment_method: payment_method) if payment_method.present?
```
maebeale, line 63: 🤖 _From Claude:_ On re-registration we only overwrite `payment_method` when the submitted form actually carried the question, so a partial re-submission can't blank out an existing choice.

maebeale:
@jmilljr24 i changed this to just the part about saving the payment method on the event registration. regardless of if we use it to change the ticket, it should be saved to the event registration so you can see it there, and to handle when the person changes their payment method after form submission.

also i edited a controller i was using for one dropdown to make it more universal.

maebeale:
image

jmilljr24 (approved):

jmilljr24:
Looks like some of this was already adding in Main
screenshot-2026-06-29_07-15-08

maebeale:
🤖 _From Claude:_ Closing as superseded by #1884, which shipped the same feature under `expected_payment_method`: the column, the capture from the form answer, and the admin-editable "Expected payment method" dropdown on the registration edit page. The hide-CC-button goal is dropped (check info is now a toggle on the payment card), so nothing here remains that main doesn't already have.

PR #1762, opened 2026-06-19
WAIT-MAYBE NOT: Hide ticket credit-card button when paying by check

maebeale:
> ⛓️ **Stacked on #1761** — base is `maebeale/hide-cc-button-check-payment` (the `payment_method` column). Review/merge that first; this PR will retarget to `main` once it lands.

- A registrant who chose **"Check"** intends to mail payment, so the digital ticket's **"Pay with Credit Card"** button is misleading.
- This hides that button for check payers and shows a "Pay by mailed check" note instead, while keeping the amount-due badge.

- Added `EventRegistration#pays_by_check?` (`payment_method == FormBuilderService::PAYMENT_METHOD_CHECK`) as the single seam the ticket gates on.
- Updated the ticket view to branch on it; covered with a model spec and a ticket system spec.

### UI Testing Checklist
- [ ] Ticket for a paid registration whose method is **Check** → shows "payment is due" + "Pay by mailed check", **no** credit-card button.
- [ ] Ticket for a card method (or no method) → still shows the **Pay with Credit Card** button.

### Anything else to add?
- Depends on the `payment_method` column from #1761; nothing here is user-editable (that lives in the base PR).

### DIFF
--- a/app/views/event_registrations/_ticket.html.erb
+++ b/app/views/event_registrations/_ticket.html.erb
@@ -86,17 +86,21 @@
             </div>
           <% end %>
 
-          <div>
-            <% if preview %>
-              <button type="button" disabled
-                      class="btn btn-accent px-6 py-2 text-sm uppercase font-telefon opacity-60 cursor-not-allowed">Pay with Credit Card</button>
-            <% else %>
-              <%= button_to "Pay with Credit Card",
-                            registration_pay_path(event_registration.slug),
-                            data: { turbo: false },
-                            class: "btn btn-accent px-6 py-2 text-sm uppercase font-telefon" %>
-            <% end %>
-          </div>
+          <% if event_registration.pays_by_check? %>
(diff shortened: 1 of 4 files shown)


app/views/event_registrations/_ticket.html.erb
```diff
@@ -91,13 +91,17 @@
...
             <% end %>
 
-            <div class="mt-3">
-              <%= button_to "Pay with Credit Card",
-                            registration_pay_path(event_registration.slug),
-                            data: { turbo: false },
-                            class: "btn px-6 py-2 text-sm uppercase font-telefon text-white hover:bg-white",
-                            style: "background-color: rgb(170, 46, 0); border: 2px solid rgb(170, 46, 0);" %>
-            </div>
+            <% if event_registration.pays_by_check? %>
```
maebeale: 🤖 _From Claude:_ Only the card button is gated here — the "payment is due" badge above still renders, so check payers see the balance without a misleading way to pay it by card.

PR #1763, opened 2026-06-19
Turn over-length registration answers into form errors, not 500s

maebeale:

- A public registration answer longer than its mapped `varchar(255)` column (e.g. a long city) raised `ActiveRecord::ValueTooLong` mid-registration and surfaced as a **500**, with no way for the registrant to recover.
- `ValueTooLong` is a `StatementInvalid`, which the service's existing `RecordInvalid`/`RecordNotUnique` rescues don't catch — so it escaped the normal "re-render the form with an error" path.

- `EventRegistrationServices::PublicRegistration` now rescues `ActiveRecord::ValueTooLong` and returns a failed result with a friendly, column-named message (`"Your city is too long. Please shorten it and try again."`). The controller already re-renders the form with that error, so it flows through as an ordinary form validation error.
- The column can't always be mapped back to a single field (both the mailing and agency address write `city`), so the message names the **column** rather than guessing a field.

### UI Testing Checklist
- [ ] Submit the public registration form with an answer longer than its column (e.g. a 256-char city) → the form re-renders with "Your city is too long…", no 500, no registration created.

### Anything else to add?
- Split out of #1760 to keep the over-length fix reviewable on its own.
- Specs (failing-first): service returns a failed result instead of raising and creates no registration; request-level confirms the 422 form error end-to-end.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -62,6 +62,8 @@ module EventRegistrationServices
 
         Result.new(success?: true, event_registration: event_registration, form_submission: submission, errors: [])
       end
+    rescue ActiveRecord::ValueTooLong => e
+      Result.new(success?: false, event_registration: nil, errors: [ too_long_message(e) ])
     rescue ActiveRecord::RecordInvalid => e
       Result.new(success?: false, event_registration: nil, errors: [ e.message ])
     rescue ActiveRecord::RecordNotUnique => e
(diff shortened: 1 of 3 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -62,6 +62,8 @@ def call
 
         Result.new(success?: true, event_registration: event_registration, form_submission: submission, errors: [])
       end
+    rescue ActiveRecord::ValueTooLong => e
+      Result.new(success?: false, event_registration: nil, errors: [ too_long_message(e) ])
```
maebeale, line 66: 🤖 _From Claude:_ `ValueTooLong` is a `StatementInvalid`, not a `RecordInvalid`, so it needs its own rescue — otherwise an answer longer than its varchar(255) column 500s mid-registration instead of re-rendering the form.

jmilljr24 (approved):

PR #1764, opened 2026-06-19
HOLD: Confirm registrants with a "You're registered!" banner

maebeale:

- After a successful public event registration, the registrant landed on the ticket page with no explicit "you're in" confirmation beyond the transient flash toast.

- The success redirect carries a `registered` flag to the ticket page.
- `registrations#show` sets `@just_registered` for that flag **or** a successful Stripe `checkout=success` return, and the ticket view renders a green "You're registered!" banner above the ticket.
- The banner is not shown on plain return visits to an existing ticket URL. The existing "You have been successfully registered!" flash toast is unchanged.

### UI Testing Checklist
- [ ] Register on a free event → land on the ticket page with a "You're registered!" banner.
- [ ] Pay by card and complete Stripe checkout → ticket page shows the banner.
- [ ] Re-open an existing ticket URL directly → no banner.

### Anything else to add?
- Split out of #1760 to keep the success banner reviewable on its own.
- Specs cover the banner showing for `registered`/`checkout=success` but not on plain ticket views.

### DIFF
--- a/app/controllers/events/registrations_controller.rb
+++ b/app/controllers/events/registrations_controller.rb
@@ -20,6 +20,8 @@ module Events
       when "cancelled"
         flash.now[:alert] = "Payment was cancelled. You are registered for this event but payment may still be due."
       end
+
+      @just_registered = params[:registered].present? || params[:checkout] == "success"
     end
 
     def invoice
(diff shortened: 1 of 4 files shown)


app/controllers/events/registrations_controller.rb
```diff
@@ -20,6 +20,8 @@ def show
       when "cancelled"
         flash.now[:alert] = "Payment was cancelled. You are registered for this event but payment may still be due."
       end
+
+      @just_registered = params[:registered].present? || params[:checkout] == "success"
```
maebeale, line 24: 🤖 _From Claude:_ Banner fires for a fresh registration (`?registered=true`) and a successful Stripe return (`checkout=success`), but not on plain ticket re-visits.

PR #1765, opened 2026-06-19
Accept bare domains for organization website URLs

maebeale:
- The organization Website URL field rejected anything without an explicit `http://`/`https://` scheme, so users who naturally typed `awbw.org` hit a validation error.
- This removes that friction: users can save whatever they type, and the clickable links still work.

- Dropped the `website_url` format validation on `Organization` so the value is stored verbatim.
- Added `Organization#website_link_url`, which prepends `https://` to a bare domain and returns `nil` for input that can't form a usable web address (requires a `URI::HTTP` with a host).
- The org profile and recipients pages now build their external links via that method, so a bare `awbw.org` links out correctly while genuine junk falls back (profile link / "Bad URL" notice).

### UI Testing Checklist
- [ ] Edit an organization, enter `awbw.org` in Website URL, save — it persists as typed and the profile page shows a working external link to `https://awbw.org`.
- [ ] An org with `awbw.org` on the event recipients page links out to the site; an org with junk text falls back to its profile page.

### Anything else to add?
- New model specs cover `website_link_url` (bare domain, existing scheme, whitespace, blank, unusable input). A weak-confidence Brakeman `LinkToHref` warning is suppressed in `config/brakeman.ignore` because the method can only ever return an `http(s)://` URL.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -38,8 +38,6 @@ class Organization < ApplicationRecord
   validates :name, presence: true
   validates :organization_status_id, presence: true
   validates :email, format: { with: URI::MailTo::EMAIL_REGEXP, message: "must be a valid email address" }, allow_blank: true
-  validates :website_url, format: { with: /\Ahttps?:\/\/\S+\z/i, message: "must be a valid web address" }, allow_blank: true
-  before_validation :normalize_website_url
   validate :affiliation_dates_locked, if: -> { affiliations.any? && !Current.user&.super_user? }
 
   # Nested attributes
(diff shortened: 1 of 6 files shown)


app/models/organization.rb
```diff
@@ -244,6 +245,15 @@ def all_additional_age_groups
 
   private
 
+  # Lets users enter a bare domain (e.g. "awbw.org") by defaulting to https://
+  # when no scheme is present, so the value passes URL validation and renders as
+  # a working link.
+  def normalize_website_url
+    return if website_url.blank?
```
maebeale, line 249: 🤖 _From Claude:_ Normalizing here (rather than relaxing the regex to accept bare domains) keeps the stored value a real URL, so the show-page link and the form's "Bad URL" warning keep working without changes.

app/models/organization.rb
```diff
@@ -242,6 +241,18 @@ def all_additional_age_groups
...
+  # Returns the website as a clickable, scheme-qualified URL — prepending
+  # https:// to a bare domain like "awbw.org" — or nil when the value is blank or
+  # not a usable web address. Drives the external links on the org profile and
+  # recipients pages without forcing users to type a scheme.
+  def website_link_url
+    return if website_url.blank?
+    candidate = website_url.strip
+    candidate = "https://#{candidate}" unless candidate.match?(/\Ahttps?:\/\//i)
+    uri = URI.parse(candidate) rescue nil
+    candidate if uri.is_a?(URI::HTTP) && uri.host.present?
```
maebeale, line 253: 🤖 _From Claude:_ This requires a parsed `URI::HTTP` with a host, so it only ever returns an http(s):// URL — a `javascript:`/`data:` value can't slip into the link href. That's why the matching Brakeman LinkToHref warning is suppressed in config/brakeman.ignore rather than worked around in the view.

PR #1767, opened 2026-06-19
Point registration confirmation email at the ticket only

maebeale:
The registration confirmation email should send registrants straight to their ticket, which already carries the event details, calendar links, and registration-management options — so the separate "View event" link was redundant noise.

Removed the "View event" link (and the now-unused `@event_url`) from both the HTML and text confirmation emails, leaving a single "View ticket" call to action. Also fixed the mailer preview, which built an unpersisted registration and therefore hid the ticket link entirely, by persisting the sample registration.

### UI Testing Checklist
- [ ] Confirmation email shows a single "View ticket" button linking to the registrant's ticket
- [ ] No "View event" link appears in the HTML or text email
- [ ] Mailer preview renders the ticket link

### Anything else to add?
Part of splitting a combined branch into focused PRs — scholarship-answer saving and the scholarship section on the view-submission page follow in their own PRs.

### DIFF
--- a/test/mailers/previews/event_mailer_preview.rb
+++ b/test/mailers/previews/event_mailer_preview.rb
@@ -41,14 +41,13 @@ class EventMailerPreview < ActionMailer::Preview
   end
 
   def sample_event_registration
-    # Try to reuse existing records to avoid duplication
+    # Try to reuse existing records to avoid duplication. Persist the registration
+    # so it has a slug and the confirmation email's "View ticket" link renders
+    # (the link is gated on a persisted registration).
     event = Event.first || create_event
     person = Person.first || create_person
 
-    EventRegistration.new(
-      event: event,
-      registrant: person
-    )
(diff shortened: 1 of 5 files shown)


test/mailers/previews/event_mailer_preview.rb
```diff
@@ -41,14 +41,13 @@ def sample_bulk_payment_submission
...
+    # so it has a slug and the confirmation email's "View ticket" link renders
+    # (the link is gated on a persisted registration).
     event = Event.first || create_event
     person = Person.first || create_person
 
-    EventRegistration.new(
-      event: event,
-      registrant: person
-    )
+    EventRegistration.find_or_create_by!(event: event, registrant: person)
```
maebeale, line 50: 🤖 _From Claude:_ Persisting the sample registration is the actual fix for "the email only links to the event" — the `View ticket` link is gated on `@event_registration.persisted?`, so the old `EventRegistration.new` preview hid it.

PR #1768, opened 2026-06-19
Lock payment-method answer options from casual edits

maebeale:

- The payment-method field's answer options (`Credit card (now)`, `Credit card (later)`, `Check`) are wired to Stripe charge logic in the registration controllers — the `Credit card (now)` option triggers an immediate charge.
- Editing, renaming, or removing these options from the form builder can silently break payments without any warning.
- This makes those options **read-only in the form editor** so they can't be changed casually, while still showing them so admins can see what's configured.

- Added `FormField#fixed_options?` (true for the `payment_method` field) and a `PAYMENT_METHOD_FIELD_IDENTIFIER` constant.
- In the form-builder "Answer options" panel, when a field has fixed options and `?admin=true` is **not** present, the options render as static read-only text with a short note explaining they're tied to system logic.
- Appending `?admin=true` to the form-editor URL restores the normal editable inputs and the "+ Add option" link for the rare case the options genuinely need to change.
- Read-only mode simply omits the nested-attribute inputs, so on save the existing options are left untouched (Rails nested attributes only modify records present in params) — they can't be accidentally wiped.

### UI Testing Checklist

- [ ] Open the form editor for a form with a payment-method field; expand "Answer options" — the three options show as read-only text with the explanatory note, no inputs or "+ Add option".
- [ ] Add `?admin=true` to the URL and reload — the options are now editable inputs with "+ Add option".
- [ ] Other multiple-choice fields remain fully editable in both cases.

### Anything else to add?

- This is UI-level gating. The `update` action still permits `form_field_answer_options_attributes`, so a hand-crafted POST could technically still submit them. Happy to add server-side enforcement if we want it locked down harder (that would also require carrying `admin=true` through the form submit).

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -40,6 +40,12 @@ class FormField < ApplicationRecord
   ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER = "additional_age_group"
   AGE_GROUP_FIELD_IDENTIFIERS = [ PRIMARY_AGE_GROUP_FIELD_IDENTIFIER, ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER ].freeze
 
+  # The payment-method field. Its answer options ("Credit card (now)", etc.) are
+  # wired to Stripe charge logic in the controllers, so they must not be edited
+  # casually from the form builder — the editor shows them read-only unless the
+  # admin override is present.
+  PAYMENT_METHOD_FIELD_IDENTIFIER = "payment_method"
+
   # The generic free-text option label that lets a respondent supply their own
   # value; a chosen "Other" answer is stored as "Other" or "Other: <text>".
   OTHER_OPTION_PREFIX = "Other"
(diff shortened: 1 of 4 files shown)


app/views/forms/_form_field_fields.html.erb
```diff
@@ -151,12 +151,27 @@
         </div>
 
         <% unless option_source %>
+          <% options_locked = field.fixed_options? && params[:admin] != "true" %>
           <div class="hidden space-y-2 pl-1 border-t border-gray-100 pt-3" data-answer-options-target="list">
-            <%= f.fields_for :form_field_answer_options do |opt| %>
-              <%= render "form_field_answer_option_fields", f: opt %>
+            <% if options_locked %>
```
maebeale, line 156: 🤖 _From Claude:_ Read-only mode just omits the nested-attribute inputs, so on save these options are left untouched — Rails nested attributes only modify records present in params, so existing options can't be accidentally wiped.

app/models/form_field.rb
```diff
@@ -163,6 +169,13 @@ def selectable?
     answer_type.in?(SELECTABLE_ANSWER_TYPES)
   end
```
maebeale, line 171: 🤖 _From Claude:_ Named this `fixed_options?` rather than hardcoding the payment_method check in the view, so other options-tied-to-backend-logic fields can opt in here later.

PR #1769, opened 2026-06-19
Persist scholarship-form answers on public registration

maebeale:
When a registrant filled out the separate scholarship form, their answers were validated but never saved — the application was silently dropped on submit. This makes scholarship applications actually persist.

Moved scholarship persistence into the `PublicRegistration` service so it runs in the same transaction as the registration, creating a `role: "scholarship"` `FormSubmission` tied to the event, and removed the dead controller method that was meant to do this but was never wired up.

### UI Testing Checklist
- [ ] Submitting a registration with the scholarship form filled in creates a scholarship submission with the answers
- [ ] A plain registration (no scholarship requested) creates no scholarship submission

### Anything else to add?
Part of splitting a combined branch into focused PRs; surfacing the saved answers to users/admins follows in a separate PR.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -15,16 +15,21 @@ module EventRegistrationServices
     ADDITIONAL_FORMS_INVOICE = "Invoice".freeze
     ADDITIONAL_FORMS_W9 = "W-9".freeze
 
-    def self.call(event:, form:, form_params:, scholarship_requested: false, person: nil)
-      new(event:, form:, form_params:, scholarship_requested:, person:).call
+    def self.call(event:, form:, form_params:, scholarship_requested: false, person: nil,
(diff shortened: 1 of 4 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -336,6 +343,34 @@ def save_form_answers(submission)
       end
     end
 
+    # Persist the answers to the separate scholarship form (when one is asked and a
+    # scholarship was requested) as its own role: "scholarship" submission tied to
+    # the event, mirroring how the registration submission is saved above.
+    def save_scholarship_submission(person)
```
maebeale, line 349: 🤖 _From Claude:_ Saving lives here (not the controller) so it runs inside the same transaction as the registration + registration-form submission. Guarded on `scholarship_requested` + a scholarship form + present params, and called from both the new and re-registration branches.

PR #1773, opened 2026-06-19
Limit bulk-payment allocation dropdown to unpaid registrants

maebeale:
- On the event bulk-payments dashboard, the "New allocation" registrant dropdown listed every active registrant, including those already fully paid.
- Admins allocating a payment had to scan past settled attendees to find who still owes; this hides registrants whose allocations already cover the event cost so only unpaid registrants remain.

- Filtered the dropdown options to registrants whose allocated total is below `event_cost_cents`, reusing the `allocated_by_registration` data and "paid in full" logic already in the card (no extra query).
- Filtered the dropdown collection specifically rather than the shared `event_registrations` list, since that collection also drives attendee matching elsewhere in the card.

### UI Testing Checklist
- [ ] Open an event's bulk-payments page with a payment that has remaining balance and confirm fully-paid registrants no longer appear in the "New allocation" dropdown.

### Anything else to add?
- Added a request spec (red-green verified) asserting a fully-covered registrant is excluded while an unpaid one is listed.

### DIFF
--- a/app/views/events/_bulk_payment_card.html.erb
+++ b/app/views/events/_bulk_payment_card.html.erb
@@ -174,9 +174,10 @@
             <%= form_tag allocate_bulk_payment_event_path(@event),
                          class: "flex items-center gap-2" do %>
               <%= hidden_field_tag :payment_id, payment.id %>
+              <% unpaid_registrations = event_registrations.reject { |r| allocated_by_registration.fetch(r.id, 0) >= event_cost_cents } %>
               <div class="w-64 shrink-0">
                 <%= select_tag :event_registration_id,
-                     options_from_collection_for_select(event_registrations, :id, ->(r) { "#{r.registrant.name} — #{r.registrant.email}" }),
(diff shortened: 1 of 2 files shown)


app/views/events/_bulk_payment_card.html.erb
```diff
@@ -174,9 +174,10 @@
             <%= form_tag allocate_bulk_payment_event_path(@event),
                          class: "flex items-center gap-2" do %>
               <%= hidden_field_tag :payment_id, payment.id %>
+              <% unpaid_registrations = event_registrations.reject { |r| allocated_by_registration.fetch(r.id, 0) >= event_cost_cents } %>
```
maebeale, line 177: 🤖 _From Claude:_ Filtering the dropdown collection here rather than the shared `event_registrations` local on purpose — that same collection drives attendee matching higher up in this card, so narrowing it globally would hide matches too. The paid-in-full check reuses `allocated_by_registration` (already preloaded by the controller), so there's no extra query.

PR #1774, opened 2026-06-19
Show submitted payment total in bulk payment confirmation email

maebeale:
- The bulk payment confirmation email showed the per-event cost and attendee count separately, but never the total owed.
- Surfacing the computed total saves payers the math and matches the figure already shown on the bulk payment ticket page.

- `EventMailer#bulk_payment_confirmation` now sets `@total_cents` via `FormSubmission#bulk_payment_amount_cents` (event cost × attendee count), guarded for event-less submissions.
- Both the HTML and text templates render a **Total** row (using `dollars_from_cents`) only when the total is positive, so free or event-less submissions don't show a `$0` line.
- Added a mailer spec asserting the total appears in the body.

### UI Testing Checklist
- [ ] Open the EventMailer preview for `bulk_payment_confirmation` and confirm the **Total** row renders in both HTML and plain-text views.
- [ ] Confirm whole-dollar totals render without trailing `.00` and amounts with cents show two decimals.

### Anything else to add?
The total reflects the expected amount (event cost × attendee count), consistent with the ticket page, so it displays even before a payment record lands.

### DIFF
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -25,6 +25,9 @@ class EventMailer < ApplicationMailer
     @event = form_submission.event&.decorate
     @answers = form_submission.answers_by_identifier
     @attendees = form_submission.bulk_payment_attendees
+    # Expected total (event cost × attendee count), shown even before a payment
+    # record lands. Mirrors the ticket page; omitted when there's no event/cost.
+    @total_cents = form_submission.event && form_submission.bulk_payment_amount_cents(form_submission.event)
 
     @notification_type = "Bulk payment confirmation"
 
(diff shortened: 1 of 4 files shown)


app/mailers/event_mailer.rb
```diff
@@ -25,6 +25,9 @@ def bulk_payment_confirmation(form_submission)
     @event = form_submission.event&.decorate
     @answers = form_submission.answers_by_identifier
     @attendees = form_submission.bulk_payment_attendees
+    # Expected total (event cost × attendee count), shown even before a payment
+    # record lands. Mirrors the ticket page; omitted when there's no event/cost.
+    @total_cents = form_submission.event && form_submission.bulk_payment_amount_cents(form_submission.event)
```
maebeale, line 30: 🤖 _From Claude:_ Total uses the **expected** amount (event cost × attendee count), matching the ticket page — so it shows even before a Stripe payment lands, not necessarily what was actually charged.

maebeale:
Screenshot 2026-06-19 at 7 21 51 AM

PR #1777, opened 2026-06-19
WAIT: Transfer a registrant to another event

maebeale:

- Admins sometimes need to move a registrant from one event to another (e.g. a reschedule) **without making them pay again**.
- Until now there was no supported way to do this — the only options were deleting the registration or manually re-registering, both of which lose history and misrepresent the payment.

- **New "Transfer" action** on the registration edit page. The admin picks a destination event (the picker excludes the current event and any the registrant is already on, since the unique `[registrant, event]` index forbids duplicates).
- Transferring **creates a new registration** on the destination, linked back to the original via a new `transferred_from_id` FK, and **marks the original "Transferred"** — a new *inactive* attendance status, so transferred registrants drop out of active rosters. The original keeps its payment history intact.
- **New "Previous reg" payment status** on the destination registration: because the fee was already collected on the original event, the new one shows nothing owed.
  - Deliberately **not** modeled as a 100% `Discount` — a discount means revenue was *forgiven*, which would corrupt discount/revenue reporting and can't record *why* it's $0. The real payment stays where it was paid; the destination only *displays* that it's covered (`payment_status_label`, the roster badge, and a banner on the payment card link back to the source event).
- **Two concepts kept distinct** (one predicate each, so they can't be conflated):
  - `transferred_out?` — the *attendance status* of the event the registrant is **leaving**. It is inactive and grants **no** access to that event's paid content, even though the original payment still lives on it (`payment_access_granted?` denies it).
  - `paid_via_previous_registration?` — the *payment status* ("Previous reg") of the event the registrant **is attending**. It's a display only and does not by itself unlock content access.
- Logic lives in `EventRegistrationServices::Transfer` (transactional: create new + mark old, all-or-nothing). Admin-only via `EventRegistrationPolicy#transfer?`/`#create_transfer?`.

### Naming

- Action/button: **Transfer** (avoided "Reallocate" — collides with the payment *Allocation* concept — and "Credit" — implies an account balance).
- Original (left) registration status: **Transferred** (inactive).
- Destination (attended) payment status: **Previous reg**.

### UI Testing Checklist

- [ ] Edit a registration → **Transfer** → pick a destination event → registrant appears on the new event; original shows **Transferred** and is hidden from the active roster.
- [ ] Destination registration's payment shows **Previous reg** (no amount due) on the roster, the edit page, and CSV export; payment card links back to the source event.
- [ ] Transfer picker omits the current event and events the registrant is already registered for; transferring to an already-registered event is rejected with an alert.
- [ ] Original registration's payment history is unchanged after transfer.

### Anything else to add?

- **Access note:** a "Previous reg" registration is a payment *display*, not an access grant — it doesn't auto-unlock the attended event's join links/materials. If a transferred registrant needs that access, an admin can toggle **Intends to pay** on the destination registration. Happy to wire "Previous reg" into the access seam instead if that's preferred.
- Tests: service spec (`EventRegistrationServices::Transfer`), model spec (status/labels/scope/access), and request specs (GET form, POST happy path + invalid target + non-admin not authorized).

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -3,6 +3,13 @@ class EventRegistration < ApplicationRecord
 
   belongs_to :registrant, class_name: "Person"
   belongs_to :event
+  # A registration created by transferring this registrant from another event.
+  # `transferred_from` is the original (where the payment still lives);
+  # `transferred_to` is the new registration created on the destination event.
+  belongs_to :transferred_from, class_name: "EventRegistration", optional: true,
+    inverse_of: :transferred_to
+  has_one :transferred_to, class_name: "EventRegistration",
+    foreign_key: :transferred_from_id, dependent: :nullify, inverse_of: :transferred_from
   has_many :comments, -> { newest_first }, as: :commentable, dependent: :destroy
   has_many :event_registration_organizations, dependent: :destroy
   has_many :notifications, as: :noticeable, dependent: :destroy
@@ -23,7 +30,11 @@ class EventRegistration < ApplicationRecord
   after_commit :send_cancellation_emails, if: :status_changed_to_cancelled?
 
   ACTIVE_STATUSES = %w[ registered attended incomplete_attendance ].freeze
-  INACTIVE_STATUSES = %w[ cancelled no_show ].freeze
+  # "transferred" marks a registration whose registrant was moved to another
+  # event via the Transfer action. It's inactive so transferred registrants drop
+  # out of active rosters; the matching new registration links back via
+  # `transferred_from`.
+  INACTIVE_STATUSES = %w[ cancelled no_show transferred ].freeze
   ATTENDANCE_STATUSES = (ACTIVE_STATUSES + INACTIVE_STATUSES).freeze
 
   # Manual onboarding checklist steps shown on the event's Onboarding tab. Each is
@@ -138,11 +149,13 @@ class EventRegistration < ApplicationRecord
       ), 0)
     SQL
   }
+  scope :transferred, -> { where.not(transferred_from_id: nil) }
   scope :payment_status, ->(value) {
     case value
     when "paid" then paid_in_full
     when "unpaid" then not_paid_in_full
     when "intends_to_pay" then where(intends_to_pay: true)
+    when "previous_registration" then transferred
     else all
     end
   }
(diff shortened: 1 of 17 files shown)


app/services/event_registration_services/transfer.rb
```diff
@@ -0,0 +1,51 @@
...
+      new_registration = nil
+      ApplicationRecord.transaction do
+        new_registration = EventRegistration.create!(
+          registrant: @source.registrant,
+          event: target_event,
+          status: "registered",
+          transferred_from: @source
+        )
+        @source.update!(status: "transferred")
+      end
```
maebeale, line 42: 🤖 _From Claude:_ The transaction is the whole point here — create the new registration and flip the original to "transferred" atomically, so a failure can't leave a registrant on both events or on neither.

app/models/event_registration.rb
```diff
@@ -201,12 +214,20 @@ def paid?
...
+  def transferred?
+    transferred_from_id.present?
   end
 
   # Human-readable payment status for rosters and CSV exports. Assumes the event
-  # has a cost — callers show nothing for free events.
+  # has a cost — callers show nothing for free events. A transferred registration
+  # reports "Previous registration" since the fee was paid on the original event.
   def payment_status_label
+    return "Previous registration" if transferred?
```
maebeale: 🤖 _From Claude:_ "Previous registration" takes precedence over paid/intends-to-pay because a transferred registration has no allocations of its own — the fee lives on the source event. This keeps `paid_in_full?` and the reporting scopes honest (they still see $0 collected here) while the label reflects that nothing is owed.

app/models/event_registration.rb
```diff
@@ -201,12 +214,20 @@ def paid?
   # surfaces (rosters, CSV exports, dashboard metrics) must keep using `paid?` /
   # `paid_in_full?` so they still reflect the real balance owed.
   def payment_access_granted?
-    paid? || intends_to_pay?
+    paid? || intends_to_pay? || transferred?
```
maebeale: 🤖 _From Claude:_ Transferred registrations get content access (join links, materials) since their fee was paid on the source event — folded into the single payment-access seam.

app/models/event_registration.rb
```diff
@@ -201,12 +214,30 @@ def paid?
   # surfaces (rosters, CSV exports, dashboard metrics) must keep using `paid?` /
   # `paid_in_full?` so they still reflect the real balance owed.
   def payment_access_granted?
+    # A registrant transferred OUT to another event isn't attending this one, so
+    # they get no access here even though the original payment still lives on it.
+    return false if transferred_out?
```
maebeale, line 233: 🤖 _From Claude:_ Correction to my earlier note — a registration left behind ("transferred" status) now gets **no** access here, even though it still holds the payment, since the registrant isn't attending. "Previous reg" on the attended registration is a payment display only and doesn't itself grant access.

PR #1778, opened 2026-06-19
Rename bulk payment form to "Pay for participant(s)"

maebeale:
- The form's internal name, "Bulk Payment," is admin jargon — it doesn't tell an event visitor what the action actually does.
- Renames the **public-facing** CTA and form heading to **"Pay for others"**, which is plainer, warmer, and frames the action as covering registration for other people.
- Chosen over "Pay for someone else" — "someone else" subtly reads as a *substitute* ("not the right person"), whereas "others" simply means *more people*, the generous framing we want. It's also shorter and plays better with a bulk (multi-person) flow.
- Tone matters for our older, artistic audience: avoids the bureaucratic "on behalf" register.

- Introduced `Form::BULK_PAYMENT_PUBLIC_NAME` as a single source of truth so the event-page CTA and the form heading can never drift apart.
- Internal/admin labels keep the "Bulk payment" form-role name; only the public-facing copy uses the new constant.
- Updated the subheading from "Pay for multiple attendees." to "Pay for one or more attendees." for accuracy.

### UI Testing Checklist
- [ ] Event page CTA button reads "Pay for others"
- [ ] Bulk payment form heading reads "Pay for others"
- [ ] Admin/internal references still say "Bulk payment"

### Anything else to add?
- Pure copy/labeling change — no behavior, routing, or schema changes.


app/models/form.rb
```diff
@@ -1,4 +1,9 @@
 class Form < ApplicationRecord
+  # Public-facing name for the "bulk payment" form — what visitors see on the
+  # event page CTA and the form heading. Internal/admin labels still say "Bulk
+  # payment" (the form role). Single source of truth so both ends stay in sync.
+  BULK_PAYMENT_PUBLIC_NAME = "Pay for others".freeze
```
maebeale: 🤖 _From Claude:_ Single source of truth for the public name — the event-page CTA and the form heading both read from this constant so they can never drift. Admin/internal labels deliberately keep the "Bulk payment" form-role name.

jmilljr24 (approved):
@maebeale I like having it set in one place for the public side of things!! I've updated per stakeholder request.

PR #1779, opened 2026-06-19
Add an admin sample-ticket preview for events

maebeale:
Admins had no way to preview a registration ticket without creating a throwaway registration. This adds a **Sample ticket** preview — a non-persisted example registration with every registrant-chosen option enabled — so admins can see exactly what registrants will see (callouts, CE, invoice, payment states, manage actions) for their event.

- New admin-only `events#sample_ticket` action (gated by `dashboard?`) builds an in-memory `EventRegistration` with all flags on and a fake "Sample Registrant" — nothing is saved.
- The shared `_ticket` partial gained a `preview` flag that disables slug-dependent buttons (pay, resend, cancel) and points the invoice card at the event's blank invoice template (the same one linked from the bulk payments index), since the sample has no slug.
- Linked from the dashboard quick links and the **Form actions** menu (the latter passing `return_to: "registrants"` so the eyebrow returns correctly); registered the new view's `page_bg_class` and added request specs covering admin render, non-admin redirect, and that no registration is created.

### UI Testing Checklist
- [ ] Dashboard → "Sample ticket" opens the preview in a new tab; eyebrow returns to Dashboard
- [ ] Registrants → Form actions → "Sample ticket" opens the preview; eyebrow returns to Registrants
- [ ] Preview shows all sections; pay/resend/cancel buttons are visibly disabled
- [ ] "View invoice" opens the event's blank invoice
- [ ] Non-admin cannot reach the preview

### Anything else to add?
The preview reflects the event's real configuration plus all registrant options, so payment sections only appear when the event actually has a cost.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -2,7 +2,7 @@ class EventsController < ApplicationController
   include AhoyTracking, TagAssignable
   skip_before_action :authenticate_user!, only: [ :index, :show, :staff, :details, :ce_hours ]
   skip_before_action :verify_authenticity_token, only: [ :preview ]
-  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder confirm_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
+  before_action :set_event, only: %i[ show edit update destroy preview dashboard sample_ticket background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder confirm_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
 
   def index
     authorize!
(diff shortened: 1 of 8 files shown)


app/controllers/events_controller.rb
```diff
@@ -42,6 +42,32 @@ def dashboard
...
 
+  # Admin preview of the registration ticket. Builds an in-memory sample
+  # registration — never saved — with every registrant-chosen option turned on,
+  # so admins can see what their event's ticket looks like with all sections
+  # present. The ticket partial renders in `preview: true` mode, which disables
+  # the slug-dependent buttons (pay, resend, cancel) and points the invoice card
+  # at the event's blank invoice template, since the sample has no slug and
+  # isn't persisted.
+  def sample_ticket
+    authorize! @event, to: :dashboard?
```
maebeale, line 53: 🤖 _From Claude:_ Authorized on `dashboard?` (admin-or-owner) rather than a new policy method — both link origins (dashboard, registrants) already gate on that level, so reusing it keeps the access rule in one place.

app/views/event_registrations/_ticket.html.erb
```diff
@@ -149,7 +156,11 @@
       <% end %>
 
       <% if event_registration.invoice_requested? %>
-        <%= link_to registration_invoice_path(event_registration.slug),
+        <%# Preview has no registration slug, so the per-registration invoice route
+            can't be built — link to the event's blank invoice template instead
+            (the same one linked from the bulk payments index). %>
+        <% invoice_href = preview ? event_invoice_path(event_registration.event) : registration_invoice_path(event_registration.slug) %>
```
maebeale, line 162: 🤖 _From Claude:_ In preview there is no registration slug, so the per-registration invoice route would raise. Falling back to the event-level blank invoice keeps the card a real, working link.

maebeale:
🤖 _From Claude:_ Superseded by #1793, which is the same change rebased cleanly onto `main` (picking up the new `MagicTicketCallouts` system). Closing this one in favor of the fresh PR.

PR #1780, opened 2026-06-19
Rename event-edit accordion label to "Form settings"

maebeale:
- The accordion toggle on the event edit page read "Registration Form", which conflated the section header with the registration-form selection field inside it.
- Renaming it to "Form settings" better describes the broader registration configuration the section reveals.

- Updated the toggle button label in `app/views/events/_form.html.erb`; the inner "Registration form" select label is unchanged since it labels the form-selection input itself.

### UI Testing Checklist
- [ ] Open an event's edit page as an admin and confirm the accordion toggle reads "Form settings".
- [ ] Confirm the accordion still expands/collapses and the inner "Registration form" field is unaffected.

### Anything else to add?
- Text-only change; no logic, IDs, or Stimulus targets were altered.


maebeale:
Screenshot 2026-06-19 at 7 23 20 AM

PR #1781, opened 2026-06-19
Rename registration form labels from Agency to Organization

maebeale:
- The person contact section of registration forms used "Agency / Organization" wording that was redundant and out of step with how staff refer to these records ("Organizations").
- Relabeling makes the public-facing form match internal terminology.

- Updated the section header and field labels in `FormBuilderService` (`person_contact_info`) from "Agency …" to "Organization …", keeping `SECTION_HEADERS`, `SECTION_FIELD_NAMES`, and the builder method in sync.
- Left field identifiers/keys (`agency_name`, `agency_street`, etc.) unchanged to avoid a data migration, and updated the matching comment in the dev seed.

### UI Testing Checklist
- [ ] On a registration form with the contact section, the section header reads "Organization Information" and fields read "Organization Name / Street Address / City / State / Province / Zip / Type / Website".

### Anything else to add?
- `FormBuilderService` is the source of these labels (seeds build forms through it), so no separate seed-data edits were needed. The "Government agency" option value under Organization Type was intentionally left as-is. Form builder spec passes (56 examples, 0 failures).

### DIFF
--- a/app/services/form_builder_service.rb
+++ b/app/services/form_builder_service.rb
@@ -61,7 +61,7 @@ class FormBuilderService
   # Header questions created by each section's builder method
   SECTION_HEADERS = {
     person_identifier: [],
-    person_contact_info: [ "Contact Information", "Mailing Address", "Agency / Organization Information" ],
+    person_contact_info: [ "Contact Information", "Mailing Address", "Organization Information" ],
     person_background: [ "Background Information" ],
     professional_info: [ "Professional Information" ],
     marketing: [ "Marketing" ],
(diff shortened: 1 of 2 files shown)


app/services/form_builder_service.rb
```diff
@@ -394,26 +394,26 @@ def build_person_contact_info_fields(form, position)
                         key: "phone_type", group: "person_contact_info", required: true,
                         width: :half, options: %w[Work Personal])
 
-    position = add_header(form, position, "Agency / Organization Information", group: "person_contact_info")
-    position = add_field(form, position, "Agency / Organization Name", :free_form_input_one_line,
+    position = add_header(form, position, "Organization Information", group: "person_contact_info")
+    position = add_field(form, position, "Organization Name", :free_form_input_one_line,
                          key: "agency_name", group: "person_contact_info", required: false, width: :half)
```
maebeale, line 399: 🤖 _From Claude:_ Labels changed to "Organization …" but the field identifiers (`agency_name`, `agency_street`, …) are deliberately left unchanged to avoid a data migration of existing `form_fields` and answers. The dev seed at line 1202 still looks them up by `agency_name`.

PR #1784, opened 2026-06-19
Distinguish scholarship registrations in event email subjects

maebeale:
- Event registration confirmation and cancellation emails read identically whether or not the registrant requested a scholarship, so staff couldn't tell an **event registration** apart from an **event registration scholarship** from the inbox or the notifications filter.
- This adds the registration type to the subject line so the two flows are distinguishable at a glance.

- Added `EventRegistration#registration_subject_noun`, driven by the `scholarship_requested` flag (set at registration time, so it's reliable before any `Scholarship` record exists) — returns `"event registration scholarship"` or `"event registration"`.
- Used it in all four event-registration email subjects (`EventMailer` registrant confirmation/cancellation, `NotificationMailer` admin FYI confirmation/cancellation).
- Changed the registrant **confirmation** wording from "confirmed" to "received" (subject + email body heading), since the email acknowledges receipt rather than implying staff have finished processing it.
- Kept the admin notification **topic filter** in sync: it matches subjects by substring, so splicing in "scholarship" would have silently dropped scholarship notifications from the existing topics. Added scholarship-specific topics and tightened the FYI-confirmed phrase so each filter maps to exactly one category.

### Anything else to add?
- Topic dropdown **labels** stay semantic ("…confirmed"); only the matched phrase strings changed — consistent with the existing FYI rows where the label and matched phrase already differ.
- Tests added/updated across the mailers, the model helper, and the notification topic filter. Full suite for the touched specs: **196 examples, 0 failures**; lint clean.


app/models/event_registration.rb
```diff
@@ -216,6 +216,16 @@ def scholarship?
...
   end
 
+  # Noun phrase distinguishing a scholarship-requested registration from a
+  # standard one in email subjects and notification labels (e.g.
+  # "event registration scholarship" vs "event registration"). Driven by the
+  # `scholarship_requested` flag, which is set at registration time, so it's
+  # reliable when the confirmation email goes out (before any Scholarship
+  # record exists).
+  def registration_subject_noun
+    scholarship_requested? ? "event registration scholarship" : "event registration"
```
maebeale: 🤖 _From Claude:_ Driven by `scholarship_requested` (set at registration time) rather than `scholarship?` (which checks for a `Scholarship` record) — at confirmation-email time no `Scholarship` record exists yet, so `scholarship?` would always be false here.

app/models/notification.rb
```diff
@@ -71,16 +71,22 @@ class Notification < ApplicationRecord
 
   EMAIL_TOPICS = [
     [ "Admin FYI (all)", "[FYI]" ],
-    [ "Admin FYI: event registration confirmed", "[FYI] New event registration" ],
+    # "by" keeps this from also matching the scholarship subject
+    # ("[FYI] New event registration scholarship by …"), which has its own topic.
+    [ "Admin FYI: event registration confirmed", "[FYI] New event registration by" ],
```
maebeale: 🤖 _From Claude:_ Topics match subjects by `LIKE %phrase%`. Splicing "scholarship" mid-phrase means scholarship subjects no longer match the plain topics, so I added scholarship-specific topics. The `by` here is what keeps this plain-confirmed topic from also matching the scholarship subject ("…New event registration scholarship by …").

app/models/notification.rb
```diff
@@ -71,16 +71,22 @@ class Notification < ApplicationRecord
...
     [ "Admin FYI: bulk payment", "[FYI] New bulk payment" ],
     [ "Admin FYI: idea submitted", "submission by" ],
     [ "Admin FYI: password reset", "[FYI] New password reset" ],
     [ "Admin FYI: workshop log submission", "New WorkshopLog submission" ],
     [ "Admin FYI: contact form submission", "contact form submission" ],
     [ "Contact: form confirmation", "We received your message" ],
     [ "Event: event registration cancelled", "Event registration cancelled" ],
-    [ "Event: event registration confirmed", "Event registration confirmed" ],
+    [ "Event: event registration scholarship cancelled", "Event registration scholarship cancelled" ],
+    [ "Event: event registration confirmed", "Event registration received" ],
```
maebeale: 🤖 _From Claude:_ Topic dropdown **labels** stay "…confirmed" (semantic) while the matched phrase is now "received" — matching the FYI rows above, where label and matched phrase already differ. Avoids changing the admin-facing dropdown text / saved filter values.

app/models/notification.rb
```diff
@@ -72,15 +72,19 @@ class Notification < ApplicationRecord
   EMAIL_TOPICS = [
     [ "Admin FYI (all)", "[FYI]" ],
     [ "Admin FYI: event registration confirmed", "[FYI] New event registration" ],
```
maebeale, line 74: 🤖 _From Claude:_ Update to my earlier "by" comment: the wording is now "event scholarship registration", so "scholarship" sits between "event" and "registration". The plain phrase "[FYI] New event registration" no longer prefixes the scholarship subject, so the `by` tightening was reverted — each topic still maps to exactly one category.

PR #1785, opened 2026-06-19
Show organization program status as a single-letter badge

maebeale:

- Surfaces each organization's AWBW **program lifecycle status** — New / Ongoing / Reinstate — as a compact single-letter badge (**N / O / R**) so staff can scan program standing at a glance.
- Reuses the same color language already used for program status on the event background dashboard (green = new, blue = ongoing, amber = reinstated), with the full label as a hover tooltip.

- **New shared helper** `OrganizationHelper#program_status_badge` renders the round single-letter badge. It accepts either the `:new`/`:ongoing`/`:reinstated` symbols (EventDashboard / index controller) or the `"New"`/`"Ongoing"`/`"Reinstate"` string from `Organization#program_status`, normalizing both.
- **Organizations index:** added an **admin-only `Program` column** (gated on `manage?`, styled `admin-only bg-blue-100`). Statuses are **bulk-computed in the controller** (`organization_program_statuses`) with two aggregate queries instead of calling `Organization#program_status` per row, avoiding an N+1 over each org's facilitator affiliations. The per-row fragment cache key now includes the status.
- **Event dashboard organizations dropdown:** badge shown next to each org name, driven by the dashboard's existing `program_status_by_organization`.
- **Organization edit:** program status shown in the affiliations section, alongside the existing "Affiliated since" / "Facilitations/program since" stats, with an info tooltip explaining the three states.

Status semantics (matching `Organization#program_status` with no recipient context):
- **N — New:** no facilitator affiliations yet
- **O — Ongoing:** has a currently-active facilitator
- **R — Reinstate:** had facilitators but all have ended

### Anything else to add?

- Tests cover all three badges on the index, the badge on the event dashboard dropdown (matched with `visible: :all` since the org list sits inside a collapsed `<details>`), and the status on the org edit page.
- No new full-page views, so no `page_bg_class` changes needed.

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -25,7 +25,7 @@ class OrganizationsController < ApplicationController
                                                 .group(:organization_id)
                                                 .distinct
                                                 .count(:person_id)
-      @program_statuses = organization_program_statuses(org_ids)
+      @program_statuses = Organization.program_statuses_by_id(org_ids)
 
       render :organization_results
     else
@@ -196,21 +196,6 @@ class OrganizationsController < ApplicationController
   end
 
   private
-
-  # Bulk program status (:new / :ongoing / :reinstated) for the listed orgs,
-  # keyed by id — mirrors Organization#program_status (no recipient context) but
-  # avoids per-row affiliation loads. An org with no facilitator affiliations is
-  # :new; with facilitators but none currently active it is :reinstated;
-  # otherwise :ongoing.
-  def organization_program_statuses(org_ids)
-    facilitator_scope = Affiliation.facilitators.where(organization_id: org_ids)
-    with_facilitators = facilitator_scope.distinct.pluck(:organization_id).to_set
-    with_active = facilitator_scope.active.distinct.pluck(:organization_id).to_set
-    org_ids.index_with do |id|
-      next :new if with_facilitators.exclude?(id)
-      with_active.include?(id) ? :ongoing : :reinstated
-    end
-  end
 
   def set_organization
     @organization = Organization.includes(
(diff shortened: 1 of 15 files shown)


app/controllers/organizations_controller.rb
```diff
@@ -196,6 +197,21 @@ def populations_served
...
+  # keyed by id — mirrors Organization#program_status (no recipient context) but
+  # avoids per-row affiliation loads. An org with no facilitator affiliations is
+  # :new; with facilitators but none currently active it is :reinstated;
+  # otherwise :ongoing.
+  def organization_program_statuses(org_ids)
+    facilitator_scope = Affiliation.facilitators.where(organization_id: org_ids)
+    with_facilitators = facilitator_scope.distinct.pluck(:organization_id).to_set
+    with_active = facilitator_scope.active.distinct.pluck(:organization_id).to_set
+    org_ids.index_with do |id|
+      next :new if with_facilitators.exclude?(id)
```
maebeale: 🤖 _From Claude:_ Statuses are bulk-computed here with two aggregate queries (any facilitator / active facilitator) rather than calling `Organization#program_status` per row, which would N+1 over each org's affiliations. Mirrors the recipient-less classification: no facilitators → new, all lapsed → reinstated, else ongoing.

app/helpers/organization_helper.rb
```diff
@@ -1,4 +1,30 @@
...
+  # Compact single-letter badge (N / O / R) for an organization's program
+  # status, with the full label as a tooltip. Accepts either a :new/:ongoing/
+  # :reinstated symbol (EventDashboard / index controller) or the "New"/"Ongoing"/
+  # "Reinstate" string from Organization#program_status. Returns nil when blank or
+  # unrecognized.
+  def program_status_badge(status)
+    return if status.blank?
+
+    normalized = status.to_s.downcase
+    key = normalized.start_with?("reinstat") ? :reinstated : normalized.to_sym
```
maebeale: 🤖 _From Claude:_ Normalized here so the badge accepts both the `:reinstated` symbol (EventDashboard/index) and the model's `"Reinstate"` string — the two callers disagree on tense, so we fold any `reinstat*` to one key.

PR #1786, opened 2026-06-19
Add credentials suffix to person profiles

maebeale:
Facilitators often hold professional credentials (e.g. LCSW, MD) that belong beside their name, but there was no way to record or display them. This adds a free-text `credentials` field and a per-person `profile_show_credentials` visibility flag.

A migration adds both columns (`profile_show_credentials` defaults to `true`, matching the existing `profile_show_*` flags). The person edit form gains a credentials text input and a "Show credentials" checkbox, both permitted in `person_params`. The public profile renders credentials as a comma suffix after the name (e.g. `Jane Doe, LCSW`), gated on the value being present and the flag enabled — mirroring the adjacent pronouns pattern.

### UI Testing Checklist
- [ ] Edit a person, enter credentials, enable "Show credentials" → suffix appears after the name on their profile
- [ ] Disable "Show credentials" → suffix is hidden even with credentials set
- [ ] Leave credentials blank → no suffix regardless of the flag

### Anything else to add?
The suffix only renders when credentials are present, so defaulting the flag to `true` is safe — existing people show nothing until they fill it in. Added request-spec coverage for the visibility flag.


db/migrate/20260619120508_add_credentials_to_people.rb
```diff
@@ -0,0 +1,14 @@
+class AddCredentialsToPeople < ActiveRecord::Migration[8.1]
+  def up
+    add_column :people, :credentials, :string unless column_exists?(:people, :credentials)
+
+    unless column_exists?(:people, :profile_show_credentials)
+      add_column :people, :profile_show_credentials, :boolean, default: true, null: false
```
maebeale, line 6: 🤖 _From Claude:_ Defaulted to `true` to match the existing `profile_show_*` flags. Safe because the suffix only renders when `credentials` is also present, so existing people show nothing until they fill it in.

maebeale:
image

maebeale:
image

PR #1790, opened 2026-06-19
WAIT: Show per-day add-to-calendar links for multi-day events

maebeale:

- Multi-day events run the same hours each day (e.g. Apr 21-23, 9 am-4:30 pm), but the add-to-calendar links generated a single calendar entry spanning the first day's start to the last day's end — which reads as one continuous overnight event.
- Emit a separate set of links per day instead, so each day is its own calendar entry running that day's hours.

**`EventDecorator#calendar_links`**
- Multi-day events now emit a separate set of links (Google · Apple · Outlook · Office 365 · Yahoo) **per day**, each prefixed with its date and running that day's hours.
- Same-day events are unchanged — one inline row.
- Extracted `calendar_links_for(start, end)` (builds the five provider links) and `calendar_day_ranges` (one start/end pair per calendar day).
- `.ics` download filenames now include the day so per-day downloads don't collide.

**View**
- `events/_registration_section.html.erb`: wrapper changed to `flex flex-wrap … justify-center` so per-day rows stack and center (matching the ticket). Each per-day row is centered within its width.

### UI Testing Checklist
- [ ] Registration ticket (`/registrations/:slug/ticket`) — per-day calendar links render centered for a multi-day event
- [ ] Event show page registration section — same
- [ ] Single-day event still shows one inline set of calendar links

### Anything else to add?

- Added decorator specs covering single-day vs. per-day link counts (with correct per-day date/time spans).
- The multi-day date-range **display** change (`EventDecorator#times`) is split out into #1791.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -62,89 +62,27 @@ class EventDecorator < ApplicationDecorator
     length ? description&.truncate(length) : description
   end
 
+  # "Add to calendar" links. A multi-day event runs the same hours each day
+  # (e.g. Apr 21-23, 9 am-4:30 pm), so we emit a separate set of links per day —
+  # one calendar entry per day — rather than a single block that would span
+  # overnight. Each day's row is prefixed with its date.
   def calendar_links
-    start_time   = object.start_date.utc.strftime("%Y%m%dT%H%M%SZ")
-    end_time     = object.end_date.utc.strftime("%Y%m%dT%H%M%SZ")
-    title_encoded = ERB::Util.url_encode(object.title)
+    ranges = calendar_day_ranges
+    return calendar_links_for(ranges.first[:start], ranges.first[:end]) if ranges.one?
 
-    has_url      = object.videoconference_url.present?
-    has_location = object.location.present?
-    location_name = has_location ? object.location.name : nil
-
-    # If both: URL in location field, physical location in description
-    # If only URL: URL in location field
-    # If only location: location in location field
-    event_description = object.rhino_description.to_plain_text
-
-    if has_url && has_location
-      cal_location = object.videoconference_url
-      base_description = "#{location_name}\n\n#{event_description}"
-    elsif has_url
-      cal_location = object.videoconference_url
-      base_description = event_description
-    elsif has_location
-      cal_location = location_name
-      base_description = event_description
-    else
-      cal_location = nil
-      base_description = event_description
+    rows = ranges.map do |range|
+      day_label = h.content_tag(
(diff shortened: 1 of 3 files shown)


app/decorators/event_decorator.rb
```diff
@@ -296,6 +235,105 @@ def bookmarkable_link
...
+  # One start/end pair per calendar day the event spans, each running the
+  # event's daily hours (start time-of-day to end time-of-day). A same-day event
+  # returns its actual start/end so single-day links are unchanged.
+  def calendar_day_ranges
+    s = object.start_date.in_time_zone(Time.zone)
+    e = (object.end_date || object.start_date).in_time_zone(Time.zone)
+    return [ { start: object.start_date, end: object.end_date || object.start_date } ] if s.to_date == e.to_date
+
+    (s.to_date..e.to_date).map do |day|
+      {
```
maebeale, line 292: 🤖 _From Claude:_ Same-day events short-circuit here and return their actual start/end, so single-day calendar links are byte-for-byte unchanged. Only multi-day events fan out into per-day ranges.

app/decorators/event_decorator.rb
```diff
@@ -296,6 +235,105 @@ def bookmarkable_link
...
+        start: Time.zone.local(day.year, day.month, day.day, s.hour, s.min),
+        end:   Time.zone.local(day.year, day.month, day.day, e.hour, e.min)
+      }
```
maebeale, line 295: 🤖 _From Claude:_ Each day reuses the event's daily hours (start time-of-day → end time-of-day), so a 3-day event yields three discrete entries rather than one overnight-spanning block.

app/decorators/event_decorator.rb
```diff
@@ -296,6 +235,105 @@ def bookmarkable_link
...
+      "https://outlook.office.com/owa/?rru=addevent" \
+        "&startdt=#{start_time}&enddt=#{end_time}" \
+        "&subject=#{title_encoded}&body=#{desc_encoded}&location=#{location_encoded}"
+
+    yahoo_link =
+      "https://calendar.yahoo.com/?v=60" \
+        "&title=#{title_encoded}&st=#{start_time}" \
+        "&et=#{end_time}&desc=#{desc_encoded}&in_loc=#{location_encoded}"
+
+    ics_date = start_date.in_time_zone(Time.zone).strftime("%b-%-d").downcase
```
maebeale, line 367: 🤖 _From Claude:_ The day is folded into the .ics filename so a user downloading every day's entry doesn't overwrite the same file.

PR #1793, opened 2026-06-19
Add an admin sample-ticket preview for events

maebeale:
Admins had no way to preview a registration ticket without creating a throwaway registration. This adds a **Sample ticket** preview — a non-persisted example registration with every registrant-chosen option enabled — so admins can see exactly what registrants will see (magic callouts, payment states, manage actions) for their event.

- New admin-only `events#sample_ticket` action (gated by `dashboard?`) builds an in-memory `EventRegistration` with all flags on, a fake "Sample Registrant", and a sentinel slug so the callout route helpers generate without raising — nothing is saved.
- The shared `_ticket` partial gained a `preview` flag that disables the state-changing buttons (pay, resend, cancel); `_callout_card` gained a `linked` flag so the magic/admin callout cards render non-navigating in the preview.
- Linked from the dashboard quick links and the **Form actions** menu (the latter passing `return_to: "registrants"` so the eyebrow returns correctly); registered the new view's `page_bg_class` and added request specs covering admin render, non-admin redirect, and that no registration is created.

### UI Testing Checklist
- [ ] Dashboard → "Sample ticket" opens the preview in a new tab; eyebrow returns to Dashboard
- [ ] Registrants → Form actions → "Sample ticket" opens the preview; eyebrow returns to Registrants
- [ ] Preview shows all sections including the magic callout cards
- [ ] Pay / resend / cancel buttons are visibly disabled; callout cards are non-navigating
- [ ] Non-admin cannot reach the preview

### Anything else to add?
Rebased onto `main` to pick up the new `MagicTicketCallouts` system, so the preview renders the current callout cards. The preview reflects the event's real configuration plus all registrant options, so payment sections only appear when the event actually has a cost. Replaces #1779.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -2,7 +2,7 @@ class EventsController < ApplicationController
   include AhoyTracking, TagAssignable
   skip_before_action :authenticate_user!, only: [ :index, :show, :staff, :details, :ce_hours ]
   skip_before_action :verify_authenticity_token, only: [ :preview ]
-  before_action :set_event, only: %i[ show edit update destroy preview dashboard background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder confirm_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
+  before_action :set_event, only: %i[ show edit update destroy preview dashboard sample_ticket background registrants details ce_hours staff edit_staff update_staff recipients bulk_payments preview_reminder confirm_reminder send_reminder copy_registration_form allocate_bulk_payment create_bulk_payment ]
 
   def index
     authorize!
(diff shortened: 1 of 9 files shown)


app/controllers/events_controller.rb
```diff
@@ -42,6 +42,33 @@ def dashboard
...
+  # the state-changing buttons (pay, resend, cancel) and renders the callout
+  # cards non-navigating. A sentinel slug lets the callout route helpers build
+  # without raising, since the sample isn't persisted.
+  def sample_ticket
+    authorize! @event, to: :dashboard?
+
+    registrant = Person.new(first_name: "Sample", last_name: "Registrant")
+    @event_registration = @event.event_registrations.new(
+      registrant: registrant,
+      slug: "sample",
```
maebeale, line 58: 🤖 _From Claude:_ Sentinel slug so `MagicTicketCallouts` can build the per-registration route helpers — they eagerly compute hrefs from `registration.slug`, which would raise on this unsaved record. The cards render non-navigating in preview, so the slug never reaches a real request.

app/views/event_registrations/_callout_card.html.erb
```diff
@@ -3,14 +3,17 @@
     locals: callout (responds to #theme, #display_icon_class, #title, #subtitle).
             href / target / trailing_icon may be passed explicitly (admin callouts
             compute href from a route) or read off the callout when it carries them
-            (MagicTicketCallouts::Card). %>
+            (MagicTicketCallouts::Card).
+            linked (default true): when false the card renders as a non-navigating
+            div — the sample-ticket preview has no real per-registration pages. %>
 <% theme = callout.theme %>
+<% linked = local_assigns.fetch(:linked, true) %>
```
maebeale, line 10: 🤖 _From Claude:_ `linked` lets the same card markup render as a non-navigating `<div>` in the preview (no real per-registration pages exist), reusing the partial for both the admin ticket and the sample preview rather than duplicating it.

PR #1794, opened 2026-06-19
Add a scholarship agreement recipients accept before it shows as awarded

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — new columns + a display-gating rule across recipient-facing surfaces

- Scholarship recipients must **accept the terms** of their scholarship. Staff record/see it, recipients accept it themselves, and — per product — a scholarship is **not shown as awarded until the agreement is signed** (the underlying record and allocation are unchanged; only the display is gated).

**Data**
- One column: `agreement_signed_at` (datetime; nil = unsigned). `agreement_signed` is a **virtual** boolean that reads the timestamp and stamps/clears it on write — single source of truth, no sync callback.

**Accept before "awarded" (display only)**
- `EventRegistration#scholarship_awarded?` = scholarship present **and** agreement signed. Drives:
  - **Ticket callout card** — amber "Review and accept your scholarship agreement", no amount chip, until signed; then the award subtitle + amount.
  - **Scholarship page** — "Amount offered" + "Pending agreement" chip until signed; "Amount awarded" + tasks chip after.

**Recipient accepts**
- Scholarship page shows the agreement above the details with a single green **Agree** button; `POST events/callouts#sign_agreement` stamps it. Once signed: "Agreement signed · <date>".

**Staff**
- Editable toggle on the scholarship form (fuchsia fill); "Agreement signed: Yes · <date>" on the show page; "Agreement signed / pending" chip on the reg-edit scholarship card.

### UI Testing Checklist

- [ ] Recipient, unsigned: ticket card is amber "Review and accept…" with no amount; scholarship page shows "Amount offered" + "Pending agreement" + green Agree button.
- [ ] After clicking Agree: card shows the award + amount; page shows "Amount awarded", tasks chip, and "Agreement signed · <date>".
- [ ] Staff toggle flips the reg-edit chip and show-page status; the allocation/amount never changes.

### Anything else to add?

- Rebased across the CE cutover / callout consolidation; the toggle now uses main's CSS fuchsia fill.
- Gating is display-only by design — no change to allocation or what "exists".

### DIFF
--- a/app/controllers/events/callouts_controller.rb
+++ b/app/controllers/events/callouts_controller.rb
@@ -64,6 +64,23 @@ module Events
       @form_responses_available = @event.registration_form&.form_submissions&.exists?(person: @event_registration.registrant)
     end
 
+    # Records the recipient agreeing, from their scholarship page, to complete the
+    # scholarship's tasks. Only "Yes" signs it; the model stamps agreement_signed_at.
+    def sign_agreement
+      scholarship = @event_registration.scholarships.first
+      unless scholarship
+        redirect_to registration_scholarship_path(@event_registration.slug)
+        return
+      end
+
+      if params[:agreement] == "yes"
+        scholarship.update!(agreement_signed: true) unless scholarship.agreement_signed?
+        redirect_to registration_scholarship_path(@event_registration.slug), notice: "Thanks — your agreement has been recorded."
+      else
+        redirect_to registration_scholarship_path(@event_registration.slug), alert: "Please select Yes to agree to complete the tasks."
+      end
+    end
+
     # CE hours status: hours, amount owed, and license number.
     def ce
       case params[:checkout]
--- a/app/controllers/events/callouts_controller.rb
+++ b/app/controllers/events/callouts_controller.rb
@@ -65,7 +65,8 @@ module Events
     end
 
     # Records the recipient agreeing, from their scholarship page, to complete the
-    # scholarship's tasks. Only "Yes" signs it; the model stamps agreement_signed_at.
+    # scholarship's tasks. The Agree button submits agreement=yes, which stamps
+    # agreement_signed_at via the model.
     def sign_agreement
       scholarship = @event_registration.scholarships.first
       unless scholarship
@@ -77,7 +78,7 @@ module Events
         scholarship.update!(agreement_signed: true) unless scholarship.agreement_signed?
         redirect_to registration_scholarship_path(@event_registration.slug), notice: "Thanks — your agreement has been recorded."
       else
-        redirect_to registration_scholarship_path(@event_registration.slug), alert: "Please select Yes to agree to complete the tasks."
+        redirect_to registration_scholarship_path(@event_registration.slug), alert: "Something went wrong recording your agreement. Please try again."
       end
     end
 
--- a/app/models/scholarship.rb
+++ b/app/models/scholarship.rb
@@ -12,6 +12,7 @@ class Scholarship < ApplicationRecord
   validate :recipient_must_match_allocation_registrant
   validate :within_grant_budget, if: :grant
 
+  before_save :stamp_agreement_signed_at, if: :will_save_change_to_agreement_signed?
   after_update :sync_allocation_amount, if: -> { saved_change_to_amount_cents? }
   after_create_commit :flag_event_registration_scholarship_requested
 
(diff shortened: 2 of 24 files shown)


app/views/scholarships/_form.html.erb
```diff
@@ -111,6 +111,22 @@
         </div>
       </div>
     <% end %>
+
+    <%# Scholarship agreement: an informational toggle shown in either layout,
+        recording whether the recipient has signed their scholarship agreement.
+        It surfaces back to the recipient on their scholarship callout page. %>
```
maebeale: 🤖 _From Claude:_ Placed the agreement toggle here, after both conditional layout branches, so it renders for event-funded *and* grant/standalone scholarships from one block — rather than duplicating it inside each branch like the amount/tasks fields.

app/views/events/callouts/scholarship.html.erb
```diff
@@ -31,15 +31,26 @@
...
+            <% if @scholarship.tasks_completed? %>
+              <span class="inline-flex items-center gap-1 rounded-full border <%= DomainTheme.border_class_for(:scholarships, intensity: 200) %> <%= DomainTheme.bg_class_for(:scholarships, intensity: 50) %> px-3 py-1 text-sm font-medium <%= DomainTheme.text_class_for(:scholarships, intensity: 700) %>">
+                <i class="fa-solid fa-circle-check"></i> Tasks completed
+              </span>
+            <% else %>
+              <span class="inline-flex items-center gap-1 rounded-full border border-amber-200 bg-amber-50 px-3 py-1 text-sm font-medium text-amber-700">
+                <i class="fa-solid fa-clock"></i> Tasks outstanding
+              </span>
+            <% end %>
+            <% if @scholarship.agreement_signed? %>
```
maebeale: 🤖 _From Claude:_ The recipient-facing chip parallels the existing tasks chip — scholarships theme when signed, amber when outstanding. Only shown on an awarded scholarship; the requested-but-not-awarded state has no scholarship record and so no agreement to surface.

app/models/scholarship.rb
```diff
@@ -44,6 +45,14 @@ def recipient_must_match_allocation_registrant
     end
   end
 
+  # Keep the signed-at timestamp in step with the flag however it's flipped — the
+  # admin toggle on the form or the recipient agreeing on their callout page.
+  # Stamp the moment it turns true (preserving an existing time), clear it when
+  # turned back off.
+  def stamp_agreement_signed_at
```
maebeale: 🤖 _From Claude:_ Stamping lives in one model callback rather than in each caller, so the timestamp stays correct whether the admin toggles the flag on the form or the recipient agrees via the public POST. `|| Time.current` preserves an existing time across unrelated saves; clearing on un-sign keeps the column honest.

app/controllers/events/callouts_controller.rb
```diff
@@ -43,6 +43,23 @@ def scholarship
...
+    # Records the recipient agreeing, from their scholarship page, to complete the
+    # scholarship's tasks. Only "Yes" signs it; the model stamps agreement_signed_at.
+    def sign_agreement
+      scholarship = @event_registration.scholarships.first
+      unless scholarship
+        redirect_to registration_scholarship_path(@event_registration.slug)
+        return
+      end
+
+      if params[:agreement] == "yes"
```
maebeale, line 77: 🤖 _From Claude:_ Only an explicit "yes" signs — anything else falls through to an alert, so a mis-submit can't silently record consent. The action is public/slug-authorized like the sibling callout pages, and guards on an awarded scholarship existing.

app/frontend/javascript/controllers/save_state_toggle_controller.js
```diff
@@ -1,9 +1,10 @@
...
-// Colors the "Requested" toggle to signal save state: amber while the choice is
-// pending (changed but not yet saved with the registration form), the
-// scholarship theme color once it matches the stored "on" value, and neutral
-// gray when stored as off.
+// Colors a toggle's track to signal save state: amber while the choice is
+// pending (changed but not yet saved with the form), the scholarship theme
+// color once it matches the stored "on" value, and neutral gray when stored as
+// off. Shared by the registration "Requested" toggle and the scholarship form's
+// tasks/agreement toggles.
 export default class extends Controller {
```
maebeale: 🤖 _From Claude:_ Renamed from `scholarship-requested` rather than cloned — the logic was already generic (checkbox + track targets, an initial-value), so the same controller now drives the registration "Requested" toggle and the scholarship form's tasks/agreement toggles. Verified the original caller still works after the rename.

PR #1795, opened 2026-06-21
Default sample ticket preview to a typical registrant, with options toggle

maebeale:
The admin sample-ticket preview previously turned on every registrant option (scholarship, CE, W-9/invoice, shoutout), which isn't indicative of what most real tickets look like. This defaults the preview to a plain, typical registrant so admins see the common case, while still allowing a full "everything on" view on demand.

`EventsController#sample_ticket` now builds the in-memory sample with those options off by default and flips them all on when `?options=all` is passed; the dashboard and Form-actions-menu links keep pointing at the default. An admin-only toggle sits in the top-right of the preview box — "Show all options" reloads with `?options=all`, "Show typical ticket" returns to the default — preserving `return_to` so the eyebrow still points home, and the banner copy adapts to the active mode.

### UI Testing Checklist
- [ ] Sample ticket from the dashboard (and Form actions menu) shows a typical registrant — no scholarship or CE callout cards.
- [ ] "Show all options" reloads the preview with scholarship, CE, and other sections present; "Show typical ticket" toggles back.
- [ ] The eyebrow returns to the correct origin (Dashboard vs Registrants) in both modes.

### Anything else to add?
Stacked on #1793 (the original sample-ticket preview), which has since merged to main, so this targets main directly.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -43,28 +43,31 @@ class EventsController < ApplicationController
   end
 
   # Admin preview of the registration ticket. Builds an in-memory sample
-  # registration — never saved — with every registrant-chosen option turned on,
-  # so admins can see what their event's ticket looks like with all sections
-  # present. The ticket partial renders in `preview: true` mode, which disables
-  # the state-changing buttons (pay, resend, cancel) and renders the callout
-  # cards non-navigating. A sentinel slug lets the callout route helpers build
-  # without raising, since the sample isn't persisted.
+  # registration — never saved. By default it models a typical registrant (no
+  # scholarship, no continuing education, no extra requests) so the preview is
+  # indicative of what most tickets look like. The "Show all options" toggle
+  # (?options=all) turns on every registrant-chosen option at once so admins can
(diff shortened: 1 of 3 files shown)


app/controllers/events_controller.rb
```diff
@@ -43,28 +43,31 @@ def dashboard
...
+  # indicative of what most tickets look like. The "Show all options" toggle
+  # (?options=all) turns on every registrant-chosen option at once so admins can
+  # see every section present. The ticket partial renders in `preview: true`
+  # mode, which disables the state-changing buttons (pay, resend, cancel) and
+  # renders the callout cards non-navigating. A sentinel slug lets the callout
+  # route helpers build without raising, since the sample isn't persisted.
   def sample_ticket
     authorize! @event, to: :dashboard?
 
+    @show_all_options = params[:options] == "all"
```
maebeale, line 57: 🤖 _From Claude:_ All registrant-chosen options (incl. W-9/invoice/shoutout, which don't surface a ticket card) are gated on this single flag so "Show all options" means *all* of them, while the default stays a plain registrant.

PR #1796, opened 2026-06-21
Link ticket callouts to resources; show docs in a registrant resource page

maebeale:

Give registrants a consistent, in-context way to view a document from their ticket: admins can attach a Resource to a custom ticket callout, and the app's hard-coded handout/forms document links now open the same registrant resource page (PDF first-page preview + download) instead of dropping the registrant out to a resource's public page or a raw download.

- Added a nullable `resource_id` FK + `belongs_to :resource` on `RegistrationTicketCallout`, a "Linked resource" select in the callout fields (with `:resource_id` permitted), and rendered the resource on the callout's detail page.
- Added a slug-scoped resource viewer (`registration_resource`) that shows a Resource in the shared `events/callouts/_callout_page` chrome with a back-to-ticket eyebrow; the handouts cards and the forms "letter to supervisors" card now link to it (same tab). A shared `_resource_body` partial renders the preview + download identically on the viewer and the admin callout page.
- Callout back links always return to the ticket (the admin callout page drops its back-to-event fallback; the eyebrow is omitted when there's no registration slug), and the registrant invoice's back link was relabeled "Back to ticket".

### UI Testing Checklist
- [ ] Link a Resource (with a PDF) to a ticket callout and confirm its page shows the PDF preview + Download, in the same chrome as the other callout pages.
- [ ] From a ticket, open Handouts and the Forms "letter to supervisors" link and confirm each opens the resource page (preview + download) with a "Back to ticket" link, in the same tab.
- [ ] Confirm a callout with a linked resource but no content still renders.
- [ ] Confirm the registrant invoice page's back link reads "Back to ticket".

### Anything else to add?
The forms W-9 (a static PDF) and View invoice (already a registrant page) links are unchanged — only Resource-backed links route through the new viewer. The shared `_callout_page` back-link default is unchanged for the existing magic callout pages.

### DIFF
--- a/app/controllers/events/callouts_controller.rb
+++ b/app/controllers/events/callouts_controller.rb
@@ -54,15 +54,25 @@ module Events
     end
 
     # Handouts page: callout-card links to the training worksheet/handout
-    # resources, in display order, each opening in a new tab.
+    # resources, in display order, each opening its own registrant resource page
+    # (PDF preview + download, with a back-to-ticket eyebrow).
     def handouts
       by_title = Resource.where(title: HANDOUT_RESOURCE_TITLES).index_by(&:title)
       @handout_cards = HANDOUT_RESOURCE_TITLES.filter_map do |title|
         resource = by_title[title]
         next unless resource
         resource_card(icon: "fa-solid fa-file-pdf", title: resource.title,
-                      subtitle: "Open this training resource", href: resource_path(resource))
+                      subtitle: "Open this training resource",
+                      href: registration_resource_path(@event_registration.slug, resource), target: nil)
       end
+    end
+
+    # Registrant-facing page for a single Resource, shown in the shared callout
+    # chrome: the PDF first-page preview and a download button. Reached from the
+    # handouts/forms callouts so a registrant views a document without leaving the
+    # ticket context.
+    def resource
+      @resource = Resource.find(params[:resource_id]).decorate
     end
 
     # Facilitator Portal access info, with a link to the home screen.
@@ -107,15 +117,17 @@ module Events
       if letter
         cards << resource_card(icon: "fa-solid fa-file-arrow-down", title: "Letter to supervisors",
                                subtitle: "Share to request release time",
-                               href: resource_download_path(letter), trailing_icon: "fa-solid fa-download")
+                               href: registration_resource_path(@event_registration.slug, letter), target: nil)
       end
       cards
     end
 
-    # A blue, new-tab callout card linking to an external/resource document.
-    def resource_card(icon:, title:, subtitle:, href:, trailing_icon: "fa-solid fa-arrow-right")
+    # A blue callout card linking to a document. External/static links open in a
+    # new tab (target: "_blank"); registrant resource pages stay in-tab so the
+    # back-to-ticket eyebrow works (pass target: nil).
+    def resource_card(icon:, title:, subtitle:, href:, target: "_blank", trailing_icon: "fa-solid fa-arrow-right")
       MagicTicketCallouts::Card.new(icon_class: icon, color: "blue", title: title, subtitle: subtitle,
-                                    href: href, target: "_blank", trailing_icon: trailing_icon)
+                                    href: href, target: target, trailing_icon: trailing_icon)
     end
   end
 end
--- a/app/controllers/registration_ticket_callouts_controller.rb
+++ b/app/controllers/registration_ticket_callouts_controller.rb
@@ -4,13 +4,13 @@ class RegistrationTicketCalloutsController < ApplicationController
 
   # Public detail page for a single registration ticket callout, linked from the
   # call-out on the registration ticket (mirrors the events#details / #ce_hours
-  # pages). When the callout has no description there is nothing to read, so fall
-  # back to the event page.
+  # pages). With no description and no linked resource there is nothing to show,
+  # so fall back to the event page.
   def show
     @callout = @event.registration_ticket_callouts.find(params[:id])
     authorize! @callout, to: :show?
 
-    if @callout.description.blank?
+    if @callout.description.blank? && @callout.resource.nil?
       redirect_to event_path(@event, reg: params[:reg].presence)
       return
     end
--- a/app/views/events/callouts/_callout_page.html.erb
+++ b/app/views/events/callouts/_callout_page.html.erb
@@ -1,9 +1,11 @@
 <%# Shared chrome for registrant callout show pages: back-to-ticket eyebrow +
     branded header, with the body supplied as a block via render(layout:). Callers
-    set their own page_bg_class / page_title and pass a `title` local. Uses
-    @event_registration and @event. %>
+    set their own page_bg_class / page_title and pass a `title` local. Uses @event.
+    The eyebrow defaults to the registrant's ticket (uses @event_registration);
+    callers without a registration can override it with a `back_path` local. %>
+<% back_path = local_assigns.fetch(:back_path) { registration_ticket_path(@event_registration.slug) } %>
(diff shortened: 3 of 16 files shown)


db/migrate/20260618202115_add_resource_to_registration_ticket_callouts.rb
```diff
@@ -0,0 +1,14 @@
+class AddResourceToRegistrationTicketCallouts < ActiveRecord::Migration[8.1]
+  def up
+    # resources.id is a legacy `int` (not bigint), so the FK column must match.
+    add_reference :registration_ticket_callouts, :resource, type: :integer, null: true, index: true
```
maebeale, line 4: 🤖 _From Claude:_ `resources.id` is a legacy `int` (not `bigint`), so the FK column is declared `type: :integer` to match — a default `add_reference` would create a `bigint` and fail the FK.

app/controllers/registration_ticket_callouts_controller.rb
```diff
@@ -4,13 +4,13 @@ class RegistrationTicketCalloutsController < ApplicationController
...
-  # pages). When the callout has no description there is nothing to read, so fall
-  # back to the event page.
+  # pages). With no description and no linked resource there is nothing to show,
+  # so fall back to the event page.
   def show
     @callout = @event.registration_ticket_callouts.find(params[:id])
     authorize! @callout, to: :show?
 
-    if @callout.description.blank?
+    if @callout.description.blank? && @callout.resource.nil?
```
maebeale, line 13: 🤖 _From Claude:_ Guard relaxed to also keep the page when a resource is linked, so a callout with a downloadable resource but no body text still renders instead of redirecting to the event.

app/views/registration_ticket_callouts/show.html.erb
```diff
@@ -32,6 +40,14 @@
       <div class="rich-label prose max-w-none text-gray-700 leading-relaxed prose-headings:text-gray-800 prose-a:text-blue-700">
         <%= form_label_html(@callout.description) %>
       </div>
+
+      <% if resource %>
```
maebeale: 🤖 _From Claude:_ Reuses the same `assets/display_assets` render as `resources/show`, so the PDF first-page preview comes for free via `DisplayImagePresenter` (`display_image` falls back to the downloadable asset).

app/views/registration_ticket_callouts/show.html.erb
```diff
@@ -1,37 +1,33 @@
...
+<% content_for(:page_title, "#{@callout.title} — #{@event.title}") %>
 <% reg_slug = params[:reg].presence %>
 <% back_path = reg_slug ? registration_ticket_path(reg_slug) : event_path(@event) %>
-<% back_label = reg_slug ? "Back to ticket" : "Back to event" %>
-<div class="max-w-3xl mx-auto px-4 pt-4">
-  <%= link_to back_path, class: "inline-flex items-center gap-1.5 text-sm text-gray-500 hover:text-gray-700 px-2 py-1" do %>
-    <i class="fa-solid fa-arrow-left text-xs"></i> <%= back_label %>
+<% resource = @callout.resource&.decorate %>
+<% downloadable = resource if resource&.downloadable_asset&.file&.attached? %>
+<%= render layout: "events/callouts/callout_page", locals: { title: @callout.title, back_path: back_path } do %>
```
maebeale, line 6: 🤖 _From Claude:_ Renders through the shared callout-page layout so this page matches the magic callout pages (handouts, forms, CE). The PDF first-page preview comes for free from `assets/display_assets` (display_image falls back to the downloadable asset via `DisplayImagePresenter`).

app/views/events/callouts/_callout_page.html.erb
```diff
@@ -1,9 +1,11 @@
 <%# Shared chrome for registrant callout show pages: back-to-ticket eyebrow +
     branded header, with the body supplied as a block via render(layout:). Callers
-    set their own page_bg_class / page_title and pass a `title` local. Uses
-    @event_registration and @event. %>
+    set their own page_bg_class / page_title and pass a `title` local. Uses @event.
+    The eyebrow defaults to the registrant's ticket (uses @event_registration);
+    callers without a registration can override it with a `back_path` local. %>
+<% back_path = local_assigns.fetch(:back_path) { registration_ticket_path(@event_registration.slug) } %>
```
maebeale, line 7: 🤖 _From Claude:_ Back link is now an optional `back_path` local; the default preserves existing behavior for the 9 magic callout pages (they pass only `title`), while this event-scoped page overrides it to fall back to the event when there is no registration slug.

db/migrate/20260618202115_add_resource_to_registration_ticket_callouts.rb
maebeale, line 4: 🤖 _From Claude:_ `resources.id` is a legacy `int` (not `bigint`), so the FK column is declared `type: :integer` to match — a default `add_reference` would create a `bigint` and fail the FK.

app/controllers/events/callouts_controller.rb
```diff
@@ -54,17 +54,27 @@ def forms
...
+    # (PDF preview + download, with a back-to-ticket eyebrow).
     def handouts
       by_title = Resource.where(title: HANDOUT_RESOURCE_TITLES).index_by(&:title)
       @handout_cards = HANDOUT_RESOURCE_TITLES.filter_map do |title|
         resource = by_title[title]
         next unless resource
         resource_card(icon: "fa-solid fa-file-pdf", title: resource.title,
-                      subtitle: "Open this training resource", href: resource_path(resource))
+                      subtitle: "Open this training resource",
+                      href: registration_resource_path(@event_registration.slug, resource), target: nil)
```
maebeale, line 66: 🤖 _From Claude:_ Handout (and the forms letter) cards now point at the slug-scoped resource viewer in the same tab (`target: nil`) so the viewer’s back-to-ticket eyebrow works, instead of opening the resource’s public page in a new tab.

app/controllers/events/callouts_controller.rb
```diff
@@ -54,17 +54,27 @@ def forms
...
       end
     end
 
+    # Registrant-facing page for a single Resource, shown in the shared callout
+    # chrome: the PDF first-page preview and a download button. Reached from the
+    # handouts/forms callouts so a registrant views a document without leaving the
+    # ticket context.
+    def resource
```
maebeale, line 74: 🤖 _From Claude:_ New registrant-scoped page that renders any Resource (PDF preview + download) in the shared callout chrome; authorization is the slug (via the existing `authorize_callout`), matching the other public callout pages.

app/views/events/callouts/_callout_page.html.erb
```diff
@@ -1,13 +1,18 @@
...
-              class: "inline-flex items-center gap-1.5 text-sm text-gray-500 hover:text-gray-700 px-2 py-1" do %>
-    <i class="fa-solid fa-arrow-left text-xs"></i> Back to ticket
-  <% end %>
-</div>
+    set their own page_bg_class / page_title and pass a `title` local. Uses @event.
+    The eyebrow defaults to the registrant's ticket (uses @event_registration);
+    callers without a registration can override it with a `back_path` local, or
+    pass `back_path: nil` to drop the eyebrow. %>
+<% back_path = local_assigns.fetch(:back_path) { registration_ticket_path(@event_registration.slug) } %>
+<% if back_path %>
```
maebeale, line 8: 🤖 _From Claude:_ Back link is rendered only when a `back_path` is present, so the admin callout page can pass `nil` (no registration slug) to omit the eyebrow rather than fall back to the event. Existing magic pages pass no `back_path` and keep the ticket default.

PR #1797, opened 2026-06-21
Remove inline affiliation-title edit from org-link editor

maebeale:
- Removes the inline affiliation-title edit from the organizations & affiliations editor (the "org popup" reached from the registrants index org pills). The inline editing surface added clutter, and admins manage affiliation titles from the person profile's affiliations editor instead.

- Non-facilitator titles now render as plain, non-editable pills (matching how facilitator pills already rendered), and the supporting `AffiliationsController#update` action, `update?` policy rule, `affiliations` PATCH route, and `affiliation_title_edit` Stimulus controller were all removed.
- Updated/removed the now-obsolete request specs and refreshed AGENTS.md (Stimulus controller list + count).

### UI Testing Checklist
- [ ] On a registrant's org-link editor, affiliation titles show as plain pills with no pencil/edit control.
- [ ] Facilitator and inactive pills still render correctly, and the "title differs from form" badge still appears when applicable.

### Anything else to add?
- Behavior-only removal; no data changes. `AffiliationsController#destroy` is unchanged.


config/routes.rb
```diff
@@ -201,7 +201,7 @@
 
   resources :refunds, only: [ :new, :create, :show ]
   resources :organization_statuses
-  resources :affiliations
+  resources :affiliations, only: :destroy
```
maebeale, line 204: 🤖 _From Claude:_ Narrowed to `only: :destroy` since `update` was the only other action and it existed solely for the inline title edit being removed here.

PR #1798, opened 2026-06-21
Default ticket-callout icon to the arrow; move icon field after colour

maebeale:
When admins add a registration-ticket callout, the **Icon** field rendered empty, giving no hint of a reasonable value. This pre-fills new callouts with the arrow icon and reorders the row so the most niche field (icon) comes last.

- Set an `attribute :icon_class` default of the "action" arrow (`fa-solid fa-arrow-right`) on `RegistrationTicketCallout`, so newly built callouts (including cocoon's "+ Add callout" template) start pre-filled. Loaded records keep their stored value, and a blank one still falls back via `#display_icon_class`, so existing callouts are unaffected.
- Reordered the callout fields row from Type → Icon → Colour to **Type → Colour → Icon**.

### Anything else to add?
The model spec still passes; its tests set `icon_class` explicitly so they're unaffected by the new default.

### DIFF
--- a/app/models/registration_ticket_callout.rb
+++ b/app/models/registration_ticket_callout.rb
@@ -11,6 +11,11 @@ class RegistrationTicketCallout < ApplicationRecord
   DEFAULT_ICONS = { "action" => "fa-solid fa-arrow-right", "reference" => "fa-solid fa-circle-info" }.freeze
   DEFAULT_COLORS = { "action" => "orange", "reference" => "indigo" }.freeze
 
+  # New callouts start with the arrow icon pre-filled (the "action" default) so
+  # admins see a sensible value rather than an empty field. Loaded records keep
+  # their stored value; a blank one still falls back via #display_icon_class.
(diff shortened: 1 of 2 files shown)


app/models/registration_ticket_callout.rb
```diff
@@ -11,6 +11,11 @@ class RegistrationTicketCallout < ApplicationRecord
   DEFAULT_ICONS = { "action" => "fa-solid fa-arrow-right", "reference" => "fa-solid fa-circle-info" }.freeze
   DEFAULT_COLORS = { "action" => "orange", "reference" => "indigo" }.freeze
 
+  # New callouts start with the arrow icon pre-filled (the "action" default) so
```
maebeale, line 14: 🤖 _From Claude:_ Using `attribute ... default:` only pre-fills **new** records — rows loaded from the DB keep their stored value (including blank, which still falls back through `#display_icon_class`), so existing callouts are unchanged.

PR #1800, opened 2026-06-21
Warn before unlinking an org from a registration

maebeale:
- The "Remove" control on a registration's *Organizations & affiliations* page only severs the registration↔organization link; it intentionally leaves the person's global Affiliations alone (those are managed on the Person record).
- That distinction wasn't visible, so an admin could reasonably expect it to delete the affiliation too.

- Renamed the button "Remove" → "Unlink" so it reads as removing a link, not deleting a record.
- Added a Turbo confirm dialog spelling out that unlinking does NOT delete any of the person's Affiliations, and that affiliations are edited on the Person record.
- Documented the intentional UX choice in the view and the `unlink_organization` controller action, and added a request spec asserting the warning is present (existing spec already covers that unlinking preserves the affiliation).

### UI Testing Checklist
- [ ] On a registration with a linked org, click **Unlink** and confirm the dialog explains affiliations are kept.
- [ ] Confirm unlinking removes the org from the registration but the person's Affiliation still appears on their Person record.

### Anything else to add?
- Uses the app's native Turbo confirm convention (no custom modal). The flash notice now reads "… unlinked from this registration."

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -217,10 +217,12 @@ class EventRegistrationsController < ApplicationController
     authorize! @event_registration, to: :unlink_organization?
     organization = Organization.find(params[:organization_id])
 
-    # Removes only the registration-scoped link; the person's global affiliation is left intact.
+    # Intentional UX choice: "Unlink" only removes the org from this registration and
+    # deliberately leaves the person's global Affiliation intact. Affiliations are managed
+    # on the Person record, not here — the Unlink button's confirm dialog warns about this.
     @event_registration.event_registration_organizations.where(organization_id: organization.id).destroy_all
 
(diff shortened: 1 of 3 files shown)


app/controllers/event_registrations_controller.rb
```diff
@@ -217,10 +217,12 @@ def unlink_organization
     authorize! @event_registration, to: :unlink_organization?
     organization = Organization.find(params[:organization_id])
 
-    # Removes only the registration-scoped link; the person's global affiliation is left intact.
+    # Intentional UX choice: "Unlink" only removes the org from this registration and
+    # deliberately leaves the person's global Affiliation intact. Affiliations are managed
```
maebeale, line 221: 🤖 _From Claude:_ Intentional: unlinking only destroys the `event_registration_organizations` join row and never touches `Affiliation`, so the person's profile-level affiliations survive. The renamed "Unlink" button + confirm dialog make that explicit to admins.

PR #1801, opened 2026-06-21
Add videoconference passcode and meeting-room details to events

maebeale:
Registrants joining a virtual event need the meeting ID and passcode — not just a join link — so they can dial in manually, recover from an expired link, or rejoin from a calendar reminder. We previously stored only the URL, so this adds a passcode field and surfaces the full connection details everywhere the link appears.

- Added a `videoconference_passcode` column (migration + admin form field + permitted param) and a decorator that parses the meeting ID (Zoom, grouped like Zoom's UI) or meeting code (Google Meet) out of the join URL.
- Surfaced the meeting ID/code + passcode wherever the link is already revealed — the registration ticket, event show page, confirmation/FYI emails, and the videoconference callout page — via a shared `_videoconference_connection` partial, keeping them behind the same join-link visibility gating so they never leak on public cards or previews.
- Embedded the join link, meeting ID/code, and passcode into the add-to-calendar links (Google/Apple/Outlook/Office 365/Yahoo) so the details travel into the registrant's calendar, and added decorator + request specs covering URL parsing and display.

### UI Testing Checklist
- [ ] Admin event form saves a videoconference passcode
- [ ] Registration ticket and event page show the meeting ID/code + passcode only when the join link is revealed
- [ ] Videoconference callout page shows the meeting ID/code + passcode
- [ ] Confirmation and FYI emails (HTML + text) include the meeting ID/code + passcode
- [ ] Add-to-calendar links carry the join link, meeting ID/code, and passcode in the event description

### Anything else to add?
Meeting ID grouping mirrors Zoom's display (e.g. `882 8541 1273`); unrecognized platforms simply omit the room line. The `?pwd=` URL param is ignored in favor of the explicit passcode field.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -5,6 +5,27 @@ class EventDecorator < ApplicationDecorator
     URI.parse(videoconference_url).host&.split(".")&.[](-2)&.capitalize rescue "video call"
   end
 
+  # The meeting room pulled out of the join URL so registrants can dial in
+  # manually: the numeric ID for Zoom (".../j/1234567890") or the meeting code
+  # for Google Meet ("meet.google.com/abc-defg-hij"). Returns a { label:, value: }
+  # hash, or nil when the URL is blank or the platform isn't recognized.
+  def videoconference_room
+    return if videoconference_url.blank?
+
+    uri = URI.parse(videoconference_url)
+    host = uri.host.to_s.downcase
+
+    if host.end_with?("zoom.us")
+      id = uri.path[%r{/(?:j|wc/join|wc)/(\d+)}, 1]
+      id && { label: "Meeting ID", value: format_zoom_meeting_id(id) }
+    elsif host == "meet.google.com"
+      code = uri.path.delete_prefix("/").presence
+      code && { label: "Meeting code", value: code }
+    end
+  rescue URI::InvalidURIError
+    nil
+  end
+
   def display_image
     return primary_asset.file if primary_asset&.file&.attached?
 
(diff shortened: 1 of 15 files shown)


app/decorators/event_decorator.rb
```diff
@@ -5,6 +5,27 @@ def videoconference_domain
...
+  # manually: the numeric ID for Zoom (".../j/1234567890") or the meeting code
+  # for Google Meet ("meet.google.com/abc-defg-hij"). Returns a { label:, value: }
+  # hash, or nil when the URL is blank or the platform isn't recognized.
+  def videoconference_room
+    return if videoconference_url.blank?
+
+    uri = URI.parse(videoconference_url)
+    host = uri.host.to_s.downcase
+
+    if host.end_with?("zoom.us")
```
maebeale, line 18: 🤖 _From Claude:_ Platform detection is host-based (`zoom.us` suffix / exact `meet.google.com`) rather than label-based, so a custom videoconference_label like "Virtual event" never changes parsing. Unrecognized hosts fall through to nil and the room line is simply omitted.

app/decorators/event_decorator.rb
```diff
@@ -57,18 +78,22 @@ def calendar_links
...
+      base_description = event_description
     else
       cal_location = nil
-      description  = event_description
+      base_description = event_description
     end
 
+    # Carry the join link, meeting ID/code, and passcode into the calendar entry
+    # so registrants have everything they need to connect straight from the event.
+    description = [ videoconference_calendar_details, base_description ].compact_blank.join("\n\n")
```
maebeale, line 95: 🤖 _From Claude:_ The join link/ID/passcode are prepended to the calendar description (URL also still fills the location field). `compact_blank` drops the block entirely for non-virtual events, leaving the prior description untouched.

app/views/events/_videoconference_connection.html.erb
```diff
@@ -0,0 +1,14 @@
+<%# Meeting room ID/code (parsed from the join URL) and passcode for the event.
+    Render only where the join link itself is revealed. `event` must be a
+    decorated Event. %>
+<% room = event.videoconference_room %>
```
maebeale, line 4: 🤖 _From Claude:_ This shared partial is rendered only inside the join-link reveal (gated by `joinable?`) and on the callout page, so the passcode never appears on public event cards or registration previews.

PR #1802, opened 2026-06-21
HOLD-MAYBE NOT: Replace native confirm() with a styled in-page modal

maebeale:
📖 Read

Replaces the browser's native `confirm()` dialog with a styled in-page modal.

**Why:** native confirms are unstyled and visually inconsistent with the app.

**How it works:**
- All 41 `data-turbo-confirm` prompts are routed through the modal via `Turbo.setConfirmMethod` — no per-call view changes needed.
- A shared `confirmModal(message)` helper (returns a `Promise<boolean>`) lets JS controllers use the same modal. The facilitator-warning controller now uses it (was `window.confirm`).
- Modal is a single instance rendered once in the layout; supports Esc/backdrop to cancel, falls back to native `confirm()` if absent.

**Specs:** native `confirm()` can't be driven by Capybara's `accept_confirm`/`dismiss_confirm` anymore, so added `accept_confirm_modal`/`dismiss_confirm_modal` helpers and updated the affected specs.

**Not converted:** `dirty_form_controller`'s `beforeunload`/`turbo:before-visit` guards stay native — those navigation events are synchronous and can't await an async modal (and `beforeunload` must use the browser's own prompt).

### DIFF
--- a/app/frontend/javascript/application.js
+++ b/app/frontend/javascript/application.js
@@ -1,4 +1,4 @@
-import "@hotwired/turbo-rails";
+import { Turbo } from "@hotwired/turbo-rails";
 import "@rails/actiontext";
 import "rhino-editor";
 import "rhino-editor/exports/styles/trix.css";
@@ -12,3 +12,8 @@ ActiveStorage.start();
 import "./controllers";
 import "./rhino/extend-editor.js";
 import "./turbo-events";
+import { confirmModal } from "./confirm_modal";
+
+// Route Turbo's data-turbo-confirm prompts through the styled in-page modal
+// instead of the browser's native confirm() dialog.
+Turbo.setConfirmMethod((message) => confirmModal(message));
--- a/app/frontend/javascript/confirm_modal.js
+++ b/app/frontend/javascript/confirm_modal.js
@@ -1,0 +1,22 @@
+// Opens the shared, styled confirmation modal and resolves to the user's choice.
+//
+// Returns a Promise<boolean> so it can stand in for window.confirm() in async
+// flows and as Turbo's confirm method (see application.js). The request is
+// dispatched as a "confirm-modal:open" window event carrying the message,
+// options, and a resolve callback; the confirm-modal Stimulus controller
+// (rendered once in the layout) handles it and calls preventDefault.
+//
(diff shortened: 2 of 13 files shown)


app/frontend/javascript/controllers/affiliation_facilitator_warning_controller.js
```diff
@@ -24,18 +26,31 @@ export default class extends Controller {
   }
 
   guardSubmit(event) {
+    // The modal resolved true on a prior pass — let this re-submission through.
+    if (this.confirmed) {
+      this.confirmed = false;
+      return;
+    }
+
     if (!this.facilitatorChangePresent()) return;
```
maebeale, line 35: 🤖 _From Claude:_ The modal is async but Turbo/native submit guards are sync, so we cancel the submit, then re-submit via `requestSubmit` if confirmed. The `confirmed` flag lets that second pass through.

app/frontend/javascript/confirm_modal.js
```diff
@@ -0,0 +1,22 @@
...
+// the browser's native confirm() so a missing modal never silently blocks an
+// action.
+export function confirmModal(message, options = {}) {
+  return new Promise((resolve) => {
+    const event = new CustomEvent("confirm-modal:open", {
+      detail: { message, options, resolve },
+      cancelable: true,
+    });
+
+    const handled = !window.dispatchEvent(event);
```
maebeale, line 19: 🤖 _From Claude:_ Dispatched as a cancelable window event; if no modal controller is mounted we fall back to native `confirm()` so an action is never silently blocked.

app/frontend/javascript/application.js
```diff
@@ -12,3 +12,8 @@ ActiveStorage.start();
 import "./controllers";
 import "./rhino/extend-editor.js";
 import "./turbo-events";
+import { confirmModal } from "./confirm_modal";
+
+// Route Turbo's data-turbo-confirm prompts through the styled in-page modal
+// instead of the browser's native confirm() dialog.
+Turbo.setConfirmMethod((message) => confirmModal(message));
```
maebeale, line 19: 🤖 _From Claude:_ This single hook styles all 41 `data-turbo-confirm` prompts at once — no view-level changes needed.

PR #1804, opened 2026-06-21
Reorganize the registration org-linking popup

maebeale:
- Admins use the link-organization popup to reconcile the organization a registrant typed on their registration form against the orgs in the database. This reorganizes that screen so the source answer, the confirmed links, and the person's other affiliations are each clearly separated and actionable.

- Reordered the popup top→bottom: the registration-form submission answer first (grey), the registration-linked orgs in a green-tinted card (middle), and the person's other affiliations as a grey bulleted list (bottom).
- Surface **every** registration-form submission (a registrant can have more than one via public resubmission or future person-dedupe), each with its org/title and a green file-icon link to that specific submission's public view (`form_submission_id`, scoped to the form + person).
- Show one **"Create and link"** row (primary action) per distinct submitted org not yet in the database, validated server-side against what the registrant actually submitted; plus fuzzy suggested matches and an existing-org search.
- Distinguish "no form submitted" vs "form submitted without an org", show affiliation duration on chips ("starting June 2025" / "from March 2024 - June 2025"), move "View <name>'s affiliations" into the card header, and stop showing affiliations in purple on the registrants index.
- Add a **"Back to registrants"** eyebrow above the person profile when reached from the registrants list (carries `event_id`, since the profile isn't event-scoped).
- Seed: org-link demo registrants **9** (fuzzy partial-name suggestions) and **10** (multiple submissions) so the new states are visible in the dev data.

### UI Testing Checklist
- [ ] Open a registrant's organizations & affiliations popup from the registrants page.
- [ ] Submission answer shows at top; a registrant with multiple submissions (demo 10) gets one green "View submission #N" link per submission.
- [ ] "Create and link" shows (solid/primary) for each submitted org missing from the DB; "Link organization" is the outline/secondary action.
- [ ] Fuzzy partial-name matches appear under "Suggested matches" (demo 9); exact-but-unlinked match offers a one-click link (demo 8).
- [ ] Linked orgs show green chips with an Unlink button; other affiliations render as a grey bulleted list with neutral chips.
- [ ] Registrants index no longer shows purple facilitator pills; clicking a registrant's name shows a "Back to registrants" eyebrow on their profile.

### Anything else to add?
- Rebased onto current `main` and reconciled with its concurrent work on this feature — kept main's "Unlink" button + confirm dialog and its removal of inline affiliation title-editing, layering this PR's reorg on top.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -155,19 +155,23 @@ class EventRegistrationsController < ApplicationController
     # whether each linked org is also an active/inactive affiliation (removing a
     # link here leaves the affiliation untouched).
     @affiliations_by_org = @person.affiliations.group_by(&:organization_id)
-    @submitted_org_name = find_submitted_agency_name(@event_registration)
-    # Whether an org with the submitted name already exists — if so, there's
-    # nothing to create (the admin links the existing one), so the "Create org
-    # & link" button is hidden.
-    @submitted_org_exists = @submitted_org_name.present? &&
-      Organization.where("LOWER(name) = ?", @submitted_org_name.strip.downcase).exists?
-    # The job title/position the registrant typed on the form, to compare against
-    # the title on any existing affiliation for a linked org.
-    @submitted_position = find_submitted_answer(@event_registration, "agency_position")
-    # The registrant's submission of the event's registration form, so we can link
-    # out to the public-facing form view.
-    reg_form = @event_registration.event.registration_form
-    @form_submission = reg_form && @person.form_submissions.find_by(form: reg_form)
+    # Every registration-form submission and the org/title the registrant entered
+    # on each. Normally there's one, but a registrant can have more (legacy or
+    # duplicate data), so we surface them all rather than silently picking one.
+    @submitted_entries = registration_submission_entries(@event_registration)
+    @form_submission = @submitted_entries.first&.fetch(:submission)
+    # The "primary" submitted org/title — the first submission that named an org
+    # (else the first submission) — drives the suggested-match and comparison logic.
+    primary = @submitted_entries.find { |entry| entry[:org_name].present? } || @submitted_entries.first
+    @submitted_org_name = primary && primary[:org_name]
+    @submitted_position = primary && primary[:position]
+    # Each distinct submitted org name that isn't already in the database gets its
+    # own "Create and link" row, so every typed-but-missing org can be resolved.
+    submitted_names = @submitted_entries.filter_map { |entry| entry[:org_name].presence&.strip }.uniq { |name| name.downcase }
+    existing_names = submitted_names.any? ?
+      Organization.where("LOWER(name) IN (?)", submitted_names.map(&:downcase)).pluck(Arel.sql("LOWER(name)")).map(&:to_s).to_set :
+      Set.new
+    @creatable_org_names = submitted_names.reject { |name| existing_names.include?(name.downcase) }
     @potential_matches = if @submitted_org_name.present?
       Organization.remote_search(@submitted_org_name).where.not(id: @linked_organizations.ids).limit(10)
     else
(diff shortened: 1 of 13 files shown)


app/controllers/event_registrations_controller.rb
```diff
@@ -193,9 +197,13 @@ def create_organization
...
-    # Build the org from the name the registrant typed on the form, so the button
-    # can't be used to create an arbitrary org — it only resolves the pending name.
-    name = find_submitted_agency_name(@event_registration)
+    # Build the org from a name the registrant actually typed on the form, so the
+    # button can't be used to create an arbitrary org — it only resolves a pending
+    # submitted name. A specific name is passed when there are several submissions;
+    # otherwise default to the first submitted name.
+    submitted_names = submitted_agency_names(@event_registration)
+    requested = params[:organization_name].presence
+    name = requested ? submitted_names.find { |submitted| submitted.casecmp?(requested) } : submitted_names.first
```
maebeale, line 206: 🤖 _From Claude:_ Security: a name can be passed in (needed when a registrant has several submissions with different orgs), but we only create from a name they actually submitted — `find { casecmp? }` against `submitted_agency_names`, so the button still can't mint an arbitrary org. Falls back to the first submitted name when none is given.

app/controllers/events/public_registrations_controller.rb
```diff
@@ -103,7 +103,15 @@ def show
         return
       end
 
-      @form_submission = @form.form_submissions.find_by(person: person)
+      # A specific submission can be requested by id (a registrant may have more
+      # than one); scope it to this form and person so the id can't reach another
+      # person's submission. Otherwise show their first submission for the form.
+      submissions = @form.form_submissions.where(person: person)
+      @form_submission = if params[:form_submission_id].present?
```
maebeale, line 110: 🤖 _From Claude:_ A registrant can have more than one submission, so each "View submission #N" link targets a specific one by id — scoped to this form + person so an id from another person's submission is rejected (falls through to the not-found redirect).

app/views/event_registrations/_org_affiliation_pills.html.erb
```diff
@@ -11,8 +11,18 @@
     <span class="text-[0.65rem] font-semibold uppercase tracking-wide text-gray-400">Affiliations:</span>
     <% affiliations.each do |a| %>
       <% a_active = active.call(a) %>
-      <% color = a_active ? "bg-green-50 text-green-700 border-green-200" : "bg-gray-100 text-gray-600 border-gray-200" %>
-      <% label = "#{a.facilitator? ? "Facilitator" : (a.title.presence || "No title")}#{" — inactive" unless a_active}" %>
+      <% color = a_active && !local_assigns[:neutral] ? "bg-green-50 text-green-700 border-green-200" : "bg-gray-100 text-gray-600 border-gray-200" %>
```
maebeale, line 14: 🤖 _From Claude:_ `neutral` (optional local) suppresses the green "active" tint so the bottom "other affiliations" list reads as plain/neutral, while the registration-linked chips above stay green.

app/views/people/show.html.erb
```diff
@@ -1,4 +1,13 @@
 <% content_for(:page_bg_class, "admin-or-owner-or-authsearchable") %>
+<%# Returned from a registrant's profile link on the event registrants page. Shown
+    above the profile topper; needs event_id since the profile isn't event-scoped. %>
+<% if params[:return_to] == "registrants" && params[:event_id].present? %>
```
maebeale, line 4: 🤖 _From Claude:_ Eyebrow branches on `return_to` and carries `event_id` because the profile page isn't event-scoped — same pattern as the existing bulk-payments return link below. Passed in from the registrants list via `person_profile_button`'s new `path_params:`.

db/seeds/dev/events_management.rb
```diff
@@ -1029,6 +1040,58 @@
...
+  unless linked_org
+    linked_org = org_answer_orgs[i % org_answer_orgs.size]
+    Affiliation.find_or_create_by!(person: person, organization: linked_org) do |aff|
+      aff.title = job_titles[i % job_titles.size]
+      aff.start_date = Date.current
+    end
+    registration.event_registration_organizations.find_or_create_by!(organization: linked_org)
+  end
+
+  mismatch = org_answer_index % 4 == 3
```
maebeale, line 1087: 🤖 _From Claude:_ The mismatch counter advances only when an org answer is actually written (not per registrant), so the ~1-in-4 "Pending" ratio holds even though registrants-with-submissions are sparse.

db/seeds/dev/events_management.rb
```diff
@@ -928,6 +935,10 @@
 
   # Fill in required text fields with sample data
   data[:form].form_fields.where(answer_type: [ :free_form_input_one_line, :free_form_input_paragraph ]).each do |field|
+    # Organization Name + Position / Title are seeded later with org-matching values
+    # (record_organization_answers), so leave them blank here.
+    next if %w[agency_name agency_position].include?(field.field_identifier)
```
maebeale, line 940: 🤖 _From Claude:_ Skipping these here is load-order sensitive: the generic filler runs before `record_organization_answers`, and its idempotent guard would otherwise lock in random Faker words for org/title instead of the org-matching values.

PR #1806, opened 2026-06-21
Move payment status detail into chip tooltips

maebeale:
👀 Skim

- The registrants payment chips spelled out "Partial payment"/"Discounted" on a second line, crowding the Payment column. This declutters them while keeping the detail accessible.

- Collapsed the payment chips to a status icon + amount due, moving the full status (and "Intends to pay") into the chip's hover tooltip.
- Unified the hover treatment (`transition hover:opacity-80 hover:shadow-sm`) across all chips in the row.

### Anything else to add?
- View-only; no logic or data changes. The seed data that exercises the Organization "Pending" chip moved to #1804.

### DIFF
--- a/app/views/events/_registrants_results.html.erb
+++ b/app/views/events/_registrants_results.html.erb
@@ -136,7 +136,7 @@
                       attention; a name matching a linked org is already resolved. %>
                   <% needs_linking = submitted_org_name.present? && linked_org_names.exclude?(submitted_org_name.downcase) %>
                   <% editor_path = link_organization_event_registration_path(registration, return_to: "registrants") %>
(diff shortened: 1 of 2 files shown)


app/views/events/_registrants_results.html.erb
```diff
@@ -232,21 +232,21 @@
...
                        end %>
+                    <% badge_title = if is_paid
+                         "Paid in full"
+                       elsif is_discounted
+                         "Discounted · #{dollars_from_cents(due_cents)} due"
+                       elsif is_partial
+                         "Partial payment · #{dollars_from_cents(due_cents)} due"
+                       else
+                         "#{dollars_from_cents(due_cents)} due"
+                       end %>
```
maebeale, line 243: 🤖 _From Claude:_ "Intends to pay" is appended to the tooltip too (the visible indicator below the chip stays), so hovering surfaces the full payment status in one place.

PR #1788, opened 2026-06-19
Add per-event Onboarding tab to track registrant setup pipeline

maebeale:
### What is the goal of this PR and why is this important?

AWBW staff hand-track each registrant's onboarding (system setup, scholarship/payment, CE, attendance, comms) in a Google Sheet that drifts out of sync with the portal. This adds a per-event **Onboarding** tab — a single, filterable matrix that surfaces that whole pipeline next to the data it describes, with inline editing where it makes sense and links out to the relevant detail pages everywhere else.

### How did you approach the change?

- **New Onboarding tab** (`events#onboarding`, linked from the Registrants *Bulk actions* menu): a full-width, click-to-sort matrix (shared `table-sort`), server-side column show/hide (a `hide` param threaded through filters/tabs — no client-side toggle), keyword + per-step filters, active/inactive tabs, and **CSV export** (with a header/row column-count guard test).
- **Columns** (a single shared descriptor list drives header + rows + CSV so order/sort indexes stay aligned): registration edit button, First/Last/Email (→ person edit), Day 1–N attendance checkboxes, Organization (→ org-link popup), Program type (→ org affiliations), Payment chip + Fees due (→ allocations card) + Paid (→ allocations), Scholarship amount/grant/tasks (→ scholarship edit, "Create" when none), CE requested/hours/amount/license (→ registration edit), Fee note, Portal invite (user confirmation), Mailchimp/CMS setup checkboxes, Flagged comments, joined Comments, Event attendance.
- **New persisted data:** `event_registration_checklist_completions` (audited: step + who + when) for the Mailchimp/CMS steps; `fee_note` and `completed_day_1..5` columns on `event_registrations`. Inline toggles/edits post to `event_registrations#update_onboarding`.
- **Program type** is computed per registration from its *linked org* as of the training month (`facilitator_status_on`, excluding the registrant's own affiliation) — not a registrant-wide rollup across unrelated orgs.
- **Bug fixes:** corrected the shared program-status classification when a registrant has no facilitator affiliation yet; fixed a polymorphic grant-preload crash; made payment reads use the preloaded association (no N+1); fixed a stray CSV header that misaligned columns.

### UI Testing Checklist

- [ ] Onboarding tab loads from the event sub-nav and the Registrants → Bulk actions menu.
- [ ] Checklist/day checkboxes persist and show "completed by … on …" on hover; fee note saves inline.
- [ ] Column show/hide, per-step filter, active/inactive tabs, and CSV export all work; CSV columns line up.
[...]

PR #1809, opened 2026-06-21
Create job + Facilitator affiliations on registration and org linking

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: contained changes to affiliation creation/dedupe; one new model scope that `active` delegates to (no behavior change today)

- When a person registers for an event — or an admin links an organization to their registration — we should automatically record their affiliation with that org, so the program roster reflects it without manual data entry.
- Each linking moment now creates up to **two** affiliations:
  - a **job affiliation** carrying the title the registrant typed on the form (only when they provided one), and
  - a standing **"Facilitator" affiliation**.

- Added `AffiliationServices::CreateFromRegistration`, a single PORO shared by every entry point so the behavior can't drift:
  - **Public registration** (`PublicRegistration`)
  - **Link existing org** (`EventRegistrationsController#select_organization`)
  - **Create & link org** (`EventRegistrationsController#create_organization`)
- **Facilitator de-dup:** the "Facilitator" affiliation is skipped only when an **active-or-pending** affiliation the model's `facilitators` scope considers canonical (exactly "Facilitator", trimmed, case-sensitive) already exists with that org — a current one **or** one dated to a future training that hasn't started yet. An **ended** Facilitator affiliation does not block a new one (a returning facilitator gets a fresh second affiliation). The job affiliation is created first, so a typed title of exactly "Facilitator" suppresses the duplicate — but a facilitator-ish title like "Lead Facilitator" still gets its own standing "Facilitator" alongside it.
- **`active_or_pending` scope:** added to `Affiliation` (not-inactive, not-ended; **includes** future-dated/pending affiliations). `active` now delegates to it, so existing behavior is unchanged — the scope just names the set the dedupe needs.
- **Atomicity:** the job and Facilitator creates run in **one transaction**, so a job affiliation is never left without its Facilitator (or vice versa).
- **Start dates:**
  - The **job** affiliation is left with **no start date** — we don't know when the person began that role (they may have been with the org for years before this training), and dating it to registration would misrepresent it.
  - The **Facilitator** affiliation starts on the **first day of the training's month** (falling back to the current month when an event has no start date).
- **Title source:** all three flows pull the job title from the registrant's position answer on the registration form, reusing the same "primary submission" the org-linking editor displays so the applied title matches what the editor shows.
- **Vocabulary:** new code refers to this as the *organization position*, not "agency". The legacy stored field identifiers (`agency_name` / `agency_position`) are localized behind organization-named constants on `PublicRegistration` pending a future form-field rename.
- **Dev seeds:** registration submissions now carry a job title most (~80%) of the time, leaving roughly one in five blank, so dev data exercises both the job + Facilitator and Facilitator-only outcomes.

### Anything else to add?

- New service spec (covers active/pending/ended dedupe, same title+org+dates no-dup, start-date rules) plus added coverage in the public-registration service spec and the event-registrations request spec.
- **Out of scope / follow-up PR:** treating organization status as derived from affiliations — reclassifying the ~20 `active` call sites between `active`/`active_or_pending`, tightening `active` to also require the start date to have arrived, and a symmetric org-status callback.
- **Rebased onto current `main`** (now includes #1815/#1814, which made facilitator matching exact + case-sensitive; #1818, which adds `organization_address` to affiliations; and #1788's Onboarding tab). Reconciliation notes for reviewers:
  - The service's facilitator de-dup now uses the model's `facilitators` scope, so it agrees with #1815's exact/case-sensitive definition instead of a looser `where(title:)`.
  - `Affiliation::FACILITATOR_TITLE` sits alongside #1818's new `organization_address` association; affiliations are created with no linked address (the new validation is a no-op when blank).
  - Controller auto-merged with #1788's `update_onboarding`; my org-linking actions and `submitted_position` (built on #1804's `registration_submission_entries`) are unchanged.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -212,7 +212,12 @@ class EventRegistrationsController < ApplicationController
     @person = @event_registration.registrant
     organization = Organization.find(params[:organization_id])
 
-    Affiliation.find_or_create_by!(person: @person, organization: organization)
+    AffiliationServices::CreateFromRegistration.call(
+      person: @person,
+      organization: organization,
+      job_title: submitted_position(@event_registration),
+      training_date: @event_registration.event.start_date
+    )
 
     @event_registration.event_registration_organizations
       .find_or_create_by!(organization: organization)
@@ -240,7 +245,12 @@ class EventRegistrationsController < ApplicationController
     existing = Organization.where("LOWER(name) = ?", name.strip.downcase).first
     organization = existing || Organization.create!(name: name.strip, organization_status: OrganizationStatus.find_by(name: "Active"))
 
-    Affiliation.find_or_create_by!(person: @person, organization: organization)
+    AffiliationServices::CreateFromRegistration.call(
+      person: @person,
+      organization: organization,
+      job_title: submitted_position(@event_registration),
+      training_date: @event_registration.event.start_date
+    )
     @event_registration.event_registration_organizations.find_or_create_by!(organization: organization)
 
     notice = existing ? "#{organization.name} linked." : "#{organization.name} created and linked."
@@ -408,22 +418,13 @@ class EventRegistrationsController < ApplicationController
       .uniq { |name| name.downcase }
   end
 
-  def find_submitted_agency_name(registration)
-    find_submitted_answer(registration, "agency_name")
-  end
-
-  def find_submitted_answer(registration, field_identifier)
-    form = registration.event.registration_form
-    return nil unless form
-
-    field = form.form_fields.find_by(field_identifier: field_identifier)
-    return nil unless field
-
-    FormAnswer
-      .joins(form_submission: :person)
-      .find_by(
-        form_submissions: { person_id: registration.registrant_id, form_id: form.id },
-        form_field_id: field.id
-      )&.submitted_answer
+  # The job title/position the registrant typed for their organization on the
+  # registration form. Uses the same "primary" submission as link_organization
+  # (the first submission that named an org, else the first), so the title applied
+  # when linking matches what the editor shows.
+  def submitted_position(registration)
+    entries = registration_submission_entries(registration)
+    primary = entries.find { |entry| entry[:org_name].present? } || entries.first
+    primary && primary[:position]
   end
 end
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -211,7 +211,7 @@ class PeopleController < ApplicationController
 
     if @person.save
       assign_associations(@person) if params.dig(:person, :category_ids)
-      redirect_to @person, notice: "Person was successfully updated."
+      redirect_to person_update_return_path, notice: "Person was successfully updated."
     else
       set_form_variables
       render :edit, status: :unprocessable_content
@@ -368,6 +368,17 @@ class PeopleController < ApplicationController
       name_match: exact,
       blocked: primary_email_match
     }
+  end
+
+  # Where to send the admin after a successful update. Defaults to the person's
+  # profile, but returns to the registration org-linking page when the edit was
+  # opened from one of its linked-org cards (preserving that page's own return_to).
+  def person_update_return_path
+    if params[:return_to] == "registration_link" && params[:event_registration_id].present?
+      link_organization_event_registration_path(params[:event_registration_id], return_to: params[:link_org_return_to].presence)
+    else
+      @person
+    end
   end
 
   # Only allow a list of trusted parameters through.
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -1,4 +1,9 @@
 class Affiliation < ApplicationRecord
+  # Standing title given to the "facilitator affiliation" we create on registration
+  # and org linking. Matches the `facilitators` scope / `facilitator?` predicate
+  # (both treat exactly "Facilitator" as canonical).
+  FACILITATOR_TITLE = "Facilitator".freeze
+
   belongs_to :organization, inverse_of: :affiliations
   belongs_to :person, touch: true
   # Which of the organization's addresses this person is affiliated with (optional).
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -13,10 +13,15 @@ class Affiliation < ApplicationRecord
   validates_presence_of :organization_id
   validate :organization_address_belongs_to_organization
 
-  scope :active, -> {
+  # Not flagged inactive and not past its end date. Includes affiliations whose
+  # start_date is still in the future (e.g. a Facilitator affiliation dated to an
+  # upcoming training's month) — they are "pending" but counted here.
+  scope :active_or_pending, -> {
     where(inactive: false)
       .where("affiliations.end_date IS NULL OR affiliations.end_date >= ?", Date.current)
   }
+
+  scope :active, -> { active_or_pending }
 
   # Only the exact, case-sensitive title "Facilitator" counts — variants like
   # "Lead Facilitator" or "facilitator" are deliberately excluded. BINARY forces
--- a/app/services/affiliation_services/create_from_registration.rb
+++ b/app/services/affiliation_services/create_from_registration.rb
@@ -1,0 +1,72 @@
+module AffiliationServices
+  # Creates the affiliations we want whenever a person registers for an event or
+  # an admin links an organization to their registration:
+  #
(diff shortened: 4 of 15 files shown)


app/services/affiliation_services/create_from_registration.rb
```diff
@@ -0,0 +1,65 @@
...
+      create_facilitator_affiliation
+    end
+
+    private
+
+    def create_job_affiliation
+      return unless @job_title
+      return if active_affiliation_with_title?(@job_title)
+
+      create_affiliation(@job_title, start_date: nil)
```
maebeale, line 49: 🤖 _From Claude:_ Job affiliation is created with no start date on purpose — registration ≠ when they began this role (they may have been at the org for years), so dating it to today would misrepresent it. The Facilitator affiliation does get a date (first of the training month).

app/services/affiliation_services/create_from_registration.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    def create_facilitator_affiliation
+      return if active_affiliation_with_title?(Affiliation::FACILITATOR_TITLE)
+
+      create_affiliation(Affiliation::FACILITATOR_TITLE, start_date: facilitator_start_date)
```
maebeale, line 55: 🤖 _From Claude:_ Ordering matters: the job affiliation is created first, so a typed title of exactly "Facilitator" satisfies this guard and suppresses a duplicate, while "Lead Facilitator" does not — it still gets its own standing Facilitator affiliation.

app/controllers/event_registrations_controller.rb
```diff
@@ -367,22 +377,13 @@ def submitted_agency_names(registration)
...
-      .joins(form_submission: :person)
-      .find_by(
-        form_submissions: { person_id: registration.registrant_id, form_id: form.id },
-        form_field_id: field.id
-      )&.submitted_answer
+  # The job title/position the registrant typed for their organization on the
+  # registration form. Uses the same "primary" submission as link_organization
+  # (the first submission that named an org, else the first), so the title applied
+  # when linking matches what the editor shows.
+  def submitted_position(registration)
```
maebeale, line 425: 🤖 _From Claude:_ Rebase reconciliation with #1804: derives the title from the same "primary" submission the org-linking editor shows (via `registration_submission_entries`), so the applied title matches the displayed one. Replaced the old `find_submitted_answer` helper.

app/models/affiliation.rb
```diff
@@ -8,11 +13,16 @@ class Affiliation < ApplicationRecord
...
-  scope :active, -> {
+  # Not flagged inactive and not past its end date. Includes affiliations whose
+  # start_date is still in the future (e.g. a Facilitator affiliation dated to an
+  # upcoming training's month) — they are "pending" but counted here.
+  scope :active_or_pending, -> {
     where(inactive: false)
       .where("affiliations.end_date IS NULL OR affiliations.end_date >= ?", Date.current)
   }
 
+  scope :active, -> { active_or_pending }
```
maebeale, line 24: 🤖 _From Claude:_ `active` delegates to `active_or_pending` so behavior is unchanged today; this just names the set the Facilitator dedupe needs. A follow-up PR will tighten `active` to also require the start date to have arrived (and reclassify call sites), at which point the two scopes diverge.

app/services/affiliation_services/create_from_registration.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def initialize(person:, organization:, job_title: nil, training_date: nil)
+      @person = person
+      @organization = organization
+      @job_title = job_title.presence&.strip
+      @training_date = training_date
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
```
maebeale, line 37: 🤖 _From Claude:_ Wrapped both creates in one transaction so a job affiliation is never persisted without its Facilitator (or vice versa) if the second `create!` raises.

app/services/affiliation_services/create_from_registration.rb
```diff
@@ -0,0 +1,79 @@
...
+      @person.affiliations.active_or_pending.where(organization: @organization, title: title).exists?
+    end
+
+    # Uses the model's canonical `facilitators` scope (exactly "Facilitator",
+    # trimmed and case-sensitive) so this agrees with #facilitator? and the rest of
+    # the app on what counts as a facilitator affiliation. Counts pending
+    # (future-dated) facilitator affiliations too, so registering for a second
+    # upcoming training doesn't mint a duplicate.
+    def active_or_pending_facilitator_affiliation?
+      @person.affiliations.active_or_pending.facilitators.where(organization: @organization).exists?
```
maebeale, line 76: 🤖 _From Claude:_ Checking `active_or_pending` (not just `active`) here prevents a duplicate pending Facilitator when someone registers for a second upcoming training before the first has started.

app/views/event_registrations/_org_affiliation_pills.html.erb
```diff
@@ -14,15 +14,17 @@
...
-             " from #{start_month} - #{end_month}"
+             "#{start_month} – #{end_month}"
            elsif end_month
-             " until #{end_month}"
+             "until #{end_month}"
            elsif start_month
-             " starting #{start_month}"
+             "since #{start_month}"
+           else
+             "no dates"
```
maebeale, line 25: 🤖 _From Claude:_ Dates now always render in parens, including `(no dates)` when an affiliation has neither start nor end — previously an undated pill showed a bare title with no signal its dates were just unknown. `starting` became `since` to read naturally inside the parens.

app/views/event_registrations/link_organization.html.erb
```diff
@@ -77,9 +77,16 @@
...
+            <%# The whole card links to the person's affiliations section (where these
+                are managed) via a stretched link on the org name; the Unlink button is
+                lifted above the overlay with relative z-10 so it stays clickable. %>
+            <li class="relative flex items-start justify-between gap-3 <%= org_swatch[:bg] %> border <%= org_swatch[:border] %> rounded-lg p-3 transition hover:border-green-300 hover:shadow-sm">
               <div class="min-w-0">
-                <p class="font-medium text-gray-800"><%= org.name %><% if org.city_state.present? %> <span class="text-xs text-gray-500">· <%= org.city_state %></span><% end %></p>
+                <p class="font-medium text-gray-800"><%= link_to org.name, person_path(@person, anchor: "affiliations"),
+                            target: "_blank", rel: "noopener",
+                            title: "View #{@person.first_name.presence || @person.full_name}'s affiliations",
+                            class: "after:absolute after:inset-0 hover:underline focus-visible:underline focus:outline-none",
```
maebeale, line 90: 🤖 _From Claude:_ Stretched-link pattern: the org-name anchor is the real link (accessible text), and `after:absolute after:inset-0` expands its hit area over the whole card. The `<li>` is `relative` and the Unlink form is `relative z-10` so it stays clickable above the overlay. Opens in a new tab so the in-progress linking flow is preserved.

app/controllers/people_controller.rb
```diff
@@ -370,6 +370,17 @@ def format_duplicate(person, exact: true, entered_email: nil)
     }
   end
 
+  # Where to send the admin after a successful update. Defaults to the person's
+  # profile, but returns to the registration org-linking page when the edit was
+  # opened from one of its linked-org cards (preserving that page's own return_to).
+  def person_update_return_path
```
maebeale, line 376: 🤖 _From Claude:_ Scoped strictly to `return_to == "registration_link"` so existing flows (incl. the onboarding eyebrow, which never redirected on save) are unaffected — default stays `redirect_to @person`. The origin is threaded via top-level params (hidden fields in the form), so it survives the POST without touching `person_params`.

app/views/event_registrations/link_organization.html.erb
```diff
@@ -1,4 +1,8 @@
 <% content_for(:page_bg_class, "admin-only bg-blue-100") %>
+<%# Every jump to the person's affiliations editor (header link, linked-org cards,
+    other-affiliations card) shares this path so a save returns here — carrying the
+    origin and this page's own return_to. %>
+<% edit_affiliations_path = edit_person_path(@person, anchor: "affiliations", return_to: "registration_link", event_registration_id: @event_registration.id, link_org_return_to: params[:return_to].presence) %>
```
maebeale, line 5: 🤖 _From Claude:_ One shared path for all three jumps to the affiliations editor (header link, linked-org cards, other-affiliations card) so they stay in sync — each carries the origin and this page's `return_to`, opens in the same tab, and a save returns here.

PR #1810, opened 2026-06-21
COORDINATE: Populate an org from the registrant's form on the linking page

maebeale:
🤖 suggested review level: 5 Inspect 🔬 overwrites org fields (website + type, Ahoy-logged) and adds/refreshes address records behind an opt-in; depends on #1886

> **Stacked on #1886** (base branch `maebeale/organization-type-model`). Merge #1886 first, then this retargets to `main`.

Resolving a registrant's submitted org on the org-linking page used to save only the name. Now the form's org info flows through, with a per-form **populate toggle** controlling when an org is written from the form.

- **Create and link** (typed org not in the DB) — toggle **on by default**: builds/populates the new org with **organization type**, **website** (stored as typed), and a primary **work address**; off creates the org bare. (Existing name → links, populating only when on.)
- **Suggested matches** (fuzzy + exact name matches) — toggle **on by default**: very likely the registrant's org, so populate it.
- **Manual search box** — toggle **off by default**: often an unrelated established org, so link only unless the admin opts in.
- Linking always creates the job + Facilitator affiliations (via `AffiliationServices::CreateFromRegistration`), regardless of the toggle.

When the toggle applies to an existing org: **website + organization type overwrite** the org's values (each logged to Ahoy `update.organization`); the **address is additive** — a new primary, or a refresh of a same-street record instead of a duplicate; demoted prior addresses stay active.

- One controller, gated on `params[:apply_form_details]`; the toggle is a shared partial rendered inside each link form with a per-form `checked` default (no JS).
- **Organization type** is set via the new `OrganizationType` association from #1886: the submitted type name is resolved to an `OrganizationType` (case-insensitive) and assigned; overwrites are Ahoy-logged.
- Overwrites go through a shared `overwrite_org_field` / `overwrite_org_type`.
- Website stored **as typed** per #1765 (`Organization#website_link_url` renders the link).
- **Fixes a pre-existing latent bug**: `PublicRegistration` passed `nil` street/zip to NOT NULL columns, 500-ing when a registrant filled a city but no zip — now coerced to `""`.

### Anything else to add?
- **Phone is intentionally not populated** — the form has no org phone field (only the registrant's personal phone).

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -215,6 +215,12 @@ class EventRegistrationsController < ApplicationController
     @person = @event_registration.registrant
     organization = Organization.find(params[:organization_id])
 
+    # Apply the org details the form captured only when the admin opted in — linking an
+    # established org shouldn't silently rewrite its address/website.
+    entries = registration_submission_entries(@event_registration)
+    answers = submitted_agency_answers(entries, organization.name)
+    apply_submitted_org_details(organization, answers) if params[:apply_form_details] == "1"
+
     AffiliationServices::CreateFromRegistration.call(
       person: @person,
       organization: organization,
@@ -236,7 +242,8 @@ class EventRegistrationsController < ApplicationController
     # button can't be used to create an arbitrary org — it only resolves a pending
     # submitted name. A specific name is passed when there are several submissions;
     # otherwise default to the first submitted name.
-    submitted_names = submitted_agency_names(@event_registration)
+    entries = registration_submission_entries(@event_registration)
+    submitted_names = entries.filter_map { |entry| entry[:org_name].presence&.strip }.uniq { |name| name.downcase }
     requested = params[:organization_name].presence
     name = requested ? submitted_names.find { |submitted| submitted.casecmp?(requested) } : submitted_names.first
     if name.blank?
@@ -244,9 +251,24 @@ class EventRegistrationsController < ApplicationController
       return
     end
 
-    # Reuse an existing org with that name rather than creating a duplicate.
+    # The other org details the registrant typed on the same submission, applied when the
+    # admin leaves the populate toggle on (checked by default for create-and-link).
+    answers = submitted_agency_answers(entries, name)
+    apply = params[:apply_form_details] == "1"
+
+    # Reuse an existing org with that name rather than creating a duplicate. A new org is
+    # built from the form when populating, else created bare; an existing org is populated
+    # only when the toggle is on (never silently).
     existing = Organization.where("LOWER(name) = ?", name.strip.downcase).first
-    organization = existing || Organization.create!(name: name.strip, organization_status: OrganizationStatus.find_by(name: "Active"))
+    organization =
+      if existing
+        existing
+      elsif apply
+        build_organization_from_answers(name, answers)
+      else
+        Organization.create!(name: name.strip, organization_status: OrganizationStatus.find_by(name: "Active"))
+      end
+    apply_submitted_org_details(organization, answers) if existing && apply
 
     AffiliationServices::CreateFromRegistration.call(
       person: @person,
@@ -414,12 +436,15 @@ class EventRegistrationsController < ApplicationController
     entries
   end
 
-  # Distinct, non-blank org names the registrant typed across their registration-form
-  # submissions (case-insensitive dedupe, first spelling wins).
-  def submitted_agency_names(registration)
-    registration_submission_entries(registration)
-      .filter_map { |entry| entry[:org_name].presence&.strip }
-      .uniq { |name| name.downcase }
+  # The agency answers (org details + position) the registrant submitted for this org,
+  # keyed by field_identifier. Uses the submission that named this org, else the first
+  # that named any org (registrants normally have a single submission), so a resolved
+  # org and affiliation can carry what they typed. {} when nothing was submitted.
+  def submitted_agency_answers(entries, name)
+    entry = entries.find { |e| e[:org_name].present? && e[:org_name].strip.casecmp?(name.to_s.strip) } ||
+            entries.find { |e| e[:org_name].present? } ||
+            entries.first
+    entry&.fetch(:submission)&.answers_by_identifier || {}
   end
 
   # The job title/position the registrant typed for their organization on the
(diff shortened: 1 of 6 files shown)


app/controllers/event_registrations_controller.rb
```diff
@@ -359,30 +367,53 @@ def registration_submission_entries(registration)
...
-        form_field_id: field.id
-      )&.submitted_answer
+  # Organization#website_url requires an http(s) scheme; registrants often type a bare
+  # host, so prepend https:// when none is present. Drops a value that still doesn't
+  # match the URL format (e.g. one with spaces) so a bad answer can't block org creation.
+  def normalize_website_url(value)
+    url = value.to_s.strip
+    return if url.blank?
+    url = "https://#{url}" unless url.match?(%r{\Ahttps?://}i)
+    url if url.match?(/\Ahttps?:\/\/\S+\z/i)
```
maebeale: 🤖 _From Claude:_ A malformed website (e.g. with spaces) returns nil rather than raising, so a bad answer can never block the whole "Create and link". The regex matches `Organization#website_url`'s own validation.

app/controllers/event_registrations_controller.rb
```diff
@@ -359,30 +367,53 @@ def registration_submission_entries(registration)
...
-
-    field = form.form_fields.find_by(field_identifier: field_identifier)
-    return nil unless field
+  # Mirrors PublicRegistration#create_agency_address: a primary "work" address with an
+  # "Unknown" locality (the form doesn't capture one). Skipped unless city and state
+  # are present, since the Address model requires both.
+  def build_agency_address(organization, answers)
+    city = answers["agency_city"].presence&.strip
+    state = answers["agency_state"].presence&.strip
+    return if city.blank? || state.blank?
```
maebeale, line 546: 🤖 _From Claude:_ Address needs both city and state (the model validates presence of both + locality), so a partial address is skipped instead of raising — matching `PublicRegistration`'s hardcoded `locality: "Unknown"`, `address_type: "work"`.

app/controllers/event_registrations_controller.rb
```diff
@@ -201,19 +201,27 @@ def create_organization
...
+    answers = submitted_agency_answers(entries, name)
+
     # Reuse an existing org with that name rather than creating a duplicate.
     existing = Organization.where("LOWER(name) = ?", name.strip.downcase).first
-    organization = existing || Organization.create!(name: name.strip, organization_status: OrganizationStatus.find_by(name: "Active"))
+    organization = existing || build_organization_from_answers(name, answers)
 
-    Affiliation.find_or_create_by!(person: @person, organization: organization)
+    Affiliation.find_or_create_by!(person: @person, organization: organization) do |affiliation|
+      affiliation.title = answers["agency_position"].presence
```
maebeale: 🤖 _From Claude:_ The block only runs when the affiliation is newly created, so an existing affiliation's title is left untouched.

app/controllers/event_registrations_controller.rb
```diff
@@ -185,7 +185,12 @@ def select_organization
     @person = @event_registration.registrant
     organization = Organization.find(params[:organization_id])
 
-    Affiliation.find_or_create_by!(person: @person, organization: organization)
+    # Linking an existing org leaves its own details alone, but a new affiliation
+    # should still carry the role the registrant typed — same as the create path.
+    Affiliation.find_or_create_by!(person: @person, organization: organization) do |affiliation|
```
maebeale: 🤖 _From Claude:_ Linking a matched/searched org leaves the org's own details alone, but a new affiliation now carries the submitted position (the block only runs on create, so an existing affiliation's title is preserved). Keeps this path consistent with create-and-link.

app/controllers/event_registrations_controller.rb
```diff
@@ -359,30 +378,62 @@ def registration_submission_entries(registration)
...
+    entry&.fetch(:submission)&.answers_by_identifier || {}
   end
 
-  def find_submitted_agency_name(registration)
-    find_submitted_answer(registration, "agency_name")
-  end
+  # Backfill onto an org we already have the details the registrant typed but the org
+  # is missing — only where its own value is blank, never overwriting. Covers type,
+  # website, and (when the org has no current address) a work address.
+  def fill_missing_org_details(organization, answers)
```
maebeale: 🤖 _From Claude:_ Backfills only blank fields on an org we already have (matched/searched/reused), so an existing record is never overwritten — the address is added only when the org has no current one. Used by both `select_organization` and the reuse branch of `create_organization`.

app/controllers/event_registrations_controller.rb
```diff
@@ -185,7 +185,17 @@ def select_organization
     @person = @event_registration.registrant
     organization = Organization.find(params[:organization_id])
 
-    Affiliation.find_or_create_by!(person: @person, organization: organization)
+    # Carry the registrant's role onto a new affiliation, and — only when the admin
+    # opted in — backfill the org details the form filled but the org is missing.
+    # Linking an established org shouldn't silently rewrite its canonical fields.
+    entries = registration_submission_entries(@event_registration)
+    answers = submitted_agency_answers(entries, organization.name)
+    fill_missing_org_details(organization, answers) if params[:apply_form_details] == "1"
```
maebeale: 🤖 _From Claude:_ Opt-in gate (default off): linking an existing org only backfills its missing type/website/address when the admin checks the box, so an established org is never silently rewritten from one registrants form. The checkbox rides along in each link form — zero JS. Affiliation title still carries over unconditionally.

app/controllers/event_registrations_controller.rb
```diff
@@ -359,30 +379,84 @@ def registration_submission_entries(registration)
...
+    organization.update!(agency_type: answers["agency_type"]) if organization.agency_type.blank? && answers["agency_type"].present?
+    add_primary_agency_address(organization, answers)
   end
 
-  def find_submitted_answer(registration, field_identifier)
-    form = registration.event.registration_form
-    return nil unless form
+  # Overwrite the org's website with the submitted one — an explicit admin opt-in — and
+  # log the change to Ahoy. No-op when the form had no website or it already matches.
+  def overwrite_org_website(organization, submitted)
```
maebeale: 🤖 _From Claude:_ Website is the one field that truly overwrites (per request), so it logs an `update.organization` Ahoy event with from/to. Address adds a new primary (old one demoted, history kept); type only fills when blank. All three only run on the explicit opt-in.

app/services/event_registration_services/public_registration.rb
```diff
@@ -225,8 +225,8 @@ def create_agency_address(organization)
 
       if existing
         existing.update!(
-          street_address: field_value("agency_street"),
-          zip_code: field_value("agency_zip"),
+          street_address: field_value("agency_street").to_s,
```
maebeale, line 330: 🤖 _From Claude:_ Latent-bug fix: street/zip are NOT NULL columns but the form leaves them optional, so a city-only address answer used to 500 with a NotNullViolation. `.to_s` stores `""` instead of nil (the model only validates city/state/locality).

app/controllers/event_registrations_controller.rb
```diff
@@ -359,30 +379,88 @@ def registration_submission_entries(registration)
...
-        form_field_id: field.id
-      )&.submitted_answer
+  # Add the submitted address as a new primary "work" record with an "Unknown" locality
+  # (the form doesn't capture one). Additive — but if the org already has an address with
+  # the same street, refresh that one (e.g. fill in a city it was missing) rather than
+  # duplicating it. The chosen record becomes primary and any others are demoted, leaving
+  # them active so prior addresses aren't lost. Skipped unless city and state are present
+  # (the Address model requires them); street_address and zip_code are NOT NULL columns,
+  # so a missing one is stored as "" rather than nil.
+  def apply_submitted_address(organization, answers)
```
maebeale, line 543: 🤖 _From Claude:_ Address is additive (new primary) but de-duped by street — a same-street record is refreshed (city/zip filled in) rather than duplicated. Demoted prior addresses stay active so history is kept. Website + type instead overwrite via overwrite_org_field, each Ahoy-logged.

app/controllers/event_registrations_controller.rb
```diff
@@ -201,19 +211,38 @@ def create_organization
...
     if name.blank?
       redirect_to link_organization_event_registration_path(@event_registration, return_to: params[:return_to].presence), alert: "No submitted organization name to create from."
       return
     end
 
-    # Reuse an existing org with that name rather than creating a duplicate.
+    # The other org details the registrant typed on the same submission, applied when the
+    # admin leaves the populate toggle on (checked by default for create-and-link).
+    answers = submitted_agency_answers(entries, name)
+    apply = params[:apply_form_details] == "1"
```
maebeale, line 257: 🤖 _From Claude:_ Create-and-link now honors the populate toggle: on (default) builds the org from the form (or populates an existing same-name org); off creates it bare / links only. Suggested-match and create-and-link toggles default on in the view, the manual search box defaults off.

PR #1813, opened 2026-06-21
WAIT: Add PeopleRemover: delete a person and all associated data

maebeale:
🔬 Inspect

### Goal
We made fake submission-form data in prod and need to remove those people cleanly — and we'll need to delete people generally going forward. Adds a service + tooling to delete a person and their **entire** associated data graph (and their User account + everything it owns) without leaving orphans.

### What's here
- **`PeopleRemover` service** — deletes a person and all owned data. `Person.destroy` cascades most of it (submissions, affiliations, registrations, scholarships, addresses, contact methods, bookmarks, comments, sector/category tags, grants, event staffing); payments/refunds/allocations, bulk-payment notifications, `Pay::Customer`, and the PaperTrail/Ahoy audit trail don't, so they're torn down explicitly.
- **User + user data** — when the account is destroyed (override), it takes everything it authored with it: workshops, logs, reports, resources, stories, ideas, variations, plus bookmarks/user_forms/notifications. Records the account merely *references* (a shared banner/event it created) are FK-restricted and block, caught gracefully (never a 500).
- **Safety model** — skips people who look real (active/admin account, grant donor, spotlighted facilitator); `delete_users` removes an unused account; `force` overrides protections and does the full delete.
- **`data:remove_people` rake task** — bulk cleanup by `PERSON_IDS` / `FORM_SUBMISSION_IDS`; dry-run unless `CONFIRM=true` (`DELETE_USERS` / `FORCE` flags).
- **Admin purge button** on person edit (`?admin=true`) → interstitial confirmation page showing the full inventory (zero counts dashed, non-zero bold, tinted/hoverable rows). For a protected person, "Preview full delete" refreshes into an override preview listing the account's content under "Also removed with the User Account" before the final delete.

### Notes for reviewers
- Person-side deletion is exhaustive: every table referencing a person cascades, is handled, or is the facilitator protection (verified against schema FKs).
- Authored content is destroyed in FK-safe order (leaf → root) inside one transaction; any unenumerated FK rolls the whole thing back and surfaces a clear message.
- Two policy calls made per discussion: user-deletion stays opt-in (`force`/`DELETE_USERS`), and references the account doesn't *own* are left to block rather than nullified.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -234,7 +234,7 @@ class PeopleController < ApplicationController
     @force = params[:force] == "true"
     # force: true so the inventory reflects the full graph that WOULD be deleted,
     # even when the person looks real — the page always shows what's at stake.
-    preview = FakeSubmissionRemover.new(person_ids: [ @person.id ], force: true)
+    preview = PeopleRemover.new(person_ids: [ @person.id ], force: true)
     @protections = preview.protection_reasons(@person)
     @counts = preview.counts
     # The linked account's own content — only relevant when overriding, since that's
@@ -247,10 +247,10 @@ class PeopleController < ApplicationController
 
   # Admin-only deletion of a person and their whole data + audit graph. Skips a
   # person who looks real unless force=true overrides the protections (see
-  # FakeSubmissionRemover).
+  # PeopleRemover).
   def purge
     authorize! @person, to: :purge?
-    remover = FakeSubmissionRemover.new(person_ids: [ @person.id ], force: params[:force] == "true")
+    remover = PeopleRemover.new(person_ids: [ @person.id ], force: params[:force] == "true")
 
     if remover.call.include?(@person.id)
       redirect_to people_path, status: :see_other,
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -237,11 +237,9 @@ class PeopleController < ApplicationController
     preview = PeopleRemover.new(person_ids: [ @person.id ], force: true)
     @protections = preview.protection_reasons(@person)
     @counts = preview.counts
-    # The linked account's own content — only relevant when overriding, since that's
-    # when the account itself gets destroyed. Blocking content makes the delete
-    # impossible; cascading content is removed with the account.
-    @account_blocking = @force ? preview.blocking_account_content : {}
-    @account_cascading = @force ? preview.cascading_account_content : {}
+    # Everything removed with the linked account — only relevant when overriding,
+    # since that's when the account itself (and its content) gets destroyed.
+    @account_content = @force ? preview.account_content : {}
     @person = @person.decorate
   end
 
@@ -262,7 +260,7 @@ class PeopleController < ApplicationController
     end
   rescue ActiveRecord::InvalidForeignKey
     redirect_to edit_person_path(@person), status: :see_other,
-                alert: "Couldn't delete #{@person.full_name} — the linked User Account owns content (e.g. workshops, reports, or stories) that must be reassigned or removed first."
+                alert: "Couldn't delete #{@person.full_name} — the linked User Account is still referenced by records it doesn't own (e.g. a shared banner or event it created). Reassign those first."
   end
 
   def check_duplicates
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -1,6 +1,6 @@
 class PeopleController < ApplicationController
   include AhoyTracking, TagAssignable
-  before_action :set_person, only: %i[ show edit update destroy workshop_logs checkout bio ]
+  before_action :set_person, only: %i[ show edit update destroy workshop_logs checkout bio purge_confirmation purge ]
 
   def index
     authorize!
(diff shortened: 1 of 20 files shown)


app/services/people_remover.rb
```diff
@@ -0,0 +1,295 @@
...
+    user.super_user? ||
+      user.sign_in_count.to_i.positive? ||
+      user.current_sign_in_at.present? || user.last_sign_in_at.present? ||
+      Ahoy::Event.where(user_id: user.id).exists? ||
+      Person.where(created_by_id: user.id).where.not(id: target_person_ids).exists? ||
+      Person.where(updated_by_id: user.id).where.not(id: target_person_ids).exists?
+  end
+
+  def collected
+    @collected ||= begin
```
maebeale, line 234: 🤖 _From Claude:_ This is the heart of the "looks real" protection — an account is shielded if it has signed in, is a super_user, has its own Ahoy activity, or authored other people. It runs before any deletion so the audit cleanup can't erase the signal we judge by.

app/services/people_remover.rb
```diff
@@ -0,0 +1,295 @@
...
+      # cascades off Person. Users are destroyed before their person so the
+      # has_one :user nullify (which the model forbids) never runs.
+      Allocation.where(id: c[:allocation_ids]).find_each(&:destroy!)
+      Refund.where(id: c[:refund_ids]).find_each(&:destroy!)
+      Payment.where(id: c[:payment_ids]).find_each(&:destroy!)
+      Notification.where(id: c[:notification_ids]).find_each(&:destroy!)
+      Pay::Customer.where(id: c[:pay_customer_ids]).find_each(&:destroy!)
+      # The user's own notifications (noticeable: User) don't cascade off the user.
+      Notification.where(noticeable_type: "User", noticeable_id: c[:users].map(&:id)).delete_all
+      c[:users].each(&:destroy!)
```
maebeale: 🤖 _From Claude:_ Users are destroyed before their person on purpose: `Person has_one :user, dependent: :nullify` would try to null `users.person_id`, which the User model forbids once set — so we remove the account first and reset the cached association.

app/controllers/people_controller.rb
```diff
@@ -227,6 +227,44 @@ def destroy
...
+
+    if remover.call.include?(@person.id)
+      redirect_to people_path, status: :see_other,
+                  notice: "#{@person.full_name} and all their data were deleted."
+    else
+      reasons = remover.skipped.find { |skip| skip.person.id == @person.id }&.reasons
+      redirect_to edit_person_path(@person), status: :see_other,
+                  alert: "Couldn't delete — this person looks real (#{reasons&.join(', ')}). Use “Delete anyway” to override."
+    end
+  rescue ActiveRecord::InvalidForeignKey
```
maebeale, line 261: 🤖 _From Claude:_ Backstop for any FK we don't enumerate (an account that authored real content is FK-restricted). The DB rolls the whole transaction back, so this turns a would-be 500 into a clear "reassign or remove that content first" message.

app/services/people_remover.rb
```diff
@@ -0,0 +1,296 @@
...
+      Allocation.where(id: c[:allocation_ids]).find_each(&:destroy!)
+      Refund.where(id: c[:refund_ids]).find_each(&:destroy!)
+      Payment.where(id: c[:payment_ids]).find_each(&:destroy!)
+      Notification.where(id: c[:notification_ids]).find_each(&:destroy!)
+      Pay::Customer.where(id: c[:pay_customer_ids]).find_each(&:destroy!)
+      # Destroy everything the account authored (FK-restricted, leaf → root) and
+      # its own notifications before the account itself — these don't cascade off
+      # the user, and the FKs would otherwise block its deletion.
+      user_ids = c[:users].map(&:id)
+      authored_content_scopes(user_ids).each { |_, relation| relation.find_each(&:destroy!) }
```
maebeale, line 148: 🤖 _From Claude:_ Force now *deletes* the account's authored content (workshops/stories/reports/…) in FK-safe leaf→root order rather than blocking on it. It runs before the account destroy because that content is `created_by_id`-owned with no `dependent:`. Anything the account only references but doesn't own (a shared banner/event) still hits the FK and rolls the transaction back — surfaced by the controller rescue.

maebeale:
@jmilljr24 this has UI and a rake task for deep person+associated data deletion. 
the ui feature is hidden behind `?admin=true`, and the rake task attributes deletions to Umberto.
this is not critical for this week, but, would be nice and is a need i think they'll have occasionally going forward.

app/controllers/people_controller.rb
```diff
@@ -227,6 +227,42 @@ def destroy
...
+    # since that's when the account itself (and its content) gets destroyed.
+    @account_content = @force ? preview.account_content : {}
+    @person = @person.decorate
+  end
+
+  # Admin-only deletion of a person and their whole data + audit graph. Skips a
+  # person who looks real unless force=true overrides the protections (see
+  # PeopleRemover).
+  def purge
+    authorize! @person, to: :purge?
```
jmilljr24, line 250:
```suggestion
    authorize! @person
```

app/controllers/people_controller.rb
```diff
@@ -227,6 +227,42 @@ def destroy
     end
   end
 
+  # Admin-only interstitial: shows exactly what would be deleted for this person
+  # (and why they're protected, if so) before anything is removed.
+  def purge_confirmation
+    authorize! @person, to: :purge?
```
jmilljr24, line 233:
```suggestion
    authorize! @person
```
I think I would prefer to use the implicit policy name `purge_confirmation` and then in the policy add alias to `purge`.

Some times using `to:` in the controller makes sense if the action name doesn't make sense in the context of the policy file but that's not a problem in this one. 

( One example I did was an action in the tags controller called `sector`. It didn't make sense in the Sector policy to have a rule called `sector`, so the authorization pointed `to: :tags_index?)

app/policies/person_policy.rb
```diff
@@ -29,6 +29,14 @@ def destroy?
     admin? && record.persisted? && !has_associated_data?
   end
 
+  # Full purge of a person and their entire data graph. Bypasses the
+  # has_associated_data? guard on destroy? on purpose — purging exists precisely
+  # to remove people who have that data. The PeopleRemover service still
+  # refuses to touch anyone who looks real unless explicitly forced.
```
jmilljr24, line 35: As I mentioned before, comments don't really bother me. I'm just wondering if we can tone them down a bit to reduce some clutter. My only concern is if everything is getting a comment, the really important comments might not get noticed in the sea of comments.

The comment doesn't tell me anything about why this rule is unique or different (with respect to policy stuff). It's like explaining what the people remover does.

app/services/people_remover.rb
```diff
@@ -0,0 +1,324 @@
...
+    end
+  end
+
+  # A User is real (and must never be deleted) if it has ever signed in, is a
+  # super_user, has Ahoy activity of its own, or authored other people. This runs
+  # before any deletion, so cleaning the fake graph can't erase the signal.
+  def real_user?(user)
+    user.super_user? ||
+      user.sign_in_count.to_i.positive? ||
+      user.current_sign_in_at.present? || user.last_sign_in_at.present? ||
```
jmilljr24, line 227:
```suggestion
      user.sign_in_count.to_i.positive? ||
```
Seems redundant but doesn't hurt.

app/services/people_remover.rb
```diff
@@ -0,0 +1,324 @@
...
+      c[:users].each { |user| user.update_column(:person_id, nil) }
+      deletable.each { |person| person.association(:user).reset }
+
+      # Non-cascading person records, FK-safe order: allocations and refunds
+      # reference payments; payments reference submissions.
+      Allocation.where(id: c[:allocation_ids]).find_each(&:destroy!)
+      Refund.where(id: c[:refund_ids]).find_each(&:destroy!)
+      Payment.where(id: c[:payment_ids]).find_each(&:destroy!)
+      Notification.where(id: c[:notification_ids]).find_each(&:destroy!)
+      Pay::Customer.where(id: c[:pay_customer_ids]).find_each(&:destroy!)
```
jmilljr24, line 144: I get that this is supposed to be used with extra care, but I can't help but wonder the consequences of anything around payments and pay gem.

app/services/people_remover.rb
```diff
@@ -0,0 +1,324 @@
...
+  rescue
+    # Roll back the lifecycle events this purge buffered so a failed delete can't
+    # leave phantom Ahoy "destroy" records for records that still exist.
+    Analytics::LifecycleBuffer.store.slice!(buffered_lifecycle_events..)
+    raise
+  end
+
+  private
+
+  attr_reader :delete_users, :force
```
jmilljr24, line 170: The initializer already sets the instance variables, this seems unneeded.
```suggestion
```

jmilljr24 (commented):
I can't really run the code until this is rebased but nothing major stands out just reading the code.

PR #1814, opened 2026-06-21
Highlight only exact "Facilitator" affiliations; refactor editor row

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: an exact-match predicate for the row highlight plus a view refactor of the affiliation editor row

### Why
- The facilitator affiliation editor card is highlighted (purple) to flag facilitators. The highlight should fire only for a plain "Facilitator", but it was driven by `facilitator?`, a fuzzy substring match shared with a lot of business logic.

### Changes
- `Affiliation#facilitator?` stays a fuzzy, case-insensitive substring match (consistent with the `.facilitators` scope and all callers: org pills, `program_organization`, staff/recipient lists, tenure badges).
- New `Affiliation#exact_facilitator?` — exact, case-sensitive `"Facilitator"` (trimmed) — drives **only** the editor-row highlight; mirrored in `inactive_toggle_controller.js` so the row recolors live while editing the title.
- Card highlight is purple (`bg-purple-100`), matching the existing facilitator title badge.
- Left accent stripe moved outside the dimmed element so an expired facilitator row keeps full-color purple instead of fading with `opacity-60`.
- Extracted shared `affiliations/_fields` partial (both org and person forms render it).
- Editor row tidy: bottom-aligned Remove inline with the checkbox, two-line "Primary org / contact" label, Title/Start/End labels sized to match the rest.


maebeale:
Finally!

Screenshot 2026-06-21 at 3 06 09 PM

PR #1815, opened 2026-06-21
Match facilitator affiliations exactly and case-sensitively

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: small, contained logic changes with low blast radius

- Facilitator status, "facilitator since" dates, tenure badges, program status, the org-active determination, the editor card highlight, and the live JS figures all keyed off a fuzzy match (`LIKE '%facilitator%'` / case-insensitive `include?`), which counted variants like "Lead Facilitator" and "Co-Facilitator".
- Restrict every facilitator decision to the exact, case-sensitive title "Facilitator" (whitespace-trimmed) so a single canonical role governs them.

- Made `Affiliation#facilitator?` and the `.facilitators` scope exact (`BINARY TRIM(title) = "Facilitator"`; BINARY is needed for case sensitivity under MySQL's default collation). Every call site flows through one of them.
- The highlight rule (`exact_facilitator?`, previously the only exact matcher) and the business-logic rule now agree, so I collapsed them into one `#facilitator?` and dropped the redundant `exact_facilitator?`.
- Updated the two Stimulus controllers (`affiliation_dates`, `affiliation_facilitator_warning`) to mirror the same trimmed exact rule.

### Anything else to add?
- "Facilitator" is confirmed to be the only canonical affiliation title in use, so no data migration is needed.
- The cosmetic title-chip sort heuristics in a few view partials are intentionally left fuzzy — they only order display pills, not the facilitator-status determination.

### DIFF
--- a/app/decorators/organization_decorator.rb
+++ b/app/decorators/organization_decorator.rb
@@ -67,11 +67,11 @@ class OrganizationDecorator < ApplicationDecorator
   end
 
   def facilitator_since_date
-    @facilitator_since_date ||= affiliations.where("title LIKE ?", "%Facilitator%").minimum(:start_date)
+    @facilitator_since_date ||= affiliations.facilitators.minimum(:start_date)
   end
 
   def facilitation_end_date
-    facilitator_affiliations = affiliations.where("title LIKE ?", "%Facilitator%")
+    facilitator_affiliations = affiliations.facilitators
     return nil if facilitator_affiliations.active.exists?
     facilitator_affiliations.maximum(:end_date)
   end
--- a/app/frontend/javascript/controllers/affiliation_dates_controller.js
+++ b/app/frontend/javascript/controllers/affiliation_dates_controller.js
@@ -54,9 +54,11 @@ export default class extends Controller {
       ? new Date(Math.max(...affiliations.map(a => new Date(a.endDate))))
       : null
 
-    // Facilitator since/end — same logic filtered by title
+    // Facilitator since/end — same logic filtered by title. Mirror
+    // Affiliation#facilitator?: an exact, case-sensitive match on "Facilitator"
+    // (trimmed), so the live figure matches what the server will render.
     const facilitatorAffiliations = affiliations.filter(a =>
-      a.title.toLowerCase().includes("facilitator")
+      a.title.trim() === "Facilitator"
     )
     const facStartDates = facilitatorAffiliations.map(a => a.startDate).filter(Boolean)
     const facilitatorSince = facStartDates.length
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -10,7 +10,11 @@ class Affiliation < ApplicationRecord
       .where("affiliations.end_date IS NULL OR affiliations.end_date >= ?", Date.current)
   }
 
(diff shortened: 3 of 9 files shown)


app/models/affiliation.rb
```diff
@@ -10,7 +10,10 @@ class Affiliation < ApplicationRecord
       .where("affiliations.end_date IS NULL OR affiliations.end_date >= ?", Date.current)
   }
 
-  scope :facilitators, -> { where("title LIKE ?", "%facilitator%") }
+  # Only the exact, case-sensitive title "Facilitator" counts — variants like
+  # "Lead Facilitator" or "facilitator" are deliberately excluded. BINARY forces
```
maebeale, line 14: 🤖 _From Claude:_ `BINARY` is required here — the DB is MySQL/Trilogy with a case-insensitive default collation, so a plain `where(title: "Facilitator")` would still match "facilitator". This makes the SQL scope agree with the case-sensitive Ruby `==` in `#facilitator?`.

app/decorators/organization_decorator.rb
```diff
@@ -67,11 +67,11 @@ def affiliation_end_date
   end
 
   def facilitator_since_date
-    @facilitator_since_date ||= affiliations.where("title LIKE ?", "%Facilitator%").minimum(:start_date)
+    @facilitator_since_date ||= affiliations.facilitators.minimum(:start_date)
```
maebeale, line 70: 🤖 _From Claude:_ These two methods previously matched fuzzily on their own (`LIKE '%Facilitator%'`); routing them through `.facilitators` keeps the exact-match rule in one place.

app/models/affiliation.rb
```diff
@@ -20,15 +24,11 @@ class Affiliation < ApplicationRecord
   after_destroy :sync_organization_affiliation_dates
 
   # Methods
+  # A facilitator affiliation is one whose title is *exactly* "Facilitator"
+  # (trimmed, case-sensitive). Variants like "Lead Facilitator" or "facilitator"
+  # are deliberately excluded. Mirrors the .facilitators scope so in-memory and
+  # SQL checks agree.
```
maebeale, line 30: 🤖 _From Claude:_ Collapsed the former `exact_facilitator?` into `#facilitator?` — now that business logic is exact too, the two predicates were identical. `BINARY TRIM` mirrors this strip + case-sensitivity in SQL.

app/frontend/javascript/controllers/affiliation_dates_controller.js
```diff
@@ -54,9 +54,11 @@ export default class extends Controller {
       ? new Date(Math.max(...affiliations.map(a => new Date(a.endDate))))
       : null
 
-    // Facilitator since/end — same logic filtered by title
+    // Facilitator since/end — same logic filtered by title. Mirror
+    // Affiliation#facilitator?: an exact, case-sensitive match on "Facilitator"
+    // (trimmed), so the live figure matches what the server will render.
     const facilitatorAffiliations = affiliations.filter(a =>
```
maebeale, line 60: 🤖 _From Claude:_ Both facilitator Stimulus controllers now mirror the server's exact, trimmed rule so the live figures match what gets rendered/saved (per the CLAUDE.md JS-mirroring convention).

PR #1816, opened 2026-06-21
Show "Events attended" on org profiles (was unimplemented "Events hosted")

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: new read-only profile section + a status filter on an existing one

The org profile's "Events hosted" section was an unimplemented stub, and the label was wrong — there's no org→event hosting relationship in the data model, only the events the org's affiliated members registered for.

- Rename the org section to **Events attended** and populate it from the org's active event registrations (`event_registration_organizations`), deduped to distinct events. Lazy-loads via a Turbo frame and reuses the person profile's card grid.
- Match the **person** profile: rename its heading "Events registered" → "Events attended", and scope both profiles to `EventRegistration.active`, so cancelled / no-show registrations drop off ("show until inactive").

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -36,6 +36,15 @@ class OrganizationsController < ApplicationController
 
   def show
     authorize! @organization
+
+    if turbo_frame_request? && params[:section] == "events"
+      events = Event.where(id: @organization.event_registrations.active.select(:event_id))
+                    .includes(:primary_asset)
+                    .order(start_date: :desc)
+                    .paginate(page: params[:page], per_page: 9)
+      return render partial: "organizations/sections/events", locals: { organization: @organization, events: events }
+    end
+
     track_view(@organization)
 
     workshop_logs = WorkshopLog.where(organization_id: @organization.id)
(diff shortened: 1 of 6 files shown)


app/controllers/organizations_controller.rb
```diff
@@ -36,6 +36,15 @@ def index
 
   def show
     authorize! @organization
+
+    if turbo_frame_request? && params[:section] == "events"
+      events = Event.where(id: @organization.event_registrations.active.select(:event_id))
+                    .includes(:primary_asset)
+                    .order(start_date: :desc)
+                    .paginate(page: params[:page], per_page: 9)
```
maebeale, line 44: 🤖 _From Claude:_ Orgs link to events only through their members' registrations, so several members can register for the same event — querying distinct `Event`s (rather than registrations) dedupes the cards. `.active` excludes cancelled/no-show.

app/controllers/people_controller.rb
```diff
@@ -54,7 +54,7 @@ def show
         @stories = Story.where(id: story_ids).order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/stories", locals: { person: @person, stories: @stories }
       when "events"
-        @event_registrations = @person.event_registrations.includes(:event).order("events.start_date DESC").references(:events).paginate(page: params[:page], per_page: per_page)
+        @event_registrations = @person.event_registrations.active.includes(:event).order("events.start_date DESC").references(:events).paginate(page: params[:page], per_page: per_page)
```
maebeale, line 57: 🤖 _From Claude:_ Added `.active` here so the person profile matches the org profile — cancelled/no-show registrations no longer appear.

PR #1818, opened 2026-06-21
Link an affiliation to one of its organization's addresses

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: new optional FK + nested-attr param + a small Stimulus dropdown

Organizations can have several addresses. This lets an editor record *which* address a person's affiliation is tied to, rather than leaving it ambiguous.

- New optional `affiliations.organization_address_id` FK → `addresses` (`on_delete: :nullify`); migration is reversible.
- `Affiliation belongs_to :organization_address` (optional) with a validation that the address actually belongs to the affiliation's organization. `Address has_many :affiliations` with `dependent: :nullify`.
- Permitted `organization_address_id` in both the organizations and people affiliations nested params (the editor row is shared).
- UI: a compact numbered picker in the shared affiliation editor row — the trigger is only wide enough for a double-digit number, the panel shows each address's full one-line text. Numbers mirror the org address editor's "Address #N".
- New `address-select` Stimulus controller (open/close, outside-click + escape, writes the id to a hidden field).

### Editor-row layout & behavior tweaks
- Narrowed the Title field so the new Address column fits and Remove stays inline.
- Grey-filled the picker trigger (`bg-gray-100`) so it reads like the other optional fields.
- When an org has no addresses, the column renders an equal-width spacer so Start/End/Primary stay vertically aligned across rows.
- All addresses are selectable; inactive ones are shown with an `[INACTIVE]` marker and keep their number.

### Performance
- Preload each org's addresses on the person edit form, and add `inverse_of` between `Organization`/`Affiliation` so the org edit form reuses the already-loaded organization — removes the per-row address queries (N+1).

### Anything else to add?
- Numbering reflects each address's real position so it matches the address section.

### DIFF
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -1,9 +1,12 @@
 class Affiliation < ApplicationRecord
   belongs_to :organization
   belongs_to :person, touch: true
+  # Which of the organization's addresses this person is affiliated with (optional).
+  belongs_to :organization_address, class_name: "Address", optional: true
 
   # Validations
   validates_presence_of :organization_id
+  validate :organization_address_belongs_to_organization
 
   scope :active, -> {
     where(inactive: false)
@@ -44,6 +47,16 @@ class Affiliation < ApplicationRecord
   end
 
   private
+
+  # The linked address must be one of this affiliation's organization's own
+  # addresses — not a stray address or another org's / person's address.
+  def organization_address_belongs_to_organization
+    return if organization_address.blank?
+
+    valid = organization_address.addressable_type == "Organization" &&
+            organization_address.addressable_id == organization_id
+    errors.add(:organization_address_id, "must be an address of this organization") unless valid
+  end
 
   def skip_if_duplicate
     scope = Affiliation.where(
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -1,5 +1,5 @@
 class Affiliation < ApplicationRecord
-  belongs_to :organization
+  belongs_to :organization, inverse_of: :affiliations
   belongs_to :person, touch: true
   # Which of the organization's addresses this person is affiliated with (optional).
   belongs_to :organization_address, class_name: "Address", optional: true
--- a/app/views/affiliations/_address_picker.html.erb
+++ b/app/views/affiliations/_address_picker.html.erb
@@ -1,13 +1,17 @@
 <%# Compact numbered picker linking an affiliation to one of its organization's
     addresses. The trigger button is only wide enough for a double-digit number;
     the panel that opens shows each address's full one-line text. Numbers mirror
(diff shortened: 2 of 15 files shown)


app/models/affiliation.rb
```diff
@@ -45,6 +48,16 @@ def name
 
   private
 
+  # The linked address must be one of this affiliation's organization's own
+  # addresses — not a stray address or another org's / person's address.
+  def organization_address_belongs_to_organization
```
maebeale, line 53: 🤖 _From Claude:_ Guards against linking a stray address — the address must be one of this affiliation's own organization's addresses (not another org's or a person's).

app/views/affiliations/_address_picker.html.erb
```diff
@@ -0,0 +1,41 @@
+<%# Compact numbered picker linking an affiliation to one of its organization's
+    addresses. The trigger button is only wide enough for a double-digit number;
+    the panel that opens shows each address's full one-line text. Numbers mirror
+    the org address editor's "Address #N" (1-based, in association order); only
+    active addresses are offered. Renders nothing when the org has none. %>
+<% org = f.object.organization %>
+<% numbered = org ? org.addresses.each_with_index.filter_map { |address, i| [ i + 1, address ] if address.active? } : [] %>
+<% if numbered.any? %>
```
maebeale: 🤖 _From Claude:_ Numbers are 1-based over the org's addresses in association order so they match the org address editor's "Address #N"; only active addresses are offered, so the visible numbers can be non-contiguous by design.

db/migrate/20260621213947_add_organization_address_to_affiliations.rb
```diff
@@ -0,0 +1,13 @@
+class AddOrganizationAddressToAffiliations < ActiveRecord::Migration[8.1]
+  def up
+    add_column :affiliations, :organization_address_id, :bigint
+    add_index :affiliations, :organization_address_id
+    add_foreign_key :affiliations, :addresses, column: :organization_address_id, on_delete: :nullify
```
maebeale, line 5: 🤖 _From Claude:_ `on_delete: :nullify` (mirrored by `dependent: :nullify` on Address) so removing an org address just clears the link instead of blocking deletion.

maebeale:
Screenshot 2026-06-21 at 6 18 44 PM

maebeale:
@jmilljr24 Christy told me they need to connect people to specific organization addresses, so I added a starter way to do it. I tried to avoid it, but it needed a js controller to keep the selection display be just the ID while dropdown is full address. 

If we don't add this now and bake it in to the registration data service, we'll need to backport. 

I'm ok w waiting, but just coming through to-do's and looking for any data gotchas and found this.

PR #1821, opened 2026-06-22
WAIT: Show recipient job title, agency, and city/state on the recipients page

maebeale:
REVIEW NEEDED: 📖 Read — light-logic: a new EventDashboard method + a view swap, no migrations or data changes

Each scholarship recipient's affiliation line(s) on the recipients page now derive from the org(s) connected to their **registration** (the snapshot taken at registration plus any an admin linked later), not from all of their affiliations.

- For each connected org: show the registrant's **active job title** with the **agency name (City, ST)** (`Organization#program_location`).
- No active job affiliation there → fall back to a **title-less** line, but only when they're a **facilitator** at that org and the org's status is **Active** or **Pending**.
- Several connected orgs → several lines (sorted by name); none qualifying → no line.

Logic lives in `EventDashboard#recipient_affiliations_by_registrant` (preloads org addresses + status to avoid N+1); the view just renders the rows.

## Why
The old line iterated every non-facilitator affiliation active at the event date. Anchoring on the registration's orgs keeps what's shown tied to what the registration is actually about, and adds the agency location + a facilitator fallback.

### DIFF
--- a/app/services/event_dashboard.rb
+++ b/app/services/event_dashboard.rb
@@ -59,6 +59,25 @@ class EventDashboard
       .sort_by(&:name)
   end
 
+  # One organization line shown beneath a recipient's name: the org and, for a
+  # job affiliation, the registrant's title there (nil for the facilitator
+  # fallback, where the title is deliberately suppressed).
+  RecipientAffiliation = Struct.new(:organization, :title, keyword_init: true)
+
+  # The organization line(s) under each recipient's name on the recipients page,
+  # keyed by Person id. Driven by the org(s) connected to the registrant's
+  # registration (the snapshot taken at registration plus any an admin linked
+  # later) rather than all of the person's affiliations. For each connected org
+  # we show the title from the registrant's active, non-facilitator job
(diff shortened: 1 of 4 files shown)


app/services/event_dashboard.rb
```diff
@@ -660,6 +679,34 @@ def scholarship_applicant_ids
     @scholarship_applicant_ids ||= active_registrations.where(scholarship_requested: true).pluck(:registrant_id)
   end
 
+  # Org statuses that still warrant showing a facilitator-only fallback line.
+  FALLBACK_FACILITATOR_STATUSES = %w[Active Pending].freeze
```
maebeale, line 683: 🤖 _From Claude:_ "active_or_pending" is read as the **organization's status** (Active/Pending) — the only Active/Pending concept in the schema (affiliations have no pending state). The facilitator fallback shows only when the registrant facilitates at the org *and* the org is Active/Pending.

maebeale:
Screenshot 2026-06-21 at 9 06 21 PM

PR #1824, opened 2026-06-22
Link registration affiliations to the org's address (+ country, backfill)

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — contained logic across three flows plus one new seed field; well covered by specs

Connect an affiliation to its org's address automatically (instead of only via the manual picker from #1818), and make sure the full agency address — including country — flows through.

## Why
A registrant types their agency address on the form (and an admin links/creates the org from that submission), but `affiliation.organization_address` was never set, and the typed **country** never reached the address at all.

## What
- **`OrganizationServices::UpsertAddress`** (new) — extracts the agency-address find-or-create/update out of `PublicRegistration` so registration and the admin org-linking actions build the org address identically (street/city/state/zip/**country**).
- **`CreateFromRegistration`** accepts `organization_address:` and sets it on every affiliation it creates; when an affiliation already exists and is skipped, it **backfills** the address onto it only if it has none (never overwrites an admin-set address).
- **Wired at all three entry points:** public registration, select-existing-org, create-new-org.
- **`agency_country`** added to the registration-form seed builder (prod already collects it; the seed was missing it) and threaded into the upsert so it lands on `addresses.country`.

## Tests
- New `UpsertAddress` spec (incl. country on create + update); affiliation-service backfill specs; public-registration + request specs asserting affiliations link to the built address and country is captured; form-builder seed spec for the new field.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -215,12 +215,7 @@ class EventRegistrationsController < ApplicationController
     @person = @event_registration.registrant
     organization = Organization.find(params[:organization_id])
 
-    AffiliationServices::CreateFromRegistration.call(
-      person: @person,
-      organization: organization,
-      job_title: submitted_position(@event_registration),
-      training_date: @event_registration.event.start_date
-    )
+    link_affiliations_for(@event_registration, organization)
 
     @event_registration.event_registration_organizations
       .find_or_create_by!(organization: organization)
@@ -248,12 +243,7 @@ class EventRegistrationsController < ApplicationController
     existing = Organization.where("LOWER(name) = ?", name.strip.downcase).first
     organization = existing || Organization.create!(name: name.strip, organization_status: OrganizationStatus.find_by(name: "Active"))
 
-    AffiliationServices::CreateFromRegistration.call(
-      person: @person,
-      organization: organization,
-      job_title: submitted_position(@event_registration),
-      training_date: @event_registration.event.start_date
-    )
+    link_affiliations_for(@event_registration, organization)
     @event_registration.event_registration_organizations.find_or_create_by!(organization: organization)
 
     notice = existing ? "#{organization.name} linked." : "#{organization.name} created and linked."
@@ -386,10 +376,8 @@ class EventRegistrationsController < ApplicationController
     return [] unless form
 
     field_ids = form.form_fields
-      .where(field_identifier: %w[agency_name agency_position])
+      .where(field_identifier: %w[agency_name agency_position agency_street agency_city agency_state agency_zip agency_country])
       .pluck(:field_identifier, :id).to_h
-    name_field_id = field_ids["agency_name"]
-    position_field_id = field_ids["agency_position"]
 
     entries = registration.registrant.form_submissions
       .where(form: form)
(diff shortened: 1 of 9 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -49,8 +49,10 @@ def call
         create_phone_contact(person) if field_value("phone").present?
 
         organization = find_organization if field_value(ORGANIZATION_NAME_IDENTIFIER).present?
-        create_affiliation(person, organization) if organization
-        create_agency_address(organization) if organization && field_value("agency_city").present?
+        if organization
+          agency_address = create_agency_address(organization)
```
maebeale, line 56: 🤖 _From Claude:_ Reordered so the agency address is created *before* the affiliation — `CreateFromRegistration` needs the address to link it. Previously the affiliation was made first and the address after, so they were never connected.

app/services/organization_services/upsert_address.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(organization:, city: nil, state: nil, street_address: nil, zip_code: nil)
+      @organization = organization
+      @city = city&.strip
+      @state = state&.strip
+      @street_address = street_address
+      @zip_code = zip_code
+    end
+
+    def call
+      return if @city.blank?
```
maebeale, line 29: 🤖 _From Claude:_ Returns nil on a blank city (addresses are keyed on city), so all three callers can invoke this unconditionally and pass the nil straight through to `organization_address:` — preserving the old "only build an address when a city was typed" guard.

app/controllers/event_registrations_controller.rb
```diff
@@ -427,4 +422,30 @@ def submitted_position(registration)
...
+  def submitted_agency_address(registration)
+    entries = registration_submission_entries(registration)
+    primary = entries.find { |entry| entry[:org_name].present? } || entries.first
+    primary && primary[:address] || {}
+  end
+
+  # Upsert the linked org's work address from the registrant's submission and
+  # create the job + facilitator affiliations linked to it. Shared by the
+  # select-existing and create-new org-linking actions.
+  def link_affiliations_for(registration, organization)
```
maebeale, line 442: 🤖 _From Claude:_ Shared by both linking actions (select existing / create new) so they upsert the org address and link the affiliations identically — mirroring what `PublicRegistration` does at registration time.

app/services/form_builder_service.rb
```diff
@@ -407,6 +407,8 @@ def build_person_contact_info_fields(form, position)
                          key: "agency_state", group: "person_contact_info", required: false, width: :third)
     position = add_field(form, position, "Organization Zip / Postal Code", :free_form_input_one_line,
                          key: "agency_zip", group: "person_contact_info", required: false, width: :third)
+    position = add_field(form, position, "Organization Country", :free_form_input_one_line,
+                         key: "agency_country", group: "person_contact_info", required: false, width: :third)
```
maebeale: 🤖 _From Claude:_ Prod already collects Organization Country; the seed builder was just missing it. Added here (and to SECTION_FIELD_IDENTIFIERS/SECTION_FIELD_NAMES, which a sync spec enforces) so newly-seeded dev forms match prod and `agency_country` resolves through to `addresses.country`.

PR #1825, opened 2026-06-22
Match returning registrants across a nickname/legal first-name swap

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — small, contained change to one matching method, covered by specs

Stop incognito registrations from creating duplicate people when the registrant's first name and nickname swap places.

## Why
Anonymous registrations match an existing `Person` by exact `first_name + last_name + email`. The registrant's "first name" is the **nickname** when they give one (`resolve_names`), so a returning person who types their legal first name where we stored their nickname — or the reverse — slips past the match and registers as a **duplicate Person**. Logged-in registrations are unaffected (they arrive with the person already resolved).

## What
`find_matching_person` now matches on **email + last name** (strong, stable identifiers) and accepts **either** the typed first name **or** the nickname against **either** the stored `first_name` **or** `legal_first_name`.

This only *widens* the previous match — every registrant that matched before still matches — so there's no risk of a returning person newly being treated as new. The last-name + email gate keeps it from merging genuinely different people who happen to share an email.

## Tests
- Matches a nickname-stored person who types their legal name, and the reverse.
- Still creates a new person when the email matches but it's a different name (guard against over-merging).

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -129,10 +129,7 @@ module EventRegistrationServices
       email = field_value("primary_email")&.strip&.downcase
       email_type = field_value("primary_email_type")&.downcase
 
-      person = Person.find_by(
-        "LOWER(first_name) = ? AND LOWER(last_name) = ? AND LOWER(email) = ?",
-        first_name&.downcase, last_name&.downcase, email&.downcase
-      )
+      person = find_matching_person(last_name: last_name, email: email)
       return person if person
 
       Person.create!(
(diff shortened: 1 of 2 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -147,6 +144,28 @@ def find_or_create_person
...
+      return if email.blank? || last_name.blank?
+
+      first_names = [ field_value("first_name"), field_value("nickname") ]
+        .filter_map { |value| value&.strip.presence&.downcase }
+        .uniq
+      return if first_names.empty?
+
+      Person
+        .where("LOWER(last_name) = ? AND LOWER(email) = ?", last_name.downcase, email.downcase)
+        .where("LOWER(first_name) IN (:names) OR LOWER(COALESCE(legal_first_name, '')) IN (:names)", names: first_names)
```
maebeale, line 165: 🤖 _From Claude:_ The `:names` set holds both the typed first name and the nickname; matching it against `first_name` OR `legal_first_name` is what tolerates the swap. Strictly a superset of the old exact match, so no previously-matched registrant stops matching.

PR #1826, opened 2026-06-22
WAIT: Match interstitial duplicates on legal first name too

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained change to the duplicate-person name query

The new-person duplicate interstitial matched the typed first name (and its nickname variants) only against the stored `first_name` column. A returning registrant whose legal name lives in `legal_first_name` (nickname in `first_name`) who types their legal name slipped past the name check — the exact duplicate the interstitial exists to surface.

This is the follow-up to the registration-side fix (`PublicRegistration#find_matching_person`), ported the other direction: high-recall instead of swapping in the conservative matcher.

- Match the first-name variants against **either** `first_name` **or** `legal_first_name` (period/space normalized, mirroring the existing column normalization).
- Count a legal-name hit as an exact name match so the "name match" badge and sort order behave like a first-name hit.
- Failing test first (red/green): added a "Legal first name matching" context to the request spec.

### Anything else to add?
Kept the interstitial's own normalization (strip `.`/spaces, nickname variants) rather than the registration matcher's plain `LOWER` — fully aligning the two normalizations is a separate, larger change and isn't needed to close this duplicate.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -293,16 +293,26 @@ class PeopleController < ApplicationController
       normalized_last = NicknameMap.normalize(last_name)
       normalized_first = NicknameMap.normalize(first_name)
 
+      # Match the typed first name (and its nickname variants) against either the
+      # stored first_name or legal_first_name, so a returning registrant whose legal
+      # name lives in legal_first_name (nickname in first_name) is still surfaced when
+      # they type their legal name. Mirrors the registration matcher in
+      # EventRegistrationServices::PublicRegistration#find_matching_person.
       name_matches = Person.includes(:user)
                            .where("REPLACE(REPLACE(LOWER(last_name), '.', ''), ' ', '') = ?", normalized_last)
-                           .where("REPLACE(REPLACE(LOWER(first_name), '.', ''), ' ', '') IN (?)", first_variants)
+                           .where(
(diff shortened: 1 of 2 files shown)


app/controllers/people_controller.rb
```diff
@@ -293,16 +293,26 @@ def find_duplicate_people(first_name, last_name, email)
...
+                           )
                            .limit(10)
 
       name_matches.each do |person|
         next if duplicate_ids.include?(person.id)
 
         duplicate_ids.add(person.id)
-        exact_name = NicknameMap.normalize(person.first_name) == normalized_first
+        exact_name = NicknameMap.normalize(person.first_name) == normalized_first ||
+          NicknameMap.normalize(person.legal_first_name) == normalized_first
```
maebeale, line 315: 🤖 _From Claude:_ A legal-name hit counts as an exact name match ("name match" badge + name-priority sort), matching how a `first_name` hit is treated. `normalized_first` is never blank here (guarded by `if first_name.presence`), so a nil/blank `legal_first_name` normalizes to `""` and can't false-match.

maebeale, line 315: 🤖 _From Claude:_ A legal-name hit counts as an exact name match (name-match badge + name-priority sort), matching how a first_name hit is treated. normalized_first is never blank here (guarded by `if first_name.presence`), so a nil/blank legal_first_name normalizes to an empty string and can't false-match.

PR #1827, opened 2026-06-22
WAIT: Apply registrant's org type & website to linked organizations

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: contained change to two org-linking actions plus the editor's "primary submission" pick

## Why

Registration captures **Organization Type** and **Website** only as `FormAnswer` rows — they were never written to the Organization's `agency_type` / `website_url` columns. Linking a registrant to an org (new or existing) now applies what they typed.

## What

- `select_organization` and `create_organization` apply the registrant's `agency_type` / `agency_website` answers to the linked org.
- Applies to **both** new orgs (`create_organization`) and existing orgs (`select_organization`).
- **Overrides** curated values — the registrant's answer wins (per request). A **blank** answer is skipped so it can't wipe a curated value.
- **Primary submission = newest one that collected an org address.** A registrant can have several submissions, and not every form asks for org info, so the newest overall may carry none. We pick the newest submission with `agency_city` present (the same signal `public_registration` uses to build the address), falling back to the newest that named an org, then the newest overall. Consolidated the editor's duplicate "primary" logic into the shared `primary_submission_entry` helper so display and applied values stay in sync.

## Safety

- Every org change is captured as an **Ahoy lifecycle event** (`AhoyTrackable#after_update`) with before/after values and `Current.user`, so overrides are auditable.

### DIFF
--- a/app/controllers/event_registrations_controller.rb
+++ b/app/controllers/event_registrations_controller.rb
@@ -187,9 +187,10 @@ class EventRegistrationsController < ApplicationController
     # duplicate data), so we surface them all rather than silently picking one.
     @submitted_entries = registration_submission_entries(@event_registration)
     @form_submission = @submitted_entries.first&.fetch(:submission)
-    # The "primary" submitted org/title — the first submission that named an org
-    # (else the first submission) — drives the suggested-match and comparison logic.
-    primary = @submitted_entries.find { |entry| entry[:org_name].present? } || @submitted_entries.first
+    # The "primary" submitted org/title — the newest submission that named an org
+    # (else the newest submission) — drives the suggested-match and comparison logic,
+    # and is the same submission whose details linking applies to the org.
+    primary = primary_submission_entry(@submitted_entries)
     @submitted_org_name = primary && primary[:org_name]
     @submitted_position = primary && primary[:position]
     # Each distinct submitted org name that isn't already in the database gets its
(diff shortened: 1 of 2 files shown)


app/controllers/event_registrations_controller.rb
```diff
@@ -418,13 +428,36 @@ def submitted_agency_names(registration)
...
+
+  # The Organization Type and Website the registrant typed on the registration form,
+  # as Organization attributes ({ agency_type:, website_url: }) — only the values the
+  # registrant actually provided, so a blank answer never clobbers a curated value.
+  def submitted_organization_attributes(registration)
+    entry = primary_submission_entry(registration)
+    return {} unless entry
+
+    { agency_type: entry[:agency_type], website_url: entry[:agency_website] }
+      .select { |_, value| value.present? }
```
maebeale, line 463: 🤖 _From Claude:_ This `presence` filter is the one judgment call: we override curated `agency_type`/`website_url` with the registrant's answer, but a *blank* answer is dropped so it can't wipe a curated value. If you ever want new-org vs. existing-org to differ (e.g. backfill-only for existing), this is the seam to branch on.

app/controllers/event_registrations_controller.rb
```diff
@@ -418,13 +429,37 @@ def submitted_agency_names(registration)
       .uniq { |name| name.downcase }
   end
 
+  # The "primary" entry among a registrant's submission entries — the one whose
+  # details we apply to a linked org and surface in the editor: the NEWEST submission
+  # that named an org, else the newest submission. Entries come oldest-first, so we
+  # scan from the end. link_organization picks the same one, so what we apply matches
+  # the editor.
+  def primary_submission_entry(entries)
+    entries.reverse.find { |entry| entry[:org_name].present? } || entries.last
```
maebeale: 🤖 _From Claude:_ Primary = the **newest** submission that named an org (entries are oldest-first, so we scan from the end). `link_organization` now calls this same helper, so the answers the editor surfaces match the ones linking applies — previously the editor had its own inline first-named copy of this logic.

app/controllers/event_registrations_controller.rb
```diff
@@ -418,13 +431,43 @@ def submitted_agency_names(registration)
...
+  # details we apply to a linked org and surface in the editor. Not every form asks
+  # for organization info, so the newest submission overall may carry none; we pick
+  # the NEWEST submission that actually collected an org address (agency_city present
+  # — the same signal public_registration uses to build the address), falling back to
+  # the newest that named an org, then the newest overall. Entries come oldest-first,
+  # so we scan from the end. link_organization picks the same one, so what we apply
+  # matches the editor.
+  def primary_submission_entry(entries)
+    newest_first = entries.reverse
+    newest_first.find { |entry| entry[:agency_city].present? } ||
```
maebeale, line 444: 🤖 _From Claude:_ Primary now keys on `agency_city` (the signal `public_registration` uses for whether a submission carries an org address), not org name — a later submission from a form that did not ask for org info no longer wins. Falls back to newest-named, then newest overall, so name-only or org-less data still degrades gracefully.

PR #1830, opened 2026-06-22
Add a paid-in-full receipt for event registrations

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: a new read-only receipt page modeled on the existing invoice, gated to paid-in-full registrations.

- Gives registrants a printable **RECEIPT** (proof of payment) once their balance is settled, alongside the existing invoice.

- New `EventReceipt` presenter mirrors `EventInvoice` but adds the allocation ledger and a summary that reconciles to **$0 balance due**.
- New `receipt` action on `Events::RegistrationsController`, gated by `EventRegistration#receipt_available?` (paid event + zero balance).
- `EventRegistration#invoice_available?` gates the invoice too — **free events get neither** document.
- Reachable from the **Forms** callout (W-9, invoice, receipt) and the **Payment** page (invoice always, receipt once paid), so both are available even after the balance is paid. Reuses main's `return_to` eyebrow convention (forms/payment/ticket).
- Ledger labels payments by method (Cash / Check / Credit card), folding in the check number when present.
- Squared the invoice + receipt boxes (`rounded-2xl` → `rounded-lg`).

### Anything else to add?
- Rebased onto main after it independently landed the `return_to` invoice work and moved payment/allocation methods into the `Registerable` concern; receipt is layered on top of that structure.
- Receipt reachable by slug (same public auth as the invoice/ticket).

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -309,6 +309,17 @@ class EventRegistration < ApplicationRecord
     event.end_date.present? && event.end_date.past? && attended? && scholarship_tasks_met?
   end
 
+  # An invoice (and receipt) only make sense for a paid event — free events have
+  # nothing to bill or receipt.
+  def invoice_available?
+    event.cost_cents.to_i.positive?
+  end
+
+  # A paid-in-full receipt is available once a paid event carries no balance.
+  def receipt_available?
+    invoice_available? && remaining_cost.zero?
+  end
+
   # Cost source for the Registerable payment interface: the event's price.
   def cost_cents
     event.cost_cents
--- a/app/presenters/event_receipt.rb
+++ b/app/presenters/event_receipt.rb
@@ -1,0 +1,139 @@
+# Normalizes the data needed to render a paid-in-full receipt for a single
+# EventRegistration. Mirrors EventInvoice (same bill-to/attention/line-item
+# shape) but adds the allocation ledger that settled the balance and a summary
+# that always reconciles to $0 due — a receipt is only generated once the
+# registration is paid in full (see EventRegistration#receipt_available?).
+class EventReceipt
+  ISSUER_NAME = "A Window Between Worlds".freeze
+  ISSUER_ADDRESS_LINES = [ "1029 1/2 W 24th St", "Los Angeles, CA 90007" ].freeze
+  ISSUER_EMAIL = "info@awbw.org".freeze
+  THANK_YOU_NOTE = "Payment received in full — thank you. Please retain this receipt for your records.".freeze
+
+  # Friendly payment-method names for the ledger. Payment is an STI base whose
+  # polymorphic source_type is always "Payment"; the concrete subclass tells us
+  # how it was paid. Non-payment allocations (scholarships, discounts) fall back
+  # to their humanized type.
+  PAYMENT_METHOD_LABELS = {
+    "CashPayment" => "Cash",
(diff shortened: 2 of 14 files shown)


app/models/event_registration.rb
```diff
@@ -279,6 +279,13 @@ def paid_in_full?
     allocations_sum >= event.cost_cents.to_i
   end
 
+  # A paid-in-full receipt is available once a paid event carries no balance.
+  # Free events have nothing to receipt, so they're excluded even though they're
+  # nominally "paid in full".
+  def receipt_available?
```
maebeale, line 319: 🤖 _From Claude:_ Free events (`cost_cents <= 0`) are excluded even though `paid_in_full?` returns true for them — there is no payment to receipt, and the ledger would render empty.

app/presenters/event_receipt.rb
```diff
@@ -0,0 +1,139 @@
...
+      method: method_label(allocation),
+      reference: reference_for(allocation),
+      amount_cents: allocation.amount
+    )
+  end
+  private_class_method :entry_for
+
+  def self.method_label(allocation)
+    if allocation.source_type == Payment.polymorphic_name
+      PAYMENT_METHOD_LABELS[allocation.source&.class&.name] || "Payment"
```
maebeale, line 81: 🤖 _From Claude:_ `source_type` is the STI base `"Payment"`, so the concrete subclass (`allocation.source.class.name`) is what tells us the method. Falls back to "Payment" for legacy/unknown types; non-payment allocations use their humanized type.

app/presenters/event_receipt.rb
```diff
@@ -0,0 +1,139 @@
...
+        LineItem.new(
+          date: registration.created_at.to_date,
+          description: event.title,
+          quantity: 1,
+          unit_price_cents: event.cost_cents.to_i
+        )
+      ],
+      entries: allocations.map { |allocation| entry_for(allocation) },
+      amount_paid_cents: amount_paid_cents,
+      balance_cents: [ event.cost_cents.to_i - amount_paid_cents, 0 ].max
```
maebeale, line 65: 🤖 _From Claude:_ Balance is floored at 0 so an overpaid/over-allocated registration still reads "$0 due" rather than a negative figure. Amount paid above shows the true allocation sum.

app/views/events/registrations/receipt.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for(:page_bg_class, "public") %>
```
maebeale, line 1: 🤖 _From Claude:_ Eyebrow adapts to origin: `return_to=payment` (set by the payment-page link) returns there; otherwise it defaults to the forms page it normally lives under. The invoice view mirrors this with a ticket default.

app/models/event_registration.rb
```diff
@@ -279,6 +279,17 @@ def paid_in_full?
     allocations_sum >= event.cost_cents.to_i
   end
 
+  # An invoice (and receipt) only make sense for a paid event — free events have
+  # nothing to bill or receipt.
+  def invoice_available?
+    event.cost_cents.to_i.positive?
```
maebeale, line 315: 🤖 _From Claude:_ `invoice_available?` (event has a cost) gates the invoice; `receipt_available?` adds the zero-balance condition. Free events return false for both, so neither document is offered or reachable.

PR #1832, opened 2026-06-22
HOLD UNTIL REQUESTED: Add "Download all" button that zips a callout's attached documents

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: one new public endpoint + a shared-chrome button, no client JS

A single **Download all** button in the callout page header downloads every attached document on the page as one **zip**, instead of opening/downloading each one individually.

### Where it shows
- The shared `_callout_page` chrome renders the button whenever the controller sets `@download_all_url` — any callout page with **more than one** downloadable linked resource: **handouts** and any built-in callout (**scholarship, certificate, videoconference**) with attached files.
- **Payment is excluded** — it manages its own invoice/receipt/W-9 documents in a dedicated Documents section.

### How it works (no JavaScript)
- Button is a plain `link_to` a per-callout `…/documents.zip` endpoint (`CalloutsController#download_all`).
- The endpoint streams a server-side zip built with **rubyzip** (now declared explicitly in the Gemfile), entries named by resource title and de-duped. 404s for fewer than two files or the payment callout.

### Notes
- Rebased onto the callout redesign + subsequent main (#1997–#1999).
- Only shows past one document — a single file is grabbed from its own card.

### Tests
`spec/requests/events/callouts_spec.rb`: button visibility (2+/1/0/payment) and the zip endpoint (streams a named-entry zip, 404s under two files, 404s for payment).

### DIFF
--- a/app/controllers/events/callouts_controller.rb
+++ b/app/controllers/events/callouts_controller.rb
@@ -150,6 +150,7 @@ module Events
       return redirect_to registration_ticket_path(@event_registration.slug) unless builtin_published?("handouts")
       callout = @event.registration_ticket_callouts.find_by(builtin_key: "handouts")
       @handout_cards = resource_cards_for(callout, icon: "fa-solid fa-file-pdf", return_to: "handouts")
+      @download_all_url = download_all_url_for(callout)
     end
 
     # Registrant-facing page for a single Resource, shown in the shared callout
@@ -176,6 +177,17 @@ module Events
       return redirect_to(registration_ticket_path(@event_registration.slug)) unless builtin_published?("faq")
       callout = @event.registration_ticket_callouts.find_by(builtin_key: "faq")
       @faq_content = callout&.description.presence || BuiltinCallouts.faq_html
+    end
+
+    # Streams every attached document on a callout as a single zip (the "Download
+    # all" button). Payment is excluded — it manages its own documents — and a
+    # callout with fewer than two files 404s, since the button only shows past one.
+    def download_all
+      callout = @event.registration_ticket_callouts.find_by(builtin_key: params[:builtin_key])
+      resources = callout && callout.builtin_key != "payment" ? downloadable_resources(callout) : []
+      raise ActiveRecord::RecordNotFound if resources.size < 2
+      send_data build_documents_zip(resources),
+        filename: documents_zip_filename(callout), type: "application/zip"
     end
 
     private
@@ -228,6 +240,7 @@ module Events
       @builtin_intro = callout&.description.presence
       callout = nil if action_name == "payment"
       @builtin_resource_cards = resource_cards_for(callout, icon: "fa-solid fa-file-lines", return_to: action_name)
+      @download_all_url = download_all_url_for(callout)
     end
 
     # This registrant's cards for a callout's linked resources (nil callout → none).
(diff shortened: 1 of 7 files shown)


app/controllers/events/callouts_controller.rb
```diff
@@ -122,6 +126,16 @@ def build_form_cards
       cards
     end
```
maebeale, line 321: 🤖 _From Claude:_ The invoice has no stored PDF (it is rendered HTML printed via `window.print()`), so it can't be bundled as a file download — `newTab: true` opens its print-to-PDF page instead. Everything else is a real file.

app/frontend/javascript/controllers/download_all_controller.js
```diff
@@ -0,0 +1,34 @@
...
+//           data-download-all-items-value="[{&quot;href&quot;:&quot;/a.pdf&quot;}]"
+//           data-action="download-all#start">Download all</button>
+export default class extends Controller {
+  static values = { items: Array, delay: { type: Number, default: 400 } }
+
+  start() {
+    this.itemsValue.forEach((item, index) => {
+      setTimeout(() => this.trigger(item), index * this.delayValue)
+    })
+  }
```
maebeale: 🤖 _From Claude:_ Staggered by `delay` ms — browsers suppress multiple downloads fired synchronously in one tick, so the burst is spread out.

app/controllers/events/callouts_controller.rb
```diff
@@ -249,6 +251,22 @@ def resource_cards_for(callout, icon:, return_to:)
       callout.decorate.resource_cards(registrant_slug: @event_registration.slug, return_to:, icon:)
     end
 
+    # Direct-download URLs for a callout's linked resources that carry a
+    # downloadable file, in display order. Backs the "Download all" button the
+    # shared callout chrome shows whenever a page has downloadable resources
+    # (handouts and any built-in callout with attachments — Payment passes a nil
+    # callout here, so its own invoice/receipt/W-9 documents are excluded).
+    def download_all_urls_for(callout)
```
maebeale: 🤖 _From Claude:_ Keying the button off a nil callout is what excludes Payment for free — `set_builtin_content` nils the payment callout before calling this, so its own invoice/receipt/W-9 documents never get bundled here.

app/frontend/javascript/controllers/download_all_controller.js
```diff
@@ -0,0 +1,28 @@
...
+//
+//   <button data-controller="download-all"
+//           data-download-all-urls-value="[&quot;/resources/1/download&quot;]"
+//           data-action="download-all#start">Download all</button>
+export default class extends Controller {
+  static values = { urls: Array, delay: { type: Number, default: 400 } }
+
+  start() {
+    this.urlsValue.forEach((url, index) => {
+      setTimeout(() => this.download(url), index * this.delayValue)
```
maebeale: 🤖 _From Claude:_ Staggered by `delay` ms — browsers suppress multiple downloads fired synchronously in one tick, so the burst is spread out.

app/controllers/events/callouts_controller.rb
```diff
@@ -178,6 +179,17 @@ def faq
       @faq_content = callout&.description.presence || BuiltinCallouts.faq_html
     end
 
+    # Streams every attached document on a callout as a single zip (the "Download
+    # all" button). Payment is excluded — it manages its own documents — and a
+    # callout with fewer than two files 404s, since the button only shows past one.
+    def download_all
+      callout = @event.registration_ticket_callouts.find_by(builtin_key: params[:builtin_key])
+      resources = callout && callout.builtin_key != "payment" ? downloadable_resources(callout) : []
+      raise ActiveRecord::RecordNotFound if resources.size < 2
```
maebeale, line 188: 🤖 _From Claude:_ Guards the endpoint independently of the button: payment is excluded and anything with fewer than two files 404s, so a hand-crafted URL can't pull a zip the UI would never offer.

app/controllers/events/callouts_controller.rb
```diff
@@ -239,6 +252,57 @@ def resource_cards_for(callout, icon:, return_to:)
...
+
+    # Builds an in-memory zip of the resources' files, named by resource title and
+    # de-duped so two documents sharing a title don't collide inside the archive.
+    def build_documents_zip(resources)
+      used = Set.new
+      Zip::OutputStream.write_buffer do |zip|
+        resources.each do |resource|
+          file = resource.downloadable_asset.file
+          zip.put_next_entry(unique_entry_name(resource, file, used))
+          zip.write(file.download)
```
maebeale, line 283: 🤖 _From Claude:_ Builds the whole zip in memory (each `file.download` loads a blob) — fine for a handful of PDFs per callout; would want streaming if these grew large.

PR #1833, opened 2026-06-22
Add CE status column and filter to registrants index

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — substantive view/JS logic across the registrants roster plus a shared helper and the public-registration eyebrow

- Give admins an at-a-glance, triage-friendly registrants roster: see CE credit standing, keep the table dense and readable, and navigate without losing their place.
- CE only applies to events that grant CE credit, so CE surfaces stay hidden elsewhere.

- **CE status**: new `EventRegistration.ce_status` scope + shared `ce_status_label`; a "CE status" column (scholarship-style pill linking to the registration's CE section) and dropdown filter, plus onboarding columns, bulk-reminder filter, and both CSV exports — all gated on `Event#ce_eligible?`.
- **Column toggles**: generalized the `column-toggle` Stimulus controller to multiple named groups. CE on by default, Attendance off by default, User confirmation off — each with its own switch.
- **Layout/density**: CE before Scholarship; Date registered moved to the far right after Edit; filters wrapped in a card; compact registrant profile button (small avatar/type, uniform width); leading chip icons removed; column widths/truncation retuned.
- **Navigation**: viewing a submission from the roster stays in the same tab and its eyebrow returns to that registrant's row (anchored + highlighted), instead of the public ticket.

### UI Testing Checklist
- [ ] CE-eligible event: CE column shows by default, toggle hides it; Attendance hidden until toggled; CE filter + CSV column present.
- [ ] Non-CE event: no CE anywhere (roster, onboarding, bulk reminders, CSVs).
- [ ] View-submission from roster: same tab, "Back to registrants" returns to the highlighted row.

### Anything else to add?
- Rebased on main; reuses main's `Event#ce_eligible?` (CE foundation PR #1916). Covered by model + request specs (scope, filter, CSV, toggle defaults, eyebrow).

### DIFF
--- a/app/frontend/javascript/controllers/column_toggle_controller.js
+++ b/app/frontend/javascript/controllers/column_toggle_controller.js
@@ -1,23 +1,26 @@
 import { Controller } from "@hotwired/stimulus"
 
 // Connects to data-controller="column-toggle"
-// Toggles visibility of table columns marked with a data attribute.
+// Each instance is a single slide switch that shows/hides the table columns
+// whose data-column-toggle-col value matches this switch's `group` value, so a
+// table can host several independent column toggles (e.g. "User confirmation",
+// "CE status"). Columns live outside the switch's element, under a shared
+// [data-column-toggle-root] ancestor.
 
 export default class extends Controller {
   static targets = ["toggle", "track", "knob"]
+  static values = { group: String }
 
   toggle() {
     const checked = this.toggleTarget.checked
-    this.element.querySelectorAll("[data-column-toggle-col]").forEach((el) => {
+    const root = this.element.closest("[data-column-toggle-root]") || document
+
+    root.querySelectorAll(`[data-column-toggle-col="${this.groupValue}"]`).forEach((el) => {
       el.classList.toggle("hidden", !checked)
     })
 
-    if (this.hasTrackTarget) {
-      this.trackTarget.classList.toggle("bg-gray-300", !checked)
-      this.trackTarget.classList.toggle("bg-blue-600", checked)
-    }
-    if (this.hasKnobTarget) {
-      this.knobTarget.style.transform = checked ? "translateX(16px)" : ""
-    }
+    this.trackTarget.classList.toggle("bg-gray-300", !checked)
+    this.trackTarget.classList.toggle("bg-blue-600", checked)
+    this.knobTarget.style.transform = checked ? "translateX(16px)" : ""
   }
 }
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -107,6 +107,13 @@ class Event < ApplicationRecord
     forms.find_by(event_forms: { role: "registration" })
   end
 
+  # True when at least one registrant for this event requested CE credit. Drives
+  # the CE status column and filter on the registrants index — there's nothing
+  # to show or filter when no one has asked for CE.
+  def any_ce_credit_requests?
+    event_registrations.exists?(ce_credit_requested: true)
+  end
+
   # Whether a signed-in user should register in one click rather than being
   # routed to the registration form. True when no registration form is linked,
   # or when an admin has explicitly opted members out of the form. A linked form
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -107,11 +107,15 @@ class Event < ApplicationRecord
     forms.find_by(event_forms: { role: "registration" })
   end
 
-  # True when at least one registrant for this event requested CE credit. Drives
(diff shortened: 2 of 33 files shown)


app/models/event_registration.rb
```diff
@@ -146,6 +146,19 @@ class EventRegistration < ApplicationRecord
...
+  # Mirrors ReminderRecipientFilter#matches_ce_status?. The "license"/"hours"
+  # sub-statuses only make sense for someone who requested CE credit, so they're
+  # gated on it. "paid" has no CE-specific payment record yet, so it falls back
+  # to the registrant being paid in full.
+  scope :ce_status, ->(value) {
+    case value
+    when "requested" then where(ce_credit_requested: true)
+    when "license_not_provided" then where(ce_credit_requested: true).where(ce_license_number: [ nil, "" ])
+    when "hours_not_provided" then where(ce_credit_requested: true).where("COALESCE(ce_hours_requested, 0) <= 0")
+    when "paid" then where(ce_credit_requested: true).merge(paid_in_full)
```
maebeale, line 159: 🤖 _From Claude:_ "paid" intentionally has no CE-specific payment record yet, so it falls back to the registrant being paid in full — kept in sync with `ReminderRecipientFilter#matches_ce_status?`.

app/views/events/_registrants_results.html.erb
```diff
@@ -40,8 +40,11 @@
     <% if @event_registrations.any? %>
       <div class="overflow-x-auto">
         <% payment_col = @event.cost_cents.to_i > 0 %>
-        <% date_index = payment_col ? 5 : 4 %>
-        <% attendance_index = payment_col ? 6 : 5 %>
+        <% ce_col = @offers_ce %>
+        <% extra_cols = (payment_col ? 1 : 0) + (ce_col ? 1 : 0) %>
+        <% ce_index = payment_col ? 4 : 3 %>
+        <% date_index = 4 + extra_cols %>
+        <% attendance_index = 5 + extra_cols %>
```
maebeale: 🤖 _From Claude:_ Sort-column indices are recomputed from `extra_cols` so the Date/Attendance headers stay aligned whether or not the optional Payment and CE columns are present.

app/models/event.rb
```diff
@@ -101,6 +101,13 @@ def registration_form
     forms.find_by(event_forms: { role: "registration" })
   end
 
+  # True when at least one registrant for this event requested CE credit. Drives
+  # the CE status column and filter on the registrants index — there's nothing
+  # to show or filter when no one has asked for CE.
+  def any_ce_credit_requests?
```
maebeale: 🤖 _From Claude:_ Column/filter visibility is gated on any registrant having requested CE, so non-CE events stay uncluttered. Easy to switch to a registration-form-field check if we'd rather surface it before anyone answers.

app/models/event.rb
```diff
@@ -101,6 +101,17 @@ def registration_form
     forms.find_by(event_forms: { role: "registration" })
   end
 
+  # True when this event's registration form includes the seeded CE-interest
+  # question — i.e. the event actually offers continuing-education credit. Gates
+  # every CE surface (registrants index, onboarding matrix, bulk reminders) and
+  # their CSV exports, so events that don't offer CE never show empty CE
+  # columns/filters even if a stray registration carries CE data.
+  def offers_ce?
```
maebeale: 🤖 _From Claude:_ Single gate for every CE surface (index column/filter, onboarding matrix + CSV, bulk-reminder filter, registrants CSV): the event offers CE only when its registration form carries the seeded `ce_credit_interest` field — not when a stray registration happens to hold CE data.

app/frontend/javascript/controllers/column_toggle_controller.js
```diff
@@ -1,23 +1,26 @@
...
+// whose data-column-toggle-col value matches this switch's `group` value, so a
+// table can host several independent column toggles (e.g. "User confirmation",
+// "CE status"). Columns live outside the switch's element, under a shared
+// [data-column-toggle-root] ancestor.
 
 export default class extends Controller {
   static targets = ["toggle", "track", "knob"]
+  static values = { group: String }
 
   toggle() {
```
maebeale, line 14: 🤖 _From Claude:_ Generalized to multiple independent switches keyed by `group` value, so the same controller drives both "User confirmation" and "CE status". Columns live under a shared `[data-column-toggle-root]` ancestor and are matched by `data-column-toggle-col="<group>"`.

PR #1834, opened 2026-06-22
Disentangle sector field identifiers and tag primary vs additional distinctly

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — substantive logic: registration sector tagging changes behavior; new shared concern method; field-identifier rename with legacy compat

## Why
The sector fields were tangled: the multi-select **"Additional sectors"** field was built with the misleading `field_identifier: "primary_sector"`, and registration marked **every** submitted sector `is_primary: true` — lumping the primary dropdown and the additional checkboxes together. A profile crowned all of them, and a seed workaround had to demote the extras. Separately, "service area" naming lingered well beyond the legacy-acceptance constants.

## What
**Phase 1 — naming cleanup** (the only remaining "service area" is now the legacy constants + the specs that exercise them)
- `Person#other_service_area_responses` → `other_sector_responses` (+ callers / view locals)
- view/seed local vars, seed data keys, and comments renamed to `sector`

**Phase 2 — disentangle identifiers + fix tagging**
- New canonical `additional_sectors` identifier for the multi-select field; `primary_sector` / `primary_service_area` (and `primary_service_area_single`) stay in `FormField`'s constants as legacy so existing form data keeps resolving.
- `SectorsTaggable#tag_sectors(primary_ids:, additional_ids:)` — mirrors `AgeGroupTaggable#tag_age_groups` but upholds the single-primary rule (demotes any prior primary not re-selected).
- `PublicRegistration#assign_tags` now splits the dropdown answer (primary) from the checkbox answers (additional); `assign_primary_sectors` removed.
- Seeds write both fields and tag via `tag_sectors`, dropping the manual demote workaround.

## Verification
- Full affected spec suite + new `sectors_taggable_spec` green; rubocop clean.
- Ran `db:seed:dev`: 0 people with >1 primary sector; legacy `primary_sector` forms still resolve.

## Out of scope / notes
- `EventDashboard` recipient header still reads the *additional* sector identifiers (unchanged display behavior); only the field-identifier set was updated. Worth a follow-up if it should show the primary sector instead.

### DIFF
--- a/app/models/concerns/sectors_taggable.rb
+++ b/app/models/concerns/sectors_taggable.rb
@@ -17,7 +17,43 @@ module SectorsTaggable
     sectorable_items.sort_by { |item| [ item.is_primary? ? 0 : 1, item.sector&.name.to_s.downcase ] }
   end
 
+  # Additively tag sectors as primary/additional without disturbing other
+  # taggings — used by registration, where a respondent names a single primary
+  # sector (the dropdown) plus any number of additional sectors (the checkboxes).
+  # Mirrors AgeGroupTaggable#tag_age_groups, but upholds the single-primary rule:
+  # at most one id is promoted, and any prior primary the respondent didn't
+  # re-select is demoted first. A sector listed as both primary and additional is
+  # treated as primary.
+  def tag_sectors(primary_ids:, additional_ids:)
+    primary = sanitize_sector_ids(primary_ids).first(1)
+    additional = sanitize_sector_ids(additional_ids) - primary
+
+    demote_unselected_primary_sectors(primary) if primary.any?
+    upsert_sector_items(primary, is_primary: true)
+    upsert_sector_items(additional, is_primary: false)
+    sectorable_items.reset
+  end
+
   private
+
+  # Demote any currently-primary sector that isn't the newly selected primary, so
+  # promoting the new pick never leaves two primaries behind.
+  def demote_unselected_primary_sectors(primary_ids)
+    sectorable_items.where(is_primary: true).where.not(sector_id: primary_ids)
+      .update_all(is_primary: false)
+  end
+
+  def upsert_sector_items(sector_ids, is_primary:)
+    Sector.where(id: sector_ids).find_each do |sector|
+      item = sectorable_items.find_or_initialize_by(sector: sector)
+      item.is_primary = is_primary
+      item.save!
+    end
+  end
+
+  def sanitize_sector_ids(ids)
+    Array(ids).reject(&:blank?).map(&:to_i)
+  end
 
   def at_most_one_primary_sector
     # Count the in-memory set (not a DB query): with nested attributes both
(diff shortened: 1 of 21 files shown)


app/models/concerns/sectors_taggable.rb
```diff
@@ -17,8 +17,44 @@ def sectorable_items_primary_first
...
 
+  # Additively tag sectors as primary/additional without disturbing other
+  # taggings — used by registration, where a respondent names a single primary
+  # sector (the dropdown) plus any number of additional sectors (the checkboxes).
+  # Mirrors AgeGroupTaggable#tag_age_groups, but upholds the single-primary rule:
+  # at most one id is promoted, and any prior primary the respondent didn't
+  # re-select is demoted first. A sector listed as both primary and additional is
+  # treated as primary.
+  def tag_sectors(primary_ids:, additional_ids:)
+    primary = sanitize_sector_ids(primary_ids).first(1)
```
maebeale, line 28: 🤖 _From Claude:_ `.first(1)` is the belt-and-suspenders for the single-primary invariant — the primary always comes from a single-select dropdown, but this guarantees we never promote two even if a forged/legacy payload sends multiple. Combined with `demote_unselected_primary_sectors`, exactly one primary survives.

app/services/event_registration_services/public_registration.rb
```diff
@@ -273,14 +273,17 @@ def create_agency_address(organization)
...
-      if sector_ids.any?
-        sectors = Sector.where(id: sector_ids)
-        assign_primary_sectors(person, sectors)
-        organization.sectors = (organization.sectors + sectors).uniq if organization
+      if primary_sector_ids.any? || additional_sector_ids.any?
+        person.tag_sectors(primary_ids: primary_sector_ids, additional_ids: additional_sector_ids)
+        # Organizations aggregate sectors across many people and have no single
+        # "primary", so union everyone's selections in as additional tags rather
+        # than churning the org's primary on each registration.
+        organization&.tag_sectors(primary_ids: [], additional_ids: primary_sector_ids + additional_sector_ids)
```
maebeale, line 286: 🤖 _From Claude:_ Deliberate asymmetry vs the person: orgs aggregate sectors across many registrants and have no meaningful single primary (the org form never permits `is_primary`), so we union everyone in as additional rather than letting each registration churn the org primary. This preserves the prior union behavior while reusing the shared concern method.

app/models/form_field.rb
```diff
@@ -15,10 +15,12 @@ class FormField < ApplicationRecord
...
-  # Multi-select "additional sectors" field identifiers. The current "sector"
-  # name and the legacy "service area" name are both accepted so existing form
-  # data keeps resolving.
-  ADDITIONAL_SECTOR_FIELD_IDENTIFIERS = %w[primary_sector primary_service_area].freeze
+  # Multi-select "additional sectors" field identifiers. "additional_sectors" is
+  # the canonical name new forms are built with; the older "primary_sector" name
+  # (a misnomer — it was always the multi-select additional field) and the legacy
+  # "service area" name are both still accepted so existing form data keeps
+  # resolving.
+  ADDITIONAL_SECTOR_FIELD_IDENTIFIERS = %w[additional_sectors primary_sector primary_service_area].freeze
```
maebeale, line 23: 🤖 _From Claude:_ `additional_sectors` is new-canonical (first); `primary_sector` was always the multi-select additional field despite the name, kept here as legacy alongside `primary_service_area` so existing form answers keep resolving with no data migration.

PR #1837, opened 2026-06-22
WAIT: Show combined primary + additional sectors on the recipients header

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: a new dashboard read method + a view swap; no writes, no migration

## Why
The recipients "Serves" strip read only the *additional*-sector answer (under a single normalized `sector` key), so a registrant's primary sector and their additional sectors weren't shown together.

## What
- `EventDashboard#header_sectors_for(person)` — combines the single-select **primary** and multi-select **additional** sector answers for the event, uniqued, primary first. Read straight from the form answers (so it's independent of the registration-derived `is_primary` profile tags, which legacy data marks all-primary), falling back to the person's profile sector tags when no sector answers are on file.
- Recipients header renders them as chips with the **primary starred**, then the additional sectors alphabetically.
- Dropped the now-unused `HEADER_SECTOR_KEY` / `header_answer_key`; `HEADER_ANSWER_IDENTIFIERS` is now age-group-only.

## Verification
- `event_dashboard_spec` (new `#header_sectors_for` cases incl. legacy identifiers + profile fallback) and `events_spec` green; rubocop clean.
- Against seeded data the recipients header renders e.g. `★Disability Services, Fundraising/Donor Engagement` — primary starred, additional following, uniqued.

### DIFF
--- a/app/services/event_dashboard.rb
+++ b/app/services/event_dashboard.rb
@@ -41,15 +41,11 @@ class EventDashboard
   # standalone scholarship form, or both).
   SCHOLARSHIP_ANSWER_IDENTIFIERS = FormBuilderService::SECTION_FIELD_IDENTIFIERS[:scholarship].freeze
 
-  # Professional-info answers shown in each recipient's header. The view falls
-  # back to the person's profile (sectors / age-range tags) when these aren't on
-  # file as form answers. Sector answers are normalized under the "sector" key
-  # (see #header_answers_by_applicant) so the view reads one key regardless of
-  # which sector field identifier the answer was stored under.
-  HEADER_ANSWER_IDENTIFIERS = (FormField::ADDITIONAL_SECTOR_FIELD_IDENTIFIERS + %w[primary_age_group]).freeze
-
-  # The key sector answers are filed under in the per-applicant header hash.
-  HEADER_SECTOR_KEY = "sector".freeze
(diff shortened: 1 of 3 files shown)


app/services/event_dashboard.rb
```diff
@@ -75,20 +71,30 @@ def scholarship_answers_by_applicant
...
+  # A recipient's sectors for the "Serves" header, as [ Sector, is_primary ] pairs
+  # with their primary sector first (starred) and additional sectors following
+  # (alphabetical), all uniqued. Read straight from this event's form answers —
+  # combining the single-select primary and multi-select additional sector fields
+  # — so it reflects what the registrant chose, independent of the profile's
+  # is_primary tags. Falls back to the person's profile sector tags when they
+  # named no sectors on a form.
+  def header_sectors_for(person)
+    ids = header_sector_ids_by_applicant[person.id]
+    return profile_sector_pairs(person) unless ids && (ids[:primary].any? || ids[:additional].any?)
```
maebeale, line 91: 🤖 _From Claude:_ Form answers are the authority here, not the profile `is_primary` tags — legacy/prod registrations were tagged all-primary by the old `assign_tags`, so reading tags would star every sector. Profile tags are only the fallback when the registrant named no sectors on a form.

PR #1838, opened 2026-06-22
Sentence-case the age-group field labels in the registration form seed

maebeale:
🤖 suggested review level: 1 Skim 👀 seed-only copy change, no logic or data

Sentence-case the two age-group field labels in the registration form seed, so the professional section reads consistently ("Primary age group served" instead of "Primary Age Group(s) Served").

- The identifier-pluralization this PR originally carried already shipped on main via #2279 / #2419 — rebasing left only this label fix.
- Changes `SECTION_FIELD_NAMES` and the `add_field` calls together, so the builder-sync spec still holds.
- Seed-only: existing in-database form fields keep their stored labels.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -37,15 +37,20 @@ class FormField < ApplicationRecord
   # (as a string). Maps the field identifier to its backing CategoryType name.
   DYNAMIC_FIELD_CATEGORY_TYPES = {
     "primary_age_group" => "AgeRange",
+    "additional_age_groups" => "AgeRange",
     "additional_age_group" => "AgeRange"
   }.freeze
 
   # The "primary" and "additional" age group fields. Both are backed by AgeRange
   # categories but omit the catch-all "Mixed-age groups" category — a respondent
   # names the concrete age groups they serve, not the mixed bucket.
+  #
+  # The "additional" field is multi-select, so its current identifier is plural;
+  # the legacy singular name is kept so existing form data keeps resolving
+  # (current name first, legacy last — matching the sector identifiers).
   PRIMARY_AGE_GROUP_FIELD_IDENTIFIER = "primary_age_group"
-  ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER = "additional_age_group"
-  AGE_GROUP_FIELD_IDENTIFIERS = [ PRIMARY_AGE_GROUP_FIELD_IDENTIFIER, ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER ].freeze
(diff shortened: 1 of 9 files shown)


app/models/form_field.rb
```diff
@@ -35,15 +35,20 @@ class FormField < ApplicationRecord
...
+  # The "additional" field is multi-select, so its current identifier is plural;
+  # the legacy singular name is kept so existing form data keeps resolving
+  # (current name first, legacy last — matching the sector identifiers).
   PRIMARY_AGE_GROUP_FIELD_IDENTIFIER = "primary_age_group"
-  ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER = "additional_age_group"
-  AGE_GROUP_FIELD_IDENTIFIERS = [ PRIMARY_AGE_GROUP_FIELD_IDENTIFIER, ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIER ].freeze
+  ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIERS = %w[additional_age_groups additional_age_group].freeze
+  AGE_GROUP_FIELD_IDENTIFIERS = [ PRIMARY_AGE_GROUP_FIELD_IDENTIFIER, *ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIERS ].freeze
 
   # The payment-method field. Its answer options ("Credit card (now)", etc.) are
```
maebeale, line 55: 🤖 _From Claude:_ Current name first, legacy singular last — same ordering convention as `ADDITIONAL_SECTOR_FIELD_IDENTIFIERS`. Both keys are also added to `DYNAMIC_FIELD_CATEGORY_TYPES` so every `.key?`/`[id]` lookup resolves either name.

app/services/event_registration_services/public_registration.rb
```diff
@@ -275,7 +275,7 @@ def create_agency_address(organization)
     def assign_tags(person, organization)
       sector_ids = FormField::SECTOR_FIELD_IDENTIFIERS.flat_map { |id| collect_ids_from_checkboxes(id) }
       primary_age_ids = collect_ids_from_checkboxes("primary_age_group")
-      additional_age_ids = collect_ids_from_checkboxes("additional_age_group")
+      additional_age_ids = FormField::ADDITIONAL_AGE_GROUP_FIELD_IDENTIFIERS.flat_map { |id| collect_ids_from_checkboxes(id) }
```
maebeale, line 279: 🤖 _From Claude:_ Collects across both identifiers (`flat_map`) exactly like the sector line above, so registrations submitted against legacy singular fields still tag correctly.

PR #1839, opened 2026-06-22
WAIT: Add organization status and location filters to registrants list

maebeale:
The event registrants filter bar had grown unwieldy and was missing controls admins need to triage registrants by how their organization is resolved and where they live.

Grouped the filters into a card with two rows (keyword · organization status · scholarship · payment · attendance, then state · locality · county · country), added an Organization status filter (Linked/Pending/None) backed by the new `EventOrganizationLinkStatus` PORO that mirrors the per-row badge logic, and added Locality and Country location filters via new `EventRegistration` scopes and an `EventDashboard#localities` option source.

### UI Testing Checklist
- [ ] Filters render as a card with two rows in the expected order
- [ ] Organization status filters to Linked, Pending, and None correctly (overlap allowed)
- [ ] Locality and Country dropdowns appear when registrants have that data and filter correctly
- [ ] "Clear filters" resets every control

### Anything else to add?
The Linked/Pending/None states intentionally overlap (a registration can be both linked and pending), so each option returns everyone matching that state. Covered by model, service, dashboard, and request specs.

### DIFF
--- a/app/services/event_organization_link_status.rb
+++ b/app/services/event_organization_link_status.rb
@@ -1,0 +1,93 @@
+# Classifies an event's registrations by how their organization is resolved,
+# mirroring the badge logic in `events/_registrants_results` so the registrants
+# filter and the displayed pills agree:
+#
+# - linked:  the registration has at least one linked organization
+# - pending: the registrant submitted an agency name that doesn't match any
+#            linked organization (needs admin attention)
+# - none:    no linked organization and no submitted agency name
+#
+# The states are not mutually exclusive — a registration can be both linked and
+# pending — so each option returns everyone matching that predicate.
+class EventOrganizationLinkStatus
+  STATUSES = %w[linked pending none].freeze
+
+  def initialize(event)
+    @event = event
+  end
+
+  # Registration ids matching the given status ("linked"/"pending"/"none").
+  # Returns [] for an unrecognized status so callers can scope to nothing.
+  def registration_ids_for(status)
+    case status.to_s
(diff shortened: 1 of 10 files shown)


app/services/event_organization_link_status.rb
```diff
@@ -0,0 +1,93 @@
...
+    registrations.select { |registration| needs_linking?(registration) }.map(&:id)
+  end
+
+  def none_registration_ids
+    registrations.select do |registration|
+      linked_names(registration).blank? && submitted_name(registration).blank?
+    end.map(&:id)
+  end
+
+  def needs_linking?(registration)
```
maebeale, line 48: 🤖 _From Claude:_ Overlap is intentional — a registration with a linked org **and** a non-matching submitted agency name counts as both Linked and Pending, matching the per-row badges. Each filter option returns everyone in that state rather than a single exclusive bucket.

spec/requests/events_spec.rb
```diff
@@ -581,15 +625,15 @@ def submit_agency_name(name)
 
         get registrants_event_path(event)
 
-        expect(response.body).to include(">Pending<")
-        expect(response.body).not_to include(">None<")
+        expect(response.body).to include("<span>Pending</span>")
+        expect(response.body).not_to include("<span>None</span>")
```
maebeale, line 629: 🤖 _From Claude:_ Tightened from `>None<` to the exact chip markup because the new Organization status dropdown legitimately renders a "None"/"Pending" `<option>`, which would otherwise match this assertion.

PR #1840, opened 2026-06-22
WAIT: Save logged_out_only answers for signed-in registrants

maebeale:
- When a signed-in registrant has identity/contact data on file, those `logged_out_only` fields are hidden from the registration form, so their answers were never recorded on the submission — leaving an incomplete snapshot that looked like the registrant skipped questions.
- This records that on-file data as form answers so each registration is a complete snapshot, and makes the field-hiding rule consistent with the bulk-payment form.

- Added a `PersonKnownFields` PORO that maps a person's on-file identity/contact data to the registration form's `logged_out_only` field identifiers, and `PublicRegistration` now backfills those values onto the submission for signed-in registrants (skipping `confirm_email`, blanks, and anything the registrant actually submitted).
- Changed `PublicRegistrationsController` to hide **all** `logged_out_only` fields for signed-in users (matching the bulk-payment rule) instead of only the ones we already had answers for, with a temporary exception keeping the Organization Information (`agency_*`) fields visible until they can be backfilled from affiliations.

### UI Testing Checklist
- [ ] As a signed-in registrant, the identity/contact fields you already have on file are hidden, but the Organization Information section still shows.
- [ ] After submitting, the saved registration shows answers for those hidden fields populated from your profile data.
- [ ] As a logged-out registrant, every field still renders as before.

### Anything else to add?
- The Organization carve-out is intentionally temporary; once org data is backfilled from `person.affiliations.active`, the `agency_*` exception in `hide_logged_out_only_fields` comes out.
- The scholarship-form rendering still applies no `logged_out_only` filter — left untouched in this PR.

### DIFF
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -210,29 +210,7 @@ module Events
 
       person = current_user&.person
       if person
-        # Always hide logged_out_only fields for logged-in users with known data
-        known_identifiers = person_known_identifiers(person)
-        if known_identifiers.any?
-          known_ids = @form.form_fields
-                           .where(visibility: :logged_out_only, field_identifier: known_identifiers)
-                           .ids
-          scope = scope.where.not(id: known_ids) if known_ids.any?
-        end
-
-        # Hide logged_out_only headers when all their non-header fields are hidden
-        logged_out_sections = @form.form_fields.where(visibility: :logged_out_only)
-                                  .where.not(answer_type: :group_header)
-                                  .pluck(:section).uniq.compact
-        logged_out_sections.each do |sect|
-          section_field_ids = @form.form_fields.where(section: sect, visibility: :logged_out_only)
-                                  .where.not(answer_type: :group_header).ids
-          if section_field_ids.any? && known_identifiers.any? && (section_field_ids - scope.where(id: section_field_ids).ids).any?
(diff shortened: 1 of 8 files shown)


app/controllers/events/public_registrations_controller.rb
```diff
@@ -294,31 +272,43 @@ def visible_form_fields
...
+    # signed-in registrant manages this data on their profile, and whatever we
+    # already have is backfilled onto the submission by PersonKnownFields. So hide
+    # the whole set for signed-in users regardless of what's on file, matching the
+    # bulk-payment form's rule.
+    #
+    # Temporary exception: the Organization Information fields (`agency_*`) stay
+    # visible until we backfill them from the person's affiliations. We keep their
+    # section header too, and drop every other logged_out_only header once nothing
+    # under it survives (walking in position order so headers that share a DB
+    # `section` with the org fields are handled individually).
```
jmilljr24, line 285: I might not understand this. We have org questions set in the form as Always Ask. Why do we need to hard code this check as well?

jmilljr24: I guess this is the part I'm not sure I understand.

> The Organization carve-out is intentionally temporary; once org data is backfilled from person.affiliations.active, the agency_* exception in hide_logged_out_only_fields comes out.

PR #1841, opened 2026-06-22
Gate field-identifier editing behind admin, enforce per-form uniqueness

maebeale:
- Make a field's `field_identifier` editable from the form builder, but only under the existing `?admin=true` URL override — identifiers wire a field to backend logic (Stripe, service areas, email validation), so they shouldn't be edited casually.
- Guarantee a `field_identifier` can't appear twice on the same form. A duplicate would make that backend logic target an ambiguous field (e.g. two `payment_method` fields).

- **View** (`_form_field_fields.html.erb`): added an admin-gated `field_identifier` text input to the field's "More options" panel, reusing the `params[:admin] == "true"` convention already used one block below for the payment-method options lock. `field_identifier` was already permitted in `forms_controller` strong params, so no controller change was needed.
- **Model** (`form_field.rb`): added `validates :field_identifier, uniqueness: { scope: :form_id }, allow_blank: true`. Blanks are exempt because ordinary fields (and group headers) carry no identifier, so any number of them may coexist.
- **Tests** (`form_field_spec.rb`): TDD — added the duplicate-rejection test first, confirmed it failed (red), then added the validation (green). Specs cover: rejecting a duplicate on the same form, allowing the same identifier on a different form, and allowing multiple blank identifiers on one form.

### Anything else to add?
- Scoped uniqueness to `form_id` rather than global, since the same identifier (e.g. `primary_email`) legitimately recurs across different forms.
- No DB-level unique index added — the validation lives at the app layer where the blank-exemption and form scoping are expressed cleanly. Can add a partial index if we want a hard guarantee.

### DIFF
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -100,6 +100,11 @@ class FormField < ApplicationRecord
   # Keeps an over-long name as a friendly validation error instead of a
   # database ValueTooLong exception (the column is text, this is the UX cap).
   validates :name, length: { maximum: 1000 }
+  # A field_identifier wires a field to backend logic (Stripe, service areas,
+  # email checks, etc.), so the same identifier must not appear twice on one form
+  # or that logic would target an ambiguous field. Ordinary fields carry no
+  # identifier, so blanks are exempt — any number of them may coexist.
+  validates :field_identifier, uniqueness: { scope: :form_id }, allow_blank: true
   validates :min_words, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, allow_nil: true
   validates :max_characters, numericality: { only_integer: true, greater_than: 0 }, allow_nil: true
   validate :max_characters_allows_min_words
--- a/app/views/forms/_form_field_fields.html.erb
+++ b/app/views/forms/_form_field_fields.html.erb
@@ -150,6 +150,17 @@
                 class: "text-sm text-gray-400 hover:text-red-600 underline shrink-0" %>
         </div>
 
+        <% if params[:admin] == "true" %>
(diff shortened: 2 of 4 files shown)


app/models/form_field.rb
```diff
@@ -100,6 +100,11 @@ class FormField < ApplicationRecord
   # Keeps an over-long name as a friendly validation error instead of a
   # database ValueTooLong exception (the column is text, this is the UX cap).
   validates :name, length: { maximum: 1000 }
+  # A field_identifier wires a field to backend logic (Stripe, service areas,
+  # email checks, etc.), so the same identifier must not appear twice on one form
+  # or that logic would target an ambiguous field. Ordinary fields carry no
+  # identifier, so blanks are exempt — any number of them may coexist.
+  validates :field_identifier, uniqueness: { scope: :form_id }, allow_blank: true
```
maebeale, line 107: 🤖 _From Claude:_ Scoped to `form_id`, not global — identifiers like `primary_email` legitimately recur across different forms. `allow_blank` exempts ordinary fields (and headers), which carry no identifier. App-layer only; no DB partial index, so concurrent saves could in theory race a duplicate through.

app/views/forms/_form_field_fields.html.erb
```diff
@@ -150,6 +150,17 @@
                 class: "text-sm text-gray-400 hover:text-red-600 underline shrink-0" %>
         </div>
 
+        <% if params[:admin] == "true" %>
```
maebeale, line 153: 🤖 _From Claude:_ Reuses the existing `params[:admin] == "true"` convention (same as the payment-method options lock below). `field_identifier` was already in `forms_controller` strong params, so no controller change was needed to persist it.

PR #1850, opened 2026-06-22
Capture submitted amount on form submission

maebeale:

- Form submissions for paid events depend on Stripe round-tripping the actual `Payment` back (via checkout metadata) before we know the amount. Until that webhook lands, tickets fall back to **recomputing** the expected total from `event.cost_cents` — which silently shifts if an admin edits the event cost after someone registers.
- This captures the amount **at submission time** into a new `form_submissions.submitted_amount_cents` column, so:
  - The ticket can show the correct figure immediately, with **no Stripe round-trip needed** (pay-later, check, or pre-webhook pay-now).
  - The figure is a **snapshot** — it stays fixed to what the registrant actually committed to, even if the event price later changes.
  - It opens the door to reconciliation (intended vs. actual paid) and expected-revenue reporting on still-pending submissions.

- **Migration** adds nullable `submitted_amount_cents` to `form_submissions` (reversible `up`/`down`).
- **Populate on save:**
  - `PublicRegistration` (registration) stores `event.cost_cents` on create and refreshes it on re-registration.
  - `BulkPayment` stores `event.cost_cents × attendee_count` (reusing `bulk_payment_amount_cents`) after the answers it derives from are saved.
- **Display precedence** on the bulk-payment ticket: actual `Payment` → captured `submitted_amount_cents` → derived total (the last only for legacy rows saved before this column existed). This matches the requested behavior of *referencing the captured value only until the Stripe payment comes in*.

### Anything else to add?

- Draft — capturing the value now; only the bulk-payment ticket consumes it so far. The registration ticket still drives its pending figure from `EventRegistration` allocations/`remaining_cost`; it can adopt the snapshot in a follow-up if we want the same decoupling there.
- `submitted_amount_cents` is the **gross** event price snapshot (before scholarships/discounts/allocations), kept consistent across registration and bulk paths.

### DIFF
--- a/app/services/event_registration_services/bulk_payment.rb
+++ b/app/services/event_registration_services/bulk_payment.rb
@@ -101,6 +101,10 @@ module EventRegistrationServices
     def create_form_submission(person)
       submission = FormSubmission.create!(person: person, form: @form, event: @event, role: "bulk_payment")
       save_form_answers(submission)
+      # Snapshot the expected total (event cost times attendee count) now that the
+      # answers it derives from are saved, so the ticket can show the right figure
+      # before Stripe reports the actual payment back.
+      submission.update!(submitted_amount_cents: submission.bulk_payment_amount_cents(@event))
       submission
     end
 
--- a/app/views/events/bulk_payments/ticket.html.erb
+++ b/app/views/events/bulk_payments/ticket.html.erb
@@ -1,7 +1,7 @@
 <% content_for(:page_bg_class, "public") %>
 <%
   attendees = @submission.bulk_payment_attendees
-  total_cents = @payment&.amount_cents || @submission.bulk_payment_amount_cents(@event)
+  total_cents = @payment&.amount_cents || @submission.submitted_amount_cents || @submission.bulk_payment_amount_cents(@event)
 %>
 
 <div class="max-w-2xl mx-auto px-4 pt-4">
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -351,7 +351,10 @@ module EventRegistrationServices
     end
 
     def create_form_submission(person)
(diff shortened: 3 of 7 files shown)


app/services/event_registration_services/bulk_payment.rb
```diff
@@ -101,6 +101,10 @@ def create_phone_contact(person)
     def create_form_submission(person)
       submission = FormSubmission.create!(person: person, form: @form, event: @event, role: "bulk_payment")
       save_form_answers(submission)
+      # Snapshot the expected total (event cost times attendee count) now that the
+      # answers it derives from are saved, so the ticket can show the right figure
+      # before Stripe reports the actual payment back.
+      submission.update!(submitted_amount_cents: submission.bulk_payment_amount_cents(@event))
       submission
```
maebeale, line 108: 🤖 _From Claude:_ Snapshot is set *after* `save_form_answers` on purpose — `bulk_payment_amount_cents` derives the count from the saved answers, so it'd read 0 if computed before.

app/views/events/bulk_payments/ticket.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for(:page_bg_class, "public") %>
 <%
   attendees = @submission.bulk_payment_attendees
-  total_cents = @payment&.amount_cents || @submission.bulk_payment_amount_cents(@event)
+  total_cents = @payment&.amount_cents || @submission.submitted_amount_cents || @submission.bulk_payment_amount_cents(@event)
```
maebeale, line 4: 🤖 _From Claude:_ Precedence: actual Stripe `Payment` wins; `submitted_amount_cents` covers the pre-webhook window; the derived total is only a fallback for legacy rows with a nil snapshot.

maebeale:
don't need this since we store temp total to show for ticket

PR #1852, opened 2026-06-22
Surface pending email-change on person form and profile

maebeale:

Neither admins nor the person themselves could tell that a primary email change was initiated but not yet confirmed, so the displayed address looked authoritative even though it may still revert. A pending change is private account state: it must be visible to the person and admins only, clearly marked as private, and never leaked to a public profile visitor.

Surfaced `User#unconfirmed_email` as an "awaiting confirmation" warning in three places — the person edit form, the profile, and the admin `person_profile_button` (a hover-titled triangle in member/search lists). Every surface is gated behind a new `PersonPolicy#show_email_change?` (admin or owner), the profile chip carries a visible eye-slash "Private" marker, and all three hovers state that only the person and admins can see it and that it is not shown on the public profile. Seeded base user **Aisha** with an in-progress reconfirmation so all states are visible via `rake db:seed:dev`.

### UI Testing Checklist
- [ ] As an admin, Aisha's edit form, profile chip, and profile-button all show the warning with `aisha.new@example.com`
- [ ] As Aisha herself, her profile shows the "Email change pending / Private" chip
- [ ] As a non-admin, non-owner, no pending-change notice appears anywhere
- [ ] Hover text on every surface explains it is private and not shown publicly
- [ ] No notice appears for a person with no pending email change

### Anything else to add?
Dev-only seed sets the reconfirmation columns directly so no confirmation email is sent on reseed. Adds specs for the policy predicate and the profile-button warning.

### DIFF
--- a/app/helpers/person_helper.rb
+++ b/app/helpers/person_helper.rb
@@ -75,7 +75,7 @@ module PersonHelper
                     tag.i(class: "fa-solid fa-triangle-exclamation"),
                     class: "flex-shrink-0 ml-auto text-yellow-500",
                     title: "Email change to #{unconfirmed_email} awaiting confirmation. " \
-                           "Only the person and admins can see this; it is not shown on the public profile.")
+                           "Only the person and admins can see this; it is hidden from other profile visitors.")
       else
         "".html_safe
       end
--- a/app/helpers/person_helper.rb
+++ b/app/helpers/person_helper.rb
@@ -22,8 +22,11 @@ module PersonHelper
     full_name = display_name || person.try(:name) || person.to_s
     hover_title = [ full_name, subtitle ].compact_blank.join(" — ")
 
-    unconfirmed_email = person.user&.unconfirmed_email
-    show_email_change = unconfirmed_email.present? && allowed_to?(:show_email_change?, person)
+    # Only touch person.user when the viewer is actually allowed to see a pending
+    # email change — this keeps the button from firing an N+1 :user lookup per row
+    # for anonymous/public viewers on people-heavy pages (e.g. organizations#show).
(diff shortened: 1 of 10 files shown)


app/views/people/_form.html.erb
```diff
@@ -22,6 +22,12 @@
                 <div>
                   <label class="block text-md font-medium text-gray-700 mb-1">Primary email</label>
                   <span class="text-gray-900 font-medium"><%= f.object.user&.email %></span>
+                  <% if f.object.user.unconfirmed_email.present? %>
```
maebeale: 🤖 _From Claude:_ Intentional safe-nav asymmetry: here `f.object.user` is non-nil (guarded by the enclosing `<% if f.object.user %>`), so `.unconfirmed_email` is direct. The profile view uses `@person.user&.unconfirmed_email` since `user` can be absent there.

app/helpers/person_helper.rb
```diff
@@ -15,6 +15,9 @@ def person_profile_button(person, truncate_at: nil, subtitle: nil, display_name:
     full_name = display_name || person.try(:name) || person.to_s
     hover_title = [ full_name, subtitle ].compact_blank.join(" — ")
 
+    unconfirmed_email = person.user&.unconfirmed_email
+    show_email_change = unconfirmed_email.present? && allowed_to?(:show_email_change?, person)
```
maebeale: 🤖 _From Claude:_ Same `show_email_change?` predicate (admin-or-owner) gates the profile chip, the form chip, and this button warning, so a public list viewer never learns a person is mid-reconfirmation. The button also renders on some non-admin pages (org/story views), where this correctly hides it per-person.

PR #1854, opened 2026-06-22
Warn (not block) on approximate email dupes; widen duplicate matching

maebeale:
🤖 suggested review level: 3 Read 📖 contained duplicate-check change — widened matching logic + reverted blocking, covered by request specs

- Two people can legitimately share an email, so an exact name + **secondary/user** email match should **warn**, not block. Only an exact name + **primary** email match blocks creation.
- Duplicate detection should catch a match across the full set of identity fields the form captures, not just first name + primary email.

- `blocked:` is now `primary_email_match` only. Exact name + secondary/user email is an `"approximate"` match: orange warning badge + note, with "Create anyway" available.
- Widened `find_duplicate_people` to consider **legal first name** (entered and stored, plus nickname variants) and the entered **secondary email** — so a match is caught regardless of which name/email pair lines up. New `exact_first_name_match?` helper decides exact-vs-nickname across first + legal forms.
- Passed the new fields through both `create` and `check_duplicates`, and carried them on the standalone page's "View details"/shareable links.

### UI Testing Checklist
- [ ] Exact name + matching **secondary** or **user** email shows the orange "approximate match" badge and warning, and still allows "Create anyway".
- [ ] Exact name + matching **primary** email still shows the red "exact match" badge and blocks.
- [ ] Entering a name whose **legal name** matches an existing record's first name (or vice versa) surfaces that record.
- [ ] Entering a **secondary email** that matches an existing record surfaces it.

### Anything else to add?
- No schema or model changes — controller/view logic only.
- The block remains UI-soft (server still honors `skip_duplicate_check`); unchanged from before, and only the approximate case was relaxed.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -348,6 +348,8 @@ class PeopleController < ApplicationController
       person.email.present? && person.email.downcase == entered_email.downcase
     secondary_email_match = exact && any_email_match && !primary_email_match
 
+    # Exact name + primary email is an "exact" match; exact name + a secondary or
+    # user email is "approximate" — distinct display, but both block creation.
     match_type = if primary_email_match
       "exact"
     elsif secondary_email_match
@@ -366,7 +368,7 @@ class PeopleController < ApplicationController
       labeled_emails: labeled_emails,
       match_type: match_type,
       name_match: exact,
-      blocked: primary_email_match
+      blocked: primary_email_match || secondary_email_match
     }
   end
 
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -171,12 +171,16 @@ class PeopleController < ApplicationController
       @duplicates = find_duplicate_people(
         @person.first_name,
         @person.last_name,
-        @person.email
+        @person.email,
+        legal_first_name: @person.legal_first_name,
+        email_2: @person.email_2
       )
       if @duplicates.any?
         @first_name = @person.first_name
         @last_name = @person.last_name
         @email = @person.email
+        @legal_first_name = @person.legal_first_name
+        @email_2 = @person.email_2
         @blocked = @duplicates.any? { |d| d[:blocked] }
         set_form_variables
         respond_to do |format|
@@ -233,7 +237,12 @@ class PeopleController < ApplicationController
     @first_name = params[:first_name]
     @last_name = params[:last_name]
     @email = params[:email]
-    @duplicates = find_duplicate_people(@first_name, @last_name, @email)
+    @legal_first_name = params[:legal_first_name]
+    @email_2 = params[:email_2]
+    @duplicates = find_duplicate_people(
+      @first_name, @last_name, @email,
+      legal_first_name: @legal_first_name, email_2: @email_2
+    )
     @blocked = @duplicates.any? { |d| d[:blocked] }
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/people_controller.rb
```diff
@@ -366,7 +368,7 @@ def format_duplicate(person, exact: true, entered_email: nil)
       labeled_emails: labeled_emails,
       match_type: match_type,
       name_match: exact,
-      blocked: primary_email_match
+      blocked: primary_email_match || secondary_email_match
```
maebeale: 🤖 _From Claude:_ Both primary and secondary/user email matches now block, but they stay separate `match_type`s (`exact` vs `approximate`) purely so the views can show which email matched. `blocked` is the OR of the two.

app/controllers/people_controller.rb
```diff
@@ -282,51 +291,67 @@ def set_form_variables
...
-      normalized_first = NicknameMap.normalize(first_name)
+      first_variants += NicknameMap.variants_for(legal_first_name) if legal_first_name.present?
+      first_variants = first_variants.uniq
 
       name_matches = Person.includes(:user)
                            .where("REPLACE(REPLACE(LOWER(last_name), '.', ''), ' ', '') = ?", normalized_last)
-                           .where("REPLACE(REPLACE(LOWER(first_name), '.', ''), ' ', '') IN (?)", first_variants)
+                           .where(
+                             "REPLACE(REPLACE(LOWER(first_name), '.', ''), ' ', '') IN (:variants) OR " \
+                             "REPLACE(REPLACE(LOWER(legal_first_name), '.', ''), ' ', '') IN (:variants)",
```
maebeale, line 321: 🤖 _From Claude:_ Name matching now also checks the stored `legal_first_name` column, and `first_variants` above includes nickname variants of both the entered first name and legal name — so e.g. entering "Roberto" surfaces a record stored as first "Bob"/legal "Roberto".

app/controllers/people_controller.rb
```diff
@@ -282,51 +291,67 @@ def set_form_variables
...
-    if email.presence && !email.downcase.end_with?("@example.com") && duplicates.size < 10
-      email_lower = email.downcase
+    if query_emails.any? && duplicates.size < 10
       email_query = Person.includes(:user)
                           .left_joins(:user)
                           .where(
-                            "(users.id IS NULL AND LOWER(people.email) = :email) OR LOWER(people.email_2) = :email OR LOWER(users.email) = :email",
-                            email: email_lower
+                            "(users.id IS NULL AND LOWER(people.email) IN (:emails)) OR " \
+                            "LOWER(people.email_2) IN (:emails) OR LOWER(users.email) IN (:emails)",
```
maebeale, line 342: 🤖 _From Claude:_ The email lookup now searches both the entered primary and secondary email (`query_emails`) against each record's email/email_2/user email, so a collision on the new person's secondary email is caught too.

app/controllers/people_controller.rb
```diff
@@ -335,19 +360,31 @@ def find_duplicate_people(first_name, last_name, email)
...
-    any_email_match = entered_email.present? &&
-      emails.any? { |e| e.downcase == entered_email.downcase }
+    any_email_match = emails.any? { |e| entered_emails.include?(e.downcase) }
     primary_email_match = exact && entered_email.present? &&
       person.email.present? && person.email.downcase == entered_email.downcase
     secondary_email_match = exact && any_email_match && !primary_email_match
 
+    # Exact name + primary email is an "exact" match and blocks creation. Exact
+    # name plus a secondary or user email is "approximate" — the admin is warned
+    # but can still create the person (they may legitimately share an email).
```
maebeale, line 387: 🤖 _From Claude:_ Blocking now keys off `primary_email_match` only. Exact name + a secondary/user email is `"approximate"` — warned but still creatable, since two people can legitimately share an email.

PR #1856, opened 2026-06-22
HOLD: Import stories from a WordPress export CSV with an admin preview UI

maebeale:
- AWBW is migrating ~300 stories off the legacy WordPress site, and staff need a safe, repeatable way to bring them in without engineering involvement.
- Each CSV row becomes a `StoryIdea` (the canonical submission record); rows that were public on WordPress also get a connected, published `Story` — mirroring the in-app idea→story promotion flow.
- **Marked `HOLD:`** pending product decisions on the few unmappable fields (below) and a review of the import-user / visibility defaults before any real import is run.

- `StoryImporter` does the mapping with **no schema changes**, logging fields that have no home rather than dropping them silently, and supports a dry run.
- Two entry points: a `rake import:stories[path,user_email]` task and an **admin-only UI** from the Stories index — upload → dry-run **preview** of what will be created (ideas/stories/skipped/warnings) → confirm.
- The uploaded CSV is stashed as an ActiveStorage blob between preview and confirm (too large for the cookie session) and purged after import; created records are attributed to the importing admin.
- Covered by service specs and request specs; `AGENTS.md` updated for the new service, rake task, and controller.

### UI Testing Checklist
- [ ] As an admin, the Stories index shows an **Import** button next to **New Story**
- [ ] Uploading a WordPress export CSV shows a preview (counts + warnings) and saves nothing
- [ ] Confirming creates the story ideas/stories and redirects to Stories with a summary notice
- [ ] Non-admins and signed-out users cannot reach the import screens
- [ ] Re-uploading after the preview blob is gone shows a friendly "choose the file again" message

### Anything else to add?
- This works with zero schema changes, at the cost of dropping (but **logging**) four things with no current home: the original WordPress publish date, the US `state`, the free-text facilitator, and the source permalink / post id.
- Recommended follow-ups: add `original_published_at` and `legacy_wp_id` columns (correct chronological sort + idempotent re-import — note the permalink can't go in `website_url` because that triggers an external redirect on `Story#show`), and seed a `StoryCategory` taxonomy for the ~131 thematic WordPress categories that don't map onto an existing `Sector`.
- _Screenshots of the upload/preview screens to be added._

### DIFF
--- a/app/controllers/story_imports_controller.rb
+++ b/app/controllers/story_imports_controller.rb
@@ -1,0 +1,63 @@
+class StoryImportsController < ApplicationController
+  # Admin-only flow for importing stories from a WordPress Posts Export CSV.
+  #
+  #   new     → upload form
+  #   create  → dry-run preview of what would be created (nothing written)
+  #   confirm → run the real import against the stashed file
+  #
+  # The uploaded CSV is stashed as an ActiveStorage blob between the preview
+  # and confirm steps (it is too large for the cookie session). The blob is
+  # purged once the import runs.
+
+  def new
+    authorize! Story, to: :import?
+  end
+
+  def create
+    authorize! Story, to: :import?
+
+    file = params[:file]
+    return redirect_to(new_story_import_path, alert: "Choose a CSV file to import.") if file.blank?
+    return redirect_to(new_story_import_path, alert: "That file is not a CSV.") unless csv?(file)
+
+    @filename = file.original_filename
+    @blob = ActiveStorage::Blob.create_and_upload!(
+      io: file.open, filename: @filename, content_type: "text/csv"
+    )
+    @result = run_import(file.path, dry_run: true)
+    render :create
+  rescue CSV::MalformedCSVError, ArgumentError => e
+    @blob&.purge
+    redirect_to new_story_import_path, alert: "Could not read that CSV: #{e.message}"
+  end
+
+  def confirm
+    authorize! Story, to: :import?
+
+    blob = ActiveStorage::Blob.find_signed!(params[:signed_id])
+    result = blob.open { |tempfile| run_import(tempfile.path, dry_run: false) }
+    blob.purge
+
+    redirect_to stories_path, notice: import_notice(result)
+  rescue ActiveSupport::MessageVerifier::InvalidSignature, ActiveRecord::RecordNotFound
+    redirect_to new_story_import_path,
+                alert: "That upload is no longer available — please choose the file again."
+  end
+
+  private
+
+  def run_import(path, dry_run:)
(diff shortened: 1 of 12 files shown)


app/controllers/story_imports_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def create
+    authorize! Story, to: :import?
+
+    file = params[:file]
+    return redirect_to(new_story_import_path, alert: "Choose a CSV file to import.") if file.blank?
+    return redirect_to(new_story_import_path, alert: "That file is not a CSV.") unless csv?(file)
+
+    @filename = file.original_filename
+    @blob = ActiveStorage::Blob.create_and_upload!(
```
maebeale, line 24: 🤖 _From Claude:_ The CSV is ~1MB, so it is stashed as an ActiveStorage blob between preview and confirm rather than the cookie session. The blob is purged in `confirm` once the import runs; an abandoned preview leaves an orphan blob — open to a periodic sweep if that matters.

app/services/story_importer.rb
```diff
@@ -0,0 +1,291 @@
...
+      elsif audiences.intersect?(CHILD_AUDIENCE)
+        "Children"
+      else
+        DEFAULT_WINDOWS_TYPE
+      end
+    @windows_type_cache[short_name] ||= WindowsType.find_by!(short_name: short_name)
+  end
+
+  # Records that have no home in the current schema — logged, not dropped.
+  def note_unmapped_fields(row)
```
maebeale, line 221: 🤖 _From Claude:_ These fields (original publish date, US state, free-text facilitator) have no home in the current schema, so they are logged as warnings instead of dropped silently — that surfaces the data-model gap in the preview. The permalink is deliberately *not* mapped to `website_url`, since that triggers an external redirect on `Story#show`.

app/services/story_importer.rb
```diff
@@ -0,0 +1,291 @@
...
+
+  def duplicate_idea?(idea)
+    StoryIdea.where(organization_id: idea.organization_id)
+             .where("LOWER(title) = ?", idea.title.downcase)
+             .exists?
+  end
+
+  # Best-effort: tag whatever WordPress categories resolve to an existing
+  # Sector. Unmatched values are surfaced as warnings, never invented.
+  def tag_taxonomies(record, row)
```
maebeale, line 179: 🤖 _From Claude:_ Best-effort: a WordPress category is tagged only when it resolves to an existing `Sector` (directly or via `SECTOR_SYNONYMS`). The ~131 thematic categories with no Sector match are warned, never invented — they would need a seeded `StoryCategory` taxonomy.

PR #1857, opened 2026-06-22
HOLD: Clarify payment Revert as removing an allocation

maebeale:
- The payment "Revert" action only offsets an allocation and returns its amount to the payment's unallocated balance, but the bare "revert" label and a silently greyed-out row left admins guessing what it actually did.
- This reframes the action around what it does — removing an allocation — so the payments admin screens read clearly.

- Renamed the action to **Remove** (trash icon) with a consequence-stating confirm that names the dollar amount returned, plus an explainer line under the Allocations heading.
- Added explicit **Removed** / **Reversal** badges so reverted originals and their offsetting rows are self-describing instead of just dimmed.
- Polished the payments index and detail views: results in a card with row hover, amber highlight on unallocated balances, an icon empty state, and sentence-case labels.

### UI Testing Checklist
- [ ] Payments index: results render in a card, rows highlight on hover, unallocated amounts show amber, empty search shows the icon state.
- [ ] Payment detail: an active allocation shows a **Remove** button; confirming returns the amount to the unallocated balance.
- [ ] After removing, the original row shows a **Removed** badge and the offsetting row shows a **Reversal** badge with a negative amount.

### Anything else to add?
- **HOLD** — not ready to merge yet; opening as draft for early visibility.
- View-only change; no controller, route, or `revert` behavior was modified. Rebased onto latest main (adopts the `dollars_from_cents` helper); Rubocop and Brakeman pass.

### DIFF
--- a/app/views/payments/show.html.erb
+++ b/app/views/payments/show.html.erb
@@ -36,7 +36,7 @@
     </div>
     <% if @payment.check_number.present? %>
       <div>
-        <dt class="text-sm font-medium text-gray-500">Check Number</dt>
+        <dt class="text-sm font-medium text-gray-500">Check number</dt>
         <dd class="mt-1 text-sm text-gray-900"><%= @payment.check_number %></dd>
       </div>
     <% end %>
@@ -48,7 +48,7 @@
     <% end %>
     <% if @payment.pay_charge_id.present? %>
       <div>
-        <dt class="text-sm font-medium text-gray-500">Pay Charge ID</dt>
+        <dt class="text-sm font-medium text-gray-500">Pay charge ID</dt>
         <dd class="mt-1 text-sm text-gray-900"><%= @payment.pay_charge_id %></dd>
       </div>
     <% end %>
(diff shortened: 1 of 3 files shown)


app/views/payments/show.html.erb
```diff
@@ -92,43 +92,57 @@
...
               <th class="px-4 py-3 text-right text-xs font-medium text-gray-500 uppercase">Actions</th>
             </tr>
           </thead>
           <tbody class="divide-y divide-gray-200">
             <% @allocations.each do |allocation| %>
-              <% row_class = if allocation.amount.to_i < 0
-                  'bg-red-50'
-                elsif allocation.reverted_id.present?
-                  'bg-gray-100 border border-gray-300 opacity-60'
+              <% is_reversal = allocation.amount.to_i < 0 %>
```
maebeale, line 111: 🤖 _From Claude:_ The two badges map to the revert mechanism: `is_removed` is the original allocation (`reverted_id` set), `is_reversal` is the auto-created negative offset row. Both hide the Remove action so a removal can't be removed again.

PR #1861, opened 2026-06-22
Tint scholarship and CE credit cards to match their toggle state

maebeale:

- On the event-registration edit screen, the Scholarship and CE credits cards previously colored only the toggle track when "Requested" was on, leaving the rest of the card neutral white.
- A requested scholarship or CE credit was easy to miss at a glance; the request status should read from across the card, not just the small switch.

- Moved each card's Stimulus controller (`scholarship-requested`, `ce-credit-requested`) up onto the `<section>` so it can tint the whole box, not just the toggle.
- `refresh()` now mirrors the existing three-state logic onto the card itself: theme color when saved-on (fuchsia for scholarship, teal for CE), amber while the choice is pending (changed but unsaved), neutral white when off.
- Rendered the initial tint server-side so the correct color shows on load with no flash before Stimulus connects.
- An already-awarded scholarship has no toggle, so the controller guards on `hasCheckboxTarget` and leaves the static themed tint in place.

### Anything else to add?

- Purely presentational — no model, controller, or behavior changes.
- `border-teal-200` is the one new Tailwind utility; the rest were already in use. It's referenced from the ERB, so production/CI asset builds compile it.
- **On HOLD** pending the registration-edit redesign settling / design sign-off — opened as a draft so it's visible but not yet up for review.

### DIFF
--- a/app/frontend/javascript/controllers/scholarship_requested_controller.js
+++ b/app/frontend/javascript/controllers/scholarship_requested_controller.js
@@ -1,9 +1,11 @@
 import { Controller } from "@hotwired/stimulus"
 
-// Colors the "Requested" toggle to signal save state: amber while the choice is
-// pending (changed but not yet saved with the registration form), the
-// scholarship theme color once it matches the stored "on" value, and neutral
-// gray when stored as off.
+// Colors the "Requested" toggle and its whole card to signal save state: amber
+// while the choice is pending (changed but not yet saved with the registration
+// form), the scholarship theme color once it matches the stored "on" value, and
+// neutral gray when stored as off. Mounted on the <section> so this.element is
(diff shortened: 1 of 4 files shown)


app/frontend/javascript/controllers/scholarship_requested_controller.js
```diff
@@ -13,11 +15,20 @@ export default class extends Controller {
   }
```
maebeale, line 16: 🤖 _From Claude:_ Controller moved from the `<label>`/inner div up to the `<section>` so it can tint the whole card. This `hasCheckboxTarget` guard matters because an already-awarded scholarship renders no toggle — the controller still mounts, so without the guard `refresh()` would throw and clobber the server-rendered tint.

PR #1864, opened 2026-06-22
Add communications/notifications section to person edit

maebeale:

- Admins managing event registrations already get a communications log on the registration edit page; the same context is valuable when working a person record directly.
- This mirrors that section on the person edit page so admins can see and log communications without bouncing through a registration.

- Added a `notifications` polymorphic association (`as: :noticeable`) and nested attributes to `Person`, matching `EventRegistration`.
- Added `Person` to `Notification::NOTICEABLE_TYPES` so person-logged communications appear in the notifications index "Record" filter and links resolve.
- Permitted `notifications_attributes` in `PeopleController#person_params` and set `recipient_email` on new manual logs in `#update` (mirrors the registration controller).
- New `app/views/people/_notifications.html.erb` partial mirroring `event_registrations/_notifications` — read-only paginated history (matched by `preferred_email`, newest first) plus inline "Add notification" logging.
- Rendered the section in the person form above comments, gated by `NotificationPolicy#index?` (admin-only — owners editing their own profile don't see it).
- Generalized the shared `notifications/_notification_fields` partial's `event_registration:` local to a neutral `record:` local; updated both the registration and person callers.

### Anything else to add?

- **Marked HOLD** pending review — confirm the desired styling/placement and admin-only gating before merging.
- Rebased onto current `main` (resolved conflicts in `Notification::NOTICEABLE_TYPES` and `comments_attributes`).
- Tests: new `spec/requests/people_notifications_spec.rb` (creates a manual log from nested attributes, ignores blank entries, lists past communications on the edit page). Verified no regressions across `people_notifications`, `event_registrations`, and the `event_registration_edit` system spec (98 examples, 0 failures). Lint clean.
- No browser screenshot captured here; the shared-partial system spec passes, but a visual spot-check of the person edit page is worth doing before un-holding.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -209,6 +209,10 @@ class PeopleController < ApplicationController
     @person.comments.select(&:new_record?).each { |c| c.created_by = current_user; c.updated_by = current_user }
     @person.comments.select { |c| c.persisted? && c.body_changed? }.each { |c| c.updated_by = current_user }
 
+    # Inline-logged notifications are addressed to the person.
+    recipient_email = @person.preferred_email.presence || "n/a"
+    @person.notifications.select(&:new_record?).each { |n| n.recipient_email = recipient_email }
+
     if @person.save
       assign_associations(@person) if params.dig(:person, :category_ids)
       redirect_to person_update_return_path, notice: "Person was successfully updated."
@@ -488,6 +492,7 @@ class PeopleController < ApplicationController
         :_destroy
       ],
       comments_attributes: [ :id, :topic, :body, :flagged, :_destroy ],
+      notifications_attributes: [ :channel, :sender_id, :email_subject, :email_body_text, :noticeable_type, :noticeable_id ]
     )
   end
 end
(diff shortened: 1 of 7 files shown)


app/views/notifications/_notification_fields.html.erb
```diff
@@ -7,10 +7,10 @@
 <% sender_users |= [ current_user ] if current_user %>
 <% sender_options = sender_users.sort_by { |u| u.full_name.to_s.downcase }.map { |u| [ u.full_name, u.id ] } %>
 <% channel_default = Notification::MANUAL_CHANNELS.include?(f.object.channel) ? f.object.channel : "email" %>
-<% record = local_assigns[:event_registration] %>
+<% record = local_assigns[:record] %>
```
maebeale, line 10: 🤖 _From Claude:_ Renamed the `event_registration:` local to a neutral `record:` so this partial can be shared by both the registration and person notification sections. Both callers updated.

app/controllers/people_controller.rb
```diff
@@ -209,6 +209,10 @@ def update
     @person.comments.select(&:new_record?).each { |c| c.created_by = current_user; c.updated_by = current_user }
     @person.comments.select { |c| c.persisted? && c.body_changed? }.each { |c| c.updated_by = current_user }
 
+    # Inline-logged notifications are addressed to the person.
+    recipient_email = @person.preferred_email.presence || "n/a"
```
maebeale, line 213: 🤖 _From Claude:_ Falls back to "n/a" when the person has no email, mirroring the registration controller — `recipient_email` is `presence`-validated on Notification, so a manual log against an email-less person still saves.

PR #1865, opened 2026-06-22
HOLD: Validate scholarship amount against event registration due

maebeale:

- Awarding a scholarship larger than the event registration's remaining amount due is invalid, but the app handled it poorly:
  - On **update**, the amount flows through `Scholarship#sync_allocation_amount` → `allocation.update!`, where the `Allocation` over-cost validation raises `ActiveRecord::RecordInvalid` — an unhandled **500**.
  - There was no **flash** message and no **field-level** error on the amount input.
- Note: `main` already validates the **grant** budget (`within_grant_budget`); this PR adds the parallel guard for the **event registration** amount due, which was not covered.

- Added a `Scholarship#amount_within_event_registration_due` validation that runs **before save**, so create/update fail gracefully (`@scholarship.save` returns false) instead of reaching the raising `update!` in the `after_update` callback.
  - Computes the due as `event.cost_cents − other allocations`, **excluding the scholarship's own allocation** so edits don't double-count.
  - Skips grant-only/standalone scholarships (no event-registration allocatable) and non-positive amounts.
  - Formats the amount with the shared `MoneyFormatter` for consistency with the rest of the UI.
- Controller: set `flash.now[:alert]` from the model errors on create/update failure.
- Form: render an inline red field error directly under the "Scholarship amount" input (the top-of-form `shared/errors` list was already present on `main`).

### Anything else to add?

- **HOLD**: opening as a draft pending confirmation this is still desired given `main`'s recent scholarship/allocation rework. The `scholarship-preview` Stimulus controller (live "Still owed" preview) is purely cosmetic and never blocks submission, so this server-side validation remains the enforcement; its "other allocated" math matches this validation's "other allocations" computation.
- Tests: model spec covers below/at/over the due, multi-allocation accounting, and own-allocation exclusion; request spec covers create + update returning `422` with the field error (no more 500), plus a within-due happy path.

### DIFF
--- a/app/models/scholarship.rb
+++ b/app/models/scholarship.rb
@@ -11,6 +11,7 @@ class Scholarship < ApplicationRecord
   validates :amount_cents, numericality: { greater_than_or_equal_to: 0 }
   validate :recipient_must_match_allocation_registrant
   validate :within_grant_budget, if: :grant
+  validate :amount_within_event_registration_due
 
   after_update :sync_allocation_amount, if: -> { saved_change_to_amount_cents? }
   after_create_commit :flag_event_registration_scholarship_requested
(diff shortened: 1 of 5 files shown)


app/models/scholarship.rb
```diff
@@ -44,6 +45,18 @@ def recipient_must_match_allocation_registrant
...
+    registration = allocation&.allocatable
+    return unless registration.is_a?(EventRegistration)
+    return if amount_cents.to_i <= 0
+
+    other_allocations_sum = registration.allocations.where.not(id: allocation.id).sum(:amount)
+    due_cents = registration.event.cost_cents.to_i - other_allocations_sum
+    return if amount_cents <= due_cents
+
+    errors.add(:amount_dollars, "cannot exceed the event registration amount due (#{MoneyFormatter.dollars_from_cents(due_cents)})")
+  end
```
maebeale, line 58: 🤖 _From Claude:_ Runs **before save** so create/update fail gracefully here — without it, an over-due amount reaches `sync_allocation_amount`'s `allocation.update!` in `after_update` and raises `RecordInvalid` (500). The `where.not(id: allocation.id)` excludes this scholarship's own allocation so edits don't double-count.

PR #1866, opened 2026-06-22
HOLD: Add related-comments overview + global comments index

maebeale:

- Comments exist on five record types (people, organizations, users, workshops, event registrations) but were only ever viewable one record at a time, so relevant context on a neighboring record was easy to miss.
- Adds one **related-comments overview** reachable from every comment box that surfaces all related comments together, plus a **global index** for previewing every comment and the record it's about.
- `HOLD:` — opening for review/discussion of the "related" definitions per type before merge.

- **`RelatedComments` service** decides what "related" means per commentable type — the commentable type *is* the signal of which comment box the button was clicked from (each box links to its own polymorphic comments path), so one place owns the mapping:
  - Registration → registrant, their user account, their active-affiliation orgs
  - Person → their user account + active orgs
  - Organization → people with an active affiliation
  - User → linked person + that person's active orgs
  - Workshop → its creator (+ creator's person/orgs)
- **`comments#index`** now serves three modes: scoped related overview (full page, styled like the allocations index, with an "About" column labeling each comment's record), a global `/comments` preview of every comment, and the existing bare list for turbo-frame requests (embed/create flows untouched).
- **Shared `View all related comments` button** added to every comment box (people, orgs, workshops, users, registrations) via a single partial, opening the overview in a new tab.
- Active-affiliation filtering uses `Affiliation.active`; admin-only via the existing `CommentPolicy`.

### Anything else to add?

- Tests: `spec/services/related_comments_spec.rb` covers the per-type related set (incl. active-vs-inactive affiliation filtering); `spec/requests/comments_spec.rb` covers the scoped overview, the global index, admin-restriction, and the turbo-frame path.
- Open question for reviewers: are the per-type "related" definitions the right scope (e.g. should an Organization also pull in its people's user-account comments)?
- Global index is reachable at `/comments` (no nav entry yet — intentionally minimal).

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,12 +1,19 @@
 class CommentsController < ApplicationController
   before_action :set_commentable
+  before_action :require_commentable, only: :create
 
   def index
     authorize!
-    @comments = @commentable.comments.newest_first.paginate(page: params[:page], per_page: 10)
 
     respond_to do |format|
-      format.html { render partial: "comments/list", locals: { commentable: @commentable, comments: @comments } }
+      format.html do
+        if turbo_frame_request? && @commentable
+          @comments = @commentable.comments.newest_first.paginate(page: params[:page], per_page: 10)
+          render partial: "comments/list", locals: { commentable: @commentable, comments: @comments }
+        else
+          @comments = index_comments.paginate(page: params[:page], per_page: 10)
+        end
+      end
     end
   end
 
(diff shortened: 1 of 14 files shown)


app/services/related_comments.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  private
+
+  def commentable_groups
+    ([ @commentable ] + related_records).compact.uniq
+      .group_by { |record| record.class.base_class.name }
+      .transform_values { |records| records.map(&:id).uniq }
+  end
+
+  def related_records
```
maebeale, line 27: 🤖 _From Claude:_ The commentable's type is the only signal of which comment box the button was clicked from — each box links to its own polymorphic `*_comments_path` — so this case statement is the single place that defines "related" per type. Add a `when` to extend it.

app/controllers/comments_controller.rb
```diff
@@ -33,20 +40,36 @@ def create
 
   private
 
+  # Scoped to a commentable: every comment related to that record (see
+  # RelatedComments). Top-level /comments: a preview of every comment in the
+  # system alongside the record each is about.
+  def index_comments
```
maebeale, line 46: 🤖 _From Claude:_ Three modes share this action: scoped related overview (full page), global `/comments` preview (no commentable), and the existing bare turbo-frame list for embeds/create — kept intact so the inline comment widgets don't change.

PR #1869, opened 2026-06-22
Fix videoconference details to withhold until a week before, for paid registrants

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: contained visibility/copy changes, privacy-sensitive so worth a careful read

Videoconference connection details (Zoom join link, meeting ID, passcode) were shared the moment someone registered — embedded in every add-to-calendar entry, shown on the videoconference page with no timing gate, and included in both the registrant confirmation and admin FYI emails (potentially months early). We only want to reveal them close to the event and only to registrants who have paid or committed to paying.

Added a per-registration gate, `EventRegistration#videoconference_details_visible?` = within 7 days of the start (`Event#videoconference_details_visible?`) AND payment access (paid or intends to pay), and threaded it through the ticket, event show page, and videoconference callout (both calendar entries via `EventDecorator#calendar_links` and on-page display). Both confirmation emails (registrant + admin FYI) now show only the platform label (e.g. "Zoom"), mirroring the reminder email — the join link lives on the now-gated ticket pages. Note: the live "join now" link keeps its stricter 30-minute window.

### UI Testing Checklist
- [ ] More than a week out: ticket / event show / videoconference page show no Zoom link, ID, or passcode; calendar entries omit them and the join URL.
- [ ] Within a week + paid/intends-to-pay: all connection details appear again.
- [ ] Within a week but unpaid: videoconference page shows the "payment is on file" message instead of the link.
- [ ] Registrant confirmation and admin FYI emails show only the platform label, no link/ID/passcode.

### Anything else to add?
Free events pass the payment check automatically.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -62,12 +62,19 @@ class EventDecorator < ApplicationDecorator
     length ? description&.truncate(length) : description
   end
 
-  def calendar_links
+  # `show_videoconference_details` controls whether the join link/ID/passcode are
+  # carried into the calendar entry. Callers with a registration pass that
+  # registrant's gate (date + paid/intends); the default falls back to the
+  # event-level date gate for registration-less contexts.
+  def calendar_links(show_videoconference_details: object.videoconference_details_visible?)
     start_time   = object.start_date.utc.strftime("%Y%m%dT%H%M%SZ")
     end_time     = object.end_date.utc.strftime("%Y%m%dT%H%M%SZ")
     title_encoded = ERB::Util.url_encode(object.title)
 
-    has_url      = object.videoconference_url.present?
+    # The join URL doubles as the calendar location, so withhold it from there
+    # too until the details may be shared — a physical location (if any) takes
+    # its place, otherwise the entry is left without a location.
+    has_url      = object.videoconference_url.present? && show_videoconference_details
     has_location = object.location.present?
     location_name = has_location ? object.location.name : nil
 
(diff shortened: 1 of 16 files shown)


app/decorators/event_decorator.rb
```diff
@@ -62,12 +62,19 @@ def detail(length: nil)
...
+  # carried into the calendar entry. Callers with a registration pass that
+  # registrant's gate (date + paid/intends); the default falls back to the
+  # event-level date gate for registration-less contexts.
+  def calendar_links(show_videoconference_details: object.videoconference_details_visible?)
     start_time   = object.start_date.utc.strftime("%Y%m%dT%H%M%SZ")
     end_time     = object.end_date.utc.strftime("%Y%m%dT%H%M%SZ")
     title_encoded = ERB::Util.url_encode(object.title)
 
-    has_url      = object.videoconference_url.present?
+    # The join URL doubles as the calendar location, so withhold it from there
```
maebeale, line 74: 🤖 _From Claude:_ The join URL doubles as the calendar `LOCATION`, so the gate is folded into `has_url` here too — otherwise the URL would still leak via the location field even with the description stripped. A physical location takes its place, or none.

app/models/event_registration.rb
```diff
@@ -314,6 +314,16 @@ def joinable?
...
 
+  # Whether this registrant may see the videoconference connection details yet
+  # (join link, meeting ID, passcode) — on the ticket, the videoconference page,
+  # and in calendar entries. Gated on payment access (paid or intends to pay) AND
+  # the event being within a week of starting, so the link isn't shared early.
+  # Distinct from #joinable?, which is the tight 30-minute "click to join now"
+  # window for the live link.
+  def videoconference_details_visible?
+    payment_access_granted? && event.videoconference_details_visible?
+  end
```
maebeale, line 325: 🤖 _From Claude:_ Deliberately separate from `joinable?` — this is the 7-day "may we share the credentials" gate, while `joinable?` stays the tight 30-minute "click to join now" window for the live link.

PR #1872, opened 2026-06-22
Add a sectors-style age-range editor to the person form

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — new editor UI, a category-assignment scoping fix, and a shared cocoon mechanism

## Why

The person edit form never had an age-range editor — AgeRange isn't a `profile_specific` category type, so it was absent from the categories section. Age groups could only be set via registration. This adds an editor, and unifies it with the sector editor.

## What

- **Age ranges edit via cocoon**, the same mechanism as sectors: a scoped `Person#age_range_categorizable_items` association + `accepts_nested_attributes_for`, rendered by `shared/_age_range_item_fields` with `link_to_add_association` / `link_to_remove_association`. The `is_primary` join-row flag is the primary marker — no separate param.
- **One shared `primary_tag` controller** (renamed from `primary_sector`) drives the single-select primary star for both pickers; highlight colors come from the Stimulus Classes API. Lighting one star clears the others; **no reorder** — chips keep alphabetical (sectors) / position (age) order. The bespoke `age_range_picker` controller + chip partials are gone.
- **Single-primary validation**: `Person#at_most_one_primary_age_range` mirrors `SectorsTaggable` — posting two primaries is a form error. Person-only, because orgs aggregate several members' primary age groups.
- **Survives a failed save**: the form reads the same association the nested attributes build into, so a range picked before a validation error (its chip + primary star) is retained on re-render.
- **Displays keep primary-first** for both sectors and age ranges (profile, recipients, dashboard).
- **Layout**: equal-height Sectors/Age boxes splitting 50/50 (`flex-wrap` + `min-w-[20rem]`), stacking as the viewport narrows; on the profile they're two columns with age ranges thinner.
- **Scoping fix** (`TagAssignable`): saving used to `categories = submitted_ids`, wiping every type the form doesn't show. The form sends `managed_category_type_ids` (workshop settings only — age ranges go through nested attributes); only those are replaced, so other taggings survive. `category_ids` dropped from strong params. The form now emits an **always-present blank `managed_category_type_ids`** field (like `category_ids`): without it, a person with no profile-specific types renders zero managed fields, the browser drops the key, and saving falls back to the replace branch — wiping the person's age ranges and other-type taggings.

## Tests

- `spec/requests/people_age_ranges_spec.rb` — cocoon render, tag/primary via nested attributes, `_destroy`, position order, **single-primary validation** (request + model), **retention after a validation error**, non-AgeRange preservation, and the **always-present blank managed field** (guards the key-drop data loss above).
- `spec/models/concerns/sectors_taggable_spec.rb` — `sectorable_items_ordered` vs `sectorable_items_primary_first`.
- `spec/requests/events/professional_field_identifiers_spec.rb` — updated to assert the cocoon chips instead of the old age checkboxes.


app/views/people/_form.html.erb
```diff
@@ -116,52 +116,73 @@
...
-    </div>
-    <div class="rounded-lg border border-gray-200 <%= DomainTheme.bg_class_for(:sectors) %> p-4 mb-4 shadow-sm
-                flex flex-wrap items-center gap-2"
-         data-controller="primary-sector">
-      <% sectors_owner = f.object.respond_to?(:object) ? f.object.object : f.object %>
-      <%= f.simple_fields_for :sectorable_items, sectors_owner.sectorable_items_primary_first do |sfi| %>
-        <%= render "shared/sectorable_item_fields", f: sfi, show_admin_flags: true %>
-      <% end %>
+  <!-- Sectors + Age ranges (side by side; age ranges is the thinner column) -->
+  <% age_type, age_cats = (@person_categories_grouped || {}).find { |type, _| type.name == "AgeRange" } %>
```
maebeale: 🤖 _From Claude:_ AgeRange is pulled out of the shared categories loop here and rejected from the `other_category_types` block below (line 174), so the primary-age hidden field and "mark the primary ones" hint move into this column with it. Other profile-specific types (Workshop Settings) keep their original full-width rendering.

app/controllers/concerns/tag_assignable.rb
```diff
@@ -7,7 +7,19 @@ def assign_associations(record, param_key: nil)
     key = param_key || record.model_name.param_key
 
     selected_category_ids = Array(params[key][:category_ids]).reject(&:blank?).map(&:to_i)
-    record.categories = Category.where(id: selected_category_ids)
+    selected = Category.where(id: selected_category_ids).to_a
+
+    if params[key].key?(:managed_category_type_ids)
+      # The form only edits certain category types (e.g. age ranges + workshop
+      # settings). Preserve taggings of every other type the form never shows so
+      # saving can't silently drop them — and assign the union so the join rows
```
maebeale, line 15: 🤖 _From Claude:_ This is the core safety fix: the form sends `managed_category_type_ids` (age ranges + workshop settings), so we only replace those types and union-preserve every other type the form never shows. Because preserved categories stay in the assigned set, their join rows aren't destroyed/recreated — `is_primary` and `legacy_id` survive. Org keeps the old full-replace path since it doesn't send the key.

app/controllers/people_controller.rb
```diff
@@ -280,6 +280,18 @@ def set_form_variables
...
+    # Age ranges edit as their own sectors-style chip picker (AgeRange isn't a
+    # profile_specific type, so it never appears in @person_categories_grouped).
+    @age_range_type = CategoryType.find_by(name: AgeGroupTaggable::AGE_RANGE_CATEGORY_TYPE)
+    @age_ranges_collection = @age_range_type ? @age_range_type.categories.published.order(:position, :name) : Category.none
+
+    # The category types this form actually edits — age ranges plus every
+    # profile-specific type shown above. assign_associations preserves taggings
+    # of any other type, so saving the form can't drop a person's non-AgeRange
+    # category connections (e.g. art types tagged elsewhere).
+    @managed_category_type_ids = ([ @age_range_type&.id ] +
```
maebeale: 🤖 _From Claude:_ Note: `category_ids` was also dropped from `person_params`. Otherwise #update's `assign_attributes(person_params)` would full-replace `categories=` and save *before* assign_associations runs its scoped/preserving assignment, destroying the non-AgeRange join rows. assign_associations reads category_ids raw from params, so the contract is unchanged.

app/frontend/javascript/controllers/age_range_picker_controller.js
```diff
@@ -0,0 +1,69 @@
...
+    chip.querySelector("[data-age-range-picker-target='chipSelect']")?.focus()
+    this.refreshOptions()
+  }
+
+  // A new chip's range was chosen: key its primary star to that id (so starring
+  // submits the right category) and re-filter the other selects.
+  choose(event) {
+    const select = event.target
+    const chip = select.closest("[data-age-range-picker-target='chip']")
+    chip.dataset.categoryId = select.value
```
maebeale: 🤖 _From Claude:_ A new chip uses a select for its id, but the primary star submits `person[primary_age_category_ids][]` keyed by id — so on select we copy the chosen id onto the star's value. That keeps the multi-primary contract identical to persisted chips (which render the id directly) without changing the controller params.

app/frontend/javascript/controllers/age_range_picker_controller.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  remove(event) {
+    event.target.closest("[data-age-range-picker-target='chip']")?.remove()
+    this.refreshOptions()
+  }
+
+  // Single-select primary: lighting one star clears the others and floats that
+  // chip to the front, mirroring primary_sector#selectPrimary.
+  togglePrimary(event) {
```
maebeale: 🤖 _From Claude:_ Single-select to match sectors. The server (`apply_primary_age_groups!`) still accepts a set, but registration's primary age group is a single-select dropdown so the data never has >1 primary — the UI now enforces the same. Persisted chips' stars carry the `primaryToggle` target too, so clearing-others sees them.

app/models/concerns/sectors_taggable.rb
```diff
@@ -12,11 +12,20 @@ module SectorsTaggable
...
+  # views, where the primary should always lead.
   def sectorable_items_primary_first
     sectorable_items.sort_by { |item| [ item.is_primary? ? 0 : 1, item.sector&.name.to_s.downcase ] }
   end
 
+  # Sectorable items in stable order for the edit form: alphabetically by sector
+  # name (sectors have no position column, so name is the position-equivalent).
+  # Unlike sectorable_items_primary_first, the primary is NOT floated to the top,
+  # so starring a sector on the form doesn't reshuffle the chips.
+  def sectorable_items_ordered
```
maebeale, line 25: 🤖 _From Claude:_ Edit-form ordering is intentionally split from display ordering: the form uses this (alpha, primary not floated) so starring doesn't reshuffle chips, while profile/recipients/dashboard keep `sectorable_items_primary_first`. Sectors have no position column, so name is the position-equivalent — same idea as the age-range chips, which do have positions.

app/models/person.rb
```diff
@@ -65,6 +65,15 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :contact_methods, allow_destroy: true, reject_if: :all_blank
   accepts_nested_attributes_for :sectorable_items, allow_destroy: true,
                                 reject_if: proc { |attrs| attrs["sector_id"].blank? }
+  # Age ranges edit through cocoon nested fields like sectors. A scoped view of
+  # categorizable_items (AgeRange categories only) so the form's add/remove and
+  # primary toggle round-trip as nested attributes — the is_primary flag splits
+  # primary vs additional, no separate primary_age_category_ids param needed.
+  has_many :age_range_categorizable_items,
```
maebeale, line 79: 🤖 _From Claude:_ Scoped association so age ranges edit via cocoon nested attributes (is_primary on the join row is the primary flag — no separate param). Heads-up: AgeGroupTaggable had a private helper of the same name; I renamed it to `age_range_items_relation` to avoid the concern shadowing this reader (it was being included below the generated association methods, so the private one won).

app/models/person.rb
```diff
@@ -243,8 +257,25 @@ def other_workshop_setting_responses
     other_form_responses(OTHER_WORKSHOP_SETTING_IDENTIFIERS)
   end
 
+  # The age-range nested items in category position order for the cocoon chip
+  # editor. Reads the same association the form's nested attributes build into, so
+  # unsaved picks survive a failed save (and aren't primary-first — starring
+  # shouldn't reshuffle them). Display surfaces lead with the primary instead.
+  def age_range_items_ordered
```
maebeale, line 293: 🤖 _From Claude:_ The retention fix: read the same age_range_categorizable_items association the form's nested attributes build into. The old version read the general categorizable_items, whose in-memory target doesn't include the freshly-built (unsaved) items after a failed save — so a picked range vanished on re-render. Sectors never had this because sectorable_items is a single association.

app/models/person.rb
```diff
@@ -243,8 +257,25 @@ def other_workshop_setting_responses
...
+    age_range_categorizable_items.sort_by { |item| [ item.category&.position || 0, item.category&.name.to_s ] }
+  end
+
   private
 
+  # Count the in-memory set (not a DB query): nested attributes build the items in
+  # one transaction, so a row-level check would see none persisted yet.
+  def at_most_one_primary_age_range
```
maebeale, line 301: 🤖 _From Claude:_ Person-only on purpose (not in the shared concern like sectors): organizations aggregate several affiliated members' primary age groups via tag_age_groups, so an org legitimately has >1 primary. Sectors avoid this because registration tags orgs with primary_ids: [] (orgs never get a primary sector).

app/models/person.rb
```diff
@@ -243,8 +261,41 @@ def other_workshop_setting_responses
...
+    primary_count = age_range_categorizable_items.reject(&:marked_for_destruction?).count(&:is_primary?)
+    return if primary_count <= 1
+
+    errors.add(:base, "Only one age range can be marked as primary")
+  end
+
+  # Keep one tagging per age-range category. Prefer the persisted row, fold any
+  # duplicate's primary flag onto the keeper, and drop the extras (destroy if
+  # persisted, otherwise remove from the unsaved set).
+  def dedupe_age_range_items
```
maebeale, line 311: 🤖 _From Claude:_ Fixes RecordNotUnique on the categorizable_items unique index. Two new rows for the same age range both pass CategorizableItem's uniqueness validation (neither is persisted yet at validation time) and then collide on INSERT. This collapses them before validation — keeper prefers the persisted row and inherits any duplicate's primary flag. Note: sectors have the same theoretical in-memory-dup gap (separate SectorableItem uniqueness); not seen in practice, left as-is unless it surfaces.

app/views/people/_form.html.erb
```diff
@@ -122,52 +122,90 @@
...
+  <!-- Sectors + Age ranges (split width; both wrap their chips) -->
+  <%# Tells the controller which category types this form edits via category_ids
+      (workshop settings); saving replaces only those and preserves any other type
+      the form never shows. Age ranges are edited as nested attributes, separately. %>
+  <%= hidden_field_tag "person[category_ids][]", "" %>
+  <%# Always submit the managed_category_type_ids key (blank included) so
+      assign_associations takes the preserve branch even when there are no managed
+      types — otherwise the key vanishes and it replaces categories, dropping the
+      person's age ranges and other-type taggings. %>
+  <%= hidden_field_tag "person[managed_category_type_ids][]", "" %>
```
maebeale, line 134: 🤖 _From Claude:_ Mirrors the blank `category_ids` field above. Without it, a person with no profile-specific category types renders zero managed fields, so the browser omits `managed_category_type_ids` entirely and `assign_associations` takes the replace branch — deleting the person's age ranges and other-type taggings. The blank keeps the key present so the preserve branch always runs.

PR #1875, opened 2026-06-22
Sync registration "Other" organization type onto the linked org

maebeale:
🤖 suggested review level: 3 Read 📖 light-logic: refines the org-profile sync plus a small form-option change and a display fallback

- When a registrant picks "Other" for **Organization Type**, the free text they typed was being dropped — the org's `agency_type_other` was only ever set from the admin edit form.
- Now "Other" behaves like every other specify option (reveals a "please specify" box, stores `"Other: <text>"`), and the answer is split onto the linked organization: option label → `agency_type`, stripped free text → `agency_type_other`.

- `public_registration.rb`: `sync_organization_profile` now calls `sync_agency_type`, which splits the label from the free text and fills/clears `agency_type_other`. Same latest-wins / never-clobber-on-blank contract as `apply_value`; `AhoyTrackable` records before/after.
- `organization.rb`: added `Organization::AGENCY_TYPES` (the classifications, with `"Other"` as catch-all), now the single source shared by the model, the org form, and the seeded registration form.
- `organization_decorator.rb`: `agency_type_option` folds any unrecognized non-blank stored value into `"Other"` so an org holding a value not offered by the dropdown can't be silently reclassified as the first option on an untouched save. Blank stays blank.
- `organizations/_form.html.erb` / `form_builder_service.rb`: both consume `Organization::AGENCY_TYPES`; the form's selected option uses the decorator fallback.

### Anything else to add?
- **No data migration needed**: prod has no legacy `"Other (please specify below)"` values; the decorator fallback is defensive only.
- Rebased on main (which introduced `sync_organization_profile`); the "Other" split lives in that method rather than a parallel path.
- Tests: full service, form-builder, and org-decorator suites green.

### DIFF
--- a/app/decorators/organization_decorator.rb
+++ b/app/decorators/organization_decorator.rb
@@ -44,6 +44,18 @@ class OrganizationDecorator < ApplicationDecorator
                   class: "inline-flex shrink-0 items-center justify-center w-5 h-5 rounded-full border text-xs font-semibold #{self.class.program_status_classes(status)}")
   end
 
+  # The org form's type dropdown offers Organization::AGENCY_TYPES. A record may
+  # still hold a value that is no longer offered (e.g. the pre-rename legacy label
+  # "Other (please specify below)"); rendered as-is the select finds no match and
+  # an untouched save would silently reclassify the org as the first option. Fold
+  # any unrecognized non-blank value into the catch-all "Other". Blank stays blank.
+  def agency_type_option
+    return object.agency_type if object.agency_type.blank?
+    return object.agency_type if Organization::AGENCY_TYPES.include?(object.agency_type)
+
+    Organization::AGENCY_TYPE_OTHER
+  end
+
   def detail(length: nil)
     length ? description&.truncate(length) : description
   end
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -187,7 +187,23 @@ module EventRegistrationServices
 
     def sync_organization_profile(organization)
       apply_value(organization, :website_url, field_value("agency_website"))
-      apply_value(organization, :agency_type, field_value("agency_type"))
+      sync_agency_type(organization)
+    end
+
+    # The "Organization Type" answer folds an "Other" choice's free text in as
+    # "Other: <text>" (a specify option). Split the option label from the typed
+    # text: the label drives agency_type and the stripped text fills
+    # agency_type_other, which is cleared for the non-"Other" classifications so
+    # no stale free text lingers. Follows the same latest-wins / never-clobber-on-
+    # blank contract as apply_value.
+    def sync_agency_type(organization)
+      raw = field_value("agency_type")
+      return if raw.blank?
+
+      label, _separator, specified = raw.strip.partition(":")
+      label = label.strip
+      other_text = FormField.other_option?(label) ? specified.strip.presence : nil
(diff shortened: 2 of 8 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -272,6 +273,21 @@ def create_agency_address(organization)
...
 
+    # Mirror the registrant's "Organization Type" answer onto the linked
+    # organization. The "Other" choice is a specify option, so its free text
+    # folds into the answer as "Other: <text>" — split the option label from the
+    # typed text, store the label on agency_type and the stripped text on
+    # agency_type_other (cleared for non-"Other" choices). Latest registration
+    # wins; AhoyTrackable records the before/after on the update.
+    def update_agency_type(organization)
+      raw = field_value("agency_type").to_s.strip
+      label, _separator, specified = raw.partition(":")
```
maebeale: 🤖 _From Claude:_ `partition(":")` is safe here because none of the agency-type option labels contain a colon, so the split always yields the option label and (for "Other") the typed free text. `agency_type_other` is cleared for non-"Other" choices to avoid leaving stale text behind.

app/services/event_registration_services/public_registration.rb
```diff
@@ -187,7 +187,23 @@ def sync_person_profile(person)
...
+      sync_agency_type(organization)
+    end
+
+    # The "Organization Type" answer folds an "Other" choice's free text in as
+    # "Other: <text>" (a specify option). Split the option label from the typed
+    # text: the label drives agency_type and the stripped text fills
+    # agency_type_other, which is cleared for the non-"Other" classifications so
+    # no stale free text lingers. Follows the same latest-wins / never-clobber-on-
+    # blank contract as apply_value.
+    def sync_agency_type(organization)
```
maebeale, line 199: 🤖 _From Claude:_ Folded into `sync_organization_profile` on rebase: `partition(":")` splits the option label from the "Other" free text (safe — no agency-type label contains a colon). `agency_type_other` is cleared for the non-"Other" classifications so no stale text lingers.

app/decorators/organization_decorator.rb
```diff
@@ -44,6 +44,18 @@ def program_status_badge(status = object.program_status)
...
+  # The org form's type dropdown offers Organization::AGENCY_TYPES. A record may
+  # still hold a value that is no longer offered (e.g. the pre-rename legacy label
+  # "Other (please specify below)"); rendered as-is the select finds no match and
+  # an untouched save would silently reclassify the org as the first option. Fold
+  # any unrecognized non-blank value into the catch-all "Other". Blank stays blank.
+  def agency_type_option
+    return object.agency_type if object.agency_type.blank?
+    return object.agency_type if Organization::AGENCY_TYPES.include?(object.agency_type)
+
+    Organization::AGENCY_TYPE_OTHER
```
maebeale, line 56: 🤖 _From Claude:_ Fallback is display-only — it never writes `agency_type_other`, so a legacy "Other" row shows the right classification but keeps whatever specify text it had (blank for legacy). Persists to "Other" only if an admin saves. Prod has no such rows today; this is defensive.

maebeale, line 56: 🤖 _From Claude:_ Fallback is display-only — it never writes agency_type_other, so a legacy "Other" row shows the right classification but keeps whatever specify text it had (blank for legacy). Persists to "Other" only if an admin saves. Prod has no such rows today; this is defensive.

PR #1876, opened 2026-06-22
Fix calendar issue and add admin-authored short description for "Add to calendar" links

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained logic changes with low blast radius

Calendar "Add to calendar" links were feeding the full rich show-page description (`rhino_description`) into each calendar entry, flattened with `.to_plain_text`. Calendars don't render that markup — ICS `DESCRIPTION` is plain text per RFC 5545, and the Google/Outlook/Yahoo URL params render HTML inconsistently — so admins had no real control over the calendar blurb.

**Why:** give admins a dedicated, plain-text short description, while keeping existing events unaffected.

- New nullable `events.short_description` text column.
- Decorator uses it when present, **falls back** to the flattened `rhino_description` when blank.
- Field added to the videoconference section of the event admin form (full width, below the paired label/passcode row).
- Permitted in `EventPolicy`.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -81,7 +81,9 @@ class EventDecorator < ApplicationDecorator
     # If both: URL in location field, physical location in description
     # If only URL: URL in location field
     # If only location: location in location field
-    event_description = object.rhino_description.to_plain_text
+    # Prefer the admin-authored, calendar-specific blurb; fall back to a
+    # flattened version of the rich show-page description when it's blank.
+    event_description = object.calendar_description.presence || object.rhino_description.to_plain_text
 
     if has_url && has_location
       cal_location = object.videoconference_url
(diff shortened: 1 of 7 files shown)


app/decorators/event_decorator.rb
```diff
@@ -81,7 +81,9 @@ def calendar_links(show_videoconference_details: object.videoconference_details_
     # If both: URL in location field, physical location in description
     # If only URL: URL in location field
     # If only location: location in location field
-    event_description = object.rhino_description.to_plain_text
+    # Prefer the admin-authored, calendar-specific blurb; fall back to a
+    # flattened version of the rich show-page description when it's blank.
+    event_description = object.calendar_description.presence || object.rhino_description.to_plain_text
```
maebeale: 🤖 _From Claude:_ `.presence` is the fallback hinge — blank `calendar_description` (nil or empty) falls back to the flattened rich description, so every pre-existing event keeps its current calendar text unchanged.

PR #1877, opened 2026-06-22
Give every seeded notification an email body

maebeale:
🤖 PR, suggested 👤 review level: 👀 Skim — dev-seed data only: adds email bodies to seeded notifications, no app/logic changes

## Why
Seeded contact-us and FYI notifications set an `email_subject` but never `email_body_html`/`email_body_text`, so opening one in the admin **Email Preview** pane (`notifications/show`) rendered the *"No email body captured."* placeholder. This is also what made the user suspect seeded emails had no body.

## What
- Render realistic HTML + plain-text bodies for every seeded notification (contact-us confirmation, contact-us FYI, and the registration/idea/workshop-log FYIs), modeled on the real mailer templates.
- Enrich `contact_us_samples` with sender names + a message so bodies read like genuine submissions.
- Backfill `email_body_*` on re-run for records that were seeded before this change (the `find_or_create_by!` block only fires on create).

## Not in scope
The "Pending, no subject" **registration confirmation** rows in the screenshot are **not** seeded — they come from the real registration flow, where `NotificationMailerJob` fills in subject/body asynchronously. On staging they stay pending if the job queue worker isn't running or the job is failing. That's a separate ops issue, called out so it isn't lost.

### DIFF
--- a/db/seeds/dev/notifications.rb
+++ b/db/seeds/dev/notifications.rb
@@ -2,13 +2,38 @@
 # as part of `rake db:seed:dev`. Seeds contact-us and a few FYI notifications.
 
 puts "Creating Notifications…"
+
+# Shared HTML wrapper so every seeded body renders like a real (mailer-rendered)
+# email in the admin "Email Preview" pane (notifications/show), rather than the
+# "No email body captured." placeholder you get when email_body_* is blank.
+quoted_block = ->(subject, message) do
+  <<~HTML.strip
+    <div style="background-color:#f3f4f6;border-radius:6px;padding:12px;margin:16px 0;text-align:left;">
+      <p><strong>Subject:</strong> #{subject}</p>
+      <p>#{message}</p>
+    </div>
+  HTML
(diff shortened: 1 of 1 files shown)


db/seeds/dev/notifications.rb
```diff
@@ -26,9 +72,30 @@
     n.recipient_role = "person"
     n.notification_type = 0
     n.delivered_at = delivered_at
+    n.email_body_html = confirmation_html
+    n.email_body_text = confirmation_text
   end
+  # Backfill bodies on records seeded before email_body_* was added here
+  confirmation.update!(email_body_html: confirmation_html, email_body_text: confirmation_text) if confirmation.email_body_html.blank?
```
maebeale, line 79: 🤖 _From Claude:_ The block in `find_or_create_by!` only runs on **create**, so records seeded before bodies were added here would stay body-less on re-run. This backfill makes re-seeding idempotently repair them.

PR #1887, opened 2026-06-23
Add FileMaker code field to person edit

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: new string column + one admin-only form field, mirrors the recent racial/ethnic identity addition

## What & why
- Adds a `filemaker_code` string column to `people` so admins can reconcile records against the legacy FileMaker system of record.
- Surfaces it on person edit **only**, in the existing admin-only box beside racial/ethnic identity. Not shown on the profile, registration forms, or anywhere else.

## Changes
- Migration: `add_column :people, :filemaker_code, :string` (reversible)
- Permitted in `person_params`
- Form field in the admin-only box (`allowed_to?(:manage?, Person)`), wrapped both fields in a flex row
- Request specs: field renders for admins + admin can update it

### DIFF
--- a/app/views/people/_form.html.erb
+++ b/app/views/people/_form.html.erb
@@ -380,11 +380,20 @@
 
     <% if allowed_to?(:manage?, Person) %>
       <div class="w-full lg:w-auto rounded-lg border border-gray-200 admin-only bg-blue-100 p-4 shadow-sm">
-        <div class="w-full sm:w-64">
-          <%= f.input :racial_ethnic_identity,
-                      label: "Racial/ethnic identity",
-                      input_html: { class: "w-full" },
-                      wrapper_html: { class: "w-full" } %>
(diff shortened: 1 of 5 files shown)


app/views/people/_form.html.erb
```diff
@@ -380,11 +380,20 @@
 
     <% if allowed_to?(:manage?, Person) %>
       <div class="w-full lg:w-auto rounded-lg border border-gray-200 admin-only bg-blue-100 p-4 shadow-sm">
```
maebeale, line 382: 🤖 _From Claude:_ Placed inside the existing `allowed_to?(:manage?, Person)` admin-only box so the FileMaker code is admin-only, matching racial/ethnic identity. Per the request it's not surfaced on the profile, registration, or anywhere else.

PR #1888, opened 2026-06-23
Validate grant amount covers scholarships already issued

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained logic changes with low blast radius

- Prevent saving a grant with an `amount_cents` lower than the total scholarships already awarded against it. Those draws are committed, so an amount below them would leave scholarships unbacked.

- Added a `Grant` model validation (`amount_covers_scholarships_already_issued`) that compares the proposed amount against the existing `scholarships_total_cents` helper.
- Mirrors `Scholarship#within_grant_budget`, which guards each new draw from the other side of the ledger. Equality is allowed; new grants with no scholarships are unaffected.
- Error message uses `MoneyFormatter.dollars_from_cents` (e.g. "can't be less than the $600 already awarded in scholarships").

### Anything else to add?
- TDD: added four model specs (below/equal/above total, plus fresh grant). All grant + scholarship specs pass; RuboCop clean.

### DIFF
--- a/app/models/grant.rb
+++ b/app/models/grant.rb
@@ -10,6 +10,7 @@ class Grant < ApplicationRecord
   validates :name, presence: true
   validates :amount_cents, numericality: { greater_than_or_equal_to: 0 }
   validates :donor_type, inclusion: { in: DONOR_TYPES }
+  validate :amount_covers_scholarships_already_issued
 
   scope :by_deadline, -> { order(Arel.sql("application_deadline IS NULL, application_deadline ASC")) }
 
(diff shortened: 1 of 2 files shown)


app/models/grant.rb
```diff
@@ -94,6 +95,18 @@ def task_list
...
 
+  # A grant's amount can't be lowered below what has already been awarded against
+  # it — the scholarships are committed, so the grant must at least cover them.
+  # Mirrors Scholarship#within_grant_budget from the other side of the ledger.
+  def amount_covers_scholarships_already_issued
+    return unless amount_cents
+
+    issued = scholarships_total_cents
+    if amount_cents < issued
+      errors.add(:amount_cents, "can't be less than the #{MoneyFormatter.dollars_from_cents(issued)} already awarded in scholarships")
```
maebeale, line 106: 🤖 _From Claude:_ Reuses `scholarships_total_cents`, which sums the in-memory association when loaded and otherwise runs a single SQL aggregate — so no extra query is forced on save. Equality passes intentionally (amount == awarded is valid).

PR #1890, opened 2026-06-23
Improve people index: primary-only sector/age columns, list affiliations

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — view-only styling, one data-source swap on "Affiliated since", and an N+1 fix

Make the people index scan better in its dense, many-column layout. Sector and age columns showed extra values and chips wrapped or squeezed long names into tall boxes, hurting readability.

- Sector/age columns now show **only the primary** value, relabeled "Primary sector" / "Primary age range"; chips no longer wrap mid-value (sector chips truncate with full name on hover, sized to match age-range chips).
- Affiliations render as a **stacked list of truncated org links** (full name on hover, role shown for non-facilitators) instead of chips that vertically stacked long org names.
- "Affiliated since" now derives from the earliest affiliation start year (falling back to legacy `member_since`) instead of the usually-empty `member_since` column, so it actually populates — and `affiliated_since_date` now computes in Ruby from the eager-loaded association, removing a per-row `MIN(start_date)` query (N+1).
- Table fills the card with rounded corners; profile button is fixed-width; User/Edit actions shrunk. Sector-chip and profile-button changes are opt-in params, so other callers are unaffected.

### Anything else to add?
Per-row `person.decorate` was kept (not hoisted to the controller) because the index passes records into `allowed_to?(:show?, person)`, and Action Policy resolves the policy from the object's class — a decorator would break that lookup. With the N+1 removed, the decorate call is just a cheap wrapper.

### DIFF
--- a/app/decorators/person_decorator.rb
+++ b/app/decorators/person_decorator.rb
@@ -84,7 +84,10 @@ class PersonDecorator < ApplicationDecorator
   end
 
   def affiliated_since_date
-    @affiliated_since_date ||= affiliations.minimum(:start_date)
+    # Compute in Ruby from the (eager-loaded) association so list pages that
+    # preload affiliations don't fire a MIN(start_date) query per row.
+    # `.minimum` would ignore the loaded records and hit the DB every time.
+    @affiliated_since_date ||= affiliations.filter_map(&:start_date).min
   end
 
   private
--- a/app/views/people/people_results.html.erb
+++ b/app/views/people/people_results.html.erb
@@ -1,7 +1,7 @@
 <%= turbo_frame_tag :people_results do %>
   <%= turbo_stream.replace("people_count", partial: "people_count") %>
 
-  <div class="rounded-xl bg-white p-6 animate-fade blur-on-submit">
+  <div class="rounded-xl bg-white overflow-hidden animate-fade blur-on-submit">
     <% if @people.any? %>
       <div class="overflow-x-auto">
         <table class="w-full border-collapse border border-gray-200">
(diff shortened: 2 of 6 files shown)


app/views/people/people_results.html.erb
```diff
@@ -34,7 +34,7 @@
                   <% end %>
                 </td>
                 <td class="px-4 py-2 text-center text-sm text-gray-800">
-                  <%= person.member_since&.year %>
+                  <%= (person.decorate.affiliated_since_date || person.member_since)&.year || "--" %>
```
maebeale, line 37: 🤖 _From Claude:_ Column header is "Affiliated since" but it previously read the manually-entered `member_since` column, which is `NULL` for nearly everyone, so it rendered blank. Now prefers the earliest affiliation start year (`affiliated_since_date`) and falls back to `member_since`. Per-row `MIN(start_date)` query is fine here — each row is wrapped in `cache [person, …]` and `Affiliation belongs_to :person, touch: true` busts it on change.

app/views/people/people_results.html.erb
```diff
@@ -69,29 +70,38 @@
...
-                              style: org_style %>
-                        <% else %>
-                          <span class="<%= org_classes %>" style="<%= org_style %>"><%= affiliation.organization.name %></span>
-                        <% end %>
+                        <li class="min-w-0">
+                          <% if allowed_to?(:show?, affiliation.organization) %>
+                            <%= link_to affiliation.organization.name, organization_path(affiliation.organization),
+                                data: { turbo_frame: "_top" },
+                                title: affiliation.organization.name,
+                                class: "block max-w-[180px] truncate font-medium #{DomainTheme.text_class_for(:organizations)} hover:underline" %>
```
maebeale, line 84: 🤖 _From Claude:_ Switched org affiliations from chips to a stacked list of links. `max-w-[180px] truncate` clamps each name (chips can't, since an auto table grows to fit unbreakable content); full name stays on hover via `title`. The old gray left-border distinguishing non-facilitator roles is replaced by showing the role title in muted text — more legible.

app/views/sectors/_tagging_label.html.erb
```diff
@@ -36,16 +45,17 @@
 
 <% if sector %>
   <%= link_to taggings_path(sector_names_all: sector.name),
+    title: sector.name,
```
maebeale, line 48: 🤖 _From Claude:_ `max_width`/`compact` default off and `title` is harmless, so every other caller of this shared chip (tag clouds, org/profile pages) renders exactly as before. `white-space` inherits, hence `truncate min-w-0` on the inner span plus `min-w-0` on the inline-flex link makes ellipsis work.

app/decorators/person_decorator.rb
```diff
@@ -84,7 +84,10 @@ def facilitator_since_date
   end
 
   def affiliated_since_date
-    @affiliated_since_date ||= affiliations.minimum(:start_date)
+    # Compute in Ruby from the (eager-loaded) association so list pages that
+    # preload affiliations don't fire a MIN(start_date) query per row.
+    # `.minimum` would ignore the loaded records and hit the DB every time.
+    @affiliated_since_date ||= affiliations.filter_map(&:start_date).min
```
maebeale, line 90: 🤖 _From Claude:_ `affiliations.minimum(:start_date)` is a SQL aggregate that ignores the eager-loaded association and fires once per row on the index (N+1). `filter_map(&:start_date).min` reads the already-loaded records in Ruby — same result (earliest start date, or nil), no per-row query. On pages that don't preload affiliations this loads the association once instead of a MIN query, which is neutral-to-better.

PR #1891, opened 2026-06-23
Allow bare domains in the agency website field

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — small, contained view + spec change with low blast radius

- The organization website field on the public registration form rendered as `<input type="url">`, so browsers blocked submission of bare domains like `awbw.org` (they require an `http://`/`https://` scheme).
- The server already accepts bare domains — `Organization#website_link_url` prepends `https://` when there's no scheme — so the browser was the only thing rejecting valid input.

- Render the `agency_website` field as `type="text"` with `inputmode="url"`, keeping the URL-optimized mobile keyboard without the strict scheme requirement.
- Added a request spec asserting the field renders as a text input with `inputmode="url"` and not `type="url"`.

### DIFF
--- a/app/views/events/public_registrations/_form_field.html.erb
+++ b/app/views/events/public_registrations/_form_field.html.erb
@@ -48,12 +48,12 @@
         html_type = case field.field_identifier
                      when /email(?!_type)/ then "email"
                      when "phone" then "tel"
-                     when "agency_website" then "url"
                      else
                        field.number_integer? ? "number" : "text"
                      end
         extra_attrs = []
         extra_attrs << 'min="1" step="1" pattern="[0-9]*" inputmode="numeric"' if field.number_integer?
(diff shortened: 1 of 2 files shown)


app/views/events/public_registrations/_form_field.html.erb
```diff
@@ -48,12 +48,12 @@
         html_type = case field.field_identifier
                      when /email(?!_type)/ then "email"
                      when "phone" then "tel"
-                     when "agency_website" then "url"
                      else
```
maebeale, line 51: 🤖 _From Claude:_ Dropped `type="url"` here because browsers reject bare domains (`awbw.org`) against it. `inputmode="url"` keeps the URL keyboard on mobile, and the server still normalizes the scheme via `Organization#website_link_url`.

PR #1892, opened 2026-06-23
Resource-callout eyebrows, W-9 as a resource, and PDF viewer hints

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: `return_to`-driven back links, a hardcoded single-page title list, and a W-9 resource lookup

## Why
Opening a document from the **Handouts**, **Forms**, or **Payment** callout dropped the registrant back to the ticket, past the callout they came from. The W-9 was also a bare direct download, and the single-page PDF preview gave no hint there were more pages or that it takes a moment to render.

## What
- **Eyebrows follow the origin.** Handouts/Forms cards and the invoice links pass `return_to`; the resource page and invoice page branch on it to return to the originating callout (handouts/forms/payment), defaulting to the ticket. `_callout_page` / invoice `_actions` take a configurable `back_label`.
- **Resource detail header** reads *Handouts detail* / *Forms detail* with the resource's own title inside the card.
- **W-9 is now a resource** (seeded, looked up by title) shown on its own resource page with the back-to-forms eyebrow and the arrow affordance, instead of a direct download.
- **Viewer hints:** the preview notes it may take a moment to load, and multi-page PDFs (everything but the W-9 and letter to supervisors — hardcoded for now) prompt a download for all pages.

## Note
Like the letter to supervisors and handouts, the W-9 card only appears where a `Resource` titled "W-9" exists — seeded for dev here; production needs the W-9 uploaded as a resource.


app/views/events/callouts/resource.html.erb
```diff
@@ -1,5 +1,15 @@
...
-<%= render layout: "events/callouts/callout_page", locals: { title: @resource.title } do %>
+<%# Eyebrow returns to the callout the registrant came from (handouts/forms),
+    falling back to the ticket when reached directly. %>
+<% back_path, back_label = case params[:return_to]
+   when "handouts"
+     [ registration_handouts_path(@event_registration.slug), "Back to handouts" ]
+   when "forms"
+     [ registration_forms_path(@event_registration.slug), "Back to forms" ]
+   else
+     [ registration_ticket_path(@event_registration.slug), "Back to ticket" ]
```
maebeale: 🤖 _From Claude:_ The `else` branch keeps the existing default (Back to ticket), so direct/bookmarked access to the resource page and the inline custom-callout resource render are unaffected — only the handouts/forms entry points pass `return_to`.

app/controllers/events/callouts_controller.rb
```diff
@@ -103,22 +103,25 @@ def set_event
...
-      cards = [
-        resource_card(icon: "fa-solid fa-file-pdf", title: "Download W-9",
-                      subtitle: "AWBW's W-9 tax form for your records",
-                      href: "/documents/awbw-w9.pdf", trailing_icon: "fa-solid fa-download"),
-        resource_card(icon: "fa-solid fa-file-invoice-dollar", title: "View invoice",
-                      subtitle: "Itemized invoice for this registration",
-                      href: registration_invoice_path(@event_registration.slug))
-      ]
+      cards = []
+      w9 = Resource.find_by(title: "W-9")
```
maebeale, line 111: 🤖 _From Claude:_ Mirrors the letter-to-supervisors guard: the W-9 card only renders when a Resource titled "W-9" exists. Seeded for dev; prod needs it uploaded, otherwise the W-9 silently drops off the forms page (the prior static download is gone).

app/models/resource.rb
```diff
@@ -12,6 +12,11 @@ def self.mentionable_rich_text_fields
   PUBLISHED_KINDS = [ "Handout", "Template", "Toolkit", "Form" ]
   KINDS = PUBLISHED_KINDS + [ "Resource", "Story", "LeaderSpotlight", "SectorImpact", "Theme", "Scholarship" ]
 
+  # Titles whose downloadable PDF is a single page. Every other resource PDF is
+  # multi-page, so its first-page preview needs a "download for all pages" note.
+  # Hardcoded for now — replace with real PDF introspection later.
+  SINGLE_PAGE_PDF_TITLES = [ "Letter to Supervisors", "W-9" ].freeze
```
maebeale, line 18: 🤖 _From Claude:_ Hardcoded per request — we know only the W-9 and letter are single-page. Swap this title list for real PDF introspection (e.g. poppler `pdfinfo`) when we want it to generalize.

PR #1893, opened 2026-06-23
Show PDF resources in the browser's inline viewer

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: a shared view partial branching on PDF vs. not, plus removal of the earlier page-count machinery

Callout resource pages and the resource detail page previewed only a PDF's first page as an image, so multi-page handouts looked like one-pagers.

- PDFs now render in the **browser's built-in inline viewer** (scrollable, all pages — every major browser ships one), falling back to the first-page/image preview for the rare client that can't.
- Non-PDF resources keep the existing preview.
- Extracted the PDF-or-preview choice into `assets/_resource_display`, shared by the callout resource viewer and the resource **show** page.

Per review feedback, dropped the page-counting approach entirely (the `PdfPageCountAnalyzer` + `metadata["pages"]` badge, the `poppler-utils` CI step, the `pdfs:backfill_page_counts` task, and the hardcoded `single_page_pdf?`) — the browser viewer shows every page, so none of it is needed. No backfill or deploy step required now.


app/views/events/callouts/_resource_body.html.erb
```diff
@@ -1,16 +1,35 @@
...
     <%= link_to resource_download_path(resource), download: true, class: "btn btn-utility" do %>
       Download <i class="fa-solid fa-download"></i>
     <% end %>
   </div>
 <% end %>
 
-<%= render "assets/display_assets",
-           resource: resource, file: resource.display_image,
-           variant: :hero, link: true %>
+<% if pages > 1 %>
```
maebeale: 🤖 _From Claude:_ Inline embed is gated on `pages > 1`: single-page PDFs keep the crisp first-page image, and if page count is unknown (poppler missing or not yet analyzed) we fall back to the image rather than embedding. The `display_assets` render inside `<object>` is the fallback for browsers that won't render inline PDFs (notably mobile Safari/Chrome).

app/analyzers/pdf_page_count_analyzer.rb
```diff
@@ -0,0 +1,41 @@
+require "open3"
+
+# Records the page count of PDF attachments in blob metadata (metadata["pages"]),
+# so views can show a "N pages" indicator and decide whether to offer a scrollable
+# multi-page preview without shelling out per request. Uses poppler's `pdfinfo`
+# (already a runtime dependency via HighResPopplerPdfPreviewer); silently skipped
+# when poppler isn't installed.
+class PdfPageCountAnalyzer < ActiveStorage::Analyzer
+  class << self
+    def accept?(blob)
```
maebeale: 🤖 _From Claude:_ `accept?` returns false when `pdfinfo` is absent, so the analyzer is a no-op (no `pages` metadata) rather than raising — the view then degrades to the first-page preview. Mirrors how `HighResPopplerPdfPreviewer` guards on `pdftoppm`.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v6
 
+      # poppler-utils provides pdftoppm/pdfinfo, used by the PDF previewer and the
+      # PDF page-count analyzer (mirrors the production runtime dependency).
+      - name: Install poppler-utils
+        run: sudo apt-get update && sudo apt-get install -y poppler-utils
```
maebeale: 🤖 _From Claude:_ poppler is already a prod runtime dependency (the PDF previewer shells out to `pdftoppm`); installing it here lets the new PDF specs actually run instead of self-skipping.

jmilljr24:
I see the pdf viewer for multi page is just using the built in browser feature. That's a good thing since all major browsers support it.

In that case, my vote would be to completely remove the code around counting the number of pages. If `PDF` use browser pdf view, else use preview. Keeps it simple and even for single page pdf's it becomes obvious the file is actually a PDF not just an image.

This would work well on the Resource show as well.

app/views/assets/_resource_display.html.erb
```diff
@@ -0,0 +1,17 @@
+<%# Renders a resource's main display. A PDF opens in the browser's built-in
+    inline viewer (scrollable, all pages) — every major browser ships one — with
+    the first-page/image preview as the fallback child for the rare client that
+    can't. Anything else (images, etc.) renders through the shared asset
+    pipeline. `resource` is a decorated Resource. %>
+<% display = resource.display_image %>
+<% if display.respond_to?(:attached?) && display.attached? && display.content_type == "application/pdf" %>
```
maebeale: 🤖 _From Claude:_ The PDF check keys off `display_image` (the resolved hero: primary_asset → downloadable_asset → …), not the downloadable file directly — so a resource with a primary image *and* a downloadable PDF still shows its image here, and only resources whose hero actually is a PDF get the inline viewer. `respond_to?(:attached?)` guards against the symbol/string defaults `display_image` can return.

db/seeds/dev/home_page_content.rb
```diff
@@ -34,7 +34,8 @@
                .first_or_create!(
                   body: body_content,
                   rhino_body: "<p>#{body_content}</p>",
-                  author_id: User.all.sample&.id,
+                  author_id: Person.all.sample&.id,
+                  user_author_id: User.all.sample&.id,
```
jmilljr24: This needed in this PR?

jmilljr24 (approved):

db/seeds/dev/home_page_content.rb
maebeale: @jmilljr24 technically no, but this seeds fix was so small i thought it fine to ride along here rather than make a whole new pr.

app/views/assets/_resource_display.html.erb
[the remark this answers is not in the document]
maebeale: @jmilljr24 this seems like a lightweight way to do the single page differently. do you still vote to just show all pdfs in the pdf viewer? i can see arguments for both.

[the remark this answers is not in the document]
jmilljr24: We don't have "pages" in our metadata. That would require all the other changes for prepossessing.

So from the technical side `if content_type = pdf` is simplest. From a personal opinion on UX, I think having the consistency that all pdf's are displayed the same (pdf viewer) and all image files are displayed the same (image at whatever size). Having some pdf's look like an image and some opening in the pdf view creates a divide that isn't 100% transparent to the user.

db/seeds/dev/home_page_content.rb
jmilljr24: Just making sure it wasn't leftover from something not needed.

PR #1894, opened 2026-06-23
Dedupe org snapshot so paid public registrations don't crash

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained fix to an after_create hook with a regression test

Fixes a production crash (`ActiveRecord::RecordInvalid: Event registration organizations is invalid`) that hit paid public registrations at Stripe checkout. A registrant can hold multiple active affiliations to the same organization (e.g. different titles), and `EventRegistration#snapshot_registrant_organizations` created one link row per affiliation — the duplicate failed the uniqueness validation and left an invalid, unsaved record in the in-memory collection, which then blew up the next registration save (recording the checkout session id).

Dedupe the organization ids before creating link rows so each org is snapshotted once (also one query instead of N). Added a regression spec that reproduced the exact crash (red first, then green).

### Anything else to add?
Only the Stripe-paying path re-saves the registration after creation, which is why this was the only surface that crashed.

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -407,9 +407,14 @@ class EventRegistration < ApplicationRecord
 
   private
 
+  # Link each organization the registrant is actively affiliated with — but only
+  # once. A registrant can hold several active affiliations to the same org (e.g.
+  # different titles), so we dedupe the organization ids first; creating a second
+  # row for the same org would fail the uniqueness validation and leave an invalid
+  # record in the in-memory collection that breaks the next save of the registration.
   def snapshot_registrant_organizations
-    registrant.affiliations.active.includes(:organization).find_each do |aff|
-      event_registration_organizations.create(organization: aff.organization)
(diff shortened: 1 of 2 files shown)


app/models/event_registration.rb
```diff
@@ -407,9 +407,14 @@ def remote_search_label
 
   private
 
+  # Link each organization the registrant is actively affiliated with — but only
+  # once. A registrant can hold several active affiliations to the same org (e.g.
+  # different titles), so we dedupe the organization ids first; creating a second
```
maebeale, line 412: 🤖 _From Claude:_ `.distinct.pluck(:organization_id)` is the crux — without dedup, a registrant with two active affiliations to the same org produced a second `.create` that failed the uniqueness validation but stayed in the in-memory collection, poisoning the later `update!(checkout_session_id:)` autosave (the actual crash site).

PR #1895, opened 2026-06-23
Connect only the form-submitted org to a registration

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small change to which orgs a registration links, with tests + demo seeds

- Registrations auto-snapshotted **every** active affiliation of the registrant on create. A returning facilitator accumulates affiliations across trainings, so each new registration got mislabeled with all their orgs.
- Now a registration connects only the **one** org submitted on the form (or none).

- Removed the `after_create :snapshot_registrant_organizations` callback from `EventRegistration`.
- `PublicRegistration` connects only the form-submitted org, on both new and repeat submissions (`find_or_create_by!`).
- Multiple orgs now accrue only deliberately: an admin links extras from the edit page, or the registrant applies again with a different org (each submission adds its single org to the same registration).
- Updated dev seeds that relied on the snapshot to link orgs explicitly, and added two demo registrants for the multi-org paths: **A5** (form-submitted org + admin-linked second org → two orgs, one submission) and **A6** (applied twice with different orgs → two orgs, two submissions).

### Anything else to add?
- Admin-created registrations (the `new` form has no org section) now start with no orgs; admins link via the existing "Connect organization" UI.
- Affiliations are still created via the shared `AffiliationServices::CreateFromRegistration` (same as admin Link Organization). The public flow only resolves *existing* orgs (`find_by(name:)`) — it does not create an org from an unmatched typed name; that stays an admin step.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -71,12 +71,14 @@ module EventRegistrationServices
             existing.update!(status: "registered")
             send_notifications(existing)
           end
+          connect_organization(existing, organization)
           submission = update_form_submission(person)
           save_scholarship_submission(person)
           return Result.new(success?: true, event_registration: existing, form_submission: submission, errors: [])
         end
 
         event_registration = create_event_registration(person)
+        connect_organization(event_registration, organization)
         submission = create_form_submission(person)
         save_scholarship_submission(person)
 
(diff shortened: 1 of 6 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -71,12 +71,14 @@ def call
             existing.update!(status: "registered")
             send_notifications(existing)
           end
+          connect_organization(existing, organization)
```
maebeale, line 74: 🤖 _From Claude:_ Repeat applications hit this path — connecting the newly-submitted org here is what lets a registrant who applies again with a different org add that single org to the same registration (now with two submissions).

PR #1906, opened 2026-06-24
Remove dead migration-audit routes that 500 on bot probes

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: removes routes pointing at non-existent controllers; verified nothing references them

Leftover "migration audit" routes (`images`, the `images` namespace, `attachments`, `media_files`) pointed at controllers that never existed, so any request matching them raised `ActionDispatch::MissingController` — a 500 instead of a 404. A bot probing `GET /images/index.php` matched `resources :images` and dispatched to the missing `ImagesController`, generating Honeybadger noise.

Removed the dead routes (keeping `primary_assets` and `rich_text_assets`, the only two backed by real controllers and actually used — no route helpers reference the removed ones). Added a request spec asserting each removed path now 404s; it fails 7/7 on the old routes and passes after the fix.

### Anything else to add?
This was purely bot/scanner traffic probing for PHP vulnerabilities — no real user or data was affected. A routing spec can't catch this since recognition masks the missing controller as a `RoutingError`; only full dispatch reproduces the 500, so the test is a request spec.

### DIFF
--- a/spec/requests/dead_migration_audit_routes_spec.rb
+++ b/spec/requests/dead_migration_audit_routes_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+# Leftover "migration audit" routes pointed at controllers that never existed
+# (ImagesController, AttachmentsController, MediaFilesController, Images::*).
+# A bot probing GET /images/index.php matched `resources :images` and dispatched
+# to the missing ImagesController, raising ActionDispatch::MissingController (a
+# 500) instead of a plain 404. These paths must be unrouted, i.e. raise a
+# routing error rather than a missing-controller error.
+RSpec.describe "Removed migration-audit routes", type: :request do
+  [
+    "/images/index.php",
+    "/images/1",
+    "/images/primary_images/1",
+    "/images/gallery_images/1",
+    "/images/rich_texts/1",
+    "/attachments/1",
+    "/media_files/1"
+  ].each do |path|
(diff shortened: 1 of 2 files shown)


spec/requests/dead_migration_audit_routes_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+# Leftover "migration audit" routes pointed at controllers that never existed
+# (ImagesController, AttachmentsController, MediaFilesController, Images::*).
+# A bot probing GET /images/index.php matched `resources :images` and dispatched
+# to the missing ImagesController, raising ActionDispatch::MissingController (a
+# 500) instead of a plain 404. These paths must be unrouted, i.e. raise a
+# routing error rather than a missing-controller error.
```
maebeale, line 8: 🤖 _From Claude:_ Deliberately a request spec, not a routing spec — `recognize_path` masks the missing controller as a `RoutingError`, so a routing spec would pass even with the broken routes still present. Only full dispatch reproduces the 500, so this hits the real path and asserts a 404.

jmilljr24 (approved):
Thanks!

If the routes no longer exist, I'm not sure we need any tests?

maebeale:
def true. but the test helps ensure bots are getting the error we want them to. i think leaving it in is good for future agents to understand we care about which status codes requests are getting partly bc of bot traffic.

PR #1909, opened 2026-06-24
add resource policy for download to match show

### LINKED ISSUE #1908: Resource - download link produces html, not document
Set Resource to the following settings

- Published
- Publicly visible
- Hidden from search

Download button: Logged in user gets the real download, non-authenticated user get an `.html` file.
Preview Image: Both work

jmilljr24:
Closes #1908 

Fix resource download link for unauthenticated-users

skip authentication check for download and add download policy

### DIFF
--- a/app/policies/resource_policy.rb
+++ b/app/policies/resource_policy.rb
@@ -18,7 +18,7 @@ class ResourcePolicy < ApplicationPolicy
   end
 
   def download?
-    true
+    show?
   end
 
   def filter_published?
--- a/spec/policies/resource_policy_spec.rb
+++ b/spec/policies/resource_policy_spec.rb
@@ -135,19 +135,51 @@ RSpec.describe ResourcePolicy, type: :policy do
   # -----------------------------------------
 
   describe "#download?" do
-    it "allows admin" do
-      expect(policy_for(record: private_resource, user: admin_user))
-        .to be_allowed_to(:download?)
+    context "admin" do
+      it "can download anything" do
+        expect(policy_for(record: private_resource, user: admin_user))
+          .to be_allowed_to(:download?)
+      end
     end
 
-    it "allows regular user" do
-      expect(policy_for(record: private_resource, user: regular_user))
-        .to be_allowed_to(:download?)
+    context "regular user" do
+      it "can download published resource" do
+        expect(policy_for(record: published_resource, user: regular_user))
+          .to be_allowed_to(:download?)
+      end
+
+      it "cannot download private resource" do
+        expect(policy_for(record: private_resource, user: regular_user))
+          .not_to be_allowed_to(:download?)
+      end
+
+      it "can download publicly visible resource" do
+        expect(policy_for(record: public_resource, user: regular_user))
+          .to be_allowed_to(:download?)
+      end
     end
 
-    it "allows guest" do
-      expect(policy_for(record: private_resource, user: guest_user))
-        .to be_allowed_to(:download?)
+    context "guest" do
+      it "can download publicly visible resource" do
+        expect(policy_for(record: public_resource, user: guest_user))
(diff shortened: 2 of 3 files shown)


jmilljr24:
@maebeale I think this is fine for a quick fix.

Question:
This highlighted a discrepency on the `Resource` show page. There are two places a user can click to download the asset. 1. the download button 2. the preview image link. The download button uses the `download` controller action. The preview link is a direct `url_for(@file)`. 

To some degree I think that both should follow the same pattern. Using the direct url makes sense because the show page is already gated to the correct policy on who can see (in turn who can download). The download button/action however taps into the the ahoy tracking which we don't have with the direct url.

I can create an issue for this, just wanted to see which direct you thought before I wrote it up. It is a low priority.

app/policies/resource_policy.rb
```diff
@@ -18,7 +18,7 @@ def update?
   end
 
   def download?
-    true
+    show?
```
maebeale, line 21: i don't think show will be true for these hidden_from_search resources. i think true is better, or, we add `show? || record.hidden_from_search?`, or, update the show policy action.

spec/policies/resource_policy_spec.rb
```diff
@@ -135,19 +135,51 @@ def policy_for(record:, user:)
...
+        expect(policy_for(record: public_resource, user: guest_user))
+          .to be_allowed_to(:download?)
+      end
+
+      it "cannot download published-only resource" do
+        expect(policy_for(record: published_resource, user: guest_user))
+          .not_to be_allowed_to(:download?)
+      end
+
+      it "can download a publicly visible resource that is hidden from search" do
```
maebeale: regular_user should be able to download it too. they may still refer to their ticket from time to time.

maebeale:
@jmilljr24 i don't think the policies are right yet

maebeale:
@jmilljr24 i'm fine w consolidating Resource download behavior, though i thought there were cases for both. def should add the ahoy tracking to direct if we keep it or pick it as the unified behavior. if we pick one i pick direct where you get the full preview.

maebeale (approved):
i don't think the policy is correct yet so surprised tests aren't catching it. i must be missing something.

spec/policies/resource_policy_spec.rb
jmilljr24: This test is actually unneeded. `hidden_from_search` is only used on the scope for search results. There is no direct policy gated on it.

app/policies/resource_policy.rb
jmilljr24: There is a mix up between the policy scope and the policies related to controller actions. 

`hidden_from_search` is not a controller action, it is just the flag the search scope looks at to filter out results.

I'm confident `show?` is set up correctly. A resource that is hidden from search would still need the correct settings for the correct users to see it regardless of if it is hidden from search or not. In turn, if a user can see the show page, they should be able to download it as well.

jmilljr24:
> i pick direct where you get the full preview.

I'm not sure I know what you mean by this.

jmilljr24:
> i don't think the policy is correct yet so surprised tests aren't catching it. i must be missing something.

I tried to explain it on the individual comments but here is the big picture:

`resources/1` hits the show action -> show? policy
``` 
def show?
    admin? || record.publicly_visible? || (authenticated? && record.published?)
  end
```
So if you are authorized to view the show page, you should be authorized to download. There is no reason to duplicate the show method so I made download? call the show? method. Realistically I could have use an alias but I know that action policy isn't always intuitive if you don't work with it much. The registration ticket routes never hit the resource show action or policy. So the user can see the registration version of the resource, the download button which was copied from there resource show still points to `resource#download`. 

`hidden_from_search` is only used on the Resource index results.

PR #1912, opened 2026-06-25
Sync org status from facilitator affiliations and surface mismatches in the form

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — small, contained model/view logic change

Org status auto-sync keyed off **all** active affiliations, while the org form's status indicator keyed off active **Facilitator** affiliations — so the two could permanently disagree (e.g. an org with an active non-facilitator affiliation would be kept Active by the model but flagged "Does not match affiliations status" in the form, with no way to reconcile).

- Made **facilitator affiliations authoritative** on both sides: `Affiliation#sync_organization_status_with_affiliations` now keys off `affiliations.facilitators.active`. A non-facilitator affiliation no longer keeps an org Active.
- Form shows the status select + red mismatch hint only when the stored status diverges from the facilitator-calculated status (and always for admins).
- Reactivation only flips orgs back from Inactive (the status the deactivation callback sets); Pending/Reinstate/Unknown are left untouched. Both auto-changes are Ahoy-tracked.
- Removed the now-redundant status badge from the org show page.

### Anything else to add?
Covered by new model specs (incl. the non-facilitator case) and edit-view specs for select visibility / mismatch hint.

### DIFF
--- a/app/models/affiliation.rb
+++ b/app/models/affiliation.rb
@@ -31,9 +31,9 @@ class Affiliation < ApplicationRecord
 
   before_validation :skip_if_duplicate
   before_save :set_inactive_from_dates
-  after_save :deactivate_organization_if_no_active_people
+  after_save :sync_organization_status_with_affiliations
   after_save :sync_organization_affiliation_dates
-  after_destroy :deactivate_organization_if_no_active_people
+  after_destroy :sync_organization_status_with_affiliations
   after_destroy :sync_organization_affiliation_dates
 
   # Methods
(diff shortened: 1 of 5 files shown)


app/models/affiliation.rb
```diff
@@ -107,9 +107,17 @@ def sync_organization_affiliation_dates
     org.update_columns(updates) if updates.any?
   end
 
-  def deactivate_organization_if_no_active_people
-    return if organization.affiliations.active.exists?
+  # Org status tracks active *Facilitator* affiliations specifically (mirroring the
+  # form's status indicator) — a non-facilitator affiliation does not keep an org active.
+  def sync_organization_status_with_affiliations
+    if organization.affiliations.facilitators.active.exists?
```
maebeale, line 113: 🤖 _From Claude:_ Switched from `affiliations.active` to `affiliations.facilitators.active` so the model and the form's status indicator share one definition of "active". This is the deliberate fix for the divergence — facilitator affiliations are authoritative, so a non-facilitator affiliation no longer keeps an org Active.

jmilljr24 (approved):

PR #1913, opened 2026-06-25
Backfill payer fields from profile on logged-in bulk payments

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained service change with new tests

- Logged-in payers don't see the `logged_out_only` payer fields, so those keys never reached the submitted params — the payer's email/name/phone/org were silently missing from the saved form answers, even though we know who they are.

- `BulkPayment#backfill_payer_fields` fills `payer_first_name`, `payer_last_name`, `payer_email`, `payer_phone`, and `payer_organization` from the known person, only when the field is empty (never clobbers a submitted value).
- Backfill runs **after** the `create_phone_contact` guard, so copying the payer's own phone doesn't rewrite their existing contact methods.
- Organization prefers the facilitator org (`program_organization`), then most recent active affiliation (`primary_organization`).

### Anything else to add?
- New service specs cover each field, the facilitator-org preference + fallback, the no-overwrite rule, and the no-contact-rewrite guarantee.

### DIFF
--- a/app/services/event_registration_services/bulk_payment.rb
+++ b/app/services/event_registration_services/bulk_payment.rb
@@ -17,7 +17,11 @@ module EventRegistrationServices
     def call
       ActiveRecord::Base.transaction do
         person = find_or_create_person
+        # Backfill after the phone-contact guard: create_phone_contact should only
+        # fire for a phone the payer actually typed, not one we copied from their
+        # own profile (which would needlessly rewrite their contact methods).
         create_phone_contact(person) if field_value("payer_phone").present?
+        backfill_payer_fields(person)
         submission = create_form_submission(person)
         send_notifications(submission, person)
         Result.new(success?: true, form_submission: submission, errors: [])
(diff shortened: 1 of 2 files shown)


app/services/event_registration_services/bulk_payment.rb
```diff
@@ -17,7 +17,11 @@ def initialize(event:, form:, form_params:, person: nil)
     def call
       ActiveRecord::Base.transaction do
         person = find_or_create_person
+        # Backfill after the phone-contact guard: create_phone_contact should only
+        # fire for a phone the payer actually typed, not one we copied from their
```
maebeale, line 21: 🤖 _From Claude:_ Backfill runs after this guard on purpose — copying the payer's own `phone_number` here would otherwise re-fire `create_phone_contact` and rewrite their existing contact methods (forcing `contact_type: "personal"`, demoting other primaries).

app/services/event_registration_services/bulk_payment.rb
```diff
@@ -50,6 +54,34 @@ def send_notifications(submission, person)
...
+        "payer_first_name" => person.first_name,
+        "payer_last_name" => person.last_name,
+        "payer_email" => person.preferred_email,
+        "payer_phone" => person.phone_number,
+        "payer_organization" => payer_organization_name(person)
+      }.each do |key, value|
+        next if value.blank?
+
+        field = @form.form_fields.find_by(field_identifier: key)
+        next unless field
```
maebeale, line 72: 🤖 _From Claude:_ `next if @form_params[...].present?` keeps a logged-out user's typed value authoritative — backfill only fills gaps, never overwrites a submitted answer.

app/services/event_registration_services/bulk_payment.rb
```diff
@@ -50,6 +54,34 @@ def send_notifications(submission, person)
...
+        @form_params[field.id.to_s] = value
+      end
+    end
+
+    # Prefer the org the payer facilitates for, falling back to their most recent
+    # active affiliation when they have no facilitator affiliation.
+    def payer_organization_name(person)
+      (person.program_organization || person.primary_organization)&.name
+    end
+
```
maebeale, line 84: 🤖 _From Claude:_ Per request: facilitator org (`program_organization`) wins, falling back to most recent active affiliation (`primary_organization`).

jmilljr24 (approved):

PR #1918, opened 2026-06-25
add combined payer field to bulk payment index

jmilljr24:

I missed this when I finished the combine search in allocations.


maebeale (approved):
that controller edit is so satisfying

PR #1920, opened 2026-06-26
Filterable, lazy-loaded grants index + scholarship theming

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — substantive logic: new filter scopes, controller branching, and a lazy turbo-frame flow

- Awarders need to find grants by name, donor, remaining funds, and task status — the flat list didn't scale.

- Filter + lazy turbo-frame pattern (mirrors people/users): the full page renders header + filters + a skeleton; the frame's `src` request returns the filtered, paginated rows so the table reloads in place.
- Funds/task scopes use flat `WHERE`/correlated subqueries instead of `GROUP BY`, so `will_paginate`'s count stays correct on the paginated index.
- `table-sort` switched to strict `Number()` so ISO dates sort correctly instead of collapsing every date in a year to one value.
- Tasks-completed indicators and the grant scholarships card adopt the fuchsia scholarships theme; card links use the grey registrants style.
- Recipients page links out to the grants index and to each funding grant.

### Anything else to add?
- New scopes (`fully_issued`, `tasks_outstanding`, `all_tasks_completed`) are NULL-safe for grant-less scholarships; covered by model + request specs.


app/frontend/javascript/controllers/table_sort_controller.js
```diff
@@ -45,8 +45,12 @@ export default class extends Controller {
   compare(rowA, rowB, index, key) {
     const a = this.cellValue(rowA, index, key)
     const b = this.cellValue(rowB, index, key)
-    const numA = parseFloat(a)
-    const numB = parseFloat(b)
+    // Strict Number(), not parseFloat(): parseFloat("2026-12-31") is 2026, which
+    // would collapse every date in a year to one value and leave date columns
+    // effectively unsorted. Number() rejects such strings (NaN) so ISO dates fall
+    // through to localeCompare, which orders them correctly since they're padded.
+    const numA = a === "" ? NaN : Number(a)
```
maebeale, line 52: 🤖 _From Claude:_ Strict `Number()` instead of `parseFloat()`: `parseFloat("2026-12-31")` is 2026, which collapsed every date in a year to one value. `Number()` returns NaN for ISO date strings so they fall through to `localeCompare`, which orders zero-padded ISO dates correctly.

app/models/grant.rb
```diff
@@ -13,12 +13,31 @@ class Grant < ApplicationRecord
...
+
+  # Task-completion filters keyed off the grant's scholarships. "Outstanding"
+  # means at least one scholarship still has incomplete tasks; "all completed"
+  # means the grant has scholarships and none are outstanding. The subqueries
+  # exclude grant-less scholarships (grant_id IS NULL) — a stray NULL in the
+  # NOT IN set below would otherwise make all_tasks_completed match nothing.
+  scope :tasks_outstanding, -> {
+    where(id: Scholarship.where(tasks_completed: false).where.not(grant_id: nil).select(:grant_id))
+  }
+  scope :all_tasks_completed, -> {
```
maebeale, line 38: 🤖 _From Claude:_ Both subqueries exclude `grant_id IS NULL`. Without that guard a grant-less, incomplete scholarship would put a NULL in the `NOT IN` set and make `all_tasks_completed` match nothing. Covered by a model spec.

app/models/grant.rb
```diff
@@ -13,12 +13,31 @@ class Grant < ApplicationRecord
...
+  ALLOCATED_CENTS_SUBQUERY =
+    "COALESCE((SELECT SUM(scholarships.amount_cents) FROM scholarships WHERE scholarships.grant_id = grants.id), 0)".freeze
+
   # Grants that still have unallocated funds (donation amount exceeds the sum of
   # scholarships drawn against them).
-  scope :with_funds_remaining, -> {
-    left_joins(:scholarships)
-      .group(:id)
-      .having("grants.amount_cents - COALESCE(SUM(scholarships.amount_cents), 0) > 0")
+  scope :with_funds_remaining, -> { where("grants.amount_cents > #{ALLOCATED_CENTS_SUBQUERY}") }
```
maebeale, line 24: 🤖 _From Claude:_ Flat `WHERE` + correlated subquery (not `GROUP BY`/`HAVING`) so `will_paginate`'s `total_entries` count stays correct on the paginated index.

app/controllers/grants_controller.rb
```diff
@@ -4,11 +4,16 @@ class GrantsController < ApplicationController
 
   def index
     authorize!
-    @grants = authorized_scope(Grant.all)
+    # The full page renders only the header, filters, and an empty results frame;
+    # the frame's src request (turbo_frame_request?) loads the filtered rows.
+    return render :index unless turbo_frame_request?
```
maebeale: 🤖 _From Claude:_ Full page renders only header/filters/skeleton; the frame's `src` request (`turbo_frame_request?`) returns the filtered rows via `index_lazy`.

app/frontend/javascript/controllers/table_sort_controller.js
maebeale, line 52: needed to update this bc it wasn't sorting dates correctly

maebeale: @jmilljr24 you ok w this?

jmilljr24: This controller is out of my league. Best option is manual testing in as many places as you can.

app/controllers/grants_controller.rb
jmilljr24: This a just a nit, but the pattern we've been using has been
```
    if turbo_frame_request?
      render :grant_results
    else
      render :index
    end
```

Functionally the same but I like keeping a consistent pattern. Not that it matters, but realistically the turbo frame path will be use predominately so having it first feels better.

app/controllers/grants_controller.rb
```diff
@@ -4,11 +4,16 @@ class GrantsController < ApplicationController
...
+
+    @grants = filter_grants(authorized_scope(Grant.all))
                 .includes(:donor, scholarships: { allocation: :allocatable })
                 .by_deadline
                 .page(params[:page])
     track_index_intent(Grant, @grants, params)
+    render :index_lazy
```
jmilljr24: Same idea with this on matching our existing patterns. Looks like this naming convention got introduction on a couple controllers at some point, however I think using the `results` naming convention that was established on `Resource` and `Workshop` makes a bit more sense. Yes, we are hitting the index action but we are really only updating the "results" section on the Index page when `turbo_frame_request`.

jmilljr24 (approved):

app/controllers/grants_controller.rb
maebeale: we have two patterns in the codebase wrt to lazy. i'll spin up a new pr to make the outliers fit also.

PR #1922, opened 2026-06-26
Default new affiliations to first of month and unchecked primary contact

maebeale:
🤖 PR, suggested 👤 review level: 👀 Skim — view-only: form default changes in a shared partial, no logic or data changes

## What & why

New affiliation rows in the people-edit and org-edit forms now default to:

- **Start date = first of the current month** (was today). Affiliations are tracked from the start of the month.
- **Primary org contact = unchecked** (was pre-checked). Most people on an org aren't its primary contact; the pre-check made admins accept a usually-wrong default.

Both are view-only changes in the shared partial `app/views/affiliations/_fields.html.erb`. Existing affiliations still render their stored values; the DB default for `primary_contact` was already `false`.

## Tests

- View spec asserting a newly-built affiliation renders the first-of-month start date and an unchecked primary-contact box.

### DIFF
--- a/app/views/affiliations/_fields.html.erb
+++ b/app/views/affiliations/_fields.html.erb
@@ -69,7 +69,7 @@
                     label_html: { class: "block text-sm font-medium text-gray-700 mb-1" },
                     input_html: {
                       type: "date",
-                      value: (f.object.start_date || (Date.current unless f.object.persisted?))&.strftime("%Y-%m-%d"),
+                      value: (f.object.start_date || (Date.current.beginning_of_month unless f.object.persisted?))&.strftime("%Y-%m-%d"),
                       class: "rounded-md border-gray-300 focus:ring-blue-500 focus:border-blue-500 text-sm",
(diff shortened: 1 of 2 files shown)


app/views/affiliations/_fields.html.erb
```diff
@@ -69,7 +69,7 @@
                     label_html: { class: "block text-sm font-medium text-gray-700 mb-1" },
                     input_html: {
                       type: "date",
-                      value: (f.object.start_date || (Date.current unless f.object.persisted?))&.strftime("%Y-%m-%d"),
+                      value: (f.object.start_date || (Date.current.beginning_of_month unless f.object.persisted?))&.strftime("%Y-%m-%d"),
```
maebeale, line 72: 🤖 _From Claude:_ Both defaults only apply to unpersisted rows (`unless f.object.persisted?` for the date; `primary_contact?` reflects the stored value for existing rows), so editing an existing affiliation is unaffected.

PR #1924, opened 2026-06-29
Standardize all lazy indexes: plural names, skeletons, debounced filters

maebeale:
🤖 suggested review level: 5 Inspect 🔬 wide-reaching: touches 13 admin index pages, incl. filter-form behavior (debounce) and one conditional render

## Why
- Every filterable lazy index should follow ONE shape so the form → frame → response chain is predictable and self-consistent. `stories` is the reference.

## What
**1. Plural naming** — frame tag id = template filename = controller `render` target = count badge id, all `<plural_resource>_results` / `<plural_resource>_count`:
- Renamed the original five index_lazy views (stories, community_news, video_recordings, monthly_reports, bookmarks) → `<x>_results`.
- Normalized the singular outliers: `story_results`→`stories_results`, and frame/template/render/count for **workshops, resources, organizations, payments, allocations** (e.g. `organization_results`→`organizations_results`, `resource_count`→`resources_count`).
- people/users already plural — untouched.

**2. Skeleton placeholder** — added `_results_skeleton.html.erb` to the 8 indexes that lacked one (workshops, resources, organizations, video_recordings, people, payments, allocations, users), each matching its own table/card columns, rendered inside the lazy frame while results load.

**3. Debounced filters** — every index filter form now routes through the `collection` Stimulus controller (400ms text debounce, submit-on-change for selects/checkboxes):
- `monthly_reports` had no `collection` controller at all.
- `video_recordings` submitted immediately on every keystroke (inline `oninput`) — replaced with the shared debounce.

## Anything else
- grants is handled in PR #1920 (already `grants_results`); not on this branch.
- `events` index left out of scope (not a filterable lazy frame in this family).
- `editor_lazy` / `personal_lazy` / `_show_lazy` are different actions — out of scope.
- Frame DOM ids only changed where a name was singular; all references (views, specs, system specs, JS) updated repo-wide, zero stale.

## Tests
- 318 examples across all 13 controllers (request + view + system specs), 0 failures. Rubocop clean.


maebeale:
This makes all indexes have the same setup w turbo, _results, skeletons, debouncing, etc

jmilljr24:
This looks good. I just want to spend some time user testing since it hits so many places in the portal.

maebeale:
@jmilljr24 happy to have you test more! i did check, but would love a 2nd confirmation.

jmilljr24 (approved):

PR #1925, opened 2026-06-29
Add Event ready / Completed readiness filters to registrants roster

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — readiness PORO + one Status column/filter on the registrants roster; no migrations

- Adds a single **Status** column (far right) + matching filter to the registrants roster, so admins see at a glance who needs attention before and after a training without manually cross-checking payment, org/facilitator, scholarship, and CE state.
- One badge, four plain lifecycle steps (chosen over multiple columns because the staff using it aren't technical):
  - 🟠 **Not ready** — pre-event to-dos outstanding; a two-word reason (e.g. "Payment due", "Org validation", "No license #") shows under the badge, full list in the tooltip
  - 🔵 **Ready** — cleared for the event
  - 🟣 **Certificate pending** — attended + tasks done, certificate still to send; a cert-type note ("Registration" / "CE" / "Reg + CE") shows which
  - 🟢 **Completed** — certificate(s) sent

- `EventRegistrationReadiness` PORO computes both checklists + a combined `status`, returning the specific outstanding reasons so the badge tooltip/subtext and the filter share one source of truth. Reads only preloaded associations (no N+1).
- Now that main has the CE + certificate system, readiness reads real state via `Registerable` — `paid_in_full?` and `certificate_sent?` on both the registration and its `continuing_education_registrations` (no more stubs).
- Filtered in Ruby on the loaded roster (no pagination here), mirroring how the existing attendance filter sits outside the active/inactive counts.

### Anything else to add?
- Rebased onto main after the CE/certificate work and the org-"Pending"-chip PR (#1926) landed. The readiness "Org validation" reason and the roster chip now agree on the "no organization linked" rule.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -82,7 +82,7 @@ class EventsController < ApplicationController
     authorize! @event, to: :registrants?
     @event = @event.decorate
     scope = @event.event_registrations
-      .includes(:comments, :organizations, { continuing_education_registrations: [ :professional_license, :allocations ] }, registrant: [ :user, :contact_methods, { avatar_attachment: :blob }, { affiliations: :organization } ])
+      .includes(:comments, :organizations, :allocations, :scholarships, { continuing_education_registrations: [ :professional_license, :allocations ] }, registrant: [ :user, :contact_methods, { avatar_attachment: :blob }, { affiliations: :organization } ])
       .joins(:registrant)
     scope = scope.keyword(params[:keyword]) if params[:keyword].present?
     scope = scope.payment_status(params[:payment_status]) if params[:payment_status].present?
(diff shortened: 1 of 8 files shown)


app/services/event_registration_readiness.rb
```diff
@@ -0,0 +1,118 @@
...
+  def attendance_issue
+    return nil if registration.attended?
+    registration.status == "incomplete_attendance" ? "Attendance incomplete" : "Did not attend"
+  end
+
+  # --- Stubs for flows not yet built. Flip each to the real check as its feature
+  # lands (CE payment, certificate-sent tracking). ---
+
+  # TODO: no CE payment is tracked yet, so nobody is CE-paid.
+  def ce_paid?
```
maebeale, line 187: 🤖 _From Claude:_ These three are deliberate stubs — CE-paid and both `certificate_sent_at` fields don't exist yet. Until they land, "Completed" reads false for everyone (by design). Flip each to the real predicate when its feature merges; completion should require **both** the registration and CE certificate-sent fields.

app/controllers/events_controller.rb
```diff
@@ -103,9 +103,22 @@ def registrants
...
 
-    @event_registrations = scope.order(Arel.sql("people.first_name, people.last_name"))
+    @event_registrations = scope.order(Arel.sql("people.first_name, people.last_name")).to_a
     @dashboard = EventDashboard.new(@event)
 
+    @submitted_org_names = submitted_org_names_for(@event_registrations)
+    @readiness = @event_registrations.to_h do |registration|
+      [ registration.id, EventRegistrationReadiness.new(registration, submitted_org_name: @submitted_org_names[registration.registrant_id]) ]
+    end
+    if params[:event_ready].in?(%w[ ready not_ready ])
```
maebeale: 🤖 _From Claude:_ Readiness is filtered in Ruby on the already-loaded roster (no pagination on this page) rather than as a SQL scope, since the same per-row computation also drives the badge tooltips. It sits after the active/inactive counts, matching how the existing `attendance_status` filter is also excluded from those tab counts.

app/services/event_registration_readiness.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  def completed?
+    completion_issues.empty?
+  end
+
+  # The registration's single lifecycle state for the roster's one Status column
+  # and its matching filter. Completion is the final state and wins; otherwise an
+  # outstanding pre-event to-do means "not ready", and a clear checklist means
+  # "ready" for the event.
+  def status
```
maebeale, line 46: 🤖 _From Claude:_ Precedence is deliberate: completion wins over the pre-event checklist (a finished registrant shouldn't read "Not ready" just because, say, a balance was never reconciled), then an outstanding pre-event item is "not ready", and a clear checklist is "ready". The roster shows one badge from this and the filter matches it word-for-word.

PR #1927, opened 2026-06-29
HOLD: Replace org-index windows-type filter/column with age-range filter

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: a new filter scope + a display fallback, contained to the organizations index

## What & why
Age-range category tags now carry audience info on the organizations index, making the separate windows-type column and "Windows audience" filter redundant.

- Removes the windows-type badge from the **Designations** column and the **Windows audience** filter.
- Adds an **Age range** filter over the 4 AgeRange categories.
- Filter matches an org when **either** its age-range tags (org's own **or** affiliated people's) include the selected range **or** its windows audience semantically covers it — so orgs with a windows audience but no/conflicting tags still surface.
- Surfaces the windows-type badge **in the age-range column** when the windows audience adds age info the tags don't already convey (no tags, or tags that omit a range the audience covers).

## Notes
- `WindowsTypeFilterable` is left intact — it's shared by many other models; only the org index stops using `windows_type_name`.
- Index eager-loads the windows type's age-range categories to avoid an N+1 behind the new badge logic.

## Test plan
- `spec/models/organization_spec.rb` covers `.age_range_names` (org tag / affiliated-person tag / windows-audience fallback / exclusion) and `#windows_audience_unreflected_by_age_tags?`.

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -8,7 +8,8 @@ class OrganizationsController < ApplicationController
     if turbo_frame_request?
       per_page = params[:number_of_items_per_page].presence || 25
       base_scope = authorized_scope(Organization.includes(
-        :windows_type, :organization_status, :sectors, :addresses,
+        :organization_status, :sectors, :addresses,
+        { windows_type: { categorizable_items: { category: :category_type } } },
         { categorizable_items: { category: :category_type } },
         logo_attachment: :blob
       ))
@@ -189,6 +190,7 @@ class OrganizationsController < ApplicationController
 
   def set_index_variables
     @organization_statuses = OrganizationStatus.all
+    @age_range_categories = Category.age_ranges.published.ordered_by_position_and_name
   end
 
   def populations_served
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -91,6 +91,26 @@ class Organization < ApplicationRecord
   scope :organization_ids, ->(organization_ids) { where(id: organization_ids.to_s.split("-").map(&:to_i)) }
   scope :project_ids, ->(project_ids) { where(id: project_ids.to_s.split("-").map(&:to_i)) }
   scope :published, -> { active }
+  # Filters by AgeRange category name(s), powering the unified "Age range" column
+  # on the index. An org matches a selected age range when EITHER it (or an
+  # affiliated person) is tagged with that age range, OR its windows audience
+  # semantically covers it. The windows-type leg keeps orgs that carry a windows
+  # audience but no age-range tags — or tags that disagree with it — in results.
+  scope :age_range_names, ->(names) do
+    return all if names.blank?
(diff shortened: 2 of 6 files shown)


app/models/organization.rb
```diff
@@ -91,14 +91,34 @@ class Organization < ApplicationRecord
...
+
+    category_ids = Category.age_ranges.where("LOWER(categories.name) IN (?)", parsed).select(:id)
+    org_tagged = CategorizableItem.where(categorizable_type: "Organization", category_id: category_ids).select(:categorizable_id)
+    person_tagged = CategorizableItem.where(categorizable_type: "Person", category_id: category_ids).select(:categorizable_id)
+    via_affiliations = Affiliation.where(person_id: person_tagged).select(:organization_id)
+    windows_type_ids = WindowsType.joins(:categorizable_items)
+                                  .where(categorizable_items: { category_id: category_ids })
+                                  .select(:id)
+
+    where(id: org_tagged).or(where(id: via_affiliations)).or(where(windows_type_id: windows_type_ids))
```
maebeale, line 112: 🤖 _From Claude:_ Three-legged OR is the heart of the filter: org-own tag, affiliated-person tag (mirrors the column's aggregation), or windows-audience semantic coverage. All three are plain `where`s on `organizations` with no joins, so `.or` stays structurally compatible.

app/models/organization.rb
```diff
@@ -265,6 +285,17 @@ def all_additional_age_groups
...
   end
 
+  # Whether the windows audience adds age information the org's age-range tags
+  # don't already convey — drives the fallback windows-type badge in the index's
+  # age range column. True when the org has a windows type and either carries no
+  # age-range tags at all, or its tags omit an age range that windows type covers.
+  def windows_audience_unreflected_by_age_tags?
+    return false unless windows_type
+    tagged = all_primary_age_groups + all_additional_age_groups
+    return true if tagged.empty?
```
maebeale, line 295: 🤖 _From Claude:_ Empty tags → `true` is intentional: an org with a windows audience but no age tags is exactly the case we want to surface the badge for. Otherwise we only show it when the audience covers a range the tags omit.

app/controllers/organizations_controller.rb
```diff
@@ -8,7 +8,8 @@ def index
     if turbo_frame_request?
       per_page = params[:number_of_items_per_page].presence || 25
       base_scope = authorized_scope(Organization.includes(
-        :windows_type, :organization_status, :sectors, :addresses,
+        :organization_status, :sectors, :addresses,
+        { windows_type: { categorizable_items: { category: :category_type } } },
         { categorizable_items: { category: :category_type } },
```
maebeale, line 13: 🤖 _From Claude:_ Eager-loads the windows type's age-range categories so `windows_audience_unreflected_by_age_tags?` (which reads `categorizable_items` in Ruby) doesn't N+1 across the 25 rows per page.

PR #1928, opened 2026-06-29
Auto-sync attendance status from onboarding day checkboxes

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — light-logic: small, contained logic change with new model method + tests

On the bulk Onboarding matrix, toggling per-day attendance checkboxes now rolls the attendance status automatically so admins don't set it in two places.

**Rules** (only applied while the status is *active* — `cancelled`/`no_show` are deliberate manual states and are never overridden):

- **0 days checked → `registered`**
- **some but not all days checked → `incomplete_attendance`**
- **all days checked → `attended`**

A one-day event has no partial state (0 → registered, 1 → attended). Derived from `event.day_count`, so it works for 1–5 day events.

When a toggle changes the status, the turbo-stream response also re-renders the attendance badge so the matrix updates live.

### DIFF
--- a/app/models/event_registration.rb
+++ b/app/models/event_registration.rb
@@ -321,6 +321,35 @@ class EventRegistration < ApplicationRecord
     DAY_FIELDS.include?("completed_day_#{day}") && public_send("completed_day_#{day}")
   end
 
+  # How many of the event's in-range days (the first event.day_count of them) are
+  # marked complete on this registration.
+  def completed_day_count
+    DAY_FIELDS.first(event.day_count).count { |field| public_send(field) }
+  end
+
+  # The attendance status implied purely by how many days are marked complete:
+  # none → registered, all → attended, some-but-not-all → incomplete_attendance.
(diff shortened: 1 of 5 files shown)


app/models/event_registration.rb
```diff
@@ -321,6 +321,35 @@ def day_completed?(day)
...
+    "incomplete_attendance"
+  end
+
+  # Realign the attendance status to match the completed-day count after a day
+  # checkbox is toggled on the Onboarding matrix. Only auto-managed while the
+  # registration is in an active status — cancelled and no_show are deliberate
+  # manual states we never override by toggling a day. Returns true if the status
+  # actually changed.
+  def sync_attendance_status_to_days!
+    return false unless status.in?(ACTIVE_STATUSES)
```
maebeale, line 346: 🤖 _From Claude:_ This guard is the key constraint: day toggles only auto-manage the active triad (registered/attended/incomplete_attendance). `cancelled`/`no_show` are set deliberately and must never be clobbered by ticking a day.

PR #1930, opened 2026-07-04
Clear pre-existing CI failures: bulk-payment icon test + bundler-audit CVEs

maebeale:
🤖 suggested review level: 3 Read 📖 spec correction + conservative transitive dep bump; mailer specs verified

Two pre-existing failures on `main` (surfaced on full-suite PR CI, not main's RuboCop-only push CI), fixed here:

### build-and-test — self-contradictory test from #1848
- #1848 correctly restricted the registrants form icon to registration-role forms, so a bulk-payment-only submission renders **no** icon.
- Its accompanying test asserted a "gray icon" the implementation never produces. Flipped it to assert no icon (empty slot reserved), matching the sibling empty-slot test and #1848's intent. Spec-only, no implementation change.

### scan_ruby — bundler-audit CVEs
- `css_parser` 2.2.0→3.0.0 (CVE-2026-53727 — SSRF/local file disclosure in `read_remote_file`)
- `msgpack` 1.8.0→1.8.3 (CVE-2026-54522 — use-after-free)
- Both transitive; bumped conservatively so nothing else moves. `premailer` requires `css_parser >= 1.19.0`, so the major bump is compatible — all 88 mailer specs pass. `bundler-audit` now reports no vulnerabilities.

### Notes for reviewers
- None of this originates from the branch's own feature work — both were red on clean `main`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -168,8 +168,9 @@ GEM
     country_select (11.0.0)
       countries (> 6.0, < 9.0)
     crass (1.0.7)
-    css_parser (2.2.0)
+    css_parser (3.0.0)
       addressable
+      ssrf_filter (~> 1.5)
     csv (3.3.5)
     date (3.5.1)
     debug (1.11.1)
@@ -311,7 +312,7 @@ GEM
     minitest (6.0.6)
       drb (~> 2.0)
       prism (~> 1.5)
-    msgpack (1.8.0)
+    msgpack (1.8.3)
     multi_xml (0.8.1)
       bigdecimal (>= 3.1, < 5)
     mutex_m (0.3.0)
@@ -713,6 +714,7 @@ GEM
       actionpack (>= 4.0)
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
+    ssrf_filter (1.5.0)
     stimulus-rails (1.3.4)
       railties (>= 6.0.0)
     stringio (3.2.0)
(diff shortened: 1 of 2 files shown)


spec/requests/events_spec.rb
```diff
@@ -794,8 +794,8 @@ def submit_agency_name(name)
         get registrants_event_path(event)
 
         expect(response).to have_http_status(:ok)
-        expect(response.body).to include('fa-regular fa-file-lines')
-        expect(response.body).not_to include('fa-solid fa-file-lines')
+        expect(response.body).not_to include('fa-file-lines')
```
maebeale, line 797: 🤖 _From Claude:_ Flipped from asserting a nonexistent gray icon to asserting no icon here — #1848's whole point is that a bulk-payment-only submission must NOT light up the registration icon. Mirrors the sibling empty-slot test, so the reserved `inline-flex w-4 justify-center` slot stays but is empty.

Gemfile.lock
```diff
@@ -168,8 +168,9 @@ GEM
     country_select (11.0.0)
       countries (> 6.0, < 9.0)
     crass (1.0.7)
-    css_parser (2.2.0)
+    css_parser (3.0.0)
```
maebeale, line 171: 🤖 _From Claude:_ Both transitive deps — bumped with `bundle update --conservative css_parser msgpack` so only these two lines move. css_parser 3.0.0 is a major bump but premailer requires only `>= 1.19.0`; verified all 88 mailer specs pass (premailer/css_parser is the email CSS-inlining path).

PR #1932, opened 2026-07-04
Surface onboarding day-attendance checkmarks on the registration edit form

maebeale:
🤖 suggested review level: 3 Read 📖 small view + Stimulus change; day fields already permitted, status derivation mirrors existing model logic

- The per-day attendance checkmarks (Day 1…N) only existed on the event's Onboarding matrix. Surface them on the single-registration edit form too, so an admin can mark days without switching pages.

- Added the day checkboxes beside the status dropdown in `event_registrations/_form.html.erb`. `DAY_FIELDS` were already permitted params, so they save with the form.
- Extended the `attendance-status` Stimulus controller: toggling a day derives the status the same way onboarding does (`registered → incomplete_attendance → attended`), and never overrides a deliberate inactive state (cancelled/no_show/transferred_out) — mirrors `EventRegistration#sync_attendance_status_to_days!`.
- Status still only persists on save ("Unsaved" hint), and the admin can override the dropdown afterward.

### Anything else to add?
- Only admins reach this edit form (`edit?` → `manage?`), matching the admin-only day-field params.
- Covered by new system specs (render/persist, status derivation, inactive-status guard).

### DIFF
--- a/app/frontend/javascript/controllers/attendance_status_controller.js
+++ b/app/frontend/javascript/controllers/attendance_status_controller.js
@@ -4,9 +4,30 @@ import { Controller } from "@hotwired/stimulus"
 // Swaps the colored icon inside the status select as the dropdown changes. The
 // select looks like an ordinary form field (not the roster's autosaving chip),
 // so an "Unsaved" hint is shown until the form is saved.
+//
+// The per-day attendance checkboxes drive the status the same way the Onboarding
+// matrix does: toggling a day rolls the status forward/back (registered →
+// incomplete_attendance → attended), but only while the status is an active one —
+// deliberate manual states (cancelled, no_show, transferred_out) are never
+// overridden. Mirrors EventRegistration#sync_attendance_status_to_days!.
 export default class extends Controller {
-  static targets = ["select", "icon", "dirty"]
-  static values = { colors: Object, icons: Object, initial: String }
+  static targets = ["select", "icon", "dirty", "day"]
+  static values = { colors: Object, icons: Object, initial: String, activeStatuses: Array, dayCount: Number }
+
+  // Recompute the status from how many day checkboxes are checked, then reflect it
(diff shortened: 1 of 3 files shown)


app/frontend/javascript/controllers/attendance_status_controller.js
```diff
@@ -4,9 +4,30 @@ import { Controller } from "@hotwired/stimulus"
...
 export default class extends Controller {
-  static targets = ["select", "icon", "dirty"]
-  static values = { colors: Object, icons: Object, initial: String }
+  static targets = ["select", "icon", "dirty", "day"]
+  static values = { colors: Object, icons: Object, initial: String, activeStatuses: Array, dayCount: Number }
+
+  // Recompute the status from how many day checkboxes are checked, then reflect it
+  // in the dropdown (which re-renders the icon and the unsaved hint via update()).
+  deriveFromDays() {
+    if (!this.activeStatusesValue.includes(this.selectTarget.value)) return
```
maebeale, line 20: 🤖 _From Claude:_ Guards the days→status derivation to active statuses so a deliberate cancelled/no_show/transferred_out isn't silently rolled back by toggling a day — mirrors the server-side `sync_attendance_status_to_days!`.

jmilljr24:
I missed when tracking for days attended was added but one thing to highlight - during one of the stakeholder meetings, it was emphasized that the majority of the time people attend both days so they wanted to avoid having to track individual days.

I'm no sure the exact solution but this does open up more questions that we should ask.

jmilljr24 (approved):
There was one thing in the meeting that came up that I didn't check in this feature.

They mentioned that a person might only attend one day as a make up from another training and should be able to have their status as Attended. I know this change auto selects incomplete if only one day is check. Is that just triggered on the change? If I have a person checked as only day 2, can I manually change their attended status after that to Attended? That would cover the need for now.

PR #1933, opened 2026-07-04
Ignore pay Paddle webhook HMAC advisory (GHSA-mjgf-xj26-9qf9)

maebeale:
🤖 suggested review level: 3 Read 📖 one-line bundler-audit ignore; the judgment call (Paddle path unreachable) is what to check

## Why

`bundler-audit` in CI fails on `GHSA-mjgf-xj26-9qf9` — a non-constant-time HMAC comparison in pay's **Paddle Billing** webhook signature verifier.

- We only run the **Stripe** processor (`config/initializers/pay.rb`); there is no Paddle code anywhere, so the vulnerable path is unreachable.
- **No patched pay release exists** — the advisory flags all versions ≤ 11.6.1 (we're on 11.4.3), so a version bump can't resolve it.

## What

- Add `GHSA-mjgf-xj26-9qf9` to `config/bundler-audit.yml` ignore list, with a comment explaining the reasoning and a note to remove it once upstream ships a fix.

Verified `bin/bundler-audit` now reports **No vulnerabilities found**.

### DIFF
--- a/config/bundler-audit.yml
+++ b/config/bundler-audit.yml
@@ -8,3 +8,8 @@ ignore:
   - CVE-2021-41183
   - CVE-2021-41184
   - CVE-2022-31160
+  # pay: non-constant-time HMAC comparison in the Paddle Billing webhook
+  # signature verifier. We only use the Stripe processor (see config/initializers/pay.rb),
+  # never Paddle, so the vulnerable code path is unreachable. No patched pay
+  # release exists (affects all versions <= 11.6.1). Remove this once upstream ships a fix.
+  - GHSA-mjgf-xj26-9qf9


config/bundler-audit.yml
```diff
@@ -8,3 +8,8 @@ ignore:
   - CVE-2021-41183
   - CVE-2021-41184
   - CVE-2022-31160
+  # pay: non-constant-time HMAC comparison in the Paddle Billing webhook
+  # signature verifier. We only use the Stripe processor (see config/initializers/pay.rb),
+  # never Paddle, so the vulnerable code path is unreachable. No patched pay
+  # release exists (affects all versions <= 11.6.1). Remove this once upstream ships a fix.
+  - GHSA-mjgf-xj26-9qf9
```
maebeale, line 15: 🤖 _From Claude:_ Ignoring by GHSA id (not CVE) because this advisory has no CVE assigned. Justification: we register only Stripe listeners in `config/initializers/pay.rb` and have zero Paddle code, so the vulnerable Paddle webhook verifier is never invoked. Delete this line once pay ships a patched release.

maebeale, line 15: 🤖 _From Claude:_ Ignoring by GHSA id (not CVE) because this advisory has no CVE assigned. We register only Stripe listeners in config/initializers/pay.rb and have no Paddle code, so the vulnerable Paddle webhook verifier is never invoked. Remove once pay ships a patched release.

PR #1934, opened 2026-07-04
Credit stories to any person, not just active users

### LINKED ISSUE #1515: New story: facilitators with unconfirmed emails missing from author dropdown

maebeale:
🤖 suggested review level: 5 Inspect 🔬 data-model change (new author FK), controller create semantics, and author-credit fallback across public/profile display

Fixes #1515.

## Why
The story author dropdown was gated to active users (`User.has_access`), so facilitators with unconfirmed logins (e.g. Courtney Koczka, Lemny Perez) couldn't be credited. Spotlighted-facilitator already pulls from all people; author now does too.

## What
- New nullable `stories.author_id` → Person FK, mirroring `spotlighted_facilitator`. The "Story author" field now lists **all people** and is **required in the form** (model stays optional).
- `created_by` stays a `User` but is now pure audit — set to `current_user` on create, no longer the author picker.
- `AuthorCreditable` gains an overridable `author_person` (default `created_by.person`, unchanged for the 5 other models); `Story` overrides it to `author || created_by.person`, so **existing stories keep their attribution with no backfill**.
- Public show page, admin index "Author" column, and person-profile stories section now reflect the author.
- Admin index **"Author" sort** keys off `author_person` too (COALESCE of author + creator's person), so the column and its sort agree even when author ≠ creator.
- Person profile **stories section** now lists only stories the person **authored or was spotlighted in** — not ones their user merely entered — matching the "Stories authored/featured" heading.
- Converting a story idea pre-fills the author from the idea's submitter.
- Dropped the redundant "Last updated by" name/timestamp block from the story form (still recorded on save; only the form display is removed).

## Not included
- #1516 (author search box) and the first-name-only free-text case — deferred per request.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -50,8 +50,9 @@ class PeopleController < ApplicationController
         @workshop_variations = @person.user&.workshop_variations_as_creator&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
         render partial: "people/sections/workshop_variations", locals: { person: @person, workshop_variations: @workshop_variations }
       when "stories"
-        story_ids = @person.user&.stories_as_creator&.pluck(:id).to_a +
-          @person.stories_as_author.pluck(:id) +
+        # Credit the person for stories they authored or were spotlighted in —
+        # not ones their user merely entered (created_by is a pure audit trail).
+        story_ids = @person.stories_as_author.pluck(:id) +
           @person.stories_as_spotlighted_facilitator.pluck(:id)
         @stories = Story.where(id: story_ids).order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/stories", locals: { person: @person, stories: @stories }
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -50,7 +50,9 @@ class PeopleController < ApplicationController
         @workshop_variations = @person.user&.workshop_variations_as_creator&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
         render partial: "people/sections/workshop_variations", locals: { person: @person, workshop_variations: @workshop_variations }
       when "stories"
-        story_ids = @person.user&.stories_as_creator&.pluck(:id).to_a + @person.stories_as_spotlighted_facilitator.pluck(:id)
+        story_ids = @person.user&.stories_as_creator&.pluck(:id).to_a +
+          @person.stories_as_author.pluck(:id) +
+          @person.stories_as_spotlighted_facilitator.pluck(:id)
         @stories = Story.where(id: story_ids).order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/stories", locals: { person: @person, stories: @stories }
       when "events"
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -9,8 +9,8 @@ class StoriesController < ApplicationController
     if turbo_frame_request?
       per_page = params[:number_of_items_per_page].presence || 12
       base_scope = authorized_scope(Story.includes(:windows_type, :organization, :workshop,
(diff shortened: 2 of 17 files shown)


app/models/story.rb
```diff
@@ -89,6 +90,12 @@ def name
     title
   end
 
+  # AuthorCreditable: prefer the explicitly chosen author, falling back to the
+  # creating user's person so stories predating the author field stay credited.
+  def author_person
+    author || created_by&.person
```
maebeale: 🤖 _From Claude:_ The `author || created_by&.person` fallback is what lets us skip the backfill: pre-existing stories (author_id nil) stay credited to the creating user exactly as before, while new stories use the chosen author.

app/controllers/stories_controller.rb
```diff
@@ -62,6 +62,7 @@ def edit
 
   def create
     @story = Story.new(story_params.except(:category_ids, :sector_ids))
+    @story.created_by = current_user
```
maebeale, line 65: 🤖 _From Claude:_ created_by is now audit-only — always the actual creator, ignoring any submitted `created_by_id` (dropped from strong params). authenticate_user! guarantees current_user here; no public-user fallback exists yet.

app/models/concerns/author_creditable.rb
```diff
@@ -19,8 +19,14 @@ module AuthorCreditable
     "Anonymous" => "anonymous"
   }.freeze
 
+  # The person whose name is credited. Defaults to the creating user's person;
+  # models with a dedicated author association (e.g. Story) override this.
+  def author_person
+    created_by&.person
```
maebeale: 🤖 _From Claude:_ Default kept as `created_by.person` so the 5 other includers (workshops, ideas) are unaffected; only Story overrides `author_person`.

app/controllers/stories_controller.rb
```diff
@@ -174,8 +174,16 @@ def apply_sort(scope, column, direction)
...
-           .reorder(Person.arel_table[:first_name].public_send(dir))
+      # Match Story#author_person, which the column renders: sort by the explicit
+      # author, falling back to the creating user's person. `created_by: :person`
+      # is the second join to `people`, so Rails aliases it `people_users`.
+      creator_person = Person.arel_table.alias("people_users")
+      author_first_name = Arel::Nodes::NamedFunction.new(
+        "COALESCE",
+        [ Person.arel_table[:first_name], creator_person[:first_name] ]
+      )
+      scope.left_joins(:author, created_by: :person)
```
maebeale, line 185: 🤖 _From Claude:_ Sort now mirrors `Story#author_person` (explicit author, falling back to the creator's person) so the column and its sort key stay in agreement even when they differ. `people_users` is the alias Rails deterministically gives the second join to `people` — the author join takes the default `people`.

app/controllers/people_controller.rb
```diff
@@ -50,7 +50,10 @@ def show
         @workshop_variations = @person.user&.workshop_variations_as_creator&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
         render partial: "people/sections/workshop_variations", locals: { person: @person, workshop_variations: @workshop_variations }
       when "stories"
-        story_ids = @person.user&.stories_as_creator&.pluck(:id).to_a + @person.stories_as_spotlighted_facilitator.pluck(:id)
+        # Credit the person for stories they authored or were spotlighted in —
+        # not ones their user merely entered (created_by is a pure audit trail).
+        story_ids = @person.stories_as_author.pluck(:id) +
```
maebeale, line 55: 🤖 _From Claude:_ Dropped `stories_as_creator` here: `created_by` is now a pure audit trail, so a story only belongs on a person's profile if they authored it or were spotlighted in it.

PR #1935, opened 2026-07-05
Credit workshops to any person, not just active users

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new author FK + create-semantics + author-credit fallback + author-name search across index/profile

Applies the story author change from #1934 to workshops, plus author-name search for both.

- The workshop "Author" dropdown listed **Users**, so facilitators without a confirmed login couldn't be credited. Author is now a **Person** chosen from all people — mirroring the story change.

- New nullable `workshops.author_id` → Person FK. The admin "Author" field now lists **all people** and **pre-selects the creator's person** (still overridable), so a blank submission credits the creator without forcing a choice.
- `created_by` stays a `User` but is now pure audit — set to `current_user` on create, no longer the author picker.
- `AuthorCreditable#author_person` (default `created_by.person`) is overridden by the `author` association when present, so **existing workshops keep their attribution with no backfill**. `author_credit`, the decorator byline, and the index "By:" line all follow it.
- Person profile **"Workshops authored"** section now lists workshops the person authored (via `author_id`); person deletion is blocked when they've authored workshops.
- Converting a workshop idea pre-fills the author from the idea's submitter.

### Author-name search (both workshops and stories)
- `WorkshopSearchService#search_by_author_name` now also matches the explicit **author** Person (aliased `author_people` join), alongside `full_name` and the creator's person.
- Stories gain a `by_credited_person_name` scope OR-ed into the free-text query via id subqueries (SearchCop can't join the `people` table twice), so story search matches the credited author or creator by name.

### Anything else to add?
- Author is **optional** on the form (defaults to the creator); the model column stays nullable.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -44,7 +44,9 @@ class PeopleController < ApplicationController
 
       case section
       when "workshops"
-        @workshops = @person.user&.workshops&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
+        # Credit the person for workshops they authored — not ones their user
+        # merely created (created_by is a pure audit trail).
+        @workshops = @person.workshops_as_author.order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/workshops", locals: { person: @person, workshops: @workshops }
       when "workshop_variations"
         # Credit the person for variations they authored — not ones their user
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -16,7 +16,7 @@ class WorkshopsController < ApplicationController
       track_index_intent(Workshop, search_service.workshops, params)
 
       @workshops = authorized_scope(search_service.workshops
-                                                  .includes(:categories, :windows_type, :bookmarks,
+                                                  .includes(:categories, :windows_type, :bookmarks, :author,
                                                             created_by: [ :person ], primary_asset: [ :file_attachment ]))
                                                   .paginate(page: params[:page], per_page: params[:per_page] || 12)
 
@@ -198,6 +198,7 @@ class WorkshopsController < ApplicationController
     potential_series = potential_series.where.not(id: @workshop.id) if @workshop.persisted?
     @potential_series_workshops = authorized_scope(potential_series).order(:title)
     @windows_types = WindowsType.all
+    @people = Person.order(Arel.sql("LOWER(first_name), LOWER(last_name)"))
     @workshop_ideas = authorized_scope(WorkshopIdea.order(created_at: :desc))
                                   .map { |wi|
                                     [ "#{wi.created_at.strftime("%Y-%m-%d")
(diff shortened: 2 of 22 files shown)


app/models/concerns/author_creditable.rb
```diff
@@ -19,8 +19,14 @@ module AuthorCreditable
     "Anonymous" => "anonymous"
   }.freeze
 
+  # The person whose name is credited: the explicitly chosen author when the
+  # model has one (e.g. Story, Workshop), otherwise the creating user's person.
+  def author_person
+    (author if respond_to?(:author)) || created_by&.person
```
maebeale, line 25: 🤖 _From Claude:_ Generic `respond_to?(:author)` so only Workshop (the sole includer with an `author` association on this branch) uses it; the other includers fall back to `created_by.person` unchanged.

app/controllers/workshops_controller.rb
```diff
@@ -225,7 +226,7 @@ def log_workshop_error(action, error)
   def workshop_params
     params.require(:workshop).permit(
       :title, :featured, :published,
```
maebeale, line 228: 🤖 _From Claude:_ Dropped `:created_by_id` from permitted params; `current_user.workshops.build` sets `created_by` so it's now a pure audit trail regardless of any submitted value.

app/controllers/people_controller.rb
```diff
@@ -44,7 +44,9 @@ def show
 
       case section
       when "workshops"
-        @workshops = @person.user&.workshops&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
+        # Credit the person for workshops they authored — not ones their user
+        # merely created (created_by is a pure audit trail).
+        @workshops = @person.workshops_as_author.order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
```
maebeale, line 49: 🤖 _From Claude:_ Now credits by `author_id`, not the creating user. Legacy workshops (no `author_id`) won't appear here — matches the story change and the "Workshops authored" heading.

app/models/story.rb
```diff
@@ -78,6 +101,14 @@ def self.search_by_params(params)
       stories = self.search(conditions)
     end
 
+    # SearchCop's free-text query only covers title + body. Also match the
+    # credited author/creator by name, OR-ed in via id subqueries so the extra
+    # `people` joins stay isolated from SearchCop's own joins.
+    if params[:query].present?
+      stories = self.where(id: stories.select("stories.id"))
```
maebeale, line 108: 🤖 _From Claude:_ SearchCop can't join `people` twice (it constantizes the table name), so author-name matching is a plain scope OR-ed in via id subqueries. `.select("stories.id")` is qualified because the joins make a bare `id` ambiguous. Search order is discarded by the controller's `apply_sort`, so dropping SearchCop's relevance order here is harmless.

app/services/workshop_search_service.rb
```diff
@@ -156,7 +156,10 @@ def search_by_author_name(workshops, author_name)
     return workshops if author_name.blank?
 
     sanitized = author_name.strip.gsub(/\s+/, "")
+    # `created_by: :person` joins `people` (the creator's person); the explicit
+    # author is a second join to the same table, so alias it `author_people`.
     workshops.left_outer_joins(created_by: :person)
+             .joins("LEFT OUTER JOIN people author_people ON author_people.id = workshops.author_id")
```
maebeale: 🤖 _From Claude:_ `created_by: :person` already occupies the `people` alias, so the explicit author is joined as `author_people` to avoid the collision. Mirrors the story `by_credited_person_name` scope.

PR #1936, opened 2026-07-05
Credit workshop variations to any person, not just active users

maebeale:
🤖 suggested review level: 3 Read 📖 mirrors the merged Story author change (#1934) — new author FK + credit fallback across form, index, profile

## Why
The variation author dropdown was gated to Users, so facilitators without a login couldn't be credited — same gap fixed for stories in #1934. Author is now a **Person** chosen from all people.

## What
- New nullable `workshop_variations.author_id` → Person FK. The admin "Variation author" picker now lists **all people**.
- `created_by` stays a `User`, now pure audit — set to `current_user` on create (via the `workshop_variations_as_creator` build), no longer the author picker.
- `AuthorCreditable#author_person` already prefers an explicit `author` over `created_by.person`, so **existing variations keep their attribution with no backfill**.
- Index "Author" column and the show page reflect the credited author.
- Person profile "Workshop variations authored" section now lists variations the person **authored**, not ones their user merely entered.
- Promoting a variation idea pre-fills the author from the idea's submitter.
- `PersonPolicy` blocks deleting a person who authored a variation.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -47,7 +47,9 @@ class PeopleController < ApplicationController
         @workshops = @person.user&.workshops&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
         render partial: "people/sections/workshops", locals: { person: @person, workshops: @workshops }
       when "workshop_variations"
-        @workshop_variations = @person.user&.workshop_variations_as_creator&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
+        # Credit the person for variations they authored — not ones their user
+        # merely entered (created_by is a pure audit trail).
+        @workshop_variations = @person.workshop_variations_as_author.order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/workshop_variations", locals: { person: @person, workshop_variations: @workshop_variations }
       when "stories"
         # Credit the person for stories they authored or were spotlighted in —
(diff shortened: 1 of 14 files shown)


app/controllers/workshop_variations_controller.rb
```diff
@@ -110,7 +112,7 @@ def set_form_variables
 
   def workshop_variation_params
```
maebeale, line 113: 🤖 _From Claude:_ `created_by_id` is intentionally dropped from permitted params — `create` sets `created_by` to `current_user` via `workshop_variations_as_creator.build`, making it a pure audit trail while `author_id` becomes the credit picker.

app/controllers/people_controller.rb
```diff
@@ -47,7 +47,9 @@ def show
         @workshops = @person.user&.workshops&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
         render partial: "people/sections/workshops", locals: { person: @person, workshops: @workshops }
       when "workshop_variations"
-        @workshop_variations = @person.user&.workshop_variations_as_creator&.order(created_at: :desc)&.paginate(page: params[:page], per_page: per_page) || []
+        # Credit the person for variations they authored — not ones their user
+        # merely entered (created_by is a pure audit trail).
+        @workshop_variations = @person.workshop_variations_as_author.order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
```
maebeale, line 52: 🤖 _From Claude:_ Now keyed off `workshop_variations_as_author` (matches the "authored" heading). Legacy variations with a null `author_id` no longer appear on the creator user's profile — same no-backfill tradeoff accepted for stories in #1934.

PR #1937, opened 2026-07-05
Credit community news to any person, not just active users

maebeale:
🤖 suggested review level: 5 Inspect 🔬 data-model change (rename + new person FK) and shared author-credit fallback across form/show/index

Ports #1934 (stories) to community news: the author is now a **Person**, separate from the audit `created_by` User.

## Why
The "Display author" dropdown was gated to `User.has_access`, so facilitators without a confirmed login couldn't be credited. Author is now a **Person** chosen from all people.

## What — no backfill
- Legacy `author_id` (User, has_access pick) is **renamed to `user_author_id`** (kept nullable, users FK preserved).
- New nullable **`author_id` → people** added; **`author_credit_preference`** column added (default `full_name`).
- `CommunityNews` `include`s **`AuthorCreditable`** and overrides `author_person` to credit: explicit person `author` → legacy `user_author.person` → `created_by.person`. Existing rows keep their attribution with **zero data migration**.
- `created_by` is forced to `current_user` on create (audit; permitted but not spoofable).
- **Author credit-preference picker** added to the form (same admin options as stories).
- Show page + index column render `author_credit`; show links the credited person when searchable.
- Index **"Author" sort** keys off `author_person` via a **COALESCE** across the same fallback chain, so column and sort always agree.
- `Person#community_news_as_author` (`restrict_with_error`) blocks deleting a person who authored community news.

## Tests
- `it_behaves_like "author_creditable"` + explicit `#author_person` fallback specs; create/audit semantics; COALESCE sort; delete-protection; picker present in form.

Rebased onto main (resolved additive conflicts with #1936 workshop-variations author work).


db/migrate/20260705005223_change_community_news_author_to_person.rb
```diff
@@ -0,0 +1,29 @@
...
+  # author_id at people, backfilling from the previously-selected user's person
+  # so existing attribution is preserved. created_by stays a User audit trail.
+  def up
+    remove_foreign_key :community_news, :users, column: :author_id if foreign_key_exists?(:community_news, :users, column: :author_id)
+    change_column_null :community_news, :author_id, true
+    execute <<~SQL.squish
+      UPDATE community_news cn
+      JOIN users u ON u.id = cn.author_id
+      SET cn.author_id = u.person_id
+    SQL
```
maebeale: 🤖 _From Claude:_ Backfill maps each row to the previously-selected user's `person_id` (not left null), so existing attribution carries over. `author_id` was `integer` (users FK); widened to bigint to match `people.id`. `down` re-points to `created_by` so the users FK is valid again on rollback.

app/models/community_news.rb
```diff
@@ -25,7 +25,8 @@ class CommunityNews < ApplicationRecord
   has_many :sectors, through: :sectorable_items
 
   # Validations
-  validates :author_id, presence: true
+  # author is required in the form; left optional at the model so legacy rows
+  # whose author-user had no person survive the person backfill.
   validates :title, presence: true, length: { maximum: 150 }
```
maebeale, line 33: 🤖 _From Claude:_ Dropped the model-level `author_id` presence validation (form still requires it) so rows whose author-user had no person survive the backfill as null — mirrors how Story keeps author optional at the model.

app/controllers/community_news_controller.rb
```diff
@@ -60,6 +60,7 @@ def edit
 
   def create
     @community_news = CommunityNews.new(community_news_params)
+    @community_news.created_by = current_user
     authorize! @community_news
```
maebeale, line 76: 🤖 _From Claude:_ `created_by` is now forced to `current_user` here and removed from permitted params, so it can't be spoofed — pure audit trail, matching #1934.

db/migrate/20260705011531_add_person_author_to_community_news.rb
```diff
@@ -0,0 +1,33 @@
...
+  # that legacy pick as user_author_id and add a nullable person author_id
+  # instead — no backfill: AuthorCreditable#author_person falls back through
+  # user_author.person then created_by.person for existing rows.
+  def up
+    # Drop and recreate the users foreign key around the rename so its
+    # auto-generated constraint name follows user_author_id, freeing the old
+    # name for the new person author reference.
+    remove_foreign_key :community_news, :users, column: :author_id
+    rename_column :community_news, :author_id, :user_author_id
+    # New rows credit a person and leave no legacy user author, so it's nullable.
```
maebeale, line 13: 🤖 _From Claude:_ No backfill: the legacy has_access pick is preserved as `user_author_id` (made nullable — new rows only set the person `author`), and `AuthorCreditable#author_person` falls back through it. The users FK is dropped/recreated around the rename so its auto-generated constraint name follows the new column, freeing the old name for the person `author` reference.

app/models/community_news.rb
```diff
@@ -25,10 +28,18 @@ class CommunityNews < ApplicationRecord
   has_many :sectors, through: :sectorable_items
 
   # Validations
-  validates :author_id, presence: true
+  # author is required in the form; left optional at the model so legacy rows
+  # (created before author became a person) remain valid.
   validates :title, presence: true, length: { maximum: 150 }
   validates :rhino_body, presence: true
```
maebeale, line 35: 🤖 _From Claude:_ Overrides `AuthorCreditable#author_person` (which defaults to `author || created_by.person`) to slot the legacy `user_author.person` in between — otherwise existing rows where an admin credited someone else would silently re-credit to the admin (`created_by`).

PR #1938, opened 2026-07-05
Credit resources to any person, not just active users

maebeale:
🤖 suggested review level: 5 Inspect 🔬 data-model change — column rename + new author FK, credit fallback, and a new person-profile section behind a preference flag

Parity with stories (#1934) and workshop variations (#1936), applied to resources.

## Why
The resource "author" was a free-text string (its old users-only picker was dead code, removed in #985), so a resource couldn't be credited to a real, linkable person.

## What
- New nullable `resources.author_id` → **Person** FK + `author_credit_preference`; `Resource` now `include`s `AuthorCreditable`. `created_by` stays a `User` (pure audit — set to `current_user` on create).
- Renamed the old free-text `resources.author` → **`legacy_author_name`** (no backfill). It's shown as a fallback credit for old resources and surfaced read-only under the picker in the form, so no existing attribution is lost.
- Resource form now lists **all people** for the author (searchable select), defaulting to the current user's person, plus a credit-preference dropdown. Author stays optional.
- Public show page credits the person via `author_credit`, links to searchable profiles, and falls back to `legacy_author_name` then the creator.
- `Person#resources_as_author` (`restrict_with_error`) blocks deleting a person credited on a resource; `PersonPolicy#has_associated_data?` respects it.
- SearchCop now searches `legacy_author_name` in place of the removed `author` column.
- **New "Resources authored" person-profile section**, behind a new `profile_show_resources` preference (default on) — mirrors the stories/workshop-variations sections so a credited person can see and share their resources.

## Notes
- Rebased on main after #1935 (workshops) and #1937 (community news) landed the same pattern; resolved the shared `person.rb` / `person_policy.rb` / schema edits so all four author associations coexist.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -60,6 +60,11 @@ class PeopleController < ApplicationController
           @person.stories_as_spotlighted_facilitator.pluck(:id)
         @stories = Story.where(id: story_ids).order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/stories", locals: { person: @person, stories: @stories }
+      when "resources"
+        # Credit the person for resources they authored — not ones their user
+        # merely entered (created_by is a pure audit trail).
+        @resources = @person.resources_as_author.order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
+        render partial: "people/sections/resources", locals: { person: @person, resources: @resources }
       when "events"
         @event_registrations = @person.event_registrations.active.includes(:event).order("events.start_date DESC").references(:events).paginate(page: params[:page], per_page: per_page)
         render partial: "people/sections/events", locals: { person: @person, event_registrations: @event_registrations }
@@ -480,6 +485,7 @@ class PeopleController < ApplicationController
       :profile_show_workshops,
       :profile_show_workshop_ideas,
       :profile_show_workshop_logs,
+      :profile_show_resources,
       :member_since,
       :linked_in_url,
       :facebook_url,
--- a/app/decorators/resource_decorator.rb
+++ b/app/decorators/resource_decorator.rb
@@ -13,7 +13,23 @@ class ResourceDecorator < ApplicationDecorator
   end
 
   def truncated_author
-    h.truncate author, length: 20
+    h.truncate credited_author_name, length: 20
+  end
+
+  # Name shown on the author/credit line: the chosen person's credit when a
+  # person author is set, otherwise the preserved legacy free-text author,
+  # otherwise the creating user's person credit (AuthorCreditable fallback).
+  def credited_author_name
+    return author_credit if author.present?
+    legacy_author_name.presence || author_credit
+  end
+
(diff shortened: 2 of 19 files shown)


db/migrate/20260705013635_add_author_to_resources.rb
```diff
@@ -0,0 +1,15 @@
+class AddAuthorToResources < ActiveRecord::Migration[8.1]
+  def up
+    # Preserve the old free-text author string as a legacy field; the new
+    # author is a Person reference, mirroring stories and workshop variations.
+    rename_column :resources, :author, :legacy_author_name if column_exists?(:resources, :author)
```
maebeale, line 5: 🤖 _From Claude:_ Rename runs first so the string column frees up the `author` name, then `add_reference :author` creates `author_id`. `down` reverses in the opposite order. Guards make both directions idempotent.

app/decorators/resource_decorator.rb
```diff
@@ -13,7 +13,23 @@ def kind_display
...
+
+  # Name shown on the author/credit line: the chosen person's credit when a
+  # person author is set, otherwise the preserved legacy free-text author,
+  # otherwise the creating user's person credit (AuthorCreditable fallback).
+  def credited_author_name
+    return author_credit if author.present?
+    legacy_author_name.presence || author_credit
+  end
+
+  # The person to link the credit to, when there is a linkable one. A legacy
```
maebeale, line 27: 🤖 _From Claude:_ Three-way fallback preserves old behavior: chosen person → preserved legacy free-text → creating user’s person (the AuthorCreditable default). Legacy resources with only a string keep displaying it.

app/views/resources/_form.html.erb
```diff
@@ -56,6 +56,39 @@
...
   </div>
 
+  <!-- Author -->
+  <div class="flex flex-col md:flex-row md:space-x-4 mb-6">
+    <div class="flex-1 mb-4 md:mb-0" data-controller="searchable-select" data-searchable-select-mode-value="person">
+      <%= f.input :author_id,
+                  as: :select,
+                  collection: @people.map { |p| [ p.full_name_with_email, p.id ] },
+                  include_blank: "Select an author",
+                  selected: (f.object.author_person&.id || current_user.person_id),
```
maebeale, line 66: 🤖 _From Claude:_ Defaults the picker to the current user’s person on new resources (matches the workshop-variations form). Author stays optional in the model, so blank is allowed and falls back to the creator’s person.

app/controllers/people_controller.rb
```diff
@@ -60,6 +60,11 @@ def show
           @person.stories_as_spotlighted_facilitator.pluck(:id)
         @stories = Story.where(id: story_ids).order(created_at: :desc).paginate(page: params[:page], per_page: per_page)
```
maebeale, line 61: 🤖 _From Claude:_ Credits the person for resources they *authored*, not ones their user merely entered — `created_by` is pure audit, same as the stories/variations sections above.

PR #1939, opened 2026-07-05
Capture sector "Other" free-text as records; curate (promote/keep/dismiss)

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new polymorphic model + capture across 3 submission paths + org-type capture + admin curation reachable from each person's chips

## Why
Free-text "Other" answers people type on forms were derived on the fly and shown as read-only chips — with nowhere to record a curation decision, and no way to hide one from the public profile. This captures the **promotable-track** "Other"s as records so admins can promote / keep / dismiss them, from the review queue or from a person's page.

## What
- **`OtherResponse` model** — polymorphic `owner`, because the answer's subject differs by question:
  - **`sector`** → owned by the **Person**; promotable into a `Sector`, shown on the profile/edit chip.
  - **`organization_type`** → owned by the **Organization**; stored now, **not promotable** until `OrganizationType` is a model (its promote button is hidden).
  - **`generic`** questions are **not captured** — that data stays searchable in the form answers. `field_identifier` is retained so any question can be switched on later.
  - Statuses: `pending` → `promoted` / `kept` / `dismissed`.
- **Capture at submission time**: person-owned "Other"s via shared `OtherResponses::CaptureFromSubmission` (registration, scholarship, bulk payment); org-type via `PublicRegistration#sync_agency_type` (where the org is known). De-duped per owner + question. No backfill.
- **Profile + edit** show the sector chips; admins **deep-link into the review queue** from a chip (anchored, with a back link to the person) to promote/keep/dismiss; edit keeps a one-click **×** dismiss. Submission view untouched (audit record).
- **Review page** (`/other_responses`): groups by kind/question with a count, filterable All / Pending / Kept. **Promote** (sectors only) maps a group to an existing sector or mints a new published one; **Keep all** / **Dismiss all** for any group (incl. org-type).

## Decisions baked in
- Dismissed wins; new promoted sector is **published**; org-owned `agency_type` stays on the org too (#1875).
- Policy rules spelled out explicitly (`index?`/`update?`/`curate?`/`promote?`, all `admin?`).

## Tests
Model (kind derivation incl. org-type, polymorphic owner, promotability, anchors, uniqueness), capture service (person/sector), org-type capture in registration, profile/edit chip + dismiss + deep-link, review page (sector/generic/org-type groups, filter, curate by kind, promote incl. new-sector minting, eyebrow return).

## Rebased on main
Resolved conflicts with #1872 (age-range editor); coexists with #1875 (org `agency_type_other`).

## Reporting (event background page)
Pulling "Other" out of real sector tags left the background report with **no** "Other" signal. Restored in **two roles**, mirroring the profile (sector chips + separate `(other)` chips):
- **Aggregate** "Other" slice in the **All sectors** chart (how many chose it), pinned last, drilling into those registrants. The **Primary** chart gets none — "Other" is an additional-sector answer only.
- **Separate "Other sector responses" detail card** listing the actual typed values (grouped by normalized text, count each), each drilling into the registrants who wrote it. Kept out of the pie on purpose: they're un-curated free text, would double-count, and make a useless long-tail.


app/services/event_registration_services/public_registration.rb
```diff
@@ -413,9 +414,30 @@ def update_form_submission(person)
...
       save_form_answers(submission)
+      capture_other_sector_responses(submission)
       submission
     end
 
+    # Materialize the free-text "Other" sector answers as OtherResponse records so
+    # they can be curated (promoted/kept/dismissed). Reuses OtherOption.texts, the
+    # same extraction used to display them, and de-dupes on the person's normalized
+    # value so a repeat registration of the same text doesn't create a second row.
+    def capture_other_sector_responses(submission)
```
maebeale: 🤖 _From Claude:_ Capture reuses `OtherOption.texts` (the same extraction the display used) and `find_or_create_by` on the person + normalized value, so a repeat registration of the same "Other" text doesn't create a second row or trip the unique index.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # Promote every non-dismissed person who typed this value into a real Sector
+  # tag — mapping to an existing sector or minting a new (published) one — and
+  # mark those responses promoted so they stop showing as free-text chips.
+  def promote
+    authorize! to: :promote?
+    sector = target_sector
+    return redirect_to other_responses_path, alert: "Pick or name a sector to promote to." unless sector
+
+    responses = OtherResponse.sectors.promotable_now
```
maebeale: 🤖 _From Claude:_ `promotable_now` excludes `dismissed` — so "dismissed wins": promoting a value never re-tags someone who deliberately hid it.

app/views/people/_other_sector_responses.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    Renders bare chips — the caller supplies the surrounding flex container. %>
+<% responses ||= [] %>
+<% dismissable = local_assigns.fetch(:dismissable, false) %>
+<% responses.each do |response| %>
+  <span class="inline-flex items-center rounded-md border border-gray-300 bg-white text-gray-600 px-3 py-1 text-sm font-medium"
+        title="Other response entered during registration">
+    <%= response.text %>
+    <span class="ml-1 text-xs text-gray-400">(other)</span>
+    <% if dismissable %>
+      <%= link_to "×",
```
maebeale, line 24: 🤖 _From Claude:_ Dismiss is a `link_to` with `turbo_method: :patch`, not `button_to` — this chip lives inside the person edit `<form>`, and a nested `<form>` (what `button_to` emits) is invalid HTML.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        status_counts: rows.each_with_object(Hash.new(0)) { |r, h| h[r.status] += 1 }
+      }
+    end.sort_by { |group| [ -group[:count], group[:display_text].downcase ] }
+
+    @sectors = Sector.excluding_other.order(:name)
+  end
+
+  # Bulk keep/dismiss every visible person who typed this value, from the review
+  # queue. Keep leaves it as a free-text chip; dismiss hides it from profiles.
+  def curate
```
maebeale, line 26: 🤖 _From Claude:_ Bulk keep/dismiss acts on the whole grouped value (all `VISIBLE_STATUSES` rows for this `normalized_text`), mirroring how `promote` fans out across everyone who typed it. `return_status` round-trips the active filter so the queue stays put after the action.

maebeale:
what we have now on profile w form submission data is super brittle. this would allow us to capture all form 'other' answer data and decide if it gets promoted, wiped, or kept as auxiliary data.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    authorize! to: :update?
```
jmilljr24: `update?` is not defined in `other_response_policy.rb` so it will default to `application_policy`. While this will work as it hits the admin check, we are explicit in every other policy instead of falling back to `application_policy`

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    else
+      redirect_to other_responses_path, status: :see_other
+    end
+  end
+
+  # Promote every non-dismissed person who typed this value into a real Sector
+  # tag — mapping to an existing sector or minting a new (published) one — and
+  # mark those responses promoted so they stop showing as free-text chips.
+  def promote
+    authorize! to: :promote?
```
jmilljr24:
```suggestion
    authorize!
```

app/policies/other_response_policy.rb
```diff
@@ -0,0 +1,8 @@
+class OtherResponsePolicy < ApplicationPolicy
+  # Curating free-text "Other" responses (reviewing, promoting, keeping,
+  # dismissing) is an admin-only task. index?/update? fall back to manage?.
```
jmilljr24: Doesn't make sense. yes is falls back to manage in application policy but that is admin only as well

```
  def manage?
    admin?
  end
```

app/controllers/other_responses_controller.rb
maebeale: 🤖 _From Claude:_ Fixed in 65d48db → 8933c0a. `other_response_policy.rb` now spells out `index?`/`update?`/`curate?`/`promote?` explicitly (all `admin?`), and `curate` uses plain `authorize!` so it resolves `curate?` instead of borrowing `update?`. Nothing falls back to `application_policy` anymore.

app/controllers/other_responses_controller.rb
maebeale: 🤖 _From Claude:_ Applied — your suggestion is in (`promote` is now plain `authorize!`, resolving the explicit `promote?` rule). Did the same for `curate`.

app/policies/other_response_policy.rb
maebeale: 🤖 _From Claude:_ Agreed it was inconsistent — a lone `promote?` while everything else leaned on the inherited `manage?`. Rewrote the policy to define every rule the controller authorizes explicitly (`index?`/`update?`/`curate?`/`promote?`, each `admin?`), matching the other policies, and added non-admin rejection specs for `curate` and `promote`.

jmilljr24 (approved):

app/services/other_responses/capture_from_submission.rb
```diff
@@ -0,0 +1,54 @@
...
+  # box are ignored. De-dupes on the person's normalized value per question, so
+  # re-submitting the same answer never creates a second row.
+  #
+  # Shared by the registration, scholarship, and bulk-payment submission paths.
+  class CaptureFromSubmission
+    # Fields whose "Other" belongs to the organization, not the person, so they
+    # must not land in the person's review queue. Organization type is synced
+    # onto the org (PublicRegistration#sync_agency_type) and will get its own
+    # promotable path when OrganizationType becomes a model.
+    EXCLUDED_FIELD_IDENTIFIERS = %w[agency_type].freeze
```
maebeale: 🤖 _From Claude:_ `agency_type` is answered on the reg form and its answer IS saved into the person's form_answers (that's how #1875 reads it), so a naive "capture all fields" would slurp the org's "Other: …" into the person's review queue. Excluded here; it gets its own org-scoped, promotable path when OrganizationType lands.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  private
+
+  def set_other_response
+    @other_response = OtherResponse.find(params[:id])
+  end
+
+  # Group key: sector "Other"s share one bucket (kind), every other question is
+  # keyed by its field so distinct questions never merge.
+  def group_bucket(response)
```
maebeale: 🤖 _From Claude:_ Sector "Other"s from different sector fields (additional_sectors vs the legacy service-area names) bucket together since they all promote into the one Sector catalog; every other question stays keyed by its own field so distinct questions never merge. curate/promote scope the same way (`group_scope`).

app/services/other_responses/capture_from_submission.rb
```diff
@@ -0,0 +1,58 @@
...
+          capture(field_identifier, text, answer)
+        end
+      end
+    end
+
+    private
+
+    # Capture only the questions whose "Other" is (or will be) promotable — the
+    # rest stay searchable in the form answers.
+    def capturable?(field_identifier)
```
maebeale, line 39: 🤖 _From Claude:_ Only promotable questions are captured — no point storing an "Other" that can never become a tag (it stays searchable in the form answers). `field_identifier` is still recorded, so enabling a new question later is just adding its kind to `PROMOTABLE_KINDS` + `kind_for`.

app/services/event_registration_services/public_registration.rb
```diff
@@ -205,6 +205,19 @@ def sync_agency_type(organization)
       return if label.blank?
       other_text = FormField.other_option?(label) ? specified.strip.presence : nil
       organization.update!(agency_type: label, agency_type_other: other_text)
+      capture_organization_type_other(organization, other_text)
+    end
+
+    # Materialize the org-type "Other" as an OtherResponse owned by the org, so it
+    # joins the curation queue alongside sector "Other"s. Not promotable yet (no
+    # OrganizationType model), but stored now so nothing is lost; de-duped per org.
+    def capture_organization_type_other(organization, text)
```
maebeale, line 226: 🤖 _From Claude:_ Org-type "Other" is captured here rather than in CaptureFromSubmission because this is where the organization (the polymorphic owner) is known — the submission only knows the person. Stored now but `promotable? == false`, so the review queue shows keep/dismiss for it and hides Promote until OrganizationType exists.

app/views/events/background.html.erb
```diff
@@ -2,6 +2,13 @@
 
 <% primary_sector_data = @dashboard.primary_sectors.map { |s| [ s.name, @dashboard.primary_sector_counts.fetch(s.id, 0) ] }.sort_by { |_, count| -count } %>
 <% sector_data = @dashboard.sectors.map { |s| [ s.name, @dashboard.sector_counts.fetch(s.id, 0) ] }.sort_by { |_, count| -count } %>
+<%# Free-text "Other" sectors ride along as one aggregate slice, pinned last (the
+    actual typed values get their own detail card below). The primary chart has no
+    such slice — "Other" is an additional-sector answer only. %>
+<% sector_data += [ [ "Other", @dashboard.other_sector_response_count ] ] if @dashboard.other_sector_response_count.positive? %>
```
maebeale, line 8: 🤖 _From Claude:_ "Other" appears in two roles: one aggregate slice here (the count) and a separate detail card below (the typed values). Kept apart on purpose — folding un-curated free-text into the sector pie would double-count and make a count-1 long tail. Primary chart gets no slice since "Other" is additional-only.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+  def promote
+    authorize!
+    sector = target_sector
+    return redirect_to other_responses_path, alert: "Pick or name a sector to promote to." unless sector
+
+    responses = group_scope.sectors.promotable_now
+    responses.includes(:owner).find_each do |response|
+      # Reuse registration's tagging so the sector lands on the person AND their
+      # organizations, always as an additional tag (never primary).
+      SectorTagging.apply(person: response.owner, organizations: response.owner.organizations,
```
maebeale: 🤖 _From Claude:_ Promotion tags the person's **current** organizations (via affiliations), since the OtherResponse doesn't record which specific org they registered under. If someone is affiliated with multiple orgs, the promoted sector lands on all of them — consistent with how orgs aggregate their members' sectors.

app/controllers/other_responses_controller.rb
```diff
@@ -0,0 +1,127 @@
...
+  def promote
+    authorize!
+    sector = target_sector
+    return redirect_to other_responses_path, alert: "Pick or name a sector to promote to." unless sector
+
+    responses = group_scope.sectors.promotable_now
+    responses.includes(:owner).find_each do |response|
+      # Reuse registration's tagging so the sector lands on the person AND the
+      # org(s) they registered with, always as an additional tag (never primary).
+      SectorTagging.apply(person: response.owner, organizations: response.registration_organizations,
```
maebeale, line 71: 🤖 _From Claude:_ Update to my earlier note: promotion now tags the exact org from the **registration** (`OtherResponse#registration_organizations` walks `source_form_answer` → submission → the person's registration for that event), not their current affiliations — so no org needs storing on the response.

PR #1941, opened 2026-07-05
Rename user-facing "Notifications" to "Communications" (i18n)

maebeale:
🤖 suggested review level: 3 Read 📖 mechanical string→i18n swap across a few views + helper; two specs updated to match

- Renames the public-facing "Notifications" feature to **"Communications"** — the term staff actually use for outbound emails and manual contact logs.
- Routes every user-facing string through a single `communications:` namespace in `config/locales/en.yml`, so the wording is a one-line change in future.

- Added `communications.*` keys to `en.yml`; replaced hard-coded strings in the index/show pages, the admin dashboard card, and the person/registration/scholarship boxes with `t(...)` lookups.
- Left the `Notification` model, routes, controller actions, and params untouched — only presentation text changed.
- One label — **"Communications"** — for the nav card, the admin index heading, and the per-record boxes.
- Updated the two specs that asserted the old copy.

### Anything else to add?
- No model/schema changes; no migration.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -35,7 +35,6 @@ en:
   # rename "Communications" everywhere it appears in the UI.
   communications:
     title: "Communications"
-    log_title: "Communication log"
     detail_title: "Communication"
     back_link: "← Back to communications"
     empty: "No communications found."
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,6 +31,20 @@
 
 en:
   hello: "Hello world"
+  # Public-facing name for the notifications feature. Change these values to
+  # rename "Communications" everywhere it appears in the UI.
+  communications:
+    title: "Communications"
+    log_title: "Communication log"
+    detail_title: "Communication"
+    back_link: "← Back to communications"
+    empty: "No communications found."
+    add: "Add communication"
+    resent: "Communication has been resent successfully."
+    registration_title: "Registration communications"
+    scholarship_title: "Scholarship communications"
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -31,6 +31,20 @@
...
+  # rename "Communications" everywhere it appears in the UI.
+  communications:
+    title: "Communications"
+    log_title: "Communication log"
+    detail_title: "Communication"
+    back_link: "← Back to communications"
+    empty: "No communications found."
+    add: "Add communication"
+    resent: "Communication has been resent successfully."
+    registration_title: "Registration communications"
```
maebeale, line 43: 🤖 _From Claude:_ Only the user-facing label moved to i18n — the `Notification` model, routes, controller actions, and params stay `notifications` on purpose (renaming those is a large, risky refactor with no user payoff). Change any wording here to rename it everywhere.

config/locales/en.yml
```diff
@@ -31,6 +31,20 @@
...
+    scholarship_title: "Scholarship communications"
+    none_for_person: "No communications for this person yet."
```
maebeale, line 45: 🤖 _From Claude:_ `log_title` ("Communication log") is used only for the standalone admin index — the full searchable history — while `title` ("Communications") labels the nav card and per-record boxes. Split so either can change independently.

maebeale: 🤖 _From Claude:_ Update: dropped the `log_title` split — the index now uses `title` ("Communications") too, so the nav card and the page it links to share one label.

PR #1392, opened 2026-03-10
HOLD: Audit Stimulus controllers for conventions (split into smaller pr's)

maebeale:
## Summary
- Audit all Stimulus controllers to follow project conventions
- Use Stimulus targets instead of `querySelector` / direct DOM queries
- Use `hidden` class instead of inline `style.display` manipulation
- Rename JS constants from `SCREAMING_SNAKE_CASE` to `camelCase`
- Use shorthand action descriptors (omit default events)
- Remove unused controllers (`nav`, `search_box`)
- Remove debug `console.log` statements
- Add Stimulus conventions to CLAUDE.md based on PR review feedback

### Column toggle (events manage)
- [ ] Toggle columns on/off in event registrations table
- [ ] Knob slides correctly (Tailwind class instead of inline transform)

### Comment edit toggle
- [ ] Event registration form — toggle inline comment editing on/off
- [ ] Organization form — toggle inline comment editing on/off
- [ ] Person form — toggle inline comment editing on/off
- [ ] User form — toggle inline comment editing on/off
- [ ] Nested field values sync correctly between view and edit modes

### Dirty form (unsaved changes warning)
- [ ] Person form — warns before navigating away with unsaved changes
- [ ] User form — warns before navigating away with unsaved changes
- [ ] Submit clears dirty state (no false warning after save)

### Password toggle (eye icon)
- [ ] Login page — toggle password visibility
- [ ] Set password page — toggle both password fields independently
- [ ] Change password page — toggle all 3 password fields independently
- [ ] Welcome/first-login page — toggle password visibility

### Tabs controller
- [ ] Tabbed interfaces switch content correctly (index-based lookup)

### Paginated fields
- [ ] Multi-page comment fields — navigate between pages

### Tags combination highlight
- [ ] Explore by combination — sector/category checkboxes highlight correctly
- [ ] Highlights sync after turbo frame load

### Print options
- [ ] Print toggle — check/x icons swap correctly

### Toggle lock controller
- [ ] Lock icon shows/hides using `hidden` class

### Toggle user icon controller
- [ ] User form — icon toggles on/off
- [ ] User edit — icon toggles on/off

### Removed controllers
- [ ] No broken references to `data-controller="nav"` or `data-controller="search-box"`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1942, opened 2026-07-05
Editable communications: inline edit, channel icons, admin-gated bodies

maebeale:
🤖 suggested review level: 5 Inspect 🔬 substantive view/model/authorization logic across the communications box on person, registration, and scholarship edit pages

Turn the read-only communications box into a working inline log admins can maintain, while keeping the right things visible to the recipient (once person edit opens to them).

- **Inline edit/remove** of hand-noted communications, mirroring the comments box (view ⇄ edit toggle, saved with the parent form).
- **Add is no longer silently lossy** — Subject is required once a row is in use (`field-required`), so a body-only entry can't be dropped by the model's `reject_if`.
- **Channel icons** before each subject (envelope / phone / mobile-for-text / video), from a `NotificationDecorator` helper; sender name is fixed-width so icons align.
- **Visibility by type**: an automated ("AWBW Portal"/autoemail) body is the recipient's own email → visible to everyone; a hand-noted body is an internal staff note → admin-only, shown with `admin-only bg-blue-100`, hidden from non-admins (no text, no hover). Only the body is hoverable. Subject/date/icon always visible.
- **Editing scope**: any hand-noted communication owned by the record is editable; autoemails never are. `autoemail` can't be chosen in the form or set on a `manual_log` (coerced to a manual channel). Setting a manual channel only defaults a **blank** kind to `manual_log` — a real notification's kind is never silently overwritten.
- **Edits survive a validation re-render** — editable rows render the record's own in-memory notification (keyed by id), not a stale recipient-email query row, so an admin's unsaved inline edit isn't reverted when the parent form fails validation elsewhere. Matches how new entries already behaved.
- Form labels spell out "Subject (visible to user)" / "Body (hidden from user)"; form laid out like the comments box.
- Friendly **`/communications`** route redirects to `/notifications` (preserving filters).
- **Index filters round-trip** — the Email contains / Subject contains boxes now echo their active params on load (the modelless form wasn't populating text fields), so the "View all" link lands with the email prefilled. Also narrowed the Email topic select and widened Email contains.

- Generalized two comment-only Stimulus controllers into shared ones: `edit-toggle` (view/edit via targets) and `field-required` (gate a nested row's required field once in use).
- Split the box into `_notification_row` (read-only display), `_notification_form` (shared fields), and a view/edit `_notification_fields`.
- Gating is currently **view-level**; a server-side guard on `notifications_attributes` is a noted follow-up for when person edit opens to non-admins.

### Anything else to add?
- Editing works through the owning record's nested attributes, so a communication is editable only on the page of the record it belongs to.


app/frontend/javascript/controllers/edit_toggle_controller.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  connect() {
+    this.editing = false;
+  }
+
+  toggle() {
+    this.editing = !this.editing;
+
+    // When leaving edit mode, sync edit textareas into their truncated view.
```
maebeale: 🤖 _From Claude:_ Generalized from `comment-edit-toggle`. The body-sync (updating the truncated view from the textarea on exit) is now opt-in via `bodyClassValue` — comments set it, communications leave it blank since their view row has no single truncated body span.

app/views/people/_notifications.html.erb
```diff
@@ -1,12 +1,15 @@
...
+    person can be edited inline (toggle "Edit communications", like the comments
+    box) and new ones added, both saved with the person form. Links open in a
+    new tab so unsaved edits aren't lost. Mirrors the registration edit
+    notifications section. ---- %>
 <% person = f.object %>
 <% email = person.preferred_email %>
 <% notifications = email.present? ? Notification.email(email).order(created_at: :desc) : Notification.none %>
-<section class="overflow-hidden rounded-xl border border-gray-200 bg-white shadow-sm">
+<% new_notifications = person.notifications.select(&:new_record?) %>
+<% editable_notifications = notifications.select { |n| n.manual_log? && n.noticeable_type == Person.name && n.noticeable_id == person.id } %>
```
maebeale: 🤖 _From Claude:_ The list is matched by recipient email (broad), but only `manual_log` entries whose `noticeable` is this record can be edited via nested attributes — automated notifications and email-matched ones owned by other records stay read-only.

app/models/person.rb
```diff
@@ -99,7 +99,7 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :affiliations, allow_destroy: true,
     reject_if: proc { |attrs| attrs["organization_id"].blank? }
   accepts_nested_attributes_for :comments, allow_destroy: true, reject_if: proc { |attrs| attrs["body"].blank? }
-  accepts_nested_attributes_for :notifications, reject_if: proc { |attrs| attrs["email_subject"].blank? }
+  accepts_nested_attributes_for :notifications, allow_destroy: true, reject_if: proc { |attrs| attrs["email_subject"].blank? }
```
maebeale, line 102: 🤖 _From Claude:_ `allow_destroy: true` (with `:_destroy` permitted in the controller) lets the inline "Remove" link delete a logged entry — the `reject_if` blank-subject guard is bypassed for destroys, matching how comments work.

app/frontend/javascript/controllers/edit_toggle_controller.js
```diff
@@ -0,0 +1,64 @@
...
+    if (!this.editing && this.bodyClassValue) {
+      this.element.querySelectorAll(".nested-fields").forEach((item) => {
```
jmilljr24: This could be a good opportunity to update this controller to use stimulus targets vs. depending on class targets.

app/views/event_registrations/_form.html.erb
```diff
@@ -337,11 +337,11 @@
             <button
               type="button"
               class="inline-flex items-center gap-1.5 rounded-md px-2 py-1 text-xs font-medium text-gray-500 transition-colors hover:bg-gray-100 hover:text-gray-700 cursor-pointer"
-              data-action="click->comment-edit-toggle#toggle"
+              data-action="click->edit-toggle#toggle"
```
jmilljr24:
```suggestion
              data-action="edit-toggle#toggle"
```
Click is implicit on buttons

app/frontend/javascript/controllers/edit_toggle_controller.js
```diff
@@ -0,0 +1,64 @@
...
+//
+// When a body class is given, leaving edit mode syncs each edit textarea back
+// into the matching truncated `.{body}` span so the view reflects unsaved edits.
+//
+export default class extends Controller {
+  static targets = ["editLabel", "viewLabel"];
+  static values = {
+    viewClass: { type: String, default: "editable-view" },
+    editClass: { type: String, default: "editable-edit" },
+    bodyClass: { type: String, default: "" },
```
jmilljr24:
```suggestion
```
I believe these could all be done with stimulus targets instead of setting a class as a selector.

jmilljr24 (approved):
Some thoughts on the stimulus controller. Not required, just might be a good time to refactor while its being worked on.

#1392

app/frontend/javascript/controllers/edit_toggle_controller.js
maebeale: 🤖 _From Claude:_ Done — the controller now uses `view`/`edit`/`body` Stimulus targets instead of class-name selectors, and the per-caller class-value attributes are gone. Pushed in a4a17a4.

app/frontend/javascript/controllers/edit_toggle_controller.js
```diff
@@ -0,0 +1,53 @@
...
+
+    // Leaving edit mode: sync each edit textarea into its truncated view span.
+    if (!this.editing && this.hasBodyTarget) {
+      this.editTargets.forEach((edit, i) => {
+        const textarea = edit.querySelector("textarea");
+        const body = this.bodyTargets[i];
+        if (textarea && body) {
+          const text = textarea.value;
+          body.textContent =
+            text.length > 135 ? text.substring(0, 132) + "..." : text;
```
jmilljr24: the comment_field erb already has tailwind truncat so this is redundant. 
```suggestion
          body.textContent = text
```

app/views/comments/_comment_fields.html.erb
```diff
@@ -50,12 +50,12 @@
...
         <% end %>
       </span>
-      <span class="text-sm <%= is_age_range_data ? "text-amber-700" : "text-gray-900" %> comment-body truncate" title="<%= [ f.object.topic, f.object.body ].compact_blank.join(" — ") %>">
+      <span class="text-sm <%= is_age_range_data ? "text-amber-700" : "text-gray-900" %> truncate" data-edit-toggle-target="body" title="<%= [ f.object.topic, f.object.body ].compact_blank.join(" — ") %>">
         <% if is_age_range_data %><i class="fa-solid fa-triangle-exclamation"></i><% end %>
         <% if f.object.topic.present? %><span class="font-bold uppercase tracking-wide"><%= f.object.topic %></span> <% end %><%= truncate(f.object.body, length: 135) %>
       </span>
     </div>
-    <div class="comment-edit flex items-start mb-1 gap-x-1 <%= "bg-amber-50 border border-amber-200 rounded-md py-1" if is_age_range_data %>" style="display:none">
+    <div class="flex items-start mb-1 gap-x-1 <%= "bg-amber-50 border border-amber-200 rounded-md py-1" if is_age_range_data %>" style="display:none" data-edit-toggle-target="edit">
```
jmilljr24:
```suggestion
    <div class="flex items-start mb-1 gap-x-1 hidden <%= "bg-amber-50 border border-amber-200 rounded-md py-1" if is_age_range_data %>"  data-edit-toggle-target="edit">
```

app/views/comments/_comment_fields.html.erb
```diff
@@ -50,12 +50,12 @@
           <span class="<%= created_color[0] %> <%= created_color[1] %> text-xs px-1.5 py-0.5 rounded cursor-default truncate w-full text-left" title="<%= created_name %>"><%= truncate(created_first_name.presence || "--", length: 10, omission: "..") %></span>
         <% end %>
       </span>
-      <span class="text-sm <%= is_age_range_data ? "text-amber-700" : "text-gray-900" %> comment-body truncate" title="<%= [ f.object.topic, f.object.body ].compact_blank.join(" — ") %>">
+      <span class="text-sm <%= is_age_range_data ? "text-amber-700" : "text-gray-900" %> truncate" data-edit-toggle-target="body" title="<%= [ f.object.topic, f.object.body ].compact_blank.join(" — ") %>">
         <% if is_age_range_data %><i class="fa-solid fa-triangle-exclamation"></i><% end %>
         <% if f.object.topic.present? %><span class="font-bold uppercase tracking-wide"><%= f.object.topic %></span> <% end %><%= truncate(f.object.body, length: 135) %>
```
jmilljr24: span already truncates
```suggestion
        <% if f.object.topic.present? %><span class="font-bold uppercase tracking-wide"><%= f.object.topic %></span> <% end %><%= f.object.body, length: 135 %>
```

app/frontend/javascript/controllers/edit_toggle_controller.js
```diff
@@ -0,0 +1,53 @@
...
+          body.title = text;
+        }
+      });
+    }
+
+    this.viewTargets.forEach(
+      (el) => (el.style.display = this.editing ? "none" : "")
```
jmilljr24: You removed used "style" for other controllers in a recent PR. This should follow the same pattern.

maebeale: 🤖 _From Claude:_ Done — the controller now toggles the `hidden` class (not `style.display`) for the view/edit blocks and labels, and every initial-hidden block uses `class="hidden"`. Kept your `textContent = text` simplification (CSS `truncate` handles the view) and fixed the stray invalid ERB. Rebased onto latest main.

app/views/notifications/_communications.html.erb
```diff
@@ -0,0 +1,92 @@
...
+    `none_key` (empty-state copy). Everything else derives from `record`. ---- %>
+<% record = f.object %>
+<% email = record.communications_email %>
+<% notifications = email.present? ? Notification.email(email).order(created_at: :desc) : Notification.none %>
+<% admin = allowed_to?(:index?, with: NotificationPolicy) %>
+<% new_notifications = record.notifications.select(&:new_record?) %>
+<%# `record` may be a Draper decorator here, so read the model name via
+    `model_name` (delegated to the underlying class) rather than `class.name`,
+    which would return the decorator (e.g. "PersonDecorator"). %>
+<% record_type = record.model_name.name %>
```
maebeale, line 20: 🤖 _From Claude:_ In the edit views `f.object` is a Draper decorator, so `class.name` returns e.g. `PersonDecorator` and never matches a notification's `noticeable_type`. `model_name.name` delegates to the real model, so the editable-comm filter works for all three records. (The old per-model wrappers avoided this by hardcoding `Person.name`/`EventRegistration.name`/`Scholarship.name`.)

app/views/notifications/_communications.html.erb
```diff
@@ -0,0 +1,96 @@
...
+      <% end %>
+    <% end %>
+  </div>
+
+  <div class="space-y-3 p-4">
+    <div data-controller="paginated-fields" data-paginated-fields-per-page-value="5" class="space-y-2">
+      <div id="<%= list_id %>" class="space-y-2">
+        <% notifications.each do |notification| %>
+          <% if editable_notifications.include?(notification) %>
+            <%= f.simple_fields_for :notifications, own_notifications_by_id.fetch(notification.id, notification) do |nf| %>
```
maebeale, line 50: 🤖 _From Claude:_ Editable rows now render the record's own in-memory notification (via `own_notifications_by_id`), not the fresh recipient-email query row. Without this, an admin's unsaved inline edit reverted to its saved value whenever the parent form failed validation elsewhere and re-rendered `:edit`. Falls back to the query row if the record doesn't own it.

app/models/notification.rb
```diff
@@ -114,6 +114,17 @@ class Notification < ApplicationRecord
   # edit page) records a contact that already happened — fill in the sensible
   # defaults so it validates without going through the delivery pipeline.
   before_validation :apply_manual_log_defaults, on: :create, if: -> { kind.blank? }
+  # The portal only ever sends autoemails; every other channel is a hand-logged
+  # communication, so a non-autoemail channel with no kind yet defaults to
+  # "manual_log". We only fill in a *blank* kind — an already-set kind is never
+  # silently overwritten, so a real notification keeps its kind even if some
+  # caller sets a manual channel on it.
+  before_validation :classify_manual_channel_as_manual_log, if: -> { kind.blank? && channel.present? && channel != "autoemail" }
```
maebeale, line 122: 🤖 _From Claude:_ Added the `kind.blank?` guard so a non-autoemail channel only defaults a *blank* kind to `manual_log`; an already-set kind is never silently overwritten. Prevents a real notification (e.g. `event_registration_confirmation`) from being reclassified if a caller sets a manual channel on it.

app/views/notifications/_search_boxes.html.erb
```diff
@@ -6,11 +6,12 @@
                       turbo_frame: "notifications_results" },
                       autocomplete: "off",
                 class: "flex items-end gap-3" do |f| %>
-    <div class="flex-1">
+    <div class="flex-[2]">
       <%= f.label :email, "Email contains",
                   class: "block text-sm font-medium text-gray-500 mb-1" %>
 
       <%= f.text_field :email,
+                       value: params[:email],
```
maebeale, line 14: 🤖 _From Claude:_ This is a modelless `form_with`, so text fields don't auto-populate from params (only the selects passed `params[...]`). Without `value:`, the "View all" link (`/notifications?email=…`) landed with an empty Email contains box. Same fix applied to `subject_line`.

PR #1943, opened 2026-07-05
Admin submission review

jmilljr24:

Admin would view form submissions via the same route as the submitter under the registration slug. Form submission have many conditionally displayed fields so a person my not have to fill out the full form if they are logged in. Admin need to see all information related to the person that registered.

This adds an admin only form submission specific route/controller. This removes the complexity of having an admin try to view the registrants form submission view. This easily allows viewing multiple form submission for a person.


app/controllers/events/form_submissions_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Events
```
maebeale, line 1: why nest this under `events`? we'll have forms submitted sometimes that aren't tied to events. if we're going to namespace or nest, we prob should do so under `forms` or `admin`? esp since the route ends up `event_event_form_submissions_path`.

spec/requests/events/registrations_spec.rb
```diff
@@ -592,7 +592,7 @@
       ).call
       EventForm.create!(event: event, form: form, role: "registration")
       form = event.registration_form
-      form.form_submissions.create!(person: user.person)
+      form.form_submissions.create!(person: user.person, event: event)
```
maebeale, line 595: ooo. nice! ty for cleaning these up too!

app/controllers/events/form_submissions_controller.rb
jmilljr24: Great point!

PR #1945, opened 2026-07-07
Gate training-only ticket callouts to facilitator trainings

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained visibility guards on built-in ticket callouts + spec updates

- The built-in **Handouts**, **FAQ**, and **Forms** ticket callouts rendered on *every* registration ticket, but their content is 2-Day AWBW Facilitator Training–specific (training worksheets, "2-day training" FAQ).
- Public, non-training events — including **live ones with no registration process** — were surfacing these irrelevant training callouts to registrants. This is the urgent fix.

- Added `Event#show_handouts_callout?` / `#show_faq_callout?` (→ `facilitator_training?`) and `#show_forms_callout?` (→ `facilitator_training?` **or** paid, since paying registrants still need their invoice/receipt).
- `MagicTicketCallouts` now returns those cards only when the predicate holds; the callout pages (`forms`/`handouts`/`faq`) redirect to the ticket when hidden.
- **No migration** — `facilitator_training` already partitions live data correctly, so this corrects every live event on deploy.
- Layer 1 of 2: a follow-up PR adds per-event Auto/Show/Hide overrides in the event's callouts section.

### Anything else to add?
- Editor preview visibility labels updated ("On facilitator trainings", "On facilitator trainings and paid events").
- The callout *content* is still hardcoded to the 2-day training even for trainings — out of scope here, worth revisiting.

### DIFF
--- a/app/controllers/events/callouts_controller.rb
+++ b/app/controllers/events/callouts_controller.rb
@@ -64,8 +64,9 @@ module Events
 
     # Forms page: callout-card links to the W-9 (when seeded) and, for paid
     # events, the invoice and the paid-in-full receipt (once settled), each
-    # returning to forms.
+    # returning to forms. Only reachable when the event shows the Forms callout.
     def forms
+      return redirect_to registration_ticket_path(@event_registration.slug) unless @event.show_forms_callout?
       @form_cards = build_form_cards
     end
 
@@ -73,6 +74,7 @@ module Events
     # resources, in display order, each opening its own registrant resource page
     # (PDF preview + download, with a back-to-handouts eyebrow).
     def handouts
+      return redirect_to registration_ticket_path(@event_registration.slug) unless @event.show_handouts_callout?
       by_title = Resource.where(title: HANDOUT_RESOURCE_TITLES).index_by(&:title)
       @handout_cards = HANDOUT_RESOURCE_TITLES.filter_map do |title|
         resource = by_title[title]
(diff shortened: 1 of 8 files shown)


app/models/event.rb
```diff
@@ -121,6 +121,28 @@ def one_click_for_signed_in?
...
+  # admins can force a callout on or off from the event's callouts section.
+  def show_handouts_callout?
+    facilitator_training?
+  end
+
+  def show_faq_callout?
+    facilitator_training?
+  end
+
+  def show_forms_callout?
```
maebeale, line 142: 🤖 _From Claude:_ Forms is gated on `facilitator_training? || paid` rather than training-only: the Forms page carries the invoice and paid-in-full receipt, which any paying registrant needs. Handouts/FAQ are pure training curriculum, so those stay training-only.

app/controllers/events/callouts_controller.rb
```diff
@@ -96,8 +98,10 @@ def videoconference
       @event = @event_registration.event.decorate
     end
 
-    # FAQ for the training, with a folded-in contact link.
+    # FAQ for the training, with a folded-in contact link. Only reachable when
+    # the event shows the FAQ callout.
     def faq
+      redirect_to registration_ticket_path(@event_registration.slug) unless @event.show_faq_callout?
```
maebeale, line 104: 🤖 _From Claude:_ These pages are reachable by slug (no auth), so hiding the card is not enough — a shared/stale link would still surface training content on a non-training event. The redirect mirrors how `certificate`/`scholarship` guard their unavailable states.

maebeale:
@jmilljr24 this keeps the hard-coded cards hard-coded. there's a world we might want to consider where these become populated data at time of event creation or opt-in to the event, but, this gets them turned off for rn while we talk through options.

jmilljr24 (approved):

PR #1946, opened 2026-07-08
Fix idea-image promotion: transfer box on new, apply on update, guard save

maebeale:
🤖 suggested review level: 3 Read 📖 small contained logic across a service, controller update action, and one form

Fixes three bugs in the promote-workshop-variation-idea → workshop-variation flow around the "transfer attachments from the idea" box.

- **Box only showed on edit, not new.** The from-idea service pre-built (but never persisted) copies of the idea's assets on the unsaved variation, so the form's blob-id comparison thought they were already transferred and hid the box on `new`. Removed the counterproductive `duplicate_assets` (it never persisted — `create` rebuilds from params).
- **Checking the box on edit did nothing.** `create` called `attach_assets_from_idea!` when checked, but `update` didn't. Added the matching call.
- **Save button re-uploaded on double-click.** The form runs `turbo: false`, so rapid clicks fired multiple full POSTs. Added the existing `submit-once` Stimulus controller (built for `turbo: false` forms) to disable the button and show "Saving…".
- Corrected the box's warning copy — it *appends* the idea's attachments, it doesn't replace.

- Wrote failing request specs first: box renders on `GET /new` when the idea has attachments; `PATCH /update` with `promote_idea_assets: "true"` attaches them. Then applied the fixes.

### Anything else to add?
No migration, no schema change. Reused the existing `submit-once` controller rather than adding JS.

### DIFF
--- a/app/controllers/workshop_variations_controller.rb
+++ b/app/controllers/workshop_variations_controller.rb
@@ -82,6 +82,7 @@ class WorkshopVariationsController < ApplicationController
     authorize! @workshop_variation
 
     if @workshop_variation.update(workshop_variation_params)
+      @workshop_variation.attach_assets_from_idea! if params[:promote_idea_assets] == "true"
       flash[:notice] = "Workshop Variation updated successfully."
       redirect_to @workshop_variation
     else
--- a/app/services/workshop_variation_from_idea_service.rb
+++ b/app/services/workshop_variation_from_idea_service.rb
@@ -7,9 +7,7 @@ class WorkshopVariationFromIdeaService
   end
 
   def call
-    WorkshopVariation.new(attributes_from_idea).tap do |workshop_variation|
-      duplicate_assets(workshop_variation)
-    end
+    WorkshopVariation.new(attributes_from_idea)
   end
 
   private
@@ -26,11 +24,5 @@ class WorkshopVariationFromIdeaService
       inactive: true,
       rhino_body: workshop_variation_idea.rhino_body
     )
-  end
-
-  def duplicate_assets(workshop_variation)
-    workshop_variation_idea.assets.each do |asset|
-      workshop_variation.assets.build(file: asset.file.blob)
-    end
   end
 end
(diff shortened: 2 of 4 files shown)


app/services/workshop_variation_from_idea_service.rb
```diff
@@ -7,9 +7,7 @@ def initialize(workshop_variation_idea, user:)
   end
 
   def call
-    WorkshopVariation.new(attributes_from_idea).tap do |workshop_variation|
-      duplicate_assets(workshop_variation)
-    end
+    WorkshopVariation.new(attributes_from_idea)
```
maebeale, line 10: 🤖 _From Claude:_ Root cause of "box only shows on edit": the old `duplicate_assets` built idea-blob asset copies here that never persisted (create rebuilds from params), but they made the form's blob-id comparison treat the assets as already-transferred and hide the box on new. Dropping it lets the box render on new.

app/controllers/workshop_variations_controller.rb
```diff
@@ -82,6 +82,7 @@ def update
     authorize! @workshop_variation
 
     if @workshop_variation.update(workshop_variation_params)
+      @workshop_variation.attach_assets_from_idea! if params[:promote_idea_assets] == "true"
```
maebeale, line 85: 🤖 _From Claude:_ `create` already did this; `update` did not, so checking the box on edit silently did nothing. Mirrors the create path.

app/views/workshop_variations/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="rounded-xl bg-white p-6">
-  <%= simple_form_for @workshop_variation, html: { multipart: true }, data: { turbo: false } do |f| %>
+  <%= simple_form_for @workshop_variation, html: { multipart: true }, data: { turbo: false, controller: "submit-once", "submit-once-submitting-text-value": "Saving…" } do |f| %>
```
maebeale, line 2: 🤖 _From Claude:_ Form is `turbo: false`, so double-clicks fired multiple full POSTs and re-uploaded the image. `submit-once` is the existing controller purpose-built for non-Turbo forms.

maebeale, line 2: we should maybe do this on all buttons, @jmilljr24 ? `controller: "submit-once", "submit-once-submitting-text-value": "Saving…"`

jmilljr24: My first thought is why the form is set to turbo false. Turbo already handles preventing double submits and allows custom text on the button. 

This feels like we are reinventing the wheel. There may be a valid reason to disable turbo but we should keep it as much as we can.

jmilljr24: @maebeale

PR #1947, opened 2026-07-08
Fix dev seeds broken by resource author association (#1938)

maebeale:
🤖 suggested review level: 3 Read 📖 dev-seed-only logic fix (association + idempotency), no app/runtime code touched

`bin/setup` / `db:seed:dev` was aborting on a fresh DB. #1938 turned `Resource#author` from a free-text string into a `Person` association but left the dev seed assigning a Faker string → `AssociationTypeMismatch`.

- `resources.rb`: credit a real `Person` (via the association), keep the old free-text in `legacy_author_name`, and set `author_credit_preference` — mirroring the other AuthorCreditable seeds.
- `events_management.rb`: #1938 also made `Person#resources_as_author` `restrict_with_error`, so a demo registrant credited on a resource could no longer be destroyed — the "recreate from scratch each run" cleanup silently failed and re-seeding crashed on the name/email uniqueness validation. Release the credit before destroying so the block stays idempotent.

Verified `db:seed:dev` runs clean on a fresh DB and idempotently across repeated runs.

### DIFF
--- a/db/seeds/dev/events_management.rb
+++ b/db/seeds/dev/events_management.rb
@@ -1442,8 +1442,14 @@ if facilitator_training && registration_form
   end
 
   # Recreate from scratch each run so re-seeding refreshes labels and link state.
-  Person.where("email LIKE ? OR email LIKE ?",
-    "orgchip.demo.%@seed.example.com", "affdemo.%@seed.example.com").find_each(&:destroy)
+  # Release any resource authorship first: resources.rb (which runs after this) can
+  # credit these demo people, and Person#resources_as_author is restrict_with_error,
+  # so the destroy below silently fails and the re-create hits the name/email
+  # uniqueness validation. Clearing author_id here keeps the block idempotent.
+  demo_people = Person.where("email LIKE ? OR email LIKE ?",
+    "orgchip.demo.%@seed.example.com", "affdemo.%@seed.example.com")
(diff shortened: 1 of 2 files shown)


db/seeds/dev/events_management.rb
```diff
@@ -1442,8 +1442,14 @@
...
-  Person.where("email LIKE ? OR email LIKE ?",
-    "orgchip.demo.%@seed.example.com", "affdemo.%@seed.example.com").find_each(&:destroy)
+  # Release any resource authorship first: resources.rb (which runs after this) can
+  # credit these demo people, and Person#resources_as_author is restrict_with_error,
+  # so the destroy below silently fails and the re-create hits the name/email
+  # uniqueness validation. Clearing author_id here keeps the block idempotent.
+  demo_people = Person.where("email LIKE ? OR email LIKE ?",
+    "orgchip.demo.%@seed.example.com", "affdemo.%@seed.example.com")
+  Resource.where(author_id: demo_people).update_all(author_id: nil)
+  demo_people.find_each(&:destroy)
```
maebeale, line 1452: 🤖 _From Claude:_ `update_all` is deliberate here (per the CLAUDE.md caveat): it releases the `restrict_with_error` FK in one statement so the `destroy` below can succeed. Without it, a demo person credited on a resource survives the cleanup and the re-create trips the name/email uniqueness validation.

PR #1948, opened 2026-07-08
Stop CSP violation-report 404s: drop report_uri + remove inline styles/scripts

maebeale:
🤖 suggested review level: 3 Read 📖 config + inline-style/script removals; no data changes

### Why
Browsers POST CSP violation reports to `policy.report_uri`, but no route served that path — so every report-only violation on person edit/create showed up as a red `404` in the console. The violations themselves came from inline `style=`/`<script>` in the views.

### What
- Remove `report_uri` from the CSP initializer so browsers send no reports (kills the 404 noise at the source; there was nothing consuming the reports).
- Remove the inline attributes/scripts that triggered the violations:
  - person form: `float:right` → `float-right`, label `display:none` → `hidden`
  - comment fields: `display:none` → `hidden`; `comment_edit_toggle_controller` toggles the `hidden` class via `classList` instead of `style.display`
  - duplicate-person Turbo Stream: inline `<script>window.scrollTo(0,0)</script>` → new `scroll-to-top` Stimulus controller on the warning box (covers both the stream and full-page re-render paths)

Split out of the address-collapse branch (#1949).


app/views/people/_duplicate_people_warning.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= tag.div id: "duplicate_people_warning" do %>
   <% if @duplicates.present? %>
-    <div class="mb-6 bg-yellow-50 border border-yellow-200 rounded-lg p-4">
+    <div class="mb-6 bg-yellow-50 border border-yellow-200 rounded-lg p-4"
+         data-controller="scroll-to-top">
```
maebeale, line 4: 🤖 _From Claude:_ Attached here (the inner box, only rendered when `@duplicates.present?`) rather than the outer `#duplicate_people_warning` container — the container renders empty on every page load, so connecting there would scroll to top on initial load too. This placement fires only when the warning actually appears, covering both the Turbo Stream and `render :new` paths.

maebeale:
@jmilljr24 we had a handful of console errors in person edit. this fixes them and adds the sink so future ones get logged.

config/initializers/content_security_policy.rb
```diff
@@ -11,7 +11,8 @@
     policy.img_src     :self, :https, :data
     policy.object_src  :none
     policy.script_src  :self, :https
-    # Specify URI for violation reports
+    # Browsers POST violation reports here; CspReportsController#create logs them
+    # and returns 204. Keep this path in sync with the route.
     policy.report_uri "/csp-violation-report-endpoint"
```
jmilljr24: Do you plan to actually look at the logs on a regular bases? If not, we can just remove this line and keep the actual security checks in place which is what is actually important in this initializer.

```suggestion
```

maebeale: what i started to do was have a page we could look at from time to time on it. but, yeah, am ok to just ditch out.

jmilljr24 (approved):

PR #1950, opened 2026-07-08
Unify sector/age-range chip colors; surface sector leaders on people index

maebeale:
🤖 suggested review level: 3 Read 📖 small view/scope changes plus one filter scope, with model spec coverage

- **Age-range chips now match the lime sector chips.** They were amber/gray, reading as an unrelated tag family beside the lime sectors.
- **Edit-form chips match the profile.** Non-primary sector/age chips were white on edit but lime-50 on the profile — now identical on view and edit.
- **Sector leadership is visible and filterable on the people index.** The `(Leader)` badge was clipped (it sat inside the truncating name span); a "Sector leaders only" filter was added.

- Retheme age-range chips (`_age_group_tags`) to lime; align edit chips (`_sectorable_item_fields`, `_age_range_item_fields`) to the profile's lime scheme.
- Move `(Leader)` out of the truncating span (add `flex-shrink-0`) so it always shows; append "(Sector leader)" to the chip's hover tooltip.
- Add a `sector_leaders` scope + `sector_leaders_only` param to `Person.search_by_params`, and a checkbox to the people search (auto-submits via the existing `collection` controller).

### Anything else to add?
- The index sector column only renders **primary** sectors, so the crown/badge shows there only when someone leads their primary sector; the filter and the profile still reflect leadership on any sector.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -149,10 +149,13 @@ class Person < ApplicationRecord
     joins(:affiliations)
       .where(affiliations: { organization_id: organization_id })
       .distinct }
+  scope :sector_leaders, -> {
+    joins(:sectorable_items).where(sectorable_items: { is_leader: true }).distinct }
 
   def self.search_by_params(params)
     results = is_a?(ActiveRecord::Relation) ? self : all
     results = results.search(params[:contact_info]) if params[:contact_info].present?
+    results = results.sector_leaders if ActiveModel::Type::Boolean.new.cast(params[:sector_leaders_only])
     results = results.sector_names_all(params[:sector_names_all]) if params[:sector_names_all].present?
     results = results.category_names_all(params[:category_names_all]) if params[:category_names_all].present?
     results = results.organization_name(params[:organization_name]) if params[:organization_name].present?
--- a/app/views/sectors/_tagging_label.html.erb
+++ b/app/views/sectors/_tagging_label.html.erb
@@ -45,7 +45,7 @@
 
 <% if sector %>
   <%= link_to taggings_path(sector_names_all: sector.name),
-    title: sector.name,
+    title: (display_leader && is_leader ? "#{sector.name} (Sector leader)" : sector.name),
     data: { turbo_frame: "_top", sector_name: sector.name, controller: "tag-link-loading", action: "click->tag-link-loading#showSpinner" },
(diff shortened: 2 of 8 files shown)


app/views/sectors/_tagging_label.html.erb
```diff
@@ -55,7 +55,7 @@
                       #{ text_color }
                       #{ chip_size } font-medium
                       transition" do %>
-    <span data-tag-link-loading-target="text" class="<%= "truncate min-w-0" if max_width %>"><% if is_hidden %><%= render "shared/hidden_tag_icon" %><% end %><% if is_primary %><i class="fa-solid fa-star mr-1 text-xs text-amber-400" aria-hidden="true" title="Primary sector"></i><% end %><%= sector.name %><% if display_leader && is_leader %><span class="ml-1 text-xs text-indigo-600">(Leader)</span><% end %></span>
+    <span data-tag-link-loading-target="text" class="<%= "truncate min-w-0" if max_width %>"><% if is_hidden %><%= render "shared/hidden_tag_icon" %><% end %><% if is_primary %><i class="fa-solid fa-star mr-1 text-xs text-amber-400" aria-hidden="true" title="Primary sector"></i><% end %><%= sector.name %></span><% if display_leader && is_leader %><span class="ml-1 flex-shrink-0 text-xs text-indigo-600">(Leader)</span><% end %>
```
maebeale, line 58: 🤖 _From Claude:_ This is the fix for the invisible badge: `(Leader)` was inside the `truncate min-w-0` span, so it got clipped off whenever the sector name filled the chip (the index sets `max-w-[200px]`). Moving it out and adding `flex-shrink-0` keeps it visible.

app/models/person.rb
```diff
@@ -149,10 +149,13 @@ class Person < ApplicationRecord
     joins(:affiliations)
       .where(affiliations: { organization_id: organization_id })
       .distinct }
+  scope :sector_leaders, -> {
+    joins(:sectorable_items).where(sectorable_items: { is_leader: true }).distinct }
```
maebeale, line 153: 🤖 _From Claude:_ `.distinct` is load-bearing here — a person leading multiple sectors would otherwise appear once per leader `sectorable_item` row after the join.

PR #1951, opened 2026-07-08
Add shareable slug URLs for stories

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained `to_param` change plus link tweaks, no schema change

- Story links need to be *shareable* — a bare `/stories/23` says nothing. This adds a readable slug so a shared link carries the story's name.

- Override `Story#to_param` to return `"<id>-<title parameterized>"` (e.g. `/stories/23-my-great-story`), stripping bad URL characters via `parameterize`.
- No DB column: the slug is derived, and Rails resolves the URL back through the leading id's `to_i`, so `/stories/23` and `/stories/23/edit` keep working.
- Edit links pass the bare `@story.id` so edit stays `/stories/23/edit`.
- All site → story links (cards, social-share `story_url`, `polymorphic_path`) pick up the slug automatically via `to_param`.

### Anything else to add?
- Considered a stored `slug` column for a fully number-free URL (`/stories/my-great-story`), but that needs a migration, backfill, and collision handling; the id-prefix gives shareable readable links with zero schema change and can't break.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -118,6 +118,15 @@ class Story < ApplicationRecord
     stories
   end
 
+  # Shareable, readable URLs (story_path, polymorphic_path, etc.): the id followed
+  # by the title slugged with hyphens and bad URL characters stripped, e.g.
+  # "23-my-great-story". Rails resolves it back via `id.to_i`, so `/stories/23`
+  # and `/stories/23/edit` (which pass the bare id) keep working.
+  def to_param
+    return id&.to_s if title.blank?
+    "#{id}-#{title.parameterize}"
+  end
+
   def name
     title
   end
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -160,6 +160,8 @@ class StoriesController < ApplicationController
   private
 
   def set_story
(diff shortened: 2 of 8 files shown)


app/models/story.rb
```diff
@@ -118,6 +118,15 @@ def self.search_by_params(params)
     stories
   end
 
+  # Shareable, readable URLs (story_path, polymorphic_path, etc.): the id followed
+  # by the title slugged with hyphens and bad URL characters stripped, e.g.
+  # "23-my-great-story". Rails resolves it back via `id.to_i`, so `/stories/23`
+  # and `/stories/23/edit` (which pass the bare id) keep working.
+  def to_param
+    return id&.to_s if title.blank?
+    "#{id}-#{title.parameterize}"
```
maebeale, line 127: 🤖 _From Claude:_ The leading id is what makes this work with zero schema change: Rails `find` casts `"23-my-great-story"` via `to_i` → `23`, so both the slug and a bare `/stories/23` resolve the same record. `set_story` stays a plain `Story.find`. Edit links deliberately pass `@story.id` (not the object) to keep `/stories/23/edit` numeric.

jmilljr24 (approved):
This is a nice simple solution.

PR #1952, opened 2026-07-08
Uniform-size image carousel on workshop variation show

maebeale:
🤖 suggested review level: 3 Read 📖 contained view/JS change on one page; carousel controller now reads its options value (touches a second carousel's gap)

## What
- Workshop variation **show** page now renders primary + gallery images as a single Swiper carousel where every image is forced into the same `aspect-[3/2]` box — no more full-width hero + 128px thumbnails.
- Made `carousel_controller.js` read `data-carousel-options-value` (was declared but ignored). New `_display_assets_carousel.html.erb` uses it for a one-at-a-time, `max-w-2xl`-centered layout.

## Why
- Hero image was too large and secondary images too small, on the variation page. Rachel suggested a carousel forcing all images to one size.

## Notes
- Side effect: the tagged-items carousel's declared `spaceBetween: 16` now actually applies (gap 30→16) since its options are no longer ignored. Its slide count is unchanged.
- Scope limited to the variation show page per request; the workshop page and index thumbnails are untouched.

### DIFF
--- a/app/frontend/javascript/controllers/carousel_controller.js
+++ b/app/frontend/javascript/controllers/carousel_controller.js
@@ -6,15 +6,18 @@ import 'swiper/css'
 import 'swiper/css/navigation'
 
 // Connects to data-controller="carousel"
+// Pass Swiper overrides via data-carousel-options-value='{ "slidesPerView": 2, ... }'.
+// Any keys provided there win over the defaults below (breakpoints replace wholesale).
 export default class extends Controller {
+  static values = { options: Object }
 
   connect() {
-    this.swiper = new Swiper(this.element, {
+    const defaults = {
       modules: [Navigation],
       loop: true,
       speed: 1400,
       spaceBetween: 30,
-      
+
       breakpoints: {
         320: {
           slidesPerView: 1,
@@ -29,12 +32,14 @@ export default class extends Controller {
           slidesPerView: 4,
         }
       },
-      
+
       navigation: {
         nextEl: '.swiper-button-next-custom',
         prevEl: '.swiper-button-prev-custom',
       }
-    })
+    }
+
+    this.swiper = new Swiper(this.element, { ...defaults, ...this.optionsValue })
   }
 
   disconnect() {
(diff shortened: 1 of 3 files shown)


app/views/assets/_display_assets_carousel.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  assets << resource.primary_asset if resource.respond_to?(:primary_asset) && resource.primary_asset&.file&.attached?
+  assets += resource.gallery_assets.select { |asset| asset.file.attached? } if resource.respond_to?(:gallery_assets)
+%>
+<% if assets.any? %>
+  <div data-controller="carousel"
+       class="relative swiper my-6 max-w-2xl mx-auto"
+       data-carousel-options-value='{ "loop": false, "spaceBetween": 16, "slidesPerView": 1, "breakpoints": {} }'>
+    <div class="swiper-wrapper">
+      <% assets.each_with_index do |asset, idx| %>
+        <div class="swiper-slide">
```
maebeale, line 14: 🤖 _From Claude:_ Primary + gallery assets are flattened into one slide list, each in the same `aspect-[3/2]` box with `object-cover`, so every image renders at one uniform size regardless of its native dimensions. Nav buttons only render when there is more than one image.

app/frontend/javascript/controllers/carousel_controller.js
```diff
@@ -29,12 +32,14 @@ export default class extends Controller {
...
-      
+
       navigation: {
         nextEl: '.swiper-button-next-custom',
         prevEl: '.swiper-button-prev-custom',
       }
-    })
+    }
+
+    this.swiper = new Swiper(this.element, { ...defaults, ...this.optionsValue })
```
maebeale, line 42: 🤖 _From Claude:_ Shallow merge on purpose — a caller that overrides `slidesPerView` must also pass `breakpoints` (even `{}`) to clear these defaults, or the per-breakpoint values here win. That is why the new image carousel sends `"breakpoints": {}`.

maebeale:
Screenshot 2026-07-08 at 7 27 18 AM

jmilljr24 (approved):

PR #1953, opened 2026-07-08
NEEDS DISCUSSION: Add admin asset library (/asset_library) with editable filename + caption

maebeale:
🤖 suggested review level: 3 Read 📖 new admin library with contained filter/search logic, a decorator, an admin-only policy, and inline edits that mutate title + Active Storage blob filename; no schema changes

- Admins had no single place to see every uploaded asset. This adds an admin-only **Asset library** (`/asset_library`) listing all assets — `PrimaryAsset`, `GalleryAsset`, `RichTextAsset`, `DownloadableAsset` — including PDFs/docs.
- Two things about a file were previously hard/impossible to edit, now both are **inline-editable** per asset:
  - **Caption** — `Asset#title` (metadata label)
  - **Filename** — the Active Storage blob filename (what downloads)
- Results can be narrowed by **type**, **attached-to record**, **file type** (jpg/png/pdf/…), and **search visibility**.
- A **"New resource"** button opens the resource form with **"Hidden from search" pre-checked**, so newly created resources start hidden.

- `Asset.search` matches a keyword against title, attached filename (join `active_storage_blobs`), and the owner model name; `present_types`/`present_owner_types`/`present_content_types` power the dropdown filters. File-type filter keys off the blob `content_type` (not the filename string), so a mislabeled extension can't slip through.
- **Search-visibility filter** keys off the owning **Resource**'s `hidden_from_search` (a Resource-only concept): "Hidden from search" returns only assets attached to a hidden resource; "Searchable" returns everything else. No asset-level column — the filter composes as an id-membership subquery so it stacks with the other filters.
- `AssetsController#index` follows the standard lazy `turbo_frame` + filter pattern. `#update` handles the two inline fields independently — each is its own turbo frame.
- The **"New resource"** button links to `new_resource_path(hidden_from_search: true)`; `resources#new` now honors that param to pre-check the box.
- Served at `/asset_library`, **not** `/assets`: the sprockets pipeline owns the `/assets/*` prefix and would shadow `/assets/:id`.
- `AssetPolicy` gates index + inline edits to admins (`with: AssetPolicy`, so it holds across STI subtypes).
- `AssetDecorator` centralizes type label, owner label/link, filename, content-type badge, and thumbnail.
- The library container spans the full content width (`max-w-none`), and its admin-home card now sits above Banners.

### Anything else to add?
- Non-image files show a file-type badge instead of a thumbnail; assets with no attached file show "No file attached" (no filename field).
- Specs: model search + `attached_to_hidden_resource` + `present_content_types`, request (auth, filters incl. file type + resource visibility, caption edit, filename rename), policy, page_bg alignment.


app/controllers/images_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  def update
+    authorize! @image, to: :update?, with: AssetPolicy
+
+    status = @image.update(image_params) ? :ok : :unprocessable_content
+    render partial: "images/title_field", locals: { image: AssetDecorator.decorate(@image) }, status: status
+  end
+
+  private
+
+  def set_image
```
maebeale: 🤖 _From Claude:_ Authorized with `with: AssetPolicy` explicitly so the rule is admin-only for every STI subtype. Without it, ActionPolicy would resolve to the subtype policy (e.g. `PrimaryAssetPolicy#update? = authenticated?`), which is looser than we want for the index.

app/models/asset.rb
```diff
@@ -64,6 +64,29 @@ def self.accepted_types_label
...
+  # Admin image index: keyword search across title, attached filename, and the
+  # type of record the asset is attached to (owner_type). One attachment per
+  # asset (has_one_attached), so the left join never multiplies rows.
+  def self.search(query)
+    return all if query.blank?
+
+    like = "%#{sanitize_sql_like(query)}%"
+    left_joins(:file_blob).where(
+      "assets.title LIKE :like OR assets.owner_type LIKE :like OR active_storage_blobs.filename LIKE :like",
+      like: like
```
maebeale, line 89: 🤖 _From Claude:_ `left_joins(:file_blob)` is safe from row multiplication here because `has_one_attached :file` means at most one attachment per asset — so no `distinct` needed, and counts/pagination stay correct.

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   resources :primary_assets
   resources :rich_text_assets
+  # Admin asset library. Served at /asset_library rather than /assets because the
```
maebeale, line 4: 🤖 _From Claude:_ Routed at `/asset_library` rather than `/assets` on purpose: the sprockets pipeline claims the `/assets/*` prefix, so a `resources :assets` member route (`/assets/:id`) would be shadowed by asset serving, and the generated `asset_path` helper would collide with ActionView's `asset_path`.

app/controllers/assets_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  # Inline edit of an asset's caption (title) and/or download filename from the
+  # library. The two fields submit as independent turbo frames; whichever field
+  # changed is present in the params.
+  def update
+    authorize! @asset, to: :update?, with: AssetPolicy
+
+    @asset.title = params.dig(:asset, :title) if params[:asset]&.key?(:title)
+    rename_file(params.dig(:asset, :filename))
```
maebeale, line 43: 🤖 _From Claude:_ Caption (`title`) and filename (blob) are edited independently — each field is its own turbo frame. The response renders both frames (`_editable_fields`) and Turbo swaps only the one that submitted, so editing the filename never clobbers an in-progress caption edit and vice versa.

maebeale:
Screenshot 2026-07-08 at 10 15 17 AM

app/controllers/assets_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  # and the model it's attached to.
+  def index
+    authorize! Asset, to: :index?, with: AssetPolicy
+
+    if turbo_frame_request?
+      per_page = params[:number_of_items_per_page].presence || 24
+      base = Asset.includes(:owner, file_attachment: :blob)
+      base = base.where(type: params[:type]) if params[:type].present?
+      base = base.where(owner_type: params[:owner_type]) if params[:owner_type].present?
+      base = base.where(id: Asset.joins(:file_blob).where(active_storage_blobs: { content_type: params[:content_type] })) if params[:content_type].present?
```
maebeale, line 15: 🤖 _From Claude:_ Filtered via a `where(id: …)` subquery on the blob join rather than chaining `.joins(:file_blob)` onto `base` — `base` already `includes(file_attachment: :blob)`, and a second join referencing `active_storage_blobs` would collide with the eager-load alias. The subquery keeps the two blob references independent.

app/controllers/assets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    status = @asset.save ? :ok : :unprocessable_content
+    render partial: "assets/editable_fields", locals: { asset: AssetDecorator.decorate(@asset.reload) }, status: status
+  end
+
+  private
+
+  # STI subclass to instantiate, constrained to the known TYPES so a stray param
+  # can't constantize into an arbitrary class.
+  def asset_class
```
maebeale: 🤖 _From Claude:_ `create` instantiates the STI subclass via `presence_in(Asset::TYPES)` before `constantize` — a stray `type` param can't `constantize` into an arbitrary class, and an unknown/blank type safely falls back to the first known type. The permitted `:type` is stripped from the attributes hash since the subclass already fixes it.

app/controllers/assets_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  def index
+    authorize! Asset, to: :index?, with: AssetPolicy
+
+    if turbo_frame_request?
+      per_page = params[:number_of_items_per_page].presence || 24
+      base = Asset.includes(:owner, file_attachment: :blob)
+      base = base.where(type: params[:type]) if params[:type].present?
+      base = base.where(owner_type: params[:owner_type]) if params[:owner_type].present?
+      base = base.where(id: Asset.joins(:file_blob).where(active_storage_blobs: { content_type: params[:content_type] })) if params[:content_type].present?
+      hidden_ids = Asset.attached_to_hidden_resource.select(:id)
```
maebeale, line 16: 🤖 _From Claude:_ Visibility keys off the owning Resource's hidden_from_search (a Resource-only flag), via an id-membership subquery so it composes with the type/owner/file-type filters. "searchable" is the complement (`where.not(id: hidden_ids)`), which correctly includes non-resource-owned assets and assets on non-hidden resources.

PR #1957, opened 2026-07-09
Unify workshop_variations index onto the shared sortable-table pattern

maebeale:
🤖 suggested review level: 3 Read 📖 contained refactor onto an existing index pattern + a helper extraction, covered by request/view/helper specs

- Workshop variations was the only columnar index still rendered full-page with a single hand-rolled Author sort; every other table index (stories, community_news, bookmarks) uses a lazy Turbo-frame + sortable column headers.
- Brings it onto that shared pattern and removes the sort-header duplication those pages carried inline.

- Extracted `SortableHeaderHelper` (`sort_icon_class` + `sort_header_link`) from the `sort_base`/`sort_icon`/`sort_link` lambdas that were copy-pasted inline in **stories**, **community_news**, and **bookmarks**; all three now call the helper.
- Converted `workshop_variations#index` to branch on `turbo_frame_request?`, added the `_results`/`_count`/skeleton partials + `workshop_variations_results` frame, and made Name / Author / Updated sortable via the shared helper. Search box now submits into the frame.
- Specs: request specs use the `Turbo-Frame` header for sort/search; added an index view spec and a helper spec.

### Anything else to add?
- **Not changed:** `workshops` and `resources` are card/list layouts (radio-button sort / none) — column-header sorting doesn't fit them, so they're intentionally left alone.
- The other framed tables (allocations, payments, organizations, people, users, monthly_reports) currently have **no** sorting at all; they could adopt this helper if/when sorting is added, but that's out of scope here.
- Grants uses a different client-side Stimulus sort (`shared/_sortable_header`) and is deliberately not folded in.

### DIFF
--- a/app/controllers/workshop_variations_controller.rb
+++ b/app/controllers/workshop_variations_controller.rb
@@ -3,19 +3,31 @@ class WorkshopVariationsController < ApplicationController
   def index
     authorize!
     @author = Person.find_by(id: params[:author_id]) if params[:author_id].present?
+    if turbo_frame_request?
+      per_page = params[:number_of_items_per_page].presence || 25
+      base_scope = WorkshopVariation.includes(:workshop, :author, :windows_type,
+                                              :workshop_variation_idea, created_by: :person)
+                                    .joins(:workshop).where(workshops: { published: true })
+      filtered = base_scope.search_by_params(params)
+      @sort = %w[name author updated_at created_at].include?(params[:sort]) ? params[:sort] : "created_at"
+      @sort_direction = params[:direction] == "asc" ? "asc" : "desc"
+      filtered = case @sort
+      when "author"
+        filtered.order_by_author(@sort_direction)
+      when "name"
+        filtered.reorder(Arel.sql("workshop_variations.name #{@sort_direction}"))
+      when "updated_at"
+        filtered.reorder(updated_at: @sort_direction.to_sym)
+      else
+        filtered.reorder(Arel.sql("workshop_variations.created_at #{@sort_direction}, workshops.title, workshop_variations.name"))
+      end
+      @workshop_variations = filtered.paginate(page: params[:page], per_page: per_page).decorate
+      @count_display = filtered.count == base_scope.count ? base_scope.count : "#{filtered.count}/#{base_scope.count}"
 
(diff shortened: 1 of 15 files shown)


app/helpers/sortable_header_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  #   sort_link.call("title", "Title")
+  #
+  # `base` carries the current search/filter params so they survive a re-sort.
+  # Clicking a fresh column sorts descending, then toggles to ascending.
+  def sort_header_link(frame:, path:, base: {})
+    ->(column, label) do
+      direction = (@sort.to_s == column.to_s && @sort_direction == "desc") ? "asc" : "desc"
+      link_to path.call(base.merge(sort: column, direction: direction, page: nil)),
+              data: { turbo_frame: frame },
+              class: "inline-flex items-center gap-1 text-gray-700 hover:text-gray-900" do
```
maebeale, line 27: 🤖 _From Claude:_ Toggle preserves the prior behavior exactly: a fresh column starts descending, and an already-active column flips desc→asc. This is the identical rule that was inlined in stories/community_news/bookmarks, now in one place.

app/controllers/workshop_variations_controller.rb
```diff
@@ -3,19 +3,31 @@ class WorkshopVariationsController < ApplicationController
...
     authorize!
     @author = Person.find_by(id: params[:author_id]) if params[:author_id].present?
+    if turbo_frame_request?
+      per_page = params[:number_of_items_per_page].presence || 25
+      base_scope = WorkshopVariation.includes(:workshop, :author, :windows_type,
+                                              :workshop_variation_idea, created_by: :person)
+                                    .joins(:workshop).where(workshops: { published: true })
+      filtered = base_scope.search_by_params(params)
+      @sort = %w[name author updated_at created_at].include?(params[:sort]) ? params[:sort] : "created_at"
+      @sort_direction = params[:direction] == "asc" ? "asc" : "desc"
```
maebeale, line 13: 🤖 _From Claude:_ Sortable columns are whitelisted (unknown/absent `sort` falls back to `created_at`), matching stories/community_news. The default branch keeps the original multi-key ordering (created_at, workshops.title, name) so unsorted output is unchanged.

PR #1958, opened 2026-07-10
Visibility checkbox definitions on story form (info popover + hover)

maebeale:
🤖 suggested review level: 3 Read 📖 shared partial + small helper rolled across 13 admin forms; view markup + copy with request-spec coverage, low blast radius

- Non-technical admins couldn't tell what the Published / Featured / Publicly visible / Publicly featured checkboxes actually do — especially the **dependency ladder** the `Publishable`/`Featureable` scopes enforce silently (publicly visible needs published; publicly featured needs both).

- Added `shared/_visibility_info` (the "Visibility" heading + an **ⓘ hover/focus popover** listing plain-language definitions — pure CSS, no JS) and `VisibilityFlagsHelper#visibility_flag_input`.
- Each checkbox now carries **three layers of help**: a short **hint under** the box, the full **definition on hover**, and the **ⓘ popover** for the whole section.
- Standardized every visibility section into a right-hand **`admin-only bg-blue-100`** card in a two-column layout, across **13 forms**: stories, community_news, resources, video_recordings, workshops, events, workshop_variations, faqs, banners, sectors, categories, quotes, category_types.
- Copy lives in `VisibilityFlagsHelper::FLAG_DEFINITIONS` (kept next to the scopes it describes). `visibility_flag_input` accepts `definition_key:` so a repurposed flag can use different copy.

### Special cases
- **category_types** — its `published` gates child categories (not portal visibility), so it uses its own `category_type_published` definition; `story_specific`/`profile_specific` added to the helper and carded too.
- **hidden_from_search** (resources), **public_registration_enabled** (events), **is_instructional/is_podcast** (video_recordings) get accurate, dependency-aware copy and request-spec coverage.
- **organization_statuses** — intentionally left out: its `published` column is a no-op (nothing reads it), so a visibility definition would be misleading.

### Testing
- Helper spec (hint + hover title + `definition_key`).
- Request specs assert the visibility checkboxes + copy render on the special-condition and previously-uncovered forms (category_types, quotes, faqs, resources, events, video_recordings, workshop_variations, stories).


app/helpers/visibility_flags_helper.rb
```diff
@@ -0,0 +1,50 @@
+module VisibilityFlagsHelper
+  # Plain-language definitions for the publish/visibility checkboxes on the admin
+  # portal edit forms. The copy mirrors what the Publishable and Featureable
+  # scopes actually enforce (see app/models/concerns/publishable.rb and
+  # featureable.rb) — note the dependency ladder: publicly visible requires
+  # published, and publicly featured requires both. Keep them in sync.
```
maebeale: 🤖 _From Claude:_ Copy is deliberately co-located with the `Publishable`/`Featureable` scopes it describes — the "requires Published" / "requires both" wording tracks the actual scope chaining, so update it here if those scopes change.

app/frontend/javascript/controllers/info_popover_controller.js
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+      this.close();
+    }
+  }
+
+  open() {
+    this.panelTarget.classList.remove("hidden");
+    document.addEventListener("click", this.closeOnOutsideClick);
+    document.addEventListener("keydown", this.closeOnEscape);
+  }
```
maebeale: 🤖 _From Claude:_ No setTimeout guard needed: the outside-click listener is added during the opening click, and the DOM spec skips listeners added mid-dispatch, so the panel does not immediately re-close.

jmilljr24 (approved):

maebeale:
Screenshot 2026-07-10 at 7 03 13 PM
Screenshot 2026-07-10 at 7 03 07 PM

PR #1940, opened 2026-07-05
Standardize author crediting across creditable models

maebeale:
🤖 suggested review level: 5 Inspect 🔬 shared credit/search/sort/filter concern across 5 models + 3 idea models, a column rename, validation changes, and view/controller rewrites

Makes Workshop, Resource, Story, CommunityNews, and WorkshopVariation (and the `*_idea` models) credit, search, sort, and filter by author identically, via `AuthorCreditable`.

**Why**
- Workshop's list linked/printed raw full names, ignoring the credit preference (an "anonymous" author leaked their name).
- Workshop/WorkshopVariation linked to author profiles without the `profile_is_searchable` gate the others use.
- Author search/sort coverage ranged from complete (Workshop) to none (WorkshopVariation), each implemented differently.

**Display**
- One byline helper (`credited_author_link`) everywhere. Text always honors the credit preference (`author_credit`).
- Links **only** to an explicit/legacy author's profile — a creator-fallback credit is shown as plain text (that person never declared authorship and the record isn't on their profile's authored-content cards). Anonymous never links.
- Precedence: explicit author → legacy free-text/user author → creator's person → per-model `missing_author_label`. Unattributed labels: community news / resources / stories → **"AWBW Staff"**, workshops / workshop variations → **"Facilitator"**.

**Search / sort / filter (all shared on the concern)**
- `by_credited_person_name` (search) matches every name shown — explicit author, legacy author, and creator. Added to Resource + WorkshopVariation, which had none; Workshop's legacy `full_name` is folded in.
- `order_by_author` (sort) by first-then-last name, matching the displayed credit. Added to Workshop + WorkshopVariation.
- `authored_by(person_id)` (filter) — one scope replacing the inline `where(author_id:)` main duplicated across five paths.
- Preserve `author_id` (main's person-scope filter) across sorts/filters, which the sort links + workshops form were dropping.

**Credit preference**
- Admin-created records default a blank preference to `full_name` (UI + on save), no backfill.
- Public idea submissions (`*_idea`) must choose — `require_author_credit_preference` restores presence and drops the default, so a submitter isn't silently opted into publishing their full name.

**Data**
- Rename legacy `community_news.user_author_id` → `legacy_author_user_id`.

**Deferred (needs a yes)**
- Renaming `workshops.full_name` → `legacy_author_name`: welded into a 14-column MySQL FULLTEXT index + the admin form. Folded in under its current name so search/sort/display already work.

PR #1959, opened 2026-07-10
Fix dev seeds broken by author-crediting standardization (#1940)

maebeale:
🤖 suggested review level: 3 Read 📖 two-line dev-seed fix, no app logic

- `db:seed:dev` (and `bin/setup`) aborts after #1940 merged. Restores a working dev seed.

- `WorkshopIdea` now calls `require_author_credit_preference` (#1940 dropped its `full_name` default) — the seed now sets `author_credit_preference`, matching the sibling idea blocks.
- #1940 flipped `CommunityNews#author_id` from a User to a Person FK — the seed now assigns a `Person` id (nil-weighted like `resources.rb`) instead of a `User` id, which would FK-violate.

### Anything else to add?
- Checked every other author reference in the seeds (`resources.rb`, `workshops.rb`, `analytics.rb`, `events_management.rb`) — all already consistent with the current schema.

### DIFF
--- a/db/seeds/dev/home_page_content.rb
+++ b/db/seeds/dev/home_page_content.rb
@@ -34,7 +34,7 @@ puts "Creating CommunityNews…"
                .first_or_create!(
                   body: body_content,
                   rhino_body: "<p>#{body_content}</p>",
-                  author_id: User.all.sample&.id,
+                  author_id: [ Person.all.sample, nil, nil ].sample&.id,
                   created_by_id: User.first&.id,
                   updated_by_id: User.first&.id,
                   organization_id: Organization.all.sample&.id,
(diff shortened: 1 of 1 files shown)


db/seeds/dev/home_page_content.rb
```diff
@@ -34,7 +34,7 @@
                .first_or_create!(
                   body: body_content,
                   rhino_body: "<p>#{body_content}</p>",
-                  author_id: User.all.sample&.id,
+                  author_id: [ Person.all.sample, nil, nil ].sample&.id,
```
maebeale, line 37: 🤖 _From Claude:_ #1940 made `CommunityNews#author_id` a Person FK (was a User reference); this now seeds a `Person` id so it doesn't FK-violate against the `people` table.

db/seeds/dev/home_page_content.rb
```diff
@@ -73,6 +73,7 @@
   "Empowerment Through Mixed Media"
 ].each do |title|
   WorkshopIdea.where(title: title).first_or_create!(
+    author_credit_preference: AuthorCreditable::AUTHOR_CREDIT_PREFERENCES.sample,
```
maebeale, line 76: 🤖 _From Claude:_ #1940 added `require_author_credit_preference` to `WorkshopIdea`, dropping the `full_name` default — so this block now needs an explicit preference like its `StoryIdea`/`WorkshopVariationIdea` siblings.

PR #1961, opened 2026-07-10
Fix 500 in rich text asset mentions for decorated records

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained controller/helper guard fix with a reproducing spec

- Fixes a prod 500 (`NoMethodError: undefined method 'rich_text_assets' for an instance of EventDecorator`) when typing `!`-mentions in the Event editor.

- The rhino helper embedded the **decorator's** signed GlobalID (Event forms decorate the record), so the mentions endpoint located an `EventDecorator` and called `.rich_text_assets` — an association Event doesn't own.
- Emit the underlying **model's** sgid from the helper, and guard the association lookup in the controller so records that don't own rich text assets return `[]` instead of raising.
- Added a request spec that reproduces the exact 500 first, then passes.

### Anything else to add?
- Honeybadger doesn't record the offending user here (no `Honeybadger.context`); a follow-up PR adds that.

### DIFF
--- a/app/controllers/rich_text_asset_mentions_controller.rb
+++ b/app/controllers/rich_text_asset_mentions_controller.rb
@@ -2,8 +2,7 @@ class RichTextAssetMentionsController < ApplicationController
   skip_verify_authorized
   def index
     # authorize!
-    record = GlobalID::Locator.locate_signed(params[:sgid])
-    record = record.object if record.respond_to?(:decorated?) && record.decorated?
+    record = Draper.undecorate(GlobalID::Locator.locate_signed(params[:sgid]))
 
     unless record.respond_to?(:rich_text_assets)
       render json: [] and return
--- a/app/controllers/rich_text_asset_mentions_controller.rb
+++ b/app/controllers/rich_text_asset_mentions_controller.rb
@@ -3,8 +3,9 @@ class RichTextAssetMentionsController < ApplicationController
   def index
     # authorize!
     record = GlobalID::Locator.locate_signed(params[:sgid])
+    record = record.object if record.respond_to?(:decorated?) && record.decorated?
 
-    unless record
+    unless record.respond_to?(:rich_text_assets)
       render json: [] and return
     end
 
--- a/app/helpers/rhino_editor_helper.rb
+++ b/app/helpers/rhino_editor_helper.rb
@@ -1,8 +1,7 @@
 module RhinoEditorHelper
   # custom rhino editor with stimulus controller attached to edit raw source html
   def rhino_editor(form, base_attribute_name, label: nil, hint: nil)
-    object = form.object
-    object = object.object if object.respond_to?(:decorated?) && object.decorated?
+    object = Draper.undecorate(form.object)
     rhino_attr = :"rhino_#{base_attribute_name}"
     field_id = form.field_id(rhino_attr)
     value = object.public_send(rhino_attr)
--- a/app/helpers/rhino_editor_helper.rb
+++ b/app/helpers/rhino_editor_helper.rb
@@ -1,9 +1,11 @@
 module RhinoEditorHelper
   # custom rhino editor with stimulus controller attached to edit raw source html
   def rhino_editor(form, base_attribute_name, label: nil, hint: nil)
+    object = form.object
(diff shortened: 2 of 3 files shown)


app/controllers/rich_text_asset_mentions_controller.rb
```diff
@@ -3,8 +3,9 @@ class RichTextAssetMentionsController < ApplicationController
   def index
     # authorize!
     record = GlobalID::Locator.locate_signed(params[:sgid])
+    record = record.object if record.respond_to?(:decorated?) && record.decorated?
 
-    unless record
+    unless record.respond_to?(:rich_text_assets)
```
maebeale, line 7: 🤖 _From Claude:_ Guarding on `respond_to?(:rich_text_assets)` (after unwrapping the decorator) rather than the class list means non-owner records like Event degrade to an empty mention list instead of 500ing.

maebeale:
@jmilljr24 this is in response to the HB, so if you approve, feel free to merge and deploy!

jmilljr24:
> @jmilljr24 this is in response to the HB, so if you approve, feel free to merge and deploy!
> 

Can you share more details of the error? 

This change seems ok on the defensive side of things but I don't see any changes related to why a decorator was passed in the first place.

maebeale:
🤖 _From Claude:_

**The error** (Honeybadger): `NoMethodError: undefined method 'rich_text_assets' for an instance of EventDecorator` at `rich_text_asset_mentions_controller.rb:11`, triggered by a `GET /rich_text_asset_mentions.json` request while someone was editing an Event.

**Why a decorator was passed in the first place:** `EventsController` decorates the record in `new`/`edit`/`create`/`update` (`@event = @event.decorate`), so the Event form's `f.object` is an `EventDecorator`. The rhino editor helper embedded `form.object.to_sgid` into the editor — i.e. the *decorator's* signed GlobalID. When the `!`-mention endpoint located that sgid back, it got an `EventDecorator` and called `.rich_text_assets`, an association Event doesn't own → 500.

**The root cause is fixed here, not just guarded:** the change in `rhino_editor_helper.rb` unwraps the decorator (`object.object if object.decorated?`) and emits the underlying **model's** sgid, so a decorator sgid is never generated. The controller `respond_to?` guard is defense-in-depth — and also handles Event legitimately (it owns no rich text assets, so it just returns an empty mention list).

(Separately, #1962 adds `user_id`/`user_email` to Honeybadger context so future faults name the affected user directly — this one didn't record it.)

app/helpers/rhino_editor_helper.rb
```diff
@@ -1,9 +1,11 @@
 module RhinoEditorHelper
   # custom rhino editor with stimulus controller attached to edit raw source html
   def rhino_editor(form, base_attribute_name, label: nil, hint: nil)
+    object = form.object
+    object = object.object if object.respond_to?(:decorated?) && object.decorated?
```
jmilljr24:
```suggestion
    record = Draper.undecorate(form.object)
```
Just a little cleaner.

jmilljr24 (changes requested):
Let's use `Draper.undecorate` for consistency in the codebase.

jmilljr24:
I missed that there were two issues going on. The AI text is accurate but some times a human touch is nice.

app/helpers/rhino_editor_helper.rb
maebeale: 🤖 _From Claude:_ Applied — much nicer, and I used it in the controller too. Thanks!

maebeale:
oops i triggered a robot reply to your human note. sorry, @jmilljr24 !

jmilljr24 (approved):

maebeale:
@jmilljr24 we had another error in prod -- i think this was you? if you're good w this pr, would you take it through to prod (or change to what you think best and ship?)? i'm going to be offline a lot of the day today and tomorrow.

jmilljr24:
> @jmilljr24 we had another error in prod -- i think this was you? if you're good w this pr, would you take it through to prod (or change to what you think best and ship?)? i'm going to be offline a lot of the day today and tomorrow.

Yes, I will deploy it. The error is silent to the user so it's no rush though.

PR #1962, opened 2026-07-10
Add signed-in user to Honeybadger fault context

maebeale:
🤖 suggested review level: 3 Read 📖 one before_action adding user context to error reports, with a spec

- Faults (e.g. the recent rich-text-mentions 500) don't name the affected user, forcing triage to decode request params to figure out who hit the error.

- Add a `set_honeybadger_context` before_action in `ApplicationController` that attaches `user_id`/`user_email` for authenticated requests only; anonymous visitors get no user context.
- Added a request spec covering both the signed-in and anonymous cases.

### Anything else to add?
- Follow-up to #1961 (the fix for the fault that surfaced this gap).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_user! # ensures only logged-in users can access pages
   before_action :track_user_with_ahoy, unless: :devise_controller?
   before_action :set_current_user # for AhoyTrackable in models
+  before_action :set_honeybadger_context # so faults name the affected user
   before_action :set_paper_trail_whodunnit
   before_action :preload_current_user_associations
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -73,4 +74,10 @@ def preload_current_user_associations
   def set_current_user
     Current.user = current_user if user_signed_in? # needed for Ahoy tracking in models
   end
+
+  def set_honeybadger_context
+    return unless user_signed_in?
```
maebeale, line 79: 🤖 _From Claude:_ Scoped to `user_signed_in?` so anonymous requests carry no user context — a spec asserts `context` is never called with a `user_id` key when logged out.

maebeale:
@jmilljr24 what do you think of doing this to help us know who to contact when a honeybadger happens?

jmilljr24 (approved):
Doesn't hurt to have a bit more info. Thanks.

PR #1917, opened 2026-06-25
CE cutover: reroute intake/display, drop ce_* columns (PR 2)

maebeale:
🤖 PR, suggested 👤 review level: 🔬 Inspect — reroutes CE intake/display onto the new models, drops the old `ce_*`/`credentials` columns, and adds the admin CE management + professional-license surfaces

> Foundation **#1916** (the CE models) and **#1833** (registrants-index CE column + filter) have both merged; this branch is rebased onto `main` and reconciled with them.

### What is the goal of this PR and why is this important?

#1916 added `ProfessionalLicense` + `ContinuingEducationRegistration` but left intake, the callout, and read sites on the flat `EventRegistration#ce_*` columns. This PR completes the cutover, drops those columns, and builds out the admin + registrant surfaces for managing CE and professional licenses.

### Cutover (intake, callout, read sites)

- **Intake** — the CE-interest answer becomes a plain Yes/No opt-in (hours come from the event); opting in creates a CE registration against a found-or-created license.
- **Public callout** — a registrant license form (type, number, US-states issuing-state dropdown, expiry), required vs `(optional)` markers, greyed placeholders.
- **Read sites** — magic-ticket card, reminders, onboarding, CSV rerouted onto the new models; `#ce_hours` page gated on `ce_eligible?`.
- **`EventRegistration` aggregators** — `ce_requested?`, `ce_hours_total`, `ce_amount_owed_cents`, `ce_license_provided?`, `ce_paid_in_full?`, `ce_status_label` over the CE registrations; `ce_status` filter scope.
- **Drop** the `ce_credit_requested` / `ce_hours_requested` / `ce_license_number` columns (+ the legacy `credentials` column on people) and dead methods.

### Admin CE management

- Dedicated **CE registration new/edit page** (`ContinuingEducationRegistrationsController` + policy): license type/number/issuing-state/expiry, hours + cost (prefilled from the event), certificate issued toggle, and delete.
- Cross-links both ways: admin "Registrant's CE page" / "Registrant's scholarship page" pills, with the registrant callouts showing an "Edit" pill and a "← Back to CE registration / scholarship" eyebrow when reached from admin.

### Professional licenses on the person form + edit gating

- Licenses (`ProfessionalLicense`: type, number, issuing-state, expiry) are managed on the person form; uniqueness scoped to `(person, kind, number)`.
[...]

PR #1867, opened 2026-06-22
Hide scholarship/CE registration boxes when they don't apply

maebeale:
### What is the goal of this PR and why is this important?

- The registration edit form's **Organizations / Scholarship / CE credits** row showed the Scholarship and CE boxes unconditionally.
- Scholarship is only meaningful when there's a fee to subsidize, so it shouldn't appear on **free events**.
- Continuing-education credits aren't offered for every event, so admins need a way to flag eligibility — and the CE box should hide when an event doesn't grant credits.

### How did you approach the change?

- Added `ce_credits_eligible` to `Event` (boolean, **default `true`**, not null) via migration; permitted it in `EventPolicy` and surfaced a checkbox under Event Cost on the event form.
- On the registration form (`event_registrations/_form`):
  - Hide the **Scholarship** box when the event is free (`cost_cents == 0`).
  - Hide the **CE credits** box when `event.ce_credits_eligible?` is false.
  - The **Organizations** box absorbs the freed columns in the 4-col grid:
    - paid + CE-eligible → orgs at `sm:col-span-2` (both boxes show)
    - one box hidden → orgs at `sm:col-span-3`
    - free + not CE-eligible → orgs at `sm:col-span-4` (full width)

### UI Testing Checklist

- [ ] Event form shows "Eligible for continuing education credits" (checked by default)
- [ ] Paid + CE-eligible registration: Organizations (half), Scholarship, CE credits all show
- [ ] Free event: Scholarship box hidden, Organizations widens
- [ ] Event with CE eligibility unchecked: CE box hidden, Organizations widens
- [ ] Free + not CE-eligible: only Organizations, full width

### Anything else to add?

- 🚧 **HOLD** — opening as a draft for review of the approach before marking ready.
- Open question: the request mentioned the **ticket** and **registrations index** displays too. Those currently render no Scholarship/CE/Organizations boxes (ticket shows cost/payment only; index is a plain table), so there was nothing to gate there. Happy to add CE/scholarship surfacing to those if that's the intent — let me know the desired design.
- Tests: model default, 4 system specs covering each visibility/width combination, plus existing edit-form specs (use the default paid + CE-eligible event) still pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1974, opened 2026-07-14
Hide the scholarship card on free-event registrations

maebeale:
🤖 PR, suggested 👤 review level: 📖 Read — small, contained view/logic change gated by a new predicate

- On the registration edit form, the **Scholarship** card rendered for every event, even free ones — where a scholarship (which offsets a fee) has nothing to do.
- The neighboring **CE credits** card already hides itself when the event grants no CE hours (`event.ce_eligible?`); scholarship had no equivalent gate.

- Added `Event#scholarship_eligible?`, mirroring the existing `#ce_eligible?` predicate: true when the event has a **cost** (a scholarship offsets a fee) **or** offers a **scholarship form** (it invites applications even when free).
- Gated the Scholarship card on it, **plus** the registration's own scholarships — so an award made before the event became free/formless stays visible instead of vanishing.
- The **Organizations** card absorbs whatever columns the hidden cards leave behind (`sm:col-span-1/2/3`), so the row never has a gap:
  - paid + CE hours → Orgs (1) · Scholarship · CE
  - paid, no CE hours → Orgs (2) · Scholarship
  - free + CE hours → Orgs (2) · CE
  - free, no CE hours, no form → Orgs (3, full width)

### UI Testing Checklist

- [ ] Paid event: Scholarship card shows
- [ ] Free event with no scholarship form: Scholarship card hidden, Organizations widens
- [ ] Free event that offers a scholarship form: card still shows
- [ ] Free event that already has an awarded scholarship: card still shows
- [ ] Row never leaves an empty column in any paid/free × CE combination

### Anything else to add?

- Supersedes the closed HOLD PR #1867, whose CE half was already implemented on main by the CE cutover (#1917).

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -294,9 +294,6 @@ class Event < ApplicationRecord
     ce_hours_offered.to_f.positive?
   end
 
-  # A scholarship offsets a registration fee, so it only applies to paid events —
-  # a free event has nothing to subsidize. Mirrors #ce_eligible? as a named
-  # "does this apply" predicate, derived from the cost rather than a stored flag.
   def scholarship_eligible?
     cost_cents.to_i.positive?
   end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -294,6 +294,13 @@ class Event < ApplicationRecord
     ce_hours_offered.to_f.positive?
   end
 
+  # A scholarship offsets a registration fee, so it only applies to paid events —
+  # a free event has nothing to subsidize. Mirrors #ce_eligible? as a named
+  # "does this apply" predicate, derived from the cost rather than a stored flag.
+  def scholarship_eligible?
+    cost_cents.to_i.positive?
+  end
+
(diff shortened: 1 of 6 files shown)


app/models/event.rb
```diff
@@ -294,6 +294,13 @@ def ce_eligible?
     ce_hours_offered.to_f.positive?
   end
 
+  # A scholarship offsets a registration fee, so it only applies to paid events —
+  # a free event has nothing to subsidize. Mirrors #ce_eligible? as a named
+  # "does this apply" predicate, derived from the cost rather than a stored flag.
+  def scholarship_eligible?
```
maebeale, line 297: 🤖 _From Claude:_ Final definition: eligible when the event has a cost **or** offers a scholarship form (there's no `scholarship_form_id` column — the form links through `event_forms` by role). The view additionally keeps the card when this registration already has an award, so an award that predates the event going free/formless isn't hidden.

PR #1975, opened 2026-07-14
Move W-9 to the payment callout and remove the Forms built-in

maebeale:
🤖 suggested review level: 5 Inspect 🔬 substantive callout logic across services/controller/views plus a data-deleting migration

- The **Forms** ticket callout was redundant — the invoice and paid-in-full receipt links already live on the **payment** page, so the W-9 was its only unique content.
- Moves the **W-9** onto the payment callout (a removable, per-event linked resource) and **removes the Forms built-in entirely**, consolidating registrant documents onto the payment page.

- `DefaultTicketCallouts`: moved the W-9 `resources:` proc from the `forms` definition to `payment`; deleted the `forms` definition.
- `MagicTicketCallouts`: removed the `forms` card, its `CARD_BUILDERS` entry, and its editor-preview card.
- Removed `forms` from `RegistrationTicketCallout::MAGIC_KEYS` and deleted `Event#show_forms_callout?`.
- `Events::CalloutsController`: dropped the `forms` action/route/view; the payment page now renders the callout's linked W-9 in its Documents section alongside invoice/receipt. Resource/invoice/receipt eyebrows return to **payment** instead of forms.
- Migration purges any already-materialized `forms` callout rows (they'd otherwise fail the `MAGIC_KEYS` inclusion validation).

### Anything else to add?
- No functionality is lost: invoice/receipt were already on the payment page; the W-9 now joins them there.


db/migrate/20260714113355_remove_forms_ticket_callouts.rb
```diff
@@ -0,0 +1,16 @@
+class RemoveFormsTicketCallouts < ActiveRecord::Migration[8.0]
+  # The "forms" built-in ticket callout has been removed — the W-9 moved onto the
+  # payment callout, and the invoice/receipt already live on the payment page. Any
+  # rows already materialized with magic_key "forms" no longer render and would
+  # fail the RegistrationTicketCallout MAGIC_KEYS inclusion validation on save, so
+  # purge them.
+  def up
+    execute("DELETE FROM registration_ticket_callouts WHERE magic_key = 'forms'")
```
maebeale, line 8: 🤖 _From Claude:_ This DELETE is irreversible and drops any per-event admin edits on materialized `forms` callouts. Only rows that were seeded before this change exist; they no longer render (the card builder is gone) and would fail `MAGIC_KEYS` validation on save, so purging is the intent — flagging so you can confirm the data loss is acceptable.

PR #1979, opened 2026-07-14
Collapsible <details> sections on every callout page (allowlisted, styled)

maebeale:
🤖 suggested review level: 5 Inspect 🔬 sanitizer allowlist change + shared rendering pipeline across 6 callout surfaces; FAQ now renders editable content

## Problem
Callout copy is admin-authored HTML run through `form_label_html`, whose allowlist dropped `<details>`/`<summary>` — so any collapsible section was **silently stripped on the next event save**.

## Approach (standard HTML, not a custom tag)
- Admins write standard `<details><summary>Title</summary>…</details>` — exactly what HTML generators / LLMs produce, so it can be pasted as-is.
- `<details>`/`<summary>` (+ the `open` attribute) are **added to the `form_label_html` allowlist**, so a disclosure is never lost on save — even on any path that skips the shared renderer.
- `CalloutContent` upgrades each disclosure into our on-brand styled toggle card (chevron, border, hover — reliable padding/animation that's awkward to get from CSS on nested prose) and honours `<details open>`. `<toggle>` and a `title` attribute remain accepted aliases.

## Uniform across every callout content page
One shared `events/callouts/_rich_content` renderer now backs: event details ("Art supplies & what to bring"), CE hours, custom callouts, behavioural-card intros, and the FAQ. Content outside a disclosure stays open; content with none renders unchanged.

## FAQ is now editable like everything else
The FAQ page renders the editable `faq` callout `description` (each question a `<details>`), falling back to `DefaultTicketCallouts.faq_html` when the card isn't materialized.

## Notes
- `<details>`/`<summary>` are inert; the same allowlist already permits far riskier tags (`iframe`, `form`, `button`, `input`).
- Editor help/placeholders and seed/default content use `<details>`/`<summary>`.

## Tests
- `form_label_html` keeps `<details>`/`<summary>` incl. the `open` state.
- `CalloutContent` unit spec (standard `<details>`, `open`, `<toggle>` alias, `title`, multiple, whitespace, fallback).
- Request specs: details renders each `<details>`; FAQ renders default toggles and, when materialized, the editable copy.
- Full events / callouts / registration / magic-callout / page_bg / helper suites green.


app/controllers/events/callouts_controller.rb
```diff
@@ -156,9 +161,14 @@ def videoconference
...
-    # the event shows the FAQ callout.
+    # the event shows the FAQ callout. Renders the editable FAQ callout copy (the
+    # admin edits it like every other callout, using the <toggle> syntax for each
+    # question), falling back to the code-defined default for events that haven't
+    # materialized the card yet.
     def faq
-      redirect_to registration_ticket_path(@event_registration.slug) unless @event.show_faq_callout?
+      return redirect_to(registration_ticket_path(@event_registration.slug)) unless @event.show_faq_callout?
+      callout = @event.registration_ticket_callouts.find_by(magic_key: "faq")
+      @faq_content = callout&.description.presence || DefaultTicketCallouts.faq_html
```
jmilljr24, line 176: When the event form is saved with faq published, don't we always have the description saved? (default or edited)

PR #1954, opened 2026-07-08
Materialize ticket callouts: editable, configurable, reorderable built-ins

maebeale:
🤖 suggested review level: 5 Inspect 🔬 schema migration with data move, seeder wired into create/edit, and ticket-render + editor changes across all built-in callouts (live registrant-facing behavior)

## Why
Follow-up to #1945. Turns every code-defined ("magic") ticket callout into an **editable / configurable per-event row** admins can edit, hide/restore, reorder, and link resources to — one unified, admin-ordered callout list.

## Model
`DefaultTicketCallouts` seeds all nine built-ins in canonical order on event create and lazily on edit (idempotent, no backfill), each gated by the event's config (`seed_if`). Two kinds:
- **Content cards — fully editable:** Handouts, FAQ (their copy + ordered resources; hide/restore).
- **Behavioral cards — app supplies live status, row governs visibility/drip/order:** Payment, Certificate, Scholarship, CE hours, Event details, Videoconference, Forms — rendered via `MagicTicketCallouts#card_for`.

`#cards` (fallback / not-yet-seeded events) and `#card_for` (row path) share one `CARD_BUILDERS` map, so a card's live state has a single source of truth.

## Configurability highlights
- **Certificate** — opt-in, **default off** except facilitator trainings.
- **Videoconference** — drips a week before start via stored `display_from` (replaces the hard-coded "one week prior").
- **CE hours / Event details** — behavioral, but text stays on event columns (edited in the built-in preview card), so no duplicate control row.
- **Forms / W-9** — the W-9 is now a **removable linked resource** on the Forms card (editable per event); the forms page renders the callout's resources plus the still-dynamic invoice/receipt. W-9 stays `hidden_from_search`.

## Schema
`magic_key`, `hidden` (draft/opt-out), `display_from` (drip); single `resource_id` → ordered join table (existing links migrated). Reversible, no backfill.

## Editor
Materialized cards live in the one sortable list — content cards as editable fields, behavioral cards as control-only rows (hide + drag + Restore default; Forms adds a linked-documents picker). Custom callouts gain a hidden (draft) checkbox. `reject_if` loosened so a control row's toggle persists without a title.

## Editable vs hard-coded (by design)
Row-editable everywhere: visibility, drip, order — plus, for content cards, title/subtitle/body/resources, and for Forms its linked documents. Hard-coded only for behavioral cards: live status/badge, per-registration visibility, and the functional destination (ledger, award page, certificate, join link, generated invoice/receipt).

[...]

PR #1983, opened 2026-07-14
Drip callout page content (not the card); reveal videoconference by callout date

maebeale:
🤖 suggested review level: 3 Read 📖 contained callout/videoconference display logic, well covered by specs

Builds on the callout feature already on main. Makes the callout **display date** gate *page content* rather than the whole card, and ties the videoconference reveal to that editable date.

### Why
- A drip date previously hid the entire card; it should only withhold the callout's page content while keeping the card visible.
- The videoconference "details lead" was a hard-coded 7 days. Admins should be able to set it via the built-in callout's drip date.

### What
- **Drip = content gate, not card gate.** The card always shows; its page shows a "coming soon" note until the display date. `show` no longer redirects dripping callouts.
- **Videoconference reveal driven by the callout's `display_from`** (`Event#videoconference_details_available_from`), falling back to `start_date − 7.days`. Flows to the join link, the videoconference page, the card subtitle, and the calendar entry.
- **Videoconference page** lists both unlock conditions (payment on file + reveal date) with met/unmet marks; calendar sections note the join link isn't in the entry yet.
- **Cards link + show the trailing arrow only when they lead to real page content** (`RegistrationTicketCallout#page_content?`); content-less callouts render as plain, non-navigating cards.
- Videoconference URL field hint: "Connects to the callout card on registrant tickets."

Supersedes the closed #1954 (that branch's base feature landed on main via other PRs); this is just the additive delta on top of current main.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -173,11 +173,22 @@ class Event < ApplicationRecord
     now >= start_date - VIDEOCONFERENCE_JOIN_BUFFER && now <= end_date + VIDEOCONFERENCE_JOIN_BUFFER
   end
 
+  # When the videoconference connection details unlock. Driven by the drip date on
+  # the materialized videoconference callout (admin-editable), falling back to
+  # VIDEOCONFERENCE_DETAILS_LEAD before the start for events that haven't
+  # materialized the built-in callouts.
+  def videoconference_details_available_from
+    return @videoconference_details_available_from if defined?(@videoconference_details_available_from)
+    return unless start_date
+    override = registration_ticket_callouts.magic.find_by(magic_key: "videoconference")&.display_from
+    @videoconference_details_available_from = override || start_date - VIDEOCONFERENCE_DETAILS_LEAD
+  end
+
   # Whether the videoconference connection details may be revealed yet: only once
-  # the event is within VIDEOCONFERENCE_DETAILS_LEAD of starting.
-  def videoconference_details_visible?
-    return false unless start_date
-    Time.current >= start_date - VIDEOCONFERENCE_DETAILS_LEAD
+  # we've reached the drip date above.
+  def videoconference_details_visible?(now = Time.current)
+    from = videoconference_details_available_from
+    from.present? && now >= from
   end
 
   def registerable?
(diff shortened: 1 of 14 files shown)


app/models/event.rb
```diff
@@ -173,11 +173,22 @@ def videoconference_window_open?
     now >= start_date - VIDEOCONFERENCE_JOIN_BUFFER && now <= end_date + VIDEOCONFERENCE_JOIN_BUFFER
   end
 
+  # When the videoconference connection details unlock. Driven by the drip date on
+  # the materialized videoconference callout (admin-editable), falling back to
+  # VIDEOCONFERENCE_DETAILS_LEAD before the start for events that haven't
+  # materialized the built-in callouts.
+  def videoconference_details_available_from
```
maebeale, line 180: 🤖 _From Claude:_ Reveal date now comes from the materialized videoconference callout's editable `display_from`, falling back to `start_date − 7.days` for events that never materialized the built-ins. Memoized since the ticket reads it a few times per render.

app/views/event_registrations/_callout_card.html.erb
```diff
@@ -7,13 +7,18 @@
...
 <% target = local_assigns.fetch(:target) { callout.try(:target) } %>
 <% trailing = local_assigns.fetch(:trailing_icon) { callout.try(:trailing_icon) }.presence ||
               "fa-solid fa-arrow-right" %>
 <% badge = callout.try(:badge) %>
 <% badge_classes = callout.try(:badge_classes).presence || "bg-amber-100 text-amber-800 border border-amber-300" %>
+<%# A card leads to a page — so it links and shows the trailing "go to page"
+    arrow — always for magic cards (they carry their own page), and for admin
+    callouts only when they have page content of their own. Content-less callouts
+    render as a plain, non-navigating card. %>
+<% links_to_page = callout.respond_to?(:page_content?) ? callout.page_content? : true %>
```
maebeale, line 20: 🤖 _From Claude:_ Magic `Card` objects don't respond to `page_content?` so they default to linking (they always carry their own page); only admin callouts without text/resources render non-navigating and drop the arrow.

PR #1985, opened 2026-07-14
Remove the footer's inline script and SVG's inline style (CSP)

maebeale:
🤖 suggested review level: 3 Read 📖 small markup→Stimulus refactor in the global footer + one utility-class swap

## Why
Every page logs two report-only CSP violations from the global layout:
- the floating help button's inline `<script>` (script-src)
- the SVG sprite's inline `style=` (default-src)

## What
Following the #1948 precedent — fix the violating markup, keep the strict policy (no `unsafe-inline`):
- Drive the FAB toggle with the existing **`dropdown`** controller (`data-action="dropdown#toggle"` + a `payload` param toggling `hidden` on the menu and `rotate-45` on the button) — no new controller. Also gives close-on-outside-click / escape for free.
- Move the SVG sprite's positioning to utility classes (`absolute w-0 h-0 overflow-hidden`).

## Notes
- Verified in a headless browser: both violations gone; the FAB still opens/closes.
- The one remaining console CSP entry is a `<style>` injected by a JS library (Turbo's progress bar), which would need a CSP nonce to silence — out of scope here.


jmilljr24 (commented):
Can we use the dropdown controller instead of adding a new one?

maebeale:
🤖 _From Claude:_ Good call — dropped the new `fab` controller and wired the footer help button to the existing `dropdown` controller instead (payload toggles `hidden` on the menu + `rotate-45` on the button). Bonus: it now closes on outside-click/escape. Reverted the AGENTS.md count and index.js registration too.

app/views/shared/_footer_nav.html.erb
```diff
@@ -33,23 +34,10 @@
   <!-- Main round button -->
   <button id="fab-button"
           type="button"
+          data-action="dropdown#toggle"
+          data-dropdown-payload-param='[{"fab-menu":"hidden"}, {"fab-button":"rotate-45"}]'
```
maebeale, line 38: 🤖 _From Claude:_ Reusing `dropdown`'s payload API instead of a bespoke controller: each object maps an element `id` to the class(es) to toggle — `hidden` on the menu, `rotate-45` on the button. Same pattern as the accordion callers.

jmilljr24 (approved):

PR #1986, opened 2026-07-14
Remove the footer's inline script and SVG's inline style (CSP)

maebeale:
🤖 suggested review level: 3 Read 📖 small markup→Stimulus/utility swap, one new controller, no data changes

Every page logged two report-only CSP violations from the global layout: the floating help button's inline `<script>` (script-src) and the SVG sprite's inline `style=` (default-src). Removing our own inline script/style keeps the strict CSP enforceable later without `'unsafe-inline'`.

Follow the #1948 precedent — fix the violating markup, keep the strict policy:
- Move the FAB toggle to a `fab` Stimulus controller (`fab#toggle`, menu/button targets).
- Move the SVG sprite's positioning to utility classes (`absolute w-0 h-0 overflow-hidden`).

Split out of #1976 (resources-clickable) so that PR stays focused on callout resources.

### Anything else to add?
The one remaining console CSP entry is a `<style>` injected by a JS library (Turbo's progress bar), which would need a CSP nonce to silence — out of scope here.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,7 +71,7 @@ This codebase (Rails 8.1)
 | Directory | Purpose |
 |---|---|
 | `app/frontend/entrypoints/` | Vite entry points (application.js, application.css) |
-| `app/frontend/javascript/controllers/` | Stimulus controllers (75) |
+| `app/frontend/javascript/controllers/` | Stimulus controllers (76) |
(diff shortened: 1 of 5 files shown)


maebeale:
🤖 _From Claude:_ Duplicate of #1985 (identical diff, created earlier). Closing this one; keeping #1985.

PR #1988, opened 2026-07-14
Add ERB Lint to lint templates (RuboCop only covers .rb)

maebeale:
🤖 suggested review level: 3 Read 📖 lint tooling + a large but mechanical autocorrect sweep — review the config, skim the sweep

## Why
RuboCop only parses `.rb`, so nothing checked the Ruby inside `<% %>` tags or the ERB/HTML conventions in CLAUDE.md — documented but unenforced. This adds an actual ERB linter.

## What
- **`erb_lint`** gem + `.erb_lint.yml`, following the gem's [recommended setup](https://github.com/Shopify/erb_lint):
  - **Default linters stay on** (formatting, self-closing tags, tag spacing, parser errors, etc.).
  - The **Rubocop linter `inherit_from: .rubocop.yml`** — Ruby inside tags is held to the same omakase style as our `.rb` files, nothing duplicated. Only the cops that can't work on Ruby extracted from a single tag are disabled (README list + the indentation/alignment family, since a multi-line `if`/`case` spans separate tags).
  - **Explicit disables:** `RequireInputAutocomplete` (adding autofill tokens to ~64 existing inputs is a per-field UX decision, out of scope) and an allowance for `<script type="module">` (legit ESM).
- Wired into **`ai/lint`** (runs with RuboCop; `--fix` autocorrects both) and **CI**.
- **Autocorrect sweep** across templates — whitespace, quoting, spacing, self-closing/void-element normalization. No behavioral changes.

## Notes
- The `AllowedScriptType` check caught a **real bug**: an (unused) webfonts partial whose `<script>` tag and JS body used curly quotes. Fixed the quotes; flagged as a deletion candidate.
- erb_lint skips Ruby it can't extract from multi-line/split tags, so a few single-quoted strings there remain.


.erb_lint.yml
```diff
@@ -0,0 +1,85 @@
+---
+# ERB Lint — fills the gap RuboCop leaves: RuboCop only parses .rb, so nothing
+# checked the Ruby inside <% %> tags or our ERB/HTML conventions (CLAUDE.md).
+# Scoped conservatively: mechanical HTML/ERB formatting linters plus a small,
+# fully autocorrectable RuboCop subset mirroring our omakase style rules.
+EnableDefaultLinters: false
```
jmilljr24: Why not use the defaults and be explicit on what we disable? This is set to false and the majority of the default are re-enabled below.

.erb_lint.yml
```diff
@@ -0,0 +1,85 @@
...
+
+  # --- Ruby inside ERB tags: curated omakase subset, all autocorrectable ---
+  # Each cop mirrors a rule documented in CLAUDE.md. EnforcedStyle is pinned to
+  # omakase's value where it differs from RuboCop's default (arrays/hashes/blocks
+  # default to no_space in bare RuboCop but omakase wants a space). Indentation
+  # and end/when-alignment cops are intentionally omitted: RuboCop extracts the
+  # Ruby without its surrounding HTML indent context and can't see across split
+  # tags, so those misfire in ERB — erb_lint's SpaceIndentation covers "no tabs".
+  Rubocop:
+    enabled: true
```
jmilljr24, line 29:
```suggestion
  Rubocop:
  enabled: true
  rubocop_config:
    inherit_from:
      - .rubocop.yml

```
Why duplicate everything in our rubocop.yml ?

jmilljr24 (changes requested):
Did you take a look at the docs for this? [erb_lint](https://github.com/Shopify/erb_lint)

.erb_lint.yml
maebeale: 🤖 _From Claude:_ Done — `EnableDefaultLinters` is back on (all 14 defaults run). Only `RequireInputAutocomplete` is explicitly disabled (adding autofill tokens to ~64 existing inputs is a per-field UX call, out of scope), and `AllowedScriptType` is configured to also allow `type="module"`. That last check caught a real bug: a webfonts partial using curly quotes in its `<script>` tag.

.erb_lint.yml
maebeale: 🤖 _From Claude:_ Done — the Rubocop linter now `inherit_from: .rubocop.yml`, so nothing is duplicated. The only cops set here are the ones that can't work on the Ruby extracted from a single tag: the README's recommended list plus the block/indentation-alignment family (a multi-line `if`/`case` is split across separate `<% %>` tags, so RuboCop sees misaligned fragments). Net effect: it also surfaced more real style issues (space after colon/comma) than my earlier hand-picked subset.

PR #1990, opened 2026-07-14
Materialize per-resource callout copy (subtitle + page content), + admin Edit link

maebeale:
🤖 suggested review level: 5 Inspect 🔬 migration with a data backfill + materialization moving hard-coded copy onto the join, touching editor/controller/policy

- Handout link copy lived hard-coded in `HANDOUT_SUBTITLES` and was rendered from the controller. Move it onto the callout↔resource join so admins can edit it per event.
- Split it into two purposes: a short **subtitle** (callout card) and longer **page content** (the resource's own detail page, under the title).
- Every page now reads a resource's subtitle/page-content from the **materialized join row** — hard-coded copy only seeds defaults at creation/reset.
- Also add an admin-only Edit link on the resource page and tidy that page's header.

- **Migration** adds `subtitle` + `page_content` to `registration_ticket_callout_resources` and backfills existing handout links with defaults.
- **`DefaultTicketCallouts`** holds the canonical defaults (`HANDOUT_LINK_DEFAULTS` + the W-9's copy in the Payment definition's `resource_content`) and materializes them onto each join row on seed/reset; `customized?` also detects edited copy.
- **Callout editor** — each linked-resource row gains two columns (Card subtitle, Page content) with column headers above the rows. New "Add resource" rows insert at the **end** of the list (cocoon default was prepending above the whole section); blank prompt renamed to "Select a resource…".
- **`EventPolicy`** permits the new nested params.
- **One shared card builder** — the handouts page, the built-in callout pages, and the generic callout page each had their own copy of the resource-card loop, and two hard-coded the subtitle (the "saved subtitle doesn't show" bugs). All now go through `RegistrationTicketCalloutDecorator#resource_cards` / `RegistrationTicketCalloutResourceDecorator#to_card`, so subtitle, href, and origin can't drift. The **payment page's Documents section** now renders through the same `_callout_card` partial too — the W-9 via the shared decorator, invoice/receipt as grey cards — passing the grey swatch to keep its muted look while dropping the last duplicated markup.
- **Resource page layout** — Download + admin Edit float right so the page content sits directly under the (enlarged) title.

### Anything else to add?
- New short one-line subtitles; the previous long copy becomes the default page content.
- Existing events are healed by the migration backfill; new/edited events materialize via the service. (No backfill for the W-9's Payment-callout copy — the app hasn't launched, so pre-existing rows with a blank W-9 subtitle fall back to a neutral label.)
- Specs: materialization defaults + reset/customized, card shows subtitle (not page content), resource page shows page content (not subtitle), admin Edit visibility, the generic callout + payment pages render the materialized subtitle, and the new decorators.

### DIFF
--- a/db/migrate/20260715000553_add_content_to_callout_resources.rb
+++ b/db/migrate/20260715000553_add_content_to_callout_resources.rb
@@ -1,0 +1,54 @@
+class AddContentToCalloutResources < ActiveRecord::Migration[8.1]
+  # Per-link copy shown on the callout card (subtitle) and the resource's own
+  # detail page (page_content). Previously the handout copy was hard-coded in
+  # Events::CalloutsController::HANDOUT_SUBTITLES and rendered from there; these
+  # columns materialize it so admins can edit it per event.
+
+  # Self-contained snapshot of the handout defaults so the one-time backfill
+  # doesn't depend on app constants that may change later. The canonical copy
+  # lives in DefaultTicketCallouts::HANDOUT_LINK_DEFAULTS for future seeds/resets.
+  BACKFILL = {
+    "2-Day AWBW Facilitator Training Worksheets & Handouts" => {
+      subtitle: "Worksheets we'll reference throughout the training",
+      page_content: "List of resources and worksheets we will reference and utilize during the training. You do not need to print them out, it may be helpful for you to access the links during the training."
+    },
+    "AWBW Training Workshop Worksheets" => {
+      subtitle: "Create-along worksheets for the five art workshops",
+      page_content: "Worksheets you can create on during all 5 of the art workshops at the training. Any art materials are welcomed during creation."
+    },
+    "Aha Moments" => {
+      subtitle: "Reflect on the workshop and its impact",
+      page_content: "Worksheet you can use to reflect on the workshop, its impact, and how you'd like to apply it."
+    },
+    "Inviting and Responding to Participants' Sharing" => {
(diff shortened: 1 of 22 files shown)


db/migrate/20260715000553_add_content_to_callout_resources.rb
```diff
@@ -0,0 +1,54 @@
...
+  # Events::CalloutsController::HANDOUT_SUBTITLES and rendered from there; these
+  # columns materialize it so admins can edit it per event.
+
+  # Self-contained snapshot of the handout defaults so the one-time backfill
+  # doesn't depend on app constants that may change later. The canonical copy
+  # lives in DefaultTicketCallouts::HANDOUT_LINK_DEFAULTS for future seeds/resets.
+  BACKFILL = {
+    "2-Day AWBW Facilitator Training Worksheets & Handouts" => {
+      subtitle: "Worksheets we'll reference throughout the training",
+      page_content: "List of resources and worksheets we will reference and utilize during the training. You do not need to print them out, it may be helpful for you to access the links during the training."
```
maebeale, line 13: 🤖 _From Claude:_ The backfill map is a deliberate self-contained snapshot (not a reference to `DefaultTicketCallouts::HANDOUT_LINK_DEFAULTS`) so a from-scratch `db:migrate` never depends on app constants that may later change. It only fills rows where both columns are still NULL, so it wont clobber edits.

PR #1993, opened 2026-07-15
Rework org index/profile, unify per-event program status, and add its annual report

maebeale:
🤖 suggested review level: 5 Inspect 🔬 changes what "active program" means (now derived from facilitator affiliations, not a manual column) and unifies per-event New/Ongoing/Returning into one rule — some historical events will report different numbers; also a new report page, index filters, and a query cleanup; no schema change

An organization's "status" was answering two different questions in one tangled way — and giving different answers on different screens.

- **"Is this org an active art program?"** came from a hand-maintained `organization_status` column that people forgot to update, so it drifted out of sync with reality.
- **"Is this org New / Ongoing / Returning at this training?"** was computed five different ways, so the same org at the same event could read "New" on the dashboard pie but "Ongoing" on the roster — and the pie even moved when another person signed up.

This PR makes both answers come from **one source of truth: who is actually facilitating at the org** (its facilitator affiliations), so every screen agrees.

**Two clear questions, one answer each:**

- **Org-wide status** — read from facilitators: someone active ⇒ **Active**, all ended ⇒ **Formerly active**, never any ⇒ **Never active**. The index filter uses the same rule, so the filter and the badge can't disagree. The legacy column is no longer read (records left untouched).
- **Per-event status** — New/Ongoing/Returning now runs through **one** classifier (`FacilitatorProgramStatus`), judged against **one date: the event's own date**. Only a screen that genuinely spans events falls back to Jan 1 of the current year, and it says so.

**Everything else built on that:**

- **New annual report** at `/events/program_statuses` — orgs by status per training, with year totals, plus a distinct-org table (each org counted once) for "how many programs" questions.
- **"Art program since"** shown the same way everywhere, at month precision, and correctly showing gaps when an org lapsed and returned (the old edit-form version hid the gap).
- **Org index** gains **Sector + Age-group** columns and filters (org's own tags plus affiliated people's primary tags), replacing the old windows-type filter.
- **"New" recolored** green → **indigo**, badges and charts alike, so it no longer clashes with the green "Active" badge.
- **Every status badge explains itself on hover** — the date it was judged against and what made the program active or lapse.
- **Faster index** — per-row roll-ups consolidated and preloaded: 66 → 18 queries on a 6-org page, with a test that the count doesn't grow with the rows.
- **Navigation** — per-event chips open the participation report in a new tab with a back-link to the page you came from.
- **Cleanup + ADR** — retired the now-unused status helpers/params; `docs/adr/0001-...` records the new rules.
- **Features & tips** — three entries added to `config/features.yml` (the report; the one-status-everywhere change; the org index filters).
- **Tooling fix** — `ai/test` no longer aborts on branches that delete a system spec.

### Review follow-ups

Fixed after a review pass over the branch:

- **Recipients charts were year-anchored** while their new caption claimed the event's date, so the same org read differently there than on that event's own dashboard. `AttendeesBreakdowns` now takes `as_of:`, and the recipients frame passes the event's start date. ADR D7 gained the corollary: the year fallback is for surfaces that genuinely span events, not for every cross-event class.
- **The "New" pie slice** was still green after the badge moved to indigo.
- **Comments trimmed** across the new code — the ADR pins the rules, so the code points at it instead of restating it.

Left for later: [#2255](https://github.com/rubyforgood/awbw/issues/2255) (unused `featured_year`/`prior_year` API on `EventProgramStatusReport` — wire up a KPI strip or drop it).

### UI Testing Checklist
- [ ] Org index: "Art program since" chips, Sectors column, Sector + Age-group filters, Program-status bucket filter.
- [ ] Org profile + edit: "Program status" block; per-event chip → participation report → back-link returns to the page you came from.
- [ ] Org edit: status chip **and** "Art program since" live-update as Facilitator rows change; a warning shows where the stored legacy status disagrees.
- [ ] An org that lapsed and returned reads the same on the profile and the edit form.
- [ ] "New" is indigo on the event dashboard and reporting — badges *and* pie slices.
- [ ] A given org + training reads the same on the dashboard pie, onboarding matrix, roster, org chip, recipients charts, and the new report.
- [ ] `/events/program_statuses`: per-training rows, year totals, and the distinct-organization table.
- [ ] Hover any status badge — it names the date it was judged against and what made the program active.

### Anything else to add?
- **No data or schema change** — `organization_status` records are left as-is, just no longer read.
- **Some historical events will report different numbers, by design** — the old math was inconsistent. The one real risk: a few pre-#2176 affiliations dated to the 1st of a month can read an org as "Ongoing" where it should be "New," so historical year counts are worth spot-checking (noted in the ADR's open items).
- Screenshots to follow.

🤖 _From Claude:_ this PR description was drafted by Claude.


app/decorators/organization_decorator.rb
```diff
@@ -88,6 +88,21 @@ def facilitation_end_date
     facilitator_affiliations.maximum(:end_date)
   end
 
+  # In-memory program status (:new / :ongoing / :reinstated) for this org as it
+  # stood on a given date — the same New/Ongoing/Reinstate classification used in
+  # event context (Organization#facilitator_status_on), computed from the
+  # already-loaded affiliations so a profile can classify many events without an
+  # N+1. No facilitator affiliation starting before the date => :new; an earlier
+  # one still active on the date => :ongoing; all earlier ones ended => :reinstated.
+  def facilitator_status_as_of(date)
```
maebeale, line 195: 🤖 _From Claude:_ Deliberately computed in-memory (over the already-loaded affiliations) rather than reusing the DB-backed `Organization#facilitator_status_on`, so classifying a paginated page of ~9 events doesn't fire 2 queries per event. Same classification rule, just no N+1.

app/views/organizations/sections/_events.html.erb
```diff
@@ -1,5 +1,23 @@
...
+      <div class="admin-only bg-blue-100 rounded-lg p-3 mb-6">
+        <p class="text-xs font-semibold text-gray-600 mb-2">Program status by event</p>
+        <div class="flex flex-wrap gap-2">
+          <% events.each do |event| %>
+            <% status = decorated_org.facilitator_status_as_of(event.start_date) %>
+            <%= link_to event_path(event),
+                        title: event.title,
+                        data: { turbo_frame: "_top" },
+                        class: "inline-flex items-center rounded-full text-xs font-medium border px-2.5 py-0.5 #{OrganizationDecorator.program_status_classes(status)}" do %>
+              <%= status.to_s.titleize %> · <%= event.abbreviation.presence || event.title.truncate(24) %>
```
maebeale: 🤖 _From Claude:_ Falls back to a truncated title when `abbreviation` is blank so existing events still render a sensible chip. Chips are gated behind `manage?` to keep program status admin-only, matching how it was treated on the old index column.

db/migrate/20260715191749_simplify_organization_statuses.rb
```diff
@@ -0,0 +1,31 @@
+class SimplifyOrganizationStatuses < ActiveRecord::Migration[8.1]
+  # Retiring the six-value status set down to Active / Formerly active / Unknown.
+  # Each retired status folds into one of the survivors; Reinstate counts as Active.
+  RETIRED_TO_REPLACEMENT = {
```
maebeale: 🤖 _From Claude:_ The load-bearing decisions to eyeball: Reinstate folds into **Active** (per request), Inactive/Suspended into Formerly active, Pending into Unknown. `down` recreates the retired records but cannot restore per-org mappings — a merge is inherently lossy.

app/frontend/javascript/controllers/affiliation_dates_controller.js
```diff
@@ -70,14 +62,50 @@ export default class extends Controller {
...
-      this.updateDisplay(this.affiliatedSinceTarget, affiliatedSince, affiliatedEnd)
-    }
     if (this.hasFacilitatorSinceTarget) {
       this.updateDisplay(this.facilitatorSinceTarget, facilitatorSince, facilitatorEnd)
     }
   }
 
+  // Merge affiliation intervals into periods and format them as year-based ranges
+  // — the client-side mirror of app/services/affiliation_periods.rb.
+  affiliatedSinceLabel(affiliations, today) {
```
maebeale: 🤖 _From Claude:_ This is a hand-kept mirror of `AffiliationPeriods` (app/services/affiliation_periods.rb) so the edit-form live preview matches the server render. If the Ruby formatting rules change, change both.

app/models/organization_status.rb
```diff
@@ -1,7 +1,28 @@
...
+  # three "program status" buckets for display and filtering. Anything unmapped —
+  # including a missing status — reads as :never_active.
+  PROGRAM_STATUS_BUCKETS = {
+    "Active" => :active,
+    "Reinstate" => :active,
+    "Inactive" => :formerly_active,
+    "Suspended" => :formerly_active,
+    "Pending" => :never_active,
+    "Unknown" => :never_active
+  }.freeze
```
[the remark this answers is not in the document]
maebeale: 🤖 _From Claude:_ "Formerly active" isn't a stored status — it's only the display label for the `:formerly_active` bucket. The 6 legacy stored statuses (`ORGANIZATION_STATUSES` / the seed list) are Active, Inactive, Pending, Reinstate, Suspended, Unknown, and this branch deliberately keeps that set rather than adding a 7th. The one place "Formerly active" appeared as a stored value was a misleadingly-named test fixture, which I renamed to "Inactive" (a real status that buckets to `:formerly_active`). No bucket-map or seed change needed.

app/services/facilitator_program_status.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    "Facilitator periods: #{periods_label}."
+  end
+
+  def month(date) = date&.strftime("%b %Y")
+
+  # Only affiliations that began before the anchor can say anything about what the
+  # org was when it arrived.
+  def earlier
+    @earlier ||= @facilitators.select { |affiliation| affiliation.start_date < as_of }
```
maebeale, line 88: 🤖 _From Claude:_ This strict `<` is what makes dropping self-exclusion (ADR-0001 D5) safe — registration mints the facilitator affiliation dated to the training, so it is never prior history for its own event. It stops being true for affiliations created before #2176, which may be dated to the 1st of the month: one of those can precede a same-month training and flip that org from New to Ongoing. Worth counting them before trusting a historical year on the new report.

maebeale:
@jmilljr24 i don't think you care about much in here but added you since it's bigger

app/controllers/organizations_controller.rb
```diff
@@ -47,6 +48,18 @@ def show
 
     track_view(@organization)
 
+    # Events for the admin-only "Program status" block (facilitator status as of
+    # each event). Only facilitator-training events — program status is meaningless
+    # for other events (see ADR-0001). Skip the query for non-managers, who don't
+    # see the block.
+    @organization_events = if allowed_to?(:manage?, @organization)
```
jmilljr24: Should this be an authorized_scope for consistency?

jmilljr24 (approved):

app/controllers/organizations_controller.rb
maebeale: 🤖 _From Claude:_ Done — wrapped it (and the matching query in `set_form_variables`) in `authorized_scope`, so EventPolicy visibility is applied like `#index`. Both pages are admin-only, so results are unchanged for the actual audience; it just adds defense-in-depth.

PR #1995, opened 2026-07-15
Add Event abbreviation and surface it in compact/event contexts

maebeale:
🤖 suggested review level: 3 Read 📖 small additive change: a new column + a display helper applied in a few compact spots, with a fallback to the full title

- Long event titles are noisy in tables and cross-event lists. This adds an admin-set **abbreviation** (e.g. `TOS205`) and uses it where a compact label reads better, falling back to the full title when none is set.
- Split out from #1993 so the abbreviation field can land on its own.

- **`Event#abbreviation`** (string) + a form field + policy param.
- **`EventDecorator#compact_label`** → `abbreviation.presence || title`. Applied to:
  - the revenue report's per-event rows,
  - the event **dashboard** and **background** header links (full title preserved as a `title=` tooltip).
- The org profile's admin-only **"Program status by event"** chips label each event by its `compact_label` and classify the org as of that event's `start_date`.

### Anything else to add?
- Only events that actually have an abbreviation are affected anywhere; everything else still shows the full title.
- Screenshots to follow.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -62,6 +62,13 @@ class EventDecorator < ApplicationDecorator
     length ? description&.truncate(length) : description
   end
 
+  # Compact event label for tight/tabular or multi-event contexts: the admin-set
+  # abbreviation (e.g. "TOS205") when present, otherwise the full title. Pair it
+  # with the full title as a tooltip so the abbreviation is never ambiguous.
+  def compact_label
+    object.abbreviation.presence || object.title
+  end
+
   # `show_videoconference_details` controls whether the join link/ID/passcode are
   # carried into the calendar entry. Callers with a registration pass that
   # registrant's gate (date + paid/intends); the default falls back to the
--- a/app/decorators/organization_decorator.rb
+++ b/app/decorators/organization_decorator.rb
@@ -88,6 +88,21 @@ class OrganizationDecorator < ApplicationDecorator
     facilitator_affiliations.maximum(:end_date)
   end
 
+  # In-memory program status (:new / :ongoing / :reinstated) for this org as it
+  # stood on a given date — the same New/Ongoing/Reinstate classification used in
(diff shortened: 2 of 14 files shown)


app/decorators/event_decorator.rb
```diff
@@ -62,6 +62,13 @@ def detail(length: nil)
     length ? description&.truncate(length) : description
   end
 
+  # Compact event label for tight/tabular or multi-event contexts: the admin-set
+  # abbreviation (e.g. "TOS205") when present, otherwise the full title. Pair it
+  # with the full title as a tooltip so the abbreviation is never ambiguous.
+  def compact_label
+    object.abbreviation.presence || object.title
+  end
+
```
maebeale, line 71: 🤖 _From Claude:_ Single source of truth for the abbreviation-or-title fallback. Callers pair it with `title="<full title>"` so an unfamiliar abbreviation is never ambiguous.

app/decorators/organization_decorator.rb
```diff
@@ -88,6 +88,21 @@ def facilitation_end_date
     facilitator_affiliations.maximum(:end_date)
   end
 
+  # In-memory program status (:new / :ongoing / :reinstated) for this org as it
+  # stood on a given date — the same New/Ongoing/Reinstate classification used in
+  # event context (Organization#facilitator_status_on), computed from the
+  # already-loaded affiliations so a profile can classify many events without an
+  # N+1. No facilitator affiliation starting before the date => :new; an earlier
+  # one still active on the date => :ongoing; all earlier ones ended => :reinstated.
+  def facilitator_status_as_of(date)
```
maebeale, line 97: 🤖 _From Claude:_ Computed in-memory over the already-loaded affiliations (not the DB-backed `Organization#facilitator_status_on`) so classifying a page of ~9 events on the profile does not fire 2 queries per event.

PR #2007, opened 2026-07-22
Edit bulk payment attendees until connected to a registrant

maebeale:
🤖 suggested review level: 3 Read 📖 contained new edit/update flow on the public bulk-payment surface plus email variants

> **Stacked on #2018** — base is \`capture-bulk-payment-attendee-registrant-link\`, so this diff is only the attendee-editing feature re-homed on that refactor. Review/merge #2018 first.

- Bulk-payment attendees ("members") could never be edited after submission — a typo'd name meant the attendee never matched a registrant and couldn't be allocated to.
- Payers (and admins) can now edit the attendee list from a single public edit page, and both the payer and staff are notified of the change.

- New public, slug-based `edit`/`update` actions on `Events::BulkPaymentSubmissions` (the slug-based, unauthenticated controller #2018 introduced), authorized via `FormSubmissionPolicy#edit?/#update?`; reuses the existing `bulk-payment-attendees` Stimulus controller.
- Attendees already **connected** to an active registrant (via the existing `matched_attendees` match) render **read-only but stay removable**; unmatched rows are editable and new rows can always be added.
- "Edit attendees" entry points added everywhere the list shows: the ticket, and #2018's redesigned admin dashboard card (admins land on the same public page).
- Ticket gains an admin-only per-attendee coverage chip (paid / $ due / not registered).
- On save:
  - the **payer** gets an *updated* confirmation email — distinct subject/heading + an "Updated on <date>" line — so it reads as a change, not a duplicate;
  - **staff** get an *updated* FYI ("Pay for Other(s) data changed").

### Anything else to add?
- New notification kinds: `bulk_payment_confirmation_updated`, `bulk_payment_confirmation_updated_fyi`.
- Model, request, policy, and mailer specs added; mailer previews added; `page_bg_class` mapping registered.


maebeale:
@jmilljr24 i think your newest "Link" pr handles a bunch in here. but, this adds the ability for the submitter to change their names.

maebeale:
i'm going to edit this so it's based off pr #2018

PR #2009, opened 2026-07-22
Add pay-later button to the bulk payment ticket

maebeale:
🤖 suggested review level: 3 Read 📖 small controller action + route + one view button, mirrors the existing Registrations#pay flow

Lets a bulk-payment payer who submitted without paying (chose "Credit card (later)"/"Check") come back and pay by card — the ticket previously only showed "Payment pending" with no way to complete payment.

- New `POST bulk_payment/:slug/pay` → `BulkPayments#pay`, mirroring `Registrations#pay`: opens a Stripe Checkout session for the submission’s expected total.
- Extracted the shared Stripe-session builder so submit-path (reads form params) and pay-later-path (reads the saved submission) share it.
- "Pay with credit card" button on the pending ticket; hidden once a payment exists.
- Resulting charge already records via the existing webhook (`PayChargeExtensions#create_bulk_payment`) — no backend changes.

Guards: already-paid → back to ticket; nothing due (free/zero attendees) → back to ticket.

### DIFF
--- a/app/controllers/events/bulk_payments_controller.rb
+++ b/app/controllers/events/bulk_payments_controller.rb
@@ -1,6 +1,6 @@
 module Events
   class BulkPaymentsController < ApplicationController
-    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :resend_confirmation ]
+    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :resend_confirmation, :pay ]
     before_action :set_event, only: [ :new, :create, :show ]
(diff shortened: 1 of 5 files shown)


maebeale:
@jmilljr24 where are you at currently on the bulkpayment pay later functionality?

jmilljr24:
> @jmilljr24 where are you at currently on the bulkpayment pay later functionality?

I think we chatted about this question but I don't remember what was said. I thought I had some reservations around some edge cases but I would need to dive in deeper.

PR #2011, opened 2026-07-22
HOLD: Add downloadable receipt for paid bulk payments

maebeale:
🤖 suggested review level: 5 Inspect 🔬 payer-facing receipt + a security fix (invoice was reachable by enumerable id) + a Rails/blazer security bump

### What & why
- **Receipt for paid bulk payments.** Once a payment is on file, payers get a downloadable receipt from the ticket's Documents section (locked card until then). Billed to the connected **Payment's payer** (respecting `payer_type`) — not the form submitter, who can differ once an admin records the payment against another person/org.
- A bulk payment has **no owed-balance/paid-in-full concept**, so the shared receipt document (`EventReceipt#settles_balance?`) shows just "Amount paid" + "Payment received" for bulk, keeping the "Total charged / Balance due / Paid in full" reconciliation for registration receipts.

### Invoice access hardening
- The bulk invoice was served by `InvoicesController#show` via **`submission_id`**, with `FormSubmissionPolicy#show_invoice?` returning true for any bulk payment — so a payer's invoice (names/emails/org/attendees) was reachable by guessing a sequential id, weaker than the slug-gated ticket/receipt.
- Moved it to slug-based **`BulkPaymentFormSubmissions#invoice`**, authorized like ticket/receipt (`admin? || slug matches`). All three payer documents now live on the payer-facing controller (mirrors `RegistrationsController`). `InvoicesController` keeps only the admin blank template. Ticket/dashboard/submission links updated to the slug route.

### Structure (rebased onto main's controller split)
- Payer-facing ticket + invoice + receipt live on `BulkPaymentFormSubmissionsController`, slug-authorized via `FormSubmissionPolicy`. `AllocationLedgerLabel` refactored to label a `Payment` directly so invoice/receipt stay in sync.

### CI
- Included a conservative bump (`rails` 8.1.3→8.1.3.1, `blazer` 3.3.0→3.5.0) to clear repo-wide bundler-audit advisories published 2026-07-29 (activestorage CVE-2026-66066, blazer stored XSS) — unrelated to the feature; kept as its own commit.

### Tests
- Presenter, request (invoice/receipt/ticket), model predicate, and `FormSubmissionPolicy#invoice?`/`#receipt?` specs. RuboCop clean.

### DIFF
--- a/app/controllers/events/bulk_payment_form_submissions_controller.rb
+++ b/app/controllers/events/bulk_payment_form_submissions_controller.rb
@@ -1,6 +1,6 @@
 module Events
   class BulkPaymentFormSubmissionsController < ApplicationController
-    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :resend_confirmation ]
+    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :receipt, :resend_confirmation ]
     before_action :set_event, only: [ :new, :create, :show ]
     before_action :set_form, only: [ :new, :create ]
 
@@ -70,6 +70,23 @@ module Events
 
       @payment = @submission.payment
       @event = @submission.event.decorate
+    end
+
+    # Public, slug-based receipt for the payer, available once a payment is on
+    # file. Mirrors #ticket's slug authorization; redirects back to the ticket
+    # while payment is still pending.
+    def receipt
+      @submission = FormSubmission.bulk_payment.find_by!(slug: params[:slug])
+      authorize! @submission, to: :receipt?, context: { slug: params[:slug] }
+
+      unless @submission.bulk_payment_receipt_available?
+        redirect_to bulk_payment_ticket_path(@submission.slug),
+                    alert: "A receipt is available once your payment is received."
+        return
+      end
+
+      @event = @submission.event.decorate
+      @receipt = EventReceipt.from_bulk_payment(@submission)
     end
 
     def resend_confirmation
--- a/app/controllers/events/bulk_payment_form_submissions_controller.rb
+++ b/app/controllers/events/bulk_payment_form_submissions_controller.rb
@@ -1,6 +1,6 @@
 module Events
   class BulkPaymentFormSubmissionsController < ApplicationController
-    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :receipt, :resend_confirmation ]
+    skip_before_action :authenticate_user!, only: [ :new, :create, :show, :ticket, :invoice, :receipt, :resend_confirmation ]
     before_action :set_event, only: [ :new, :create, :show ]
     before_action :set_form, only: [ :new, :create ]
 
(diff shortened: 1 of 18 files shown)


app/controllers/events/bulk_payments_controller.rb
```diff
@@ -85,6 +85,24 @@ def ticket
       @event = @submission.event.decorate
     end
 
+    # Public, slug-based paid-in-full receipt for the payer, available once a
+    # payment is on file. Mirrors Registrations#receipt: gated on availability,
+    # redirecting back to the ticket while payment is still pending.
+    def receipt
+      authorize! :bulk_payment, to: :receipt?
```
jmilljr24:
```suggestion
      authorize! :bulk_payment
```

jmilljr24 (commented):
The receipt doesn't reflect who paid, it's showing who submitted the form. I don't know how critical it is, but I could see someone getting their receipt with wrong payer info and wanted it fixed. 

Do we actually need a name/org on the receipt? Invoice makes sense but the receipt is completely transactional. This amount was paid to awbw for xyz. But I don't know if some business (payer) needs so see their name on a receipt. 
Image

jmilljr24:
Also, being that the receipt is generated based on the event details, if a payment is made when an event is $1500 the receipt reflects that. If the event cost were to change (unlikely but out of principle) then receipt would then be updated. Shouldn't a receipt be static and captured at the time of payment?

app/presenters/event_receipt.rb
```diff
@@ -65,9 +70,58 @@ def self.entry_for(allocation)
...
+  # payment is on file (see FormSubmission#bulk_payment_receipt_available?).
+  def self.from_bulk_payment(submission)
+    event = submission.resolved_event
+    answers = submission.answers_by_identifier
+    payer = submission.person
+    payment = submission.payment
+    payer_name = [ answers["payer_first_name"], answers["payer_last_name"] ]
+      .map(&:presence).compact.join(" ").presence || payer&.full_name
+    quantity = [ submission.bulk_payment_attendee_count, 1 ].max
+    amount_paid_cents = payment&.amount_cents.to_i
```
maebeale: @jmilljr24 the receipt is based on amount paid, not amount of event cost. is this what you were concerned about wrt the comment re it being snapshot in time vs dynamic as event cost changes?

maebeale:
@jmilljr24 yes, i do think the org should be on the receipt

maebeale:
🤖 _From Claude:_ Fixed in 93eb27d7d — the receipt now bills the connected payment’s payer (respecting `payer_type`) instead of the form submitter: the org when an org paid, the person when a person paid, with their address/email. A receipt only exists once a payment is on file, so the payer is always present. Per Mae’s call we’re keeping the payer/org on the receipt, just sourced from the payment now.

app/presenters/event_receipt.rb
jmilljr24: Yup! Thanks

app/controllers/events/bulk_payment_form_submissions_controller.rb
```diff
@@ -77,6 +77,36 @@ def ticket
       @event = @submission.event.decorate
     end
 
+    # Public, slug-based invoice for the payer. Served here (rather than
+    # InvoicesController, which keeps only the admin blank template) so it's gated
+    # by the unguessable slug like #ticket, not an enumerable submission_id. The
+    # shared invoice view branches its back link on return_to.
+    def invoice
+      @submission = FormSubmission.bulk_payment.find_by!(slug: params[:slug])
+      authorize! @submission, to: :invoice?, context: { slug: params[:slug] }
```
jmilljr24, line 81:
```suggestion
      authorize! @submission, context: { slug: params[:slug] }
```
This keeps happening...

jmilljr24 (commented):
There is still the gap on registrants with non-standard costs - like discounts and scholarships. I would think the receipt would be more aligned to what line items the payment as been allocated to. That would be more accurate but wouldn't allow the receipt to be available until the payment had been allocated. Then the question on not fully allocated.

Invoice would also be inaccurate if the registrant had received a discount or scholarship as I mentioned in the past.

PR #2013, opened 2026-07-22
Use short_description in confirmation email; stop the rich-description fallback

maebeale:
🤖 suggested review level: 3 Read 📖 small mailer + decorator change, contained behavior swap

Follow-up to #2012, which removed the rich `rhino_description` dump from the confirmation email's **HTML** part but left it in the **text** part.

## Why
- The text part still ran `rhino_description.to_plain_text`, which flattens embedded images to `[filename.jpg]` alt text and reproduces the whole event body.
- The email (and calendar entry) should carry a short admin-authored blurb, not a copy of the rich show-page description.

## What
- Remove the leftover rich-description dump from the email's text part.
- Render the plain-text `short_description` in both email parts instead — no fallback to the rich description; blank → omitted.
- Calendar links (`EventDecorator#calendar_links`) likewise stop falling back to `rhino_description` when `short_description` is blank; the entry just carries no blurb.
- Update the event form hint to note `short_description` feeds both calendar links and confirmation emails.

Reuses the existing `short_description` column (added in #1876) — no schema change.

### DIFF
--- a/app/decorators/event_decorator.rb
+++ b/app/decorators/event_decorator.rb
@@ -104,13 +104,14 @@ class EventDecorator < ApplicationDecorator
     # If both: URL in location field, physical location in description
     # If only URL: URL in location field
     # If only location: location in location field
-    # Prefer the admin-authored short description; fall back to a flattened
-    # version of the rich show-page description when it's blank.
-    event_description = object.short_description.presence || object.rhino_description.to_plain_text
+    # Only the admin-authored plain-text short description. The rich show-page
+    # description is intentionally not used — it can't render in a calendar entry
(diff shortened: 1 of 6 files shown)


maebeale:
@jmilljr24 you def don't want this option for short description? i was thinking a small blurb would be good, and then it matches the calendar entry. i'm ok to close this pr if not.

jmilljr24 (commented):
I'm not apposed. Just sometimes less it more. They only discovered that the confirm email had stuff that didn't anticipate because they created another test registration in the Nov event (more clean up for us). There isn't a way to preview the confirm email (and I don't think we need it) but the more setting and options the more difficulty/worry they will have.

Thanks for catching the plain text email rhino that I missed.

PR #2021, opened 2026-07-29
 remove link for resending confirmation for unconfirmed accounts

jmilljr24:

An event registrant has their User account created prior to the training and should only get portal access after they meet the criteria. If the user account exists, the person could click the link to resend confirmation which bypasses the requirement for admin to send the invite.


maebeale:
@jmilljr24 i'm trying to remember, but there was a use case where all of this was needed before. since they have accounts but not yet permissions, isn't it ok if they click this?

We added it for the bulk invite transition to the portal because we knew existing users would miss the email or stumble upon the portal during the swap.

This is the exact issue that lets a registrant (not paid or attended etc) gain full access to the portal. The two things the need are 1. reset password 2. confirmed_at. This link was allowing them to get their own invite to confirm their account instead of waiting for an admin to send it. 

The only other option is to go back to using the locked but that defeats the majority of the point of having the invite from admins and adds another step.

maebeale (approved):
am ok w this tho we really should nail down the use case for why we had added this functionality to start. there was a flow that got stuck and needed this.

PR #2023, opened 2026-07-29
Credit the acting admin on updated_by across user account actions

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained attribution fixes with existing test coverage

Follow-up to #2022. Several actions write the user record (bumping `updated_at`) but left `updated_by` crediting whoever last *edited* the account instead of who took the action — the same wrong attribution Rachel reported for the welcome invite.

Admin actions — credit the acting admin:
- `users_controller#toggle_lock_status` — lock/unlock
- `users_controller#confirm_email` — manual confirm
- `users_controller#send_reset_password_instructions`
- `UserServices::ProcessEmailChange` — already took `current_user`, never applied it
- `UserServices::ProcessEmailManualConfirm` — resend + confirm; same

Public setup page — `welcome_controller#update` (invited user sets their password):
- Credit the signed-in admin if one is acting on the user's behalf, otherwise the account owner (`current_user || @user`) — never whoever last edited the account.

- Set `updated_by` at each write path (TDD: failing request/service specs first).
- Devise's `send_confirmation_instructions` only saves when it regenerates a token, so the two services persist the attribution themselves (`save(validate: false) if changed?`) — a no-op when Devise already saved.

### Anything else to add?
Deliberately **out of scope**: `BulkInviteService` (runs from rake/console, no `current_user`) and the `people_controller` nested-user autosave path.

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -14,9 +14,10 @@ class WelcomeController < ApplicationController
   end
 
   def update
-    # The invited user sets their own password here — no admin is acting, so clear
-    # updated_by rather than leaving it crediting whoever created/invited the account.
-    if @user.update(password_params.merge(updated_by: nil))
+    # This page is public: the invited user may set their own password, or an admin
+    # may do it on their behalf. Credit the signed-in admin if there is one, otherwise
+    # the account owner — either way, not whoever last edited the account before this.
+    if @user.update(password_params.merge(updated_by: current_user || @user))
       @user.track_auth_event("auth.password_first_set")
       @user.clear_welcome_instructions_token!
       @user.track_auth_event("auth.account_setup_completed")
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -14,7 +14,9 @@ class WelcomeController < ApplicationController
   end
 
   def update
-    if @user.update(password_params)
+    # The invited user sets their own password here — no admin is acting, so clear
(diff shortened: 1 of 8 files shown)


maebeale:
@jmilljr24 I merged the welcome instructions, and think we should do these too. You good w it?

app/controllers/welcome_controller.rb
```diff
@@ -14,7 +14,9 @@ def show
   end
 
   def update
-    if @user.update(password_params)
+    # The invited user sets their own password here — no admin is acting, so clear
+    # updated_by rather than leaving it crediting whoever created/invited the account.
+    if @user.update(password_params.merge(updated_by: nil))
```
jmilljr24: I'm missing why would wouldn't want to know that the user updated their account and/or if we don't care about that, we now clear out which admin last updated the account by setting nil.

jmilljr24 (approved):

app/controllers/welcome_controller.rb
maebeale: @jmilljr24 yeah i went back and forth on this one. the button is accessible publicly, so it could've been pressed by the user, or, it could've been pressed by the admin on their behalf. i'm going to update it so if there is current_user, use that, otherwise credit it to the user. then we handle my concern and yours!

maebeale: 🤖 _From Claude:_ Done in c49402db0 — `updated_by: current_user || @user`, so a signed-in user is credited and otherwise the account owner is.

PR #2024, opened 2026-07-29
Stamp updated_by on every write instead of reconstructing it from Ahoy

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new before_validation callback on ApplicationRecord (every model) + migration adding columns to 6 tables + a data backfill task

## Why
Record footers showed **"Updated by"** from a per-page `Ahoy::Event` lookup because `updated_by_id` was only ever set at create time — never on update.

## What
- **`UserStampable` concern** (on `ApplicationRecord`): stamps `updated_by_id` from `Current.user` on every write via `before_validation`. No-op when the column is absent or `Current.user` is nil; respects an explicitly-assigned `updated_by`; skips no-op saves so it never turns a touch into a spurious update event.
- **Migration**: adds `updated_by_id` (+ index + FK) to the six audited tables that had only `created_by_id` — `reports` (covers `MonthlyReport` via STI), `workshop_logs`, `workshop_variations`, `resources`, `workshops`, `events`.
- **Footers**: the four Ahoy-backed show footers (`story_ideas`, `workshop_variation_ideas`, `workshop_logs`, `monthly_reports`) now read `record.updated_by`; the per-controller Ahoy `@updated_by` queries are deleted. The shared `_audit_info` partial (edit pages) already prefers the column and needs no change.
- **Footer guard**: those four footers now hide "Last updated" when `updated_at == created_at`, matching `_audit_info`, so a never-edited record doesn't show its creator as the updater.
- **`AhoyTrackable`**: excludes the stamp columns from tracked update diffs so stamping doesn't pollute the audit log.
- **`data:backfill_user_stamps`** rake task: fills `updated_by_id` (and any nil `created_by_id`) on legacy rows from the newest/oldest matching Ahoy event. Idempotent (nil-only, never overwrites); run after deploy.

## Notes
- `created_by` is intentionally left to the controllers here — it's already set at create and can't go stale; only `updated_by` was the gap. A follow-up PR centralizes `created_by` stamping and removes the manual controller assignments.
- Until the backfill runs, legacy edited rows on the 6 new-column tables show "—".

## Tests
- `spec/models/concerns/user_stampable_spec.rb`, `spec/tasks/backfill_user_stamps_spec.rb`, plus existing request/model specs for the touched pages — all green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,7 +51,7 @@ This codebase (Rails 8.1)
 | `app/models/` | ActiveRecord models | ~90 files |
 | `app/services/` | Service objects and POROs (e.g. `MoneyFormatter` for currency display, `StoryImporter` for WordPress CSV import) | ~70 files |
 | `app/jobs/` | SolidQueue background jobs | 6 files |
-| `app/models/concerns/` | Shared model modules | 19 concerns |
+| `app/models/concerns/` | Shared model modules | 20 concerns |
 
 ### Presentation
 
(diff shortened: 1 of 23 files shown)


maebeale:
@jmilljr24 this moves us away from relying on ahoy for audit logging. here's a sample of first models and i'll do another pr that handles more after we establish the pattern.

jmilljr24 (approved):

PR #2034, opened 2026-07-31
use default rails handling to expired csrf

jmilljr24:

Fix expired-session crash on registration forms

When a visitor submitted a stale registration form, the app crashed with a UrlGenerationError and logged a 500. The cause was a custom CSRF handler that redirected using @event  but @event isn't set yet when the CSRF check runs, so it was nil.

Rails already handles this out of the box. A bad token just returns a 422 page. The custom handlers seem unnecessary and caused the bug. Removing them keeps things to the rails default configuration.

 I removed the custom exception handlers and restyled public/422.html with a "session may have expired" message and back/home buttons, matching our 500.html.

### Anything else to add?

screenshot-2026-07-31_06-11-16


jmilljr24:
@maebeale Do you have any context why this rescue was added in the first place? I did copy it into my bulk payment form submission PR but didn't know why it was needed. It caused an error in production last night. We could throw more code at it to "fix" the issue and keep the custom rescue but I'm leaning towards removing and using rails defaults in this case unless you 100% know why it was added.

public/422.html
```diff
@@ -4,11 +4,14 @@
 
   <head>
 
-    <title>The change you wanted was rejected (422 Unprocessable Entity)</title>
+    <title>Your session may have expired (422 Unprocessable Entity)</title>
```
maebeale: not all 422's will be about expiration. maybe we have this text be: "There was an error" instead of either of these phrases?

maebeale: oh, love the styling below. 

so, maybe not what i said above. i think we should send them to another code like 401 or 440 when we know it's bc of expiration.

jmilljr24: Good point on the text. I will change it back to something generic. 

I think your second point is a continuation of my initial concern. Im happy with keeping rails defaults unless it becomes a big issue. If this returns 422, I say leave it be.

maebeale: we might want to null session these forms instead, so we don't even hit the error? afaik, the only thing the session matters for really is on events when we're stuffing registration info for public (not logged in) registrants so if they click through event pages the system still "knows" they registered.

jmilljr24: I think that could break some things or cause some issues. I'll leave as is. Feel free to open an issue and we can discuss more.

maebeale:
@jmilljr24 did these discussions land the code where you wanted? anything else you want me to do here?

jmilljr24:
> @jmilljr24 did these discussions land the code where you wanted? anything else you want me to do here?

Yup, this seems to be working fine as is. Thanks

PR #2043, opened 2026-08-02
Load grant donor via remote-select instead of loading all people/orgs

### LINKED ISSUE #1971: Fix - Grants controller loads all people/orgs for form selection
We shouldn't load all people/organizations for a field selection. This should use remote-select.

If Grants (donor) is updated to follow the Payment (payer) pattern and restrict donor to only person and org instead of the polymorphic association, this could follow the same remote-select pattern that's used for payments and keep consistency with someone (person or organization) giving money towards something.

maebeale:
🤖 suggested review level: 3 Read 📖 small form/controller change reusing the existing remote-select + search pattern

Closes #1971

- The grant form loaded **every** `Organization` and `Person` into a grouped `<select>` just to pick a donor — doesn't scale.

- Swapped the donor `select_tag` for the same `remote-select` pattern the payment payer picker uses (donors are searched on demand).
- Dropped `GrantsController#set_form_variables` (`@donor_options`) and its calls — no longer needed.
- `grant[donor_sgid]` already resolves via signed GlobalID, so the compound search labels drop straight in; no model changes.
- The shared `person_or_organization` search now lists **organizations before people** everywhere (both the grant donor picker and the payment payer picker).

### Anything else to add?
- Keeps the two "someone (person or org) gives money toward something" flows (grants/donor and payments/payer) consistent.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,19 +1,14 @@
 class SearchController < ApplicationController
   skip_before_action :preload_current_user_associations, raise: false
 
-  # Virtual "models" that search people and organizations together. Both search
-  # the same two tables; they differ only in result order. The payer /
-  # additional-designation pickers on the payment form list people first; the
-  # grant donor picker lists organizations first (organizations are the more
-  # common donor).
-  COMPOUND_MODELS = {
-    "person_or_organization" => [ Person, Organization ],
-    "organization_or_person" => [ Organization, Person ]
-  }.freeze
+  # Virtual "model" that searches people and organizations together for the
+  # compound payer / donor pickers. The optional `order` param picks which kind
+  # is listed first ("organization" → organizations first; anything else →
+  # people first) — both search the same two tables.
+  COMPOUND_MODEL = "person_or_organization".freeze
 
   def index
-    compound_classes = COMPOUND_MODELS[params[:model]]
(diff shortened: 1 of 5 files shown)


app/controllers/search_controller.rb
```diff
@@ -39,17 +41,18 @@ def index
...
   private
 
-  def compound_payer_results
-    authorize! Person, to: :search?
-    authorize! Organization, to: :search?
+  # Search people and organizations, concatenating in the order the caller
+  # asked for so each picker can lead with its most common kind.
+  def compound_results
+    model_classes = params[:order] == "organization" ? [ Organization, Person ] : [ Person, Organization ]
+    model_classes.each { |klass| authorize! klass, to: :search? }
```
jmilljr24: I think this change is better in the sense it could be extended easier to different model.

I don't think there is any value in adding the order. It adds complexity and the returned results are small enough for the order to not matter. Also most people names are different enough from organization names that when typing, the results are filtered down even more.

maebeale: i just am someone who likes to see my options before i start typing. typing into the void is hard. when i clicked it open the preview only shows people bc they're first. are you ok w changing it to just be organizations first everywhere?

jmilljr24: Yup. Org first is good.

jmilljr24 (approved):
Thank you!

PR #2044, opened 2026-08-02
Track event-reg transfers via a back-link (fix lost attendance on transfer-in)

### LINKED ISSUE #1944: Event Reg - transferred in / transferred out
#1820 added transferred to an event reg attendance status. This works for transferred out. However, if we mark someone as transferred in we lose the ability to track their actual attendance for this event they transferred in to. If I change an event_reg from transferred in to attended, I lose the inform that this person transferred in.

Without getting to complicated, I think transferred_out_to_id should be added to event_reg. An attendance can be set to transferred out. Once admin know which event the person intends to go to they can add the new event_reg_id to transferred_out_to_id. The new event_reg `has_one transferred_in_from`. This keeps the historical record without braking the attendance. If in the odd chance the person transfers out again, the new event_reg would again be set to transferred_out and the FK added. Essentially a linked list you could work back to the original registration if needed. 

Anything else related to transferring out can be discussed later (forms, payments etc).

maebeale:
🤖 suggested review level: 5 Inspect 🔬 self-FK + data migration, status-list change, transfer flow, two-record CE split, and financials-on-source semantics across reporting

Closes #1944

- Marking a registration `transferred_in` overwrote its attendance status, losing whether the person actually attended the event they transferred into.
- Fix: track the transfer as a **relationship**, not a status. A `transferred_from_registration_id` self-FK lives on the **incoming** record; the **out** stays identifiable by its terminal `transferred_out` status. Chained transfers form a linked list back to the original.

### The transfer flow
- After a reg is marked **Transferred out** (from the edit-form save **or** the inline roster/onboarding Turbo badge), the admin lands on a transfer screen to pick the destination event (same format — on-demand ↔ on-demand, scheduled ↔ scheduled), enforced both in the picker and at the POST endpoint.
- **Double transfers collapse to two live regs.** Transferring out a reg that is itself a transfer-in (A→B→C) points the new reg straight at the original source and destroys the middle stop; the deletion is captured in an Ahoy destroy lifecycle log (attributes + associated records).
- **Org links carry forward.** The registrant's linked organizations are copied onto the new reg (and kept on the source), so agency context follows them to the new event.
- **Unfinished transfers are flagged.** A reg marked transferred-out with no destination recorded yet shows a ⚠ warning on its status badge (roster/onboarding), and reads as the top "Not ready" reason on the readiness Status column, so it can't be lost.
- **Informed consent.** The transfer screen spells out, in plain language, exactly what happens to the registrant's data before you confirm — with an extra warning in the chain-collapse (2nd→3rd) case.

### Editing & undoing a transfer (Manage-transfer hub)
- The transfer screen doubles as a **"Manage transfer" hub**, reached from a **"Manage" chip** on the roster status badge (amber ⚠ until a destination is set, purple after) and an edit-form link — the status dropdown stays a pure status control.
- **Change destination:** re-point a recorded transfer to a different event; the old destination is unlinked and its CE re-merged before re-splitting to the new one.
- **Undo transfer** (`RevertTransfer`): restores the reg to its pre-transfer status; for a completed transfer it also **unlinks the destination** (it becomes a normal standalone reg, nothing deleted) and **re-merges the split CE** back onto the source (`TransferContinuingEducation#revert`).

### Money & records resolve to the source registration
- An incoming reg carries no **registration** money of its own — balance, payments, invoice, receipt, and scholarship all live on the source (the old event, where they paid).
  - `payment_status_label` = "Transferred in", readiness stops flagging "Payment due", payment access derives from the source.
  - Balance/receipt/invoice mirror the source: the ticket never re-bills a paid transfer, and the invoice/receipt (reached through the new ticket) document the **old** event's cost/payments.
  - The roster **unpaid filter** buckets a transfer by the **source's** balance.
  - **Scholarship creation** on a transfer-in is blocked at the URL (redirects to the source), not just hidden.

### Continuing education splits into two records (issue #1944)
- CE is the exception to "money stays on the source": a transfer **splits** it so each event holds its own.
  - The source keeps a paid, **zero-hours stub** (cost = amount already paid) — its payments count at the original event and it still shows in that event's CE searches.
  - The destination gets a **live record** carrying the hours and the outstanding balance, where new payments are received and the certificate is earned. Its cost is snapshotted from the source's balance and **admin-locked**, with a "from original / paid on original →" link back to the stub.
  - Collapsing a double transfer **relocates** the middle record forward (not destroyed); transferring back to the origin **merges** it into the stub.
  - Manual CE creation on a transfer-in is blocked at the URL (the transfer's system-created record is exempt).
  - CE reporting follows the record: the stub counts at the original event, the carried record at the new one.

### Transferred-out ticket
- The original (withdrawn) ticket keeps its financial/credit records (payment/invoice/receipt, scholarship, CE, certificate) but hides participation material (videoconference join, staff, handouts, FAQ, custom callouts).

### Reporting
- Registration totals stay on the **original** event; the **new** event excludes transferred-in regs from its registration totals (dashboard billable basis + revenue) and the bulk-payment allocation list — while still counting them in headcount/attendance. CE is counted per-record as above.

### Surfaced as a dimension (it's no longer a status)
- Roster badge "In" marker, dashboard breakdown row (FK-counted) with drill-down, an "Attendance status → Transferred in" filter (roster + reminders), and a CSV "(transferred in)" annotation.

### Locking a transferred-out registration
- A transferred-out reg is historical, so its fields are **locked except comments and communications**. Enforced server-side (restricted `update` params + guards on the inline onboarding / certificate / attendance / org-link endpoints and the recipients shout-out), so a locked field is **refused with a warning, not silently dropped**.
- Surfaced in the UI: a lock banner + disabled fields on the edit form (disabled `<fieldset>`s), disabled onboarding/roster inline controls, and hidden "Create user".
- The scholarship / CE / allocations pages stay **reachable** but show a **caution banner** — an admin may still need to edit those deliberately.

### Transferred-in editing & the consent screen
- **CE card:** when the source had no CE, the transferred-in reg now offers the normal **Add CE** control (CE decided now attaches to this reg) instead of a dead-end link back to the original.
- **Scholarship card:** shown on a transferred-in reg when the source had no scholarship, so an admin can award one here.
- **Financials summary:** now also surfaces the **source CE payment** (No CE / No CE payments / the amount), next to the payment and scholarship it already showed.
- **Consent screen:** tightened the per-card copy, added the event's **month + year** to disambiguate same-titled trainings, and renamed the action to **"Perform transfer"**.

### Transferred-out ticket: read-only + banner everywhere
- The public ticket for a transferred-out reg no longer shows the **"Pay with Credit Card"** button (join link and manage block were already hidden) — the person is pointed to their new registration instead.
- The transfer banner (links to the paired registration) is now rendered once from the shared callout wrapper (`_callout_page`), so it appears on **every** callout page a transferred in/out registrant can reach, not just the financial ones. The four duplicate explicit renders were removed.


app/models/event_registration.rb
```diff
@@ -310,30 +340,30 @@ def self.scholarship_allocatable_ids(scholarships)
...
+    ), 0)
+  SQL
+  BILLING_COST_CENTS_SQL = <<~SQL.squish.freeze
+    COALESCE((
+      SELECT events.cost_cents FROM events
+      INNER JOIN event_registrations billing_reg ON billing_reg.id = #{BILLING_REGISTRATION_ID_SQL}
+      WHERE events.id = billing_reg.event_id
+    ), 0)
+  SQL
   scope :paid_in_full, -> {
```
maebeale, line 359: 🤖 _From Claude:_ These shared `paid_in_full`/`not_paid_in_full` scopes are also used by the revenue drilldown, so a transferred-in reg now appears under paid/unpaid there based on its **source's** balance (money totals are unaffected — they come from the billable basis, which still excludes transfer-ins).

app/services/event_dashboard.rb
```diff
@@ -1171,15 +1228,32 @@ def allocated_by_registration
...
 
-  # Active continuing-education registrations for this event: those tied to an
-  # active event registration. The basis for every CE money figure and for the
-  # CE registrant counts / pie.
+  # CE money/counts follow the CE record, not the registration billing basis: they
+  # count every CE record on a registration of this event that wasn't cancelled or
+  # a no-show — including a transferred-out reg's paid stub (counted here, where it
+  # was paid) and a transferred-in reg's own carried record (counted at the event it
+  # now credits). (#1944)
+  def ce_basis_registration_ids
```
maebeale, line 1211: 🤖 _From Claude:_ Deliberate departure from the registration-money model: CE is counted **per record** on any non-cancelled/no-show reg of the event, so a transferred-out reg's paid stub counts here and a transfer-in's carried record counts at the destination. That's why CE uses this basis (and its own registrant map) instead of the `not_transferred_in` billable basis registration money uses — no double count, since the stub holds only the paid portion and the carried record the balance.

maebeale:
@jmilljr24 i'm still combing through this one but happy to have your thoughts while it's in flight! i don't think you care about much in here but added you since it's bigger

jmilljr24:
How do I change which event someone transfers to if i select the wrong one by mistake or if the person said they are transferring from Event 1 to Event 2, but then decides they actually are going from Event 1 to Event 3.

maebeale:
@jmilljr24 you do a 2nd transfer from the new reg, and the interstitial page tells you it's going to delete the middle records. i'll improve it tho.

jmilljr24:
> @jmilljr24 you do a 2nd transfer from the new reg, and the interstitial page tells you it's going to delete the middle records. i'll improve it tho.

I didn't know how to get back to the interstitial page. My assumption would be to have a button/link on the event_reg edit page near where it shows the transfer.

PR #2058, opened 2026-08-02
Collapse secondary registrant filters behind a CSS-only More filters toggle

maebeale:
🤖 suggested review level: 1 Skim 👀 view-only filter reorg on the registrants roster; no logic or query changes

## What
- Rework the registrants roster filter bar into one always-visible row (Keyword, Status, Organization linking, Clear filters) plus a collapsible **More filters** section for the rest.
- The collapse/expand uses a native `<details>`/`<summary>` — **pure CSS, no JavaScript**. The caret and "More filters" ↔ "Fewer filters" label track state via Tailwind `group-open` variants.

## Why
- The single flat grid of ~9 filters was noisy; most sessions only touch keyword + a status.

## Note
- The advanced section **auto-opens** when one of its filters is already active (via a `params` check), so a hidden filter is never silently applied.

### DIFF
--- a/app/views/events/_registrants_search.html.erb
+++ b/app/views/events/_registrants_search.html.erb
@@ -1,6 +1,11 @@
 <% field_class = "w-full rounded-lg border border-gray-300 px-3 py-2 text-gray-800 shadow-sm
             focus:border-blue-500 focus:ring focus:ring-blue-200 focus:outline-none" %>
 <% label_class = "block text-sm font-medium text-gray-700 mb-1" %>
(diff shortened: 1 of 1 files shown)


maebeale:
🤖 _From Claude:_ Closing as redundant — superseded by #2143 (merged), which replaced `_registrants_search.html.erb` with the shared `events/_registrant_filters` + `_filter_bar` collapsible "More filters" bar (pure-CSS `:has()` disclosure, auto-open on active filter, shared with the bulk-reminder picker).

PR #2070, opened 2026-08-03
Topic subscriptions — track interest in upcoming trainings (+ admin index)

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new models + migration, an admin CRUD, a topic-dimension refactor (string → lookup model), and small Stimulus/JS; wide surface though contained to this feature

## Why
Registrants often want to hear about a future training before there's a scheduled event to register for. This was hacked via placeholder "Training Interest" events, which polluted attendance statuses, capacity, reporting, and payment. Model it as a standing **subscription to a topic** instead — and make the topic itself an admin-managed list so it can grow past trainings.

## What
**`TopicSubscription`** — a lightweight signal, **not** a placeholder registration:
- `belongs_to :topic_subscription_type` (the topic) + optional `interested_event` (null = topic broadly)
- state is timestamp-driven — `unsubscribed_at IS NULL` = active (`active?`/`unsubscribe!`/`resubscribe`); `subscribed_at` + `source` mirror `mailing_list_consent_*`
- kept distinct from the consent flag ("you may email me") and from `EventRegistration` (an enrollment); one active sub per (person, type, event), enforced on update as well as create — so an edit or a resubscribe can't quietly produce a duplicate

**`TopicSubscriptionType`** — admin-editable list of topics (full CRUD, nav via Subscriptions → "Manage topics"):
- editable `name` + immutable derived `key` slug (stable for code lookups); `archived_at` retires a topic without deleting it (types in use can't be destroyed — archive instead)
- seeded for all envs from `CANONICAL` (facilitator_trainings / news / resources)
- index defaults to **active** topics with an active/archived segmented toggle (counts); each row's name is a clickable card (light for active, dark for archived) that opens edit; archive/restore + delete live on the edit page
- `event_selector` flag marks topics tied to events (only facilitator trainings) so the subscription form shows the event picker for those and hides it for the rest

**Admin subscriptions index** (`/topic_subscriptions`, admin-only):
- Person (remote-select) + Topic filters in one live-searching row; **status is a segmented Active/Unsubscribed toggle with counts** above the results, defaulting to active — the page answers "who can I still email" without picking a filter first. The toggle navigates the page rather than the results frame, so the out-of-frame filter form re-renders holding the selected status; it also renders above the empty check, so a status with no rows still offers a way out
- **Training registrations** column showing each person's `facilitator_training` registrations (eager-loaded)
- status reads at a glance: active is a soft "live" pill, **unsubscribed** is a solid slate chip with a bell-slash (legible even in a de-emphasized row); the person cell and the status badge both link to the subscription edit and carry the active filters back
- event picker on the form lists **all upcoming events**, not just facilitator trainings
- **Email addresses** page (eyebrow link) for pasting into a mail client: reuses the index filters, shows the applied ones as grey chips, and narrows to who's actually mailable — drops unsubscribed + already-registered people, each restorable via a toggle. Counts report the *addresses* a toggle would add, not the subscriptions it hides, so someone reachable via another subscription isn't reported as left out

**Entry point from an event** — "Interest form" item in the event Forms dropdown:
- opens the subscription form prefilled to that event + the canonical trainings topic (resolved by stable key), returning there on save/cancel
- topic + event fields show a read-only value with a pencil that reveals a picker, reusing the existing `edit-toggle` controller; `dirty-form` guards unsaved changes
- the eyebrow, the form's Cancel, and the controller's post-save redirect all read one `TopicSubscriptionHelper` mapping, so they can't diverge

**Person associated records** — reorganized into Administrative / Contributed content clusters, with a Subscriptions card:
- the three idea cards now match on the **person** via a new `AuthorCreditable#created_by_person` scope (ideas carry no `author_id`; the credited person is the creating user's person). Previously they keyed on `person.user&.id`, and a person with no user account got a card reading "0" that opened the unfiltered index of everyone's ideas

## Follow-ups
- #2071 — feed active training subscriptions into the bulk email reminder recipient list
- #2080 — (fixed here) idea cards falling through to the unfiltered index
- Auto-capture from the registration form's `interested_in_more` answer (→ `INTERESTED_IN_MORE_KEY`) + backfill historical answers


app/views/topic_subscriptions/topic_subscriptions_results.html.erb
```diff
@@ -0,0 +1,120 @@
...
+    <%# Outside the empty check: with nothing to list, the toggle is the only way
+        back to the other status. It navigates the whole page rather than this
+        frame so the filter form — which lives outside the frame and carries the
+        status through every other filter change — is re-rendered with it. %>
+    <% total = @topic_subscriptions.respond_to?(:total_entries) ? @topic_subscriptions.total_entries : @topic_subscriptions.size %>
+    <% toggle_params = request.query_parameters.except("page") %>
+    <div class="flex items-center justify-between px-6 py-4 border-b border-gray-100">
+      <nav class="inline-flex rounded-lg bg-gray-100 p-0.5 text-sm font-medium" aria-label="Status filter">
+        <%= link_to topic_subscriptions_path(toggle_params.merge(status: "active")),
+              data: { turbo_frame: "_top" },
```
maebeale, line 12: 🤖 _From Claude:_ Deliberate trade-off: the toggle navigates `_top` instead of swapping this frame, costing a full page load on status switch. A frame swap never advances the URL, and the filter form lives outside the frame and is never re-rendered — so its hidden `status` field went stale and the next filter change silently reset the list to Active. The alternatives were moving the filter form inside the frame (against the lazy index/filter frame convention) or syncing the field with JS.

PR #2072, opened 2026-08-03
Report real CE payments in the dashboard, revenue report, and CSVs

maebeale:
🤖 suggested review level: 5 Inspect 🔬 money aggregation logic feeding the CEO dashboard + revenue report; verify CE paid/due sums

CE payments are a real feature now (`ContinuingEducationRegistration` + polymorphic allocations), but reporting still treated CE as unbuilt or uncollected, so CE revenue was silently under-reported.

- **Event dashboard** — `EventDashboard#cont_ed_*` were zero stubs, so the CE fee card, grand total, collected, due, and monies-made all showed **$0**. Implemented them from CE registrations' cost and CE allocations, mirroring the registration-fee methods: paid = CE cash collected (payment allocations), due = CE cost still owed after all allocations, total = paid + due. Added per-registrant CE paid/due maps and paid/unpaid registrant lists.
- **Revenue report** — CE was counted as *always outstanding*. CE cash collected now counts as fees / money-in / net, and drops out of outstanding once paid. A **comped CE fee counts as org subsidy** — otherwise it's neither collected nor owed and disappears from the report; CE cost is only ever paid, discounted, or owed.
- **`EventRevenueFigures`** — the report built an `EventDashboard` per event (~8 queries each). All components for every event now load in **five grouped queries**, with a parity spec holding the batched figures to the dashboard's definitions so a row can't drift from that event's dashboard.
- **CSV exports** — registration, onboarding, and cross-event registration CSVs omitted CE dollars. Added **CE paid** / **CE due** columns via a new `EventRegistration#ce_amount_paid_cents`. Normalized every CSV dollar cell (including the legacy "Payment total") through a shared `ApplicationController#csv_dollars` → `dollars_from_cents`, so a single export reads consistently (`$50`, not `50.00`). Preloaded what those cells read, and taught `Person#phone_number` to use the loaded `contact_methods` instead of querying past them. All three exports are now **flat — zero queries per row** (registrations index was 13/row), pinned by query-count specs.
- **Free events** — the dashboard no longer hides the whole money section on a free-registration event that still charges for CE.
- **CE filter** — `ce_status` gains a `registered` value ("signed up for CE, whatever its state") so the CE card's registrant chip and total actually drill into the CE roster instead of the unfiltered list.

### Anything else to add?
- Comped CE fees → org subsidy is a live accounting question: #2084
- Bulk-payment → CE allocation deferred to #2082 (needs UI work; not a small change). The bulk-payments "already allocated" figure stays registration-fee-scoped by design.
- `EventDashboard#discount_cents` removed — the revenue report was its only caller, and `EventRevenueFigures` now owns the (broader) definition.
- Tests: dashboard CE money (paid/partial/unpaid/cancelled-ignored), revenue report collected + comped CE, figures↔dashboard parity and query count, `ce_status("registered")`, and CSV CE-dollar columns.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -239,14 +239,18 @@ class Person < ApplicationRecord
     organizations.pluck(:id)
   end
 
+  # The primary active phone, or the first one on file. Reads the loaded
+  # contact_methods when a caller has preloaded it (rosters, CSV exports), so a
+  # page or export of people doesn't pay a query per row.
   def phone_number
-    primary_phone = contact_methods.find_by(primary: true, inactive: false, kind: :phone)
-    return primary_phone.value if primary_phone.present?
+    phones =
+      if contact_methods.loaded?
+        contact_methods.to_a.select { |method| method.phone? && !method.inactive? }
+      else
+        contact_methods.where(kind: :phone, inactive: false).to_a
+      end
 
-    first_phone = contact_methods.where(kind: :phone, inactive: false).first
-    return first_phone.value if first_phone.present?
-
-    nil
+    (phones.find(&:primary?) || phones.first)&.value
   end
 
   def has_liasion_position_for?(organization_id)
--- a/app/services/event_revenue_figures.rb
+++ b/app/services/event_revenue_figures.rb
@@ -1,0 +1,126 @@
+# Per-event money components for the revenue report, loaded for every event at
+# once in a fixed number of grouped queries instead of an EventDashboard (and its
+# ~8 queries) per event.
+#
+# The definitions mirror EventDashboard's — active registrations only, cost net of
+# every allocation and floored at zero, payment allocations for cash collected —
+# so a row in the report and that event's dashboard can't disagree. A parity spec
+# holds the two together.
+#
+# One definition is broader than the dashboard's: discounts cover CE registrations
+# too, not just event registrations. CE cost is only ever paid, discounted, or
+# owed, so counting CE discounts as org subsidy is what keeps a discounted CE fee
+# from vanishing from the report entirely.
+class EventRevenueFigures
+  Figures = Struct.new(
+    :registration_payments_cents,
+    :registration_outstanding_cents,
+    :funded_scholarship_cents,
+    :unfunded_scholarship_cents,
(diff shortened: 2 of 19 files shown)


app/services/event_revenue_figures.rb
```diff
@@ -0,0 +1,126 @@
...
+#
+# The definitions mirror EventDashboard's — active registrations only, cost net of
+# every allocation and floored at zero, payment allocations for cash collected —
+# so a row in the report and that event's dashboard can't disagree. A parity spec
+# holds the two together.
+#
+# One definition is broader than the dashboard's: discounts cover CE registrations
+# too, not just event registrations. CE cost is only ever paid, discounted, or
+# owed, so counting CE discounts as org subsidy is what keeps a discounted CE fee
+# from vanishing from the report entirely.
```
maebeale, line 13: 🤖 _From Claude:_ This is the one place the batched figures deliberately diverge from `EventDashboard`: discounts now cover CE registrations too. Without it a comped CE fee is neither collected nor owed, so it drops out of the report entirely — worth a sanity check that treating it as org subsidy (which lowers net and total expected by the comped amount, exactly like a discounted registration fee) is the accounting you want.

app/models/person.rb
```diff
@@ -239,14 +239,18 @@ def organization_ids
...
+  # page or export of people doesn't pay a query per row.
   def phone_number
-    primary_phone = contact_methods.find_by(primary: true, inactive: false, kind: :phone)
-    return primary_phone.value if primary_phone.present?
-
-    first_phone = contact_methods.where(kind: :phone, inactive: false).first
-    return first_phone.value if first_phone.present?
+    phones =
+      if contact_methods.loaded?
+        contact_methods.to_a.select { |method| method.phone? && !method.inactive? }
```
maebeale, line 248: 🤖 _From Claude:_ Widest-reaching change in the PR — `phone_number` is called from a lot of places, not just the CSVs. Behavior is identical (primary active phone, else first active phone, else nil, ignoring non-phone kinds); the only difference is reading already-loaded records instead of re-querying. New model specs assert the loaded and unloaded branches agree on every case.

PR #2073, opened 2026-08-03
CE callout: surface fee/deadline; show pending certificate card

maebeale:
🤖 suggested review level: 5 Inspect 🔬 CE deadline column type change (date→datetime) + certificate card visibility/gating change across the ticket, callout page, and controller

## CE callout page
- Registered w/ balance: a **Payment due** row sits with Cost / Amount paid, showing the deadline in the viewer's zone (e.g. "9:00 AM PDT on July 22, 2026"). Hidden once paid in full.
- Opt-in (not yet registered): the vague "available for an additional fee" becomes the concrete **fee** ("CE hours are available for $150.").
- `ce_payment_due_deadline` becomes a **datetime** (was date-only) so admins can set the time due; edited via Date + Time inputs; formatted by `EventDecorator#ce_payment_due_deadline_display`.
- License-locked note's **"Contact us"** links to the contact page, breaking out of the `license_section` Turbo frame (`data-turbo-frame="_top"`) so it doesn't trip the frame-missing handler.

## Certificate of completion card
- A **published** certificate row now shows even before it unlocks — a pending, gray-badged "Available after the event" card — so admins who publish it actually see it. Its page lists the outstanding unlock conditions (event ended / attended / scholarship tasks) with met/unmet checkmarks, mirroring the videoconference card+page. The controller no longer redirects away when locked.
- The code-defined fallback set (unseeded events) is unchanged: still surfaces the certificate only once downloadable.

## Notes
- Existing CE deadline values (dates) migrate to midnight; admins re-set the time going forward (no backfill — per discussion). Migration is reversible.
- Also tightened the AI comment guideline to prefer one-line comments.

## CE invite card & request flow
- A **not-yet-requested** registrant now sees an invite CE card (previously none) prompting "Request continuing education credit", shown only while the request window is open (no `ce_hours_request_deadline`, or it hasn't passed).
- Its badge names the request deadline — **"Request CE by Jan 10"** — with no fee, since they haven't opted in and owe nothing yet (the fee still surfaces on the card once requested). Nil when no request deadline is set.
- Requesting CE now flips **in place** inside a `ce_request_section` Turbo frame: the POST redirects back and Turbo swaps in the license-entry section, no full-page reload.


db/migrate/20260803155552_change_ce_payment_due_deadline_to_datetime.rb
```diff
@@ -0,0 +1,12 @@
+class ChangeCePaymentDueDeadlineToDatetime < ActiveRecord::Migration[8.0]
+  # The CE payment deadline now carries a time of day (e.g. "9:00 AM PT") so the
+  # CE callout page can state exactly when payment is due, matching how the event's
+  # own start/end times are stored and displayed.
+  def up
+    change_column :events, :ce_payment_due_deadline, :datetime
```
maebeale, line 4: 🤖 _From Claude:_ Heads-up for existing data: any event that already had a payment-due **date** set becomes midnight UTC here, which renders as the *previous* day in the app's Pacific zone until an admin re-saves the deadline with a time. New form entries parse in Pacific so they're fine — but pre-existing deadlines may want a one-time re-set (or a small backfill) to avoid showing a day early.

PR #2075, opened 2026-08-04
CE sign-in/out tracking + attendance report

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new model + migration, public sign-in/out and self-edit actions, three edit surfaces sharing one nested-attribute write, a certificate gate, and a per-event report across the CE callout + participation/registrants surfaces

Replaces AWBW's paper CE-hour sign-in sheet with in-portal sign-in/out, kept invisible to the general training group — surfaced only on each registrant's private CE callout, and only once CE is paid in full.

## Why
- CE hours must live in a durable, auditable system of record for the CE-accrediting body; a shared Google sheet isn't that.
- Staff need one screen to see who's signed in/out during a training to chase reminders, and to fix a missed sign-in without leaving it — the attendance report.
- The buttons only ever record "now". Someone who arrived before signing in, forgot to tap them, or is writing the training up afterwards has to be able to fix their own hours rather than ask staff.

## What
- **`EventAttendanceTimeEntry`** — generic per-day sign-in/out pairs (many/day for breaks), audited by `created_by`/`updated_by`. `Event#attendance_sign_in_open?` derives a per-day window from the event's single start/end time-of-day. Guards reject >24h logged on a day and any same-day overlap (checking persisted rows + the in-memory nested batch, so self-service and every edit surface are covered). An entry left open occupies the rest of its day, so a later session can't be recorded underneath it and make the eventual sign-out unsaveable.
- **CE callout** — one button by state: **Sign in** (inside the day's window) / **Sign out** (any time that day, while an entry is open), plus a separate catch-up button to close a day they forgot, stamped with that day's scheduled end rather than "now". The callout **card** shows a live nudge mirroring the payment-due badge.
- **The registrant's own sign-in sheet** — every training day, its sessions and total, each with **Edit**. Deliberately not windowed and kept up after the training ends: filling in a day afterwards is the whole point. Uses the same clock-time editor staff use on the report.
- **One sheet per licence** — a registrant claiming CE against two licences gets a sheet per `ContinuingEducationRegistration`, headed by that licence's number and awarded hours, each gated on its own payment, because each board is shown only its own. The times underneath are shared (one person, one room, one set of hours), so the sheets can't disagree; the sign-in controls stay singular above them. `RegistrantAttendanceSheet` builds the state once and every sheet renders from it.
- **`EventAttendanceReport`** — per event, grouped by day then reported line. Leads with a teal **Totals** table (Day N logged columns, per-day + hours-awarded totals in the All row); a line's Total logged is scoped to the event's days so it always equals its day columns. `?ce=true` scopes to CE registrants and adds license # / awarded hours. Reached from CE-eligible events on the participation report **and** the registrants **bulk-actions dropdown**.
- **A report line per licence** — the reported unit is a `Row` (registration + the `ContinuingEducationRegistration` its hours are certified against), so the report follows the same split as the callout: two licences, two lines, each showing only its own number and hours. Cells are keyed by `Row#key`, so two lines for one person don't share a DOM id and open both editors at once. Hours awarded sum across lines (each board awards its own); logged minutes deliberately don't, since the lines share one set of times — the All row counts each person once.
- **Correcting times** — three surfaces, one write: the registrant's own day table, the report's in-place per-day cell (both clock times, the day coming from the section it sits in, both rendering `events/_attendance_day_form` and parsing through `AttendanceDayRows`), and the CE edit page's full datetime editor spanning every day. All go through `EventAttendanceEntriesUpdate` and hand submitted values back on a rejected save.
- **Certificate gate** — when attendance was tracked, the CE certificate also requires logged time to cover ~90% of the awarded contact hours; events that never tracked time are unaffected (no retroactive blocking).

## Notes
- Everything the registrant does — signing in, signing out, correcting a time — is **public and unattributed** (`created_by`/`updated_by` nil), so those columns keep meaning "a staff correction happened here".
- **An open entry never crosses midnight.** A forgotten sign-out can be closed any time that day, or later via the catch-up button (which uses the day's scheduled end); it can't block the next day's sign-in or bank a ~24h session. The report's `signed in` chip is per-day, so one dangling entry doesn't light up every later day.
- Sign-in/out and self-edits are allowed once *any* CE registration is paid (`EventRegistration#ce_attendance_offered?`), not all of them — previously the view gated on `.first` while the controller required every CE paid, so a part-paid second licence rendered a section whose buttons refused.
- `EventAttendanceEntriesUpdate` loads the entries collection explicitly before saving — nested attributes only add the submitted rows to the target, and the cross-entry guards need the rest to compare against.
- Nested-association validation failures reach the flash as the child model phrases them, not prefixed with the humanized association name.
- Events store only one start/end datetime, so multi-day windows assume each day runs the same hours (same assumption `day_count` already makes). Times use the request zone (Pacific), matching the paper sheet.
- Rebased on main (incl. #2073's CE-card rework — the live sign-in nudge composes with its new fee/deadline/request-card logic).

## Tests
Model/service/request/policy/view specs across every part, including the two entry guards, the day-boundary behaviour of an entry left open, the registrant's self-edit round trip (add/correct/remove, rejected save, unpaid CE refused), the report editor's round-trip on a rejected save, and the total/column consistency. Full suite green.


app/services/event_attendance_entries_update.rb
```diff
@@ -0,0 +1,58 @@
...
+  # report. A no-op when there's nothing left to apply.
+  def save!
+    applicable = applicable_rows
+    return if applicable.blank?
+
+    @registration.assign_attributes(event_attendance_time_entries_attributes: applicable)
+    # Nested attributes only add the submitted rows to the association's target;
+    # loading it merges in the rest, so each entry's cross-entry guards (overlap, the
+    # daily limit) compare against its siblings-in-progress and their unsaved edits.
+    @registration.event_attendance_time_entries.load
```
maebeale, line 31: 🤖 _From Claude:_ This `.load` is load-bearing and looks removable — nested attributes only add the submitted rows to the association target, so without it `EventAttendanceTimeEntry#same_day_siblings` compares each entry against the stale persisted rows and rejects valid batches (e.g. moving one session into a slot another session in the same submit just vacated). It was previously an accidental side effect of walking the collection to stamp the editor, which no longer happens on registrant edits.

PR #2089, opened 2026-08-04
Attribute admin-sent emails to the sending person (invites, reminders, resends) + show From

maebeale:
🤖 suggested review level: 3 Read 📖 sender attribution threaded through several admin-triggered send paths; one shared DeviseMailer/CreateNotification change + a show-page field

## Why
- Admin-triggered emails were recorded with no sender, so the communications index/show labeled them "**AWBW Portal**" as if the portal sent them automatically. Attribute them to the acting person instead. A person's name should appear whenever a staff member sent it; "AWBW Portal" only for truly automated messages.

## What
- **Invites** — the sender rides on the `User` record: `send_confirmation_instructions(sender:)` exposes `confirmation_sender`, which `DeviseMailer` reads when it logs the notification. Devise sends with `deliver_now`, so the mailer holds the same instance the caller does. `CreateNotification` accepts `sender:`.
- **Event reminders** — the bulk reminders page (`events#send_reminder`) now passes `sender: current_user` per recipient.
- **Resends** — `notifications#resend` attributes the resent copy to the admin who resent it.
- **Show page** — added a **From** row, backed by `NotificationDecorator#sender_name` (also used by the index and the notification row, which previously disagreed on capitalization).

## Tests
- Reminders: each notification's `sender` is the admin, and no `sender_id` is nil (guards the "AWBW Portal" regression).
- Resend: resent copy attributed to the acting admin.
- Show page: the **From** row names the sender when set, "AWBW Portal" when not (scoped to the row's `<dd>`).
- Bulk invite: `sender_id` threaded to the job, and the job passes it into `send_confirmation_instructions`.
- Decorator: `sender_name` for both branches.

## Notes
- Deliberately **not** using `Current.user`: a background job shouldn't mutate request-scoped state, and `Current.user` is what gates `AhoyTrackable#track_lifecycle_event` — setting it in a job made every invite push an event onto `LifecycleBuffer`, a thread-local only `ApplicationController` flushes. Those piled up unflushed and undelivered.
- The five call sites (bulk invite job, `users#send_welcome_instructions`, `ProcessEmailChange`, `ProcessEmailManualConfirm`, `ProcessConfirmation`) pass the sender explicitly; the last four sit right next to an existing `updated_by = @current_user` "credit the acting admin" line.
- `sender:` on the console bulk invite is optional — omitting it falls back to "AWBW Portal" and logs a warning.
- The existing `auth.confirmation_email_sent` ahoy event is attributed to the sender too, when there is one.


app/models/user.rb
```diff
@@ -208,11 +208,17 @@ def gallery_assets # method needed for idea_submitted_fyi mailer
     []
   end
 
+  # The staff member who triggered this confirmation email, when one did. Not
+  # persisted — DeviseMailer reads it off the record to attribute the
+  # notification it logs, since it has no request and no current_user.
+  attr_reader :confirmation_sender
```
maebeale: 🤖 _From Claude:_ This relies on Devise's default `send_devise_notification` using `deliver_now`, so `DeviseMailer` receives the same in-memory `User` the caller set the sender on. If anyone ever overrides it to `deliver_later`, the record round-trips through GlobalID and this ivar is silently lost — the notification would go back to reading "AWBW Portal".

maebeale: 🤖 _From Claude:_ Good catch — fixed in 76c798699. The sender no longer rides on a `@confirmation_sender` ivar of the `User` record; it's passed as a plain `sender_id` in the mailer opts (`send_devise_notification(:confirmation_instructions, token, to:, sender_id:)`). `DeviseMailer` pulls it out of opts (deleting it before `super` so Devise doesn't fold it into the headers) and looks up the sender there. A hash arg survives GlobalID serialization, so the "From: <admin>" attribution now holds under `deliver_later` too, not just the current `deliver_now` default.

PR #2093, opened 2026-08-04
Make the profile the source of truth for author credit

maebeale:
🤖 suggested review level: 5 Inspect 🔬 changes how every author credit renders app-wide, narrows search results, and adds an admin reconciliation page

Name display was asked in two unrelated places with two different value sets — `people.display_name_preference` (people index) and `<item>.author_credit_preference` (every credit, on 8 forms). `AuthorCreditable` deliberately called `person.full_name` rather than `person.name`, so the per-item answer won and the profile was ignored. One facilitator could render three different ways with no single place to fix it.

**Now:** the credited person's profile decides how every credit renders.

### Two axes on the profile
| | Column | Governs |
|---|---|---|
| Name format | `display_name_preference` (4 values) | `Person#name` **and** author credits |
| Anonymity | `anonymous_contributions` | Author credits only — the person stays listed and formatted normally on the people index |

`anonymous_contributions` already existed on main but was inert; nothing read it. This makes it govern credits, so `StoryImporter` flagging an anonymous import now genuinely suppresses that credit.

### Interim: forms still collect a per-item preference
Facilitators **cannot edit their own profile** (`PersonPolicy#edit?` is `admin?`), so all 8 forms still collect `author_credit_preference`. It records what the submitter wants — it does **not** drive display or search. A choice that differs from the profile surfaces on the divergences page, and an admin applies it to the profile. Admin forms show the field alongside the existing note and divergence warning, not instead of them.

Follow-up to hide the field from non-admins once profiles are self-service: **#2261**.

### The one asymmetry
Anonymity is a one-way latch — true if the profile says so *or* the item says so, and neither can strip it from the other. Anonymity is inherently per-item, admins need a per-item lever, and there is no backfill, so without it every item currently storing `anonymous` would start rendering its author's real name on deploy.

### Search follows the profile too
Anonymous matches nothing; `first_name_only` isn't findable by last name; `first_name_last_initial` matches the initial only. This meant dropping the `person_first`/`person_last` SearchCop attributes on `Story`/`CommunityNews`, which bypassed the gate. **Tradeoff:** an author name can no longer be one term of an AND full-text query — honoring the preference needs per-person branching a flat index can't express.

### How credit works now, per model

**Collection is unchanged across all 8.** Every form still submits `author_credit_preference`. If blank, `before_create` snapshots the credited person's profile value. The stored value is a consent record, not a formatter — only `anonymous` has live effect.

**Display is one method, `AuthorCreditable#author_credit`.** Precedence: person author → legacy free-text column → `"AWBW Staff"`. The person's `display_name_preference` decides the format; anonymity suppresses to `"AWBW Facilitator"`. No creator fallback — entering a record never credits the enterer.

| Model | Author source | Legacy column | Credit renders |
|---|---|---|---|
| Story | `author` | — | person → "AWBW Staff" |
| CommunityNews | `author` | — | person → "AWBW Staff" |
| WorkshopVariation | `author` | — | person → "AWBW Staff" |
| Workshop | `author` | `full_name` | person → legacy → "AWBW Staff" |
| Resource | `author` | `legacy_author_name` | person → legacy → "AWBW Staff" |
| StoryIdea | *none* | — | always "AWBW Staff" |
| WorkshopIdea | *none* | — | always "AWBW Staff" |
| WorkshopVariationIdea | *none* | — | always "AWBW Staff" |

The three idea models have no `author_id`. Their pages show the submitter via `created_by.name` — admin-facing attribution, not a credit.

**Keep in mind**
- The per-item field is interim — collected, not honored for display, until profiles are self-service (#2261).
- No backfill, so legacy rows carry a NULL preference. That shape is what the leaks below exploited.
- `Person#name` is format-only. Anonymity governs credits, never the people index or profile header.

### Legacy author names no longer escape the gate

`workshops.full_name` and `resources.legacy_author_name` were still full-text indexed, so a name the credit suppressed to a generic label was findable by typing it. Verified both, then fixed:

- Workshop needed more than dropping the SearchCop attribute — its free-text query **never used SearchCop at all**, building its own `LIKE` over a field list that included `full_name`.
- A legacy column is now only searchable when no person author outranks it for display. Otherwise a stale column leaks past an anonymous profile even when the record's own snapshot looks innocent.
- `WorkshopIdea.author_name` went through the credited-person path, which is empty for a model that credits nobody, so the admin filter silently matched nothing. It now matches the submitter the page shows.

### Reconciliation page (`/author_credit_divergences`, linked from admin home)
Four sections, in the order to work them, each naming the code it retires once cleared:

1. **Preference drift** — stored request no longer matches the profile, grouped by person
2. **Legacy name** — credited by a free-text column, reported per column so each field can be retired on its own, with a suggested person matched from the name
3. **Creator fallback** — no `author_id`, so the credit falls back to the creating user's person (idea models excluded — that's their only credit path)
4. **No author** — nothing to credit, renders a generic placeholder

Sections 2–4 all resolve by assigning an `author_id`, the only credit path that follows a profile, links to it, and lists the record there.

### Also swept
Four places where a real name escaped past an anonymized credit, two phantom strong params referencing columns that don't exist, and the fallback labels moved to constants so views stop hardcoding them.

### Verified
7403 examples, 0 failures. Rubocop clean on 1564 files, brakeman 0 warnings. Rebased onto main (conflict-free).

### DIFF
--- a/app/controllers/author_credit_divergences_controller.rb
+++ b/app/controllers/author_credit_divergences_controller.rb
@@ -33,6 +33,15 @@ class AuthorCreditDivergencesController < ApplicationController
     return redirect_to(author_credit_divergences_path(filters), alert: "Unknown record type.") unless model
 
     record = model.find(params[:record_id])
+
+    # Anonymity is a one-way latch: clearing the snapshot of an item submitted
+    # anonymously would silently de-anonymize it. A deliberate re-credit still works
+    # by picking an explicit preference.
+    if params[:author_credit_preference].blank? && record.author_credit_preference == AuthorCreditable::ANONYMOUS
+      return redirect_to(author_credit_divergences_path(filters),
+                         alert: "Can't clear the consent for an item submitted anonymously — pick an explicit preference instead.")
+    end
+
     record.author_credit_preference = params[:author_credit_preference]
     record.updated_by = current_user if record.respond_to?(:updated_by=)
 
--- a/app/controllers/author_credit_divergences_controller.rb
+++ b/app/controllers/author_credit_divergences_controller.rb
@@ -36,13 +36,9 @@ class AuthorCreditDivergencesController < ApplicationController
 
     record = model.find(params[:record_id])
 
-    # Anonymity is a one-way latch: clearing the snapshot of an item submitted
-    # anonymously would silently de-anonymize it. A deliberate re-credit still works
-    # by picking an explicit preference.
-    if params[:author_credit_preference].blank? && record.author_credit_preference == AuthorCreditable::ANONYMOUS
-      return render_divergence_change("Can't clear the consent for an item submitted anonymously — pick an explicit preference instead.", :alert)
-    end
-
+    # Clearing an "anonymous" snapshot hands the item back to the profile, which may
+    # well credit it. That's the point: a per-item anonymous flag is the legacy state
+    # this page exists to drain, and the person's profile is the source of truth.
     record.author_credit_preference = params[:author_credit_preference]
     record.updated_by = current_user if record.respond_to?(:updated_by=)
 
--- a/app/controllers/author_credit_divergences_controller.rb
+++ b/app/controllers/author_credit_divergences_controller.rb
@@ -91,7 +91,7 @@ class AuthorCreditDivergencesController < ApplicationController
   end
 
   def person_params
-    params.require(:person).permit(:display_name_preference, :contributions_anonymous)
+    params.require(:person).permit(:display_name_preference, :anonymous_contributions)
   end
 
   # Carried through every redirect so the admin lands back on the same filtered list.
--- a/app/controllers/author_credit_divergences_controller.rb
+++ b/app/controllers/author_credit_divergences_controller.rb
@@ -3,8 +3,6 @@ class AuthorCreditDivergencesController < ApplicationController
 
   FILTER_KEYS = %i[person_id type preference include_reconciled].freeze
 
-  # The full page renders only the header, filters, and an empty results frame;
-  # the frame's src request builds the divergences.
   def index
     return unless turbo_frame_request?
 
@@ -12,8 +10,7 @@ class AuthorCreditDivergencesController < ApplicationController
     render :author_credit_divergences_results
   end
 
-  # Resolve a whole person: point their profile at one preference and stamp them
-  # reconciled so a deliberate divergence stops reappearing on the worklist.
+  # Stamped reconciled so a deliberate divergence stops reappearing on the worklist.
   def update_person
     person = Person.find(params[:id])
     person.assign_attributes(person_params)
@@ -27,18 +24,16 @@ class AuthorCreditDivergencesController < ApplicationController
     end
   end
 
-  # Resolve a single item by rewriting its stored consent snapshot. Setting
-  # "anonymous" here is a live override that suppresses that item's credit alone;
-  # any other value only re-records history (see AuthorCreditable).
+  # "anonymous" here suppresses this item's credit; any other value only re-records
+  # history (see AuthorCreditable).
   def update_item
     model = AuthorCreditDivergenceQuery.model_for(params[:record_type])
     return render_divergence_change("Unknown record type.", :alert) unless model
 
     record = model.find(params[:record_id])
 
-    # Clearing an "anonymous" snapshot hands the item back to the profile, which may
-    # well credit it. That's the point: a per-item anonymous flag is the legacy state
-    # this page exists to drain, and the person's profile is the source of truth.
+    # Clearing "anonymous" hands the item back to the profile, which may credit it —
+    # that's what this page exists to do.
     record.author_credit_preference = params[:author_credit_preference]
     record.updated_by = current_user if record.respond_to?(:updated_by=)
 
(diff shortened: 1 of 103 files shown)


spec/models/community_news_spec.rb
```diff
@@ -97,15 +88,36 @@
...
-        results = CommunityNews.search('John technology')
+      # Deliberate tradeoff: an author name can no longer be one term of an AND
+      # query, because honoring the credit preference needs per-person branching
+      # that a flat SearchCop index can't express.
+      it 'does not combine an author name with a content term' do
+        expect(CommunityNews.search('John technology')).to be_empty
+      end
+    end
+
+    context 'via search_by_params (the user-facing path)' do
```
maebeale, line 105: 🤖 _From Claude:_ Behavior change worth knowing: an author name can no longer be one term of an AND full-text query (`search("John technology")` used to work). Honoring the credit preference needs per-person branching that a flat SearchCop index cannot express, so person-name search moved entirely to `by_credited_person_name`, OR-ed in by `search_by_params`. Single-term author search still works.

app/views/author_credit_divergences/author_credit_divergences_results.html.erb
```diff
@@ -0,0 +1,112 @@
...
+              </tr>
+            </thead>
+            <tbody class="divide-y divide-gray-100">
+              <% group.records.each do |record| %>
+                <tr>
+                  <td class="px-4 py-2">
+                    <%= link_to divergence_record_title(record), polymorphic_path(record),
+                                target: "_blank", rel: "noopener",
+                                title: "Opens in a new tab",
+                                class: "text-blue-700 hover:underline" %>
```
maebeale: 🤖 _From Claude:_ Deliberate deviation from the eyebrow convention. These link to 8 different destinations, none of which has a `return_to` mechanism today (they use hardcoded utility-link rows), so retrofitting all eight was out of scope here — using `target="_blank"` instead. Happy to do the eyebrow work as a follow-up if you would rather link in-page.

spec/models/community_news_spec.rb
maebeale, line 105: 🤖 _From Claude:_ Behavior change worth knowing: an author name can no longer be one term of an AND full-text query (`search("John technology")` used to work). Honoring the credit preference needs per-person branching a flat SearchCop index cannot express, so person-name search moved entirely to `by_credited_person_name`, OR-ed in by `search_by_params`. Single-term author search still works.

app/views/author_credit_divergences/author_credit_divergences_results.html.erb
maebeale: 🤖 _From Claude:_ Deliberate deviation from the eyebrow convention. These rows link to 8 different destinations, none of which has a `return_to` mechanism today (they use hardcoded utility-link rows), so retrofitting all eight was out of scope — using `target="_blank"` instead. Happy to do the eyebrow work as a follow-up if you would rather link in-page.

maebeale: 🤖 _From Claude:_ Deliberate deviation from the eyebrow convention. These rows link to 8 different destinations, none of which has a `return_to` mechanism today (they use hardcoded utility-link rows), so retrofitting all eight was out of scope — using `target="_blank"` instead. Happy to do the eyebrow work as a follow-up if you would rather link in-page.

maebeale:
@jmilljr24 i don't think you care about much in here but added you since it's bigger

app/controllers/author_credit_divergences_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+        @result = AuthorCreditDivergenceQuery.new(**filters.symbolize_keys).call
+        render :divergence_change
+      end
+      format.html { redirect_to author_credit_divergences_path(filters), flash: { type => message } }
+    end
+  end
+
+  def authorize_page
+    authorize! :author_credit_divergence, to: :"#{action_name}?", with: AuthorCreditDivergencePolicy
+  end
```
jmilljr24, line 89:
```suggestion
  def authorize_page
    authorize! :author_credit_divergence, with: AuthorCreditDivergencePolicy
  end
```

I might be wrong, but I think action policy can infer the action name even when using a `before_action` to call the policy.

jmilljr24 (approved):

maebeale:
@jmilljr24 since profiles aren't launched yet, i'm walking back some of the full removal of the per-item preferences. we still want things to come in w the request and now we have a page for admins to do the reconciling. and there's a new gh issue for removing the per-item preferences once profiles are launched.

app/services/workshop_search_service.rb
```diff
@@ -139,20 +139,27 @@ def filter_by_query
     )
     return if spaced.blank?
 
-    # Match against spaced variant (punctuation → space) and spaceless variant (punctuation + spaces removed)
-    fields = %w[workshops.title workshops.full_name action_text_rich_texts.plain_text_body]
+    # Match against spaced variant (punctuation → space) and spaceless variant (punctuation + spaces removed).
+    # The legacy `full_name` is excluded — matching it raw would surface a workshop
```
maebeale, line 143: 🤖 _From Claude:_ Dropping the SearchCop attribute alone would not have closed this — Workshop's free-text query never used SearchCop, building its own LIKE over a field list that included `full_name`. That is why the fix lands here rather than only in the model.

PR #2094, opened 2026-08-05
Add dues subscription and registration

jmilljr24:

This is a pretty big chunk but mostly standard crud and a little UI. If its the right directions I can wire up the allocations in this or another PR. I will have to comment out the job and `invite` generation of the initial subscriptions.

This adds a `Subscription` that a person is given on `invite`. Subscriptions are the parent of `Registrations` which run for a year. When a subscription is cancelled, no new registrations are added. A person can resume a subscription or create a new one if the old one is in the past (gap).

Nightly job, 30 days out creates new registrations if the subscription is not cancelled. Admin can CRUD subscriptions and registrations but generally shouldn't have to.

TODO:

- Add local and stripe payments.
- Decide on how registrations/subscriptions are searched/surfaced to admin
-


app/models/concerns/certifiable.rb
jmilljr24, line 1: I pulled this out of Registerable as that had a good bit shared with this PR and these were specific to certificates.

app/models/allocation.rb
jmilljr24, line 1: Refactored as this would have been the 3rd with similar validations

app/models/dues.rb
jmilljr24, line 1: Do we want these here or as env's

app/controllers/dues_subscriptions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+  def edit
+    authorize! @dues_subscription
+  end
+
+  def update
+    authorize! @dues_subscription
+
+    if @dues_subscription.update(rate_params)
+      redirect_to person_dues_subscriptions_path(@person),
+        notice: "Rate updated. It applies to future dues years.", status: :see_other
```
maebeale: will Rate always be the only thing updated? should we say `Dues autorenewal updated. Changes will only be applied to future years.` instead?

app/decorators/dues_registration_decorator.rb
```diff
@@ -0,0 +1,50 @@
+class DuesRegistrationDecorator < ApplicationDecorator
+  Badge = Struct.new(:label, :icon, :classes, keyword_init: true)
+
+  BADGE_CLASSES = {
```
maebeale, line 4: i thought we already have central themes so we don't need to define in decorators?

app/controllers/dues_subscriptions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      render :edit, status: :unprocessable_content
+    end
+  end
+
+  def cancel
+    authorize! @dues_subscription, to: :update?
+    @dues_subscription.update!(cancelled_at: Time.current)
+
+    redirect_to person_dues_subscriptions_path(@person),
+      notice: "Subscription cancelled. They keep the dues year they have already paid for.",
```
maebeale: will this always be admin-facing? how about something like: `Dues autorenewal is cancelled. Annual dues already paid will remain active until [date].`

app/controllers/dues_subscriptions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      status: :see_other
+  end
+
+  def resume
+    authorize! @dues_subscription, to: :update?
+    @dues_subscription.update!(cancelled_at: nil)
+
+    redirect_to person_dues_subscriptions_path(@person),
+      notice: "Subscription resumed. Dues years will renew again.", status: :see_other
```
maebeale: Idk that they'll know what a dues year is. `Dues autorenewal turned back on.`

app/controllers/dues_registrations_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    )
+  end
+
+  def create
+    authorize!
+    @dues_registration = @dues_subscription.dues_registrations.new(dues_registration_params)
+
+    if @dues_registration.save
+      redirect_to person_dues_subscriptions_path(@dues_subscription.person),
+        notice: "Dues year added.", status: :see_other
```
maebeale: How about `Annual dues log created.`?

app/controllers/dues_registrations_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  def edit
+    authorize! @dues_registration
+  end
+
+  def update
+    authorize! @dues_registration
+
+    if @dues_registration.update(dues_registration_params)
+      redirect_to person_dues_subscriptions_path(@dues_registration.person),
+        notice: "Dues year updated.", status: :see_other
```
maebeale: `Annual dues log updated.` ?

app/controllers/dues_subscriptions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    )
+  end
+
+  def create
+    authorize!
+    @dues_subscription = @person.dues_subscriptions.new(dues_subscription_params)
+
+    if @dues_subscription.save
+      redirect_to person_dues_subscriptions_path(@person),
+        notice: "Dues subscription created.", status: :see_other
```
maebeale: `Dues autorenewal turned on`.

app/jobs/renew_dues_terms_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private
+
+  def expiring_terms
+    DuesRegistration
+      .joins(:dues_subscription)
+      .preload(:dues_subscription)
+      .where(dues_subscriptions: { cancelled_at: nil })
+      .expiring_between(Date.current, Date.current + Dues::RENEWAL_WINDOW_DAYS)
```
maebeale, line 21: we may want to consider an env var so renewal window can be changed in the future without a code change

app/models/concerns/certifiable.rb
maebeale: nice!!!

app/models/concerns/registerable.rb
```diff
@@ -1,19 +1,11 @@
...
 
-  # Shared behaviour for the two registration records that money is allocated to
-  # and that grant a certificate: EventRegistration and
-  # ContinuingEducationRegistration. Keeps their payment + certificate interfaces
-  # identical so they can't drift.
-  #
   # Includers must:
   #   - have an `allocations` association (as: :allocatable)
   #   - respond to `cost_cents` (EventRegistration delegates to event.cost_cents;
   #     ContinuingEducationRegistration has its own column)
```
maebeale, line 7: should we add DuesRegistration field note here too?

app/models/allocation.rb
```diff
@@ -10,9 +10,26 @@ class Allocation < ApplicationRecord
...
+      no_cost: "Cannot allocate to a dues year with no cost.",
+      fully_paid: "Dues year is already fully paid.",
+      too_much: "Cannot allocate more than the remaining dues."
+    }
+  }.freeze
+
   validate :reverted_requires_positive_amount, :negative_cannot_be_reverted
-  validate :validate_event_registration_cost, if: -> { allocatable_type == "EventRegistration" }
-  validate :validate_ce_registration_cost, if: -> { allocatable_type == "ContinuingEducationRegistration" }
+  validate :within_allocatable_cost, if: -> { allocatable.is_a?(Registerable) }
```
maebeale, line 32: way better here to consolidate validation. love this.

app/models/dues.rb
maebeale: i think we have env vars, AND keep these as defaults? then if no env var or env var editing messes up, we still have something?

app/models/dues_registration.rb
```diff
@@ -0,0 +1,113 @@
...
+  def self.cost_covered_by_allocations
+    arel_table[:cost_cents].lteq(allocated_cents)
+  end
+
+  def self.allocated_cents
+    allocations = Allocation.arel_table
+    allocated = allocations.project(allocations[:amount].sum)
+      .where(allocations[:allocatable_type].eq(polymorphic_name))
+      .where(allocations[:allocatable_id].eq(arel_table[:id]))
+    Arel::Nodes::NamedFunction.new("COALESCE", [ allocated, Arel.sql("0") ])
```
maebeale, line 46: nice on arel usage

app/models/dues_registration.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  def end_date_not_before_start_date
+    return if start_date.blank? || end_date.blank?
+    return if end_date >= start_date
+
+    errors.add(:end_date, "can't be before the start date")
+  end
+
+  # Spans every subscription the person has, not just this one: a cancelled
```
maebeale: "every dues subscription the person has"

app/models/dues_registration.rb
```diff
@@ -0,0 +1,113 @@
...
+  # produce two subscriptions covering the same day.
+  def no_overlapping_term_for_person
+    return if start_date.blank? || end_date.blank? || dues_subscription&.person.blank?
+
+    overlapping = dues_subscription.person.dues_registrations
+      .where(start_date: ..end_date, end_date: start_date..)
+    overlapping = overlapping.where.not(id: id) if persisted?
+    return unless overlapping.exists?
+
+    errors.add(:base, "This person already has a dues year overlapping #{start_date} to #{end_date}")
```
maebeale, line 103: we need to move names of dues subscription and dues registration to en.yml so we can change them all at once. i added comments but even those i'm rethinking. the language here is def hard to get right.

app/models/dues_subscription.rb
```diff
@@ -0,0 +1,38 @@
+class DuesSubscription < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :person
+  has_many :dues_registrations, -> { order(start_date: :desc) }, dependent: :destroy
+
+  accepts_nested_attributes_for :dues_registrations
+
+  # When nil, charge standard rate. A valude is an override
```
maebeale: value

app/models/dues_subscription.rb
```diff
@@ -0,0 +1,38 @@
...
+  validates :rate_cents, numericality: { greater_than_or_equal_to: 0 }, allow_nil: true
+
+  validate :one_uncancelled_subscription_per_person
```
maebeale, line 12: why not `one_active_subscription_per_person` ?

app/views/dues_subscriptions/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  </div>
+
+  <h1 class="text-2xl font-bold mb-1">Dues rate</h1>
+  <p class="text-sm text-gray-500 mb-6"><%= @person.full_name %></p>
+
+  <%= simple_form_for @dues_subscription do |f| %>
+    <%= render "shared/errors", resource: f.object %>
+
+    <%= f.input :rate_dollars,
+          label: "Rate ($ per year)",
```
maebeale: rate usually means percentage in fintech, so we should avoid it for meaning amount. how about `cost_dollars` or `amount_dollars` or `price_dollars` or `fee_dollars`?

app/views/dues_subscriptions/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+          required: false,
+          hint: "Leave blank for the standard rate (#{dollars_from_cents(Dues::ANNUAL_COST_CENTS)}). Applies to future dues years only — years already created keep the price they were billed at.",
+          input_html: { min: 0, step: 0.01, placeholder: "Standard", class: "bg-white" } %>
+
+    <div class="flex items-center gap-3 border-t border-gray-200 pt-6 mt-6">
+      <div class="ml-auto flex items-center gap-3">
+        <%= link_to "Cancel", person_dues_subscriptions_path(@person), class: "btn btn-secondary-outline" %>
+        <%= f.button :submit, "Save rate", class: "btn btn-primary" %>
```
maebeale: same comment re the word "rate". i think we should use "amount" in code and "amount" or "fee" in UI.

app/controllers/dues_registrations_controller.rb
```diff
@@ -0,0 +1,68 @@
+class DuesRegistrationsController < ApplicationController
+  before_action :set_dues_subscription, only: [ :new, :create ]
+  before_action :set_dues_registration, only: [ :edit, :update ]
+
+  def index
+    authorize!
+    @dues_registrations = DuesRegistration
+      .includes(dues_subscription: :person)
```
maebeale: i think this should be: `.includes(:allocations, dues_subscription: :person)` so allocations don't n+1?

app/models/dues.rb
jmilljr24: Makes sense to me.

One more thought is that eventually we might need an model to hold admin editable app configurations. This is the first example I know of but I wouldn't add it just for this. Do you know of anything else currently that would benefit from being app wide that admin should be able to configure?

maebeale (approved):
code looks great! :shipit: 

for followups:
1. we should think through the language a little more re "rate" and "dues year" and "subscription"
2. we already have badge theming in DomainTheme. i'm going to do a pr to make sure we're using that consistently and then we can incorporate that into this code
3. i've got some ux and navigation thoughts for going from Person to subscriptions, from Admin home to list of active dues to managing subscriptions etc.

jmilljr24:
> code looks great! :shipit:
> 
> for followups:
> 
>     1. we should think through the language a little more re "rate" and "dues year" and "subscription"
> 
>     2. we already have badge theming in DomainTheme. i'm going to do a pr to make sure we're using that consistently and then we can incorporate that into this code
> 
>     3. i've got some ux and navigation thoughts for going from Person to subscriptions, from Admin home to list of active dues to managing subscriptions etc.

Good stuff! 
1. yes
2. I couldn't finding anything the was usable directly in this. Here is ai recap. CE was the only pattern I followed. 
```
dues_registration_decorator.rb:4 — BADGE_CLASSES (green/blue/amber/red/gray, -50/-700/-200)
event_registration_decorator.rb:8 — CE_BADGE_CLASSES (green/blue/amber, same intensities)
scholarship_decorator.rb:36 — program_status_classes (blue/green/purple/gray, -50/-700/-200)
organization_decorator.rb:23 — .program_status_classes, built from DomainTheme keys (used by 4 views)
topic_subscription_decorator.rb:9 — status_badge, inline green/slate classes
person_decorator.rb:115 / organization_decorator.rb:132 — badge(label, key) via DomainTheme.bg/text/border_class_for
```
3. Yes, some thoughts on the UX would be good. I barely touched it yet. 

Regarding notices etc. I just went generic for now. There will probably need to be tweaked anyways depending on how the ux ends up and how stripe etc integrate.

app/controllers/dues_subscriptions_controller.rb
jmilljr24: For now yes. I don't have a need yet for editing dates on the subscription. 

 I pulled out cancel/resume intentional to their own actions but don't necessarily have to. Not in the controller per say but in the UI I think i prefer not to have cancel/resume be in the edit page. It acts more like a toggle and also prevents admin going to the edit page. I'm guessing it will be extremely rare for them to have to go in and change someones rate (and understand the consequences) vs on occasion have to cancel someones subscription.

app/decorators/dues_registration_decorator.rb
jmilljr24: Not that I could find. CE is where I replicated but it wasn't worth extracting for two instances in this PR

```
dues_registration_decorator.rb:4 — BADGE_CLASSES (green/blue/amber/red/gray, -50/-700/-200)
event_registration_decorator.rb:8 — CE_BADGE_CLASSES (green/blue/amber, same intensities)
scholarship_decorator.rb:36 — program_status_classes (blue/green/purple/gray, -50/-700/-200)
organization_decorator.rb:23 — .program_status_classes, built from DomainTheme keys (used by 4 views)
topic_subscription_decorator.rb:9 — status_badge, inline green/slate classes
person_decorator.rb:115 / organization_decorator.rb:132 — badge(label, key) via DomainTheme.bg/text/border_class_for
```

app/controllers/dues_subscriptions_controller.rb
jmilljr24: It is for now. I have to see how the flow with work for regular users once stripe in integrated.

app/models/concerns/registerable.rb
jmilljr24: I would rather not have to list every model that includes this concern as that is more to maintain. This already lists two examples.

app/models/dues_subscription.rb
jmilljr24: The model doesn't define active. It tracks `cancelled_at`. So the record is either cancelled or uncancelled.

Active is used for display purposes but that mixes some things. From the persons point of view their subscription could be cancelled but still "active" until their end day but that's a combination of subscription and registration.

This keeps the validation tightly coupled to the model's column name.

maebeale: ohhhh. yup, totally. this is way better.

app/views/dues_subscriptions/edit.html.erb
jmilljr24: You are right. I was initially thinking to differentiate between what a subscription cost represented vs a registration. Like the subscription cents is the value that will be used to populate the registration .  The registration cents is whats actually own.

I think its ok if they are the same. `cost_cents` is used on things that require money so I will go with that. `amount_cents` is used on things that have value (payments, scholarships, discounts)

app/controllers/dues_subscriptions_controller.rb
maebeale: we could do a dropdown chip similar to on events/1/registrants ?
Screenshot 2026-08-05 at 8 51 52 AM

app/models/dues_registration.rb
jmilljr24: Lets come up with a plan once I have the stripe flows worked out. There should be some consistency between what stripe calls things, what a users sees, and what admin see. We will be using the stripe checkout so some text we have no control over.

app/controllers/dues_subscriptions_controller.rb
jmilljr24: I don't understand? Rate is the only thing currently that will be via the update action/edit page. Wherever it ends up, the cancel/resume is more of a toggle but I think a conditionally displayed button will be better. It follows stripe which some users will be presented with. 

What other options where you thinking in a dropdown?

PR #2104, opened 2026-08-07
Remove avatar section from reset_password_fyi mailer

maebeale:
🤖 suggested review level: 3 Read 📖 removes one section from a mailer's two templates + a spec

- Honeybadger: `reset_password_fyi` raised `undefined method 'file' for an instance of ActiveStorage::Attached::One` whenever a person with an attached avatar triggered a password-reset FYI email.
- The avatar was incidental to a password-reset FYI notification and was the sole source of the crash, so it's removed from the mailer rather than gated/fixed.

- Dropped the avatar block from both the HTML and text templates.
- Updated the spec to attach an avatar and assert the mail still renders without an avatar section (the old code crashed here).

### DIFF
--- a/app/views/notification_mailer/reset_password_fyi.html.erb
+++ b/app/views/notification_mailer/reset_password_fyi.html.erb
@@ -24,19 +24,6 @@
     (<%= mail_to @user.email %>)
     requested a password reset.
   </p>
-
-  <% if @person&.avatar&.attached? %>
-    <h2 style="margin-bottom: 8px; text-align: left;">
-      Avatar
-    </h2>
-
-    <p style="margin: 0;">
-      <%= image_tag @person.avatar,
-                    width: 96,
-                    height: 96,
-                    style: "border-radius: 50%; display: block;" %>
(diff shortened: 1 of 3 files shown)


app/views/notification_mailer/reset_password_fyi.html.erb
```diff
@@ -25,13 +25,13 @@
...
 
-  <% if @person&.avatar&.file&.attached? %>
+  <% if @person&.avatar&.attached? %>
     <h2 style="margin-bottom: 8px; text-align: left;">
       Avatar
     </h2>
 
     <p style="margin: 0;">
-      <%= image_tag @person.avatar.file,
+      <%= image_tag @person.avatar,
```
jmilljr24:
```suggestion
```
Maybe we should just remove adding the avatar in the email all together. Not sure how much value we gain from that at the risk of other issues.

maebeale: i had that thought too

PR #2117, opened 2026-08-09
Don't 500 the homepage when a DB blip breaks the banner query

maebeale:
🤖 suggested review level: 3 Read 📖 small contained rescue on a decorative helper, one new spec

- A dropped DB connection (`Trilogy::EOFError: TRILOGY_CLOSED_CONNECTION`, wrapped as `ActiveRecord::ConnectionFailed`) during the banner query 500'd the entire page — including the public homepage that uptime probes hit. Honeybadger flagged it on `home#index`.
- The banner is purely decorative; a transient connection blip shouldn't take down the page.

- `ApplicationHelper#display_banner` now rescues `ActiveRecord::ConnectionNotEstablished`/`ConnectionFailed` and returns `nil` (renders nothing) instead of raising.
- Scoped to those two transient connection errors so real query bugs still surface.
- Added `#display_banner` specs (red→green): renders published content, and returns `nil` without raising on a connection drop.

### Anything else to add?
- This only stops the banner from crashing the page. The one-off `TRILOGY_CLOSED_CONNECTION` itself (1 occurrence) points at DB pool connection reaping/keepalive — out of scope here; worth watching if it recurs.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -436,6 +436,9 @@ module ApplicationHelper
     content_tag(:div, id: "banner-news", class: "bg-yellow-200 text-black text-center px-4 py-2") do
       content_tag(:div, safe_content.html_safe, class: "font-medium")
     end
+  rescue ActiveRecord::ConnectionNotEstablished, ActiveRecord::ConnectionFailed
+    # A transient DB connection drop shouldn't 500 the whole page over a decorative banner
+    nil
   end
 
   def sortable_field_display_name(name)
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -588,11 +588,13 @@ RSpec.describe ApplicationHelper, type: :helper do
       expect(helper.display_banner).to include("Hello world")
     end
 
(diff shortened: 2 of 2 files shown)


maebeale:
🤖 _From Claude:_ Follow-up for the underlying connection drop (this PR only stops the banner from 500ing): #2118

app/helpers/application_helper.rb
```diff
@@ -436,6 +436,9 @@ def display_banner
     content_tag(:div, id: "banner-news", class: "bg-yellow-200 text-black text-center px-4 py-2") do
       content_tag(:div, safe_content.html_safe, class: "font-medium")
     end
+  rescue ActiveRecord::ConnectionNotEstablished, ActiveRecord::ConnectionFailed
```
jmilljr24, line 439: Isn't this just kicking the can down the road? This is just the first connection to the db on page load.

jmilljr24: I just saw #2118 

I personally wouldn't add this change. Its not fixing the issue and just adds more code. The issue seems minor. I would track and see how often it occurs before throwing code at it.

PR #2120, opened 2026-08-09
Load MySQL time-zone tables in bin/setup

maebeale:
🤖 suggested review level: 3 Read 📖 small, isolated, non-fatal addition to the dev setup script; no app/runtime code touched

- Fresh dev machines with a stock local MySQL (e.g. Homebrew `mysql@8.0`) ship the `mysql.time_zone_*` tables empty.
- The admin activity charts convert timestamps to a named zone (`America/Los_Angeles`) via `CONVERT_TZ`/Groupdate, which returns NULL and raises `Groupdate::Error - Database missing time zone support` when those tables aren't loaded — taking down the whole `spec/requests/admin/ahoy_activities_spec.rb` suite locally.
- CI's MySQL already has them loaded, so this only ever bit local setups.

- Added an idempotent, **non-fatal** `load_mysql_time_zones` step to `bin/setup` that pipes `mysql_tzinfo_to_sql` into the server-global `mysql` schema after the test DB is prepared.
- Skips quietly if the tables are already populated or the tz tooling / zoneinfo isn't available, so it never breaks setup.

### Anything else to add?
- One-time per machine — the tables live in the server-global `mysql` schema, so they persist across `db:drop`/`db:reset` and cover all workspaces on that machine.
- No app/runtime code changed.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -40,6 +40,32 @@ def db_client
   end
 end
 
+# Load MySQL's named time-zone tables into the server-global `mysql` schema.
+# Analytics queries convert to a named zone (e.g. America/Los_Angeles) via
+# CONVERT_TZ; a stock server (Homebrew mysql@8.0) ships these tables empty, so
+# those queries raise Groupdate::Error until they're populated. Idempotent and
+# non-fatal — skip quietly if already loaded or the tooling isn't available.
+def load_mysql_time_zones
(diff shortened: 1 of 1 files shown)


maebeale:
@jmilljr24 would this mess w your setup at all? i hit a one-time issue bc i'm on a new machine so wanted to prevent same issue for future us/others

jmilljr24 (approved):
Looks harmless and only run during setup.

PR #2136, opened 2026-08-09
Add a file-upload answer type to the form engine

### LINKED ISSUE #2109: Form engine: add a file-upload answer type (photo uploads in forms)
🤖 _From Claude:_ Filed while spiking the post-event survey feature.

## Context
The AWBW post-event/post-training evaluation (currently an external form) includes a **"upload a photo of one of your creations"** question. When bringing that survey in-house as a `Form`, there is **no way to represent a file upload**: `FormField#answer_type` only supports one-line text, paragraph, single-select radio, single-select dropdown, multi-select checkbox, section header, and informational-only (`app/models/form_field.rb:114-122`). `FormAnswer` stores a text `submitted_answer` only — no attachment.

For the spike the photo question is **dropped** (see the post-event survey PR). This issue tracks adding real support.

## What's needed
- A new `answer_type` (e.g. `file_upload`) on `FormField`.
- Active Storage-backed attachment on `FormAnswer` (or a dedicated answer-attachment model), with content-type/size validation.
- Rendering in the public form (`events/public_registrations/_form_field` and the new survey callout page), the form builder editor, and read-only submission views.
- Handling in the submission service, `FormAnswerValidator`, and submission-notification emails.
- Consider image-sharing consent (the source form pairs the photo with a "how may we share this" consent) — see the post-event survey's content-sharing consent field.

## Not doing now
Just tracking. No work required until the form engine needs uploads.

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new answer type wired through the model enum, submission service, validator, and 6 answer-display sites

Closes #2109

- The form engine had no way to represent a file upload, so the post-event survey's "upload a photo of one of your creations" question had to be dropped.
- Adds a `file_upload` `answer_type` so form authors can ask for a file (photo, PDF, doc).

- **Reuse, no migration.** A file-upload answer stores its blob on the existing polymorphic `Asset` (`FormAnswer has_one :asset, as: :owner`) — the same attachment, content-type validation (`Asset::ACCEPTED_CONTENT_TYPES`), and image/PDF display machinery story ideas use. No schema change.
- **Direct upload.** The public form renders a `file` input wired to the existing `file_preview` Stimulus controller; JS streams the file to storage and submits a signed blob id as a normal param — so the form stays non-multipart and the existing params/validator flow is unchanged.
- **Persistence.** The three near-identical answer-save loops in `PublicRegistration` were folded into one `persist_answer`, which attaches the blob for file fields and stores the filename in `submitted_answer` (so text-only views/exports still read).
- **Display.** All six answer-display sites now go through one `shared/_form_answer_value` partial: file answers show an inline image thumbnail (or a link for non-images) plus a download link; everything else is unchanged.
- **Validation.** `FormAnswerValidator` treats a required file question as present/blank; `Asset` enforces the content type on attach.

### Anything else to add?
- **Consent pairing deferred** — the source survey pairs the photo with a "how may we share this" consent; that's a normal select question an author can add alongside, tracked separately.
- **No-JS fallback:** without JS the file isn't uploaded (the form is non-multipart by design); the whole form is Turbo/Stimulus-driven already.
- On a validation error elsewhere, the file input clears (browsers never re-populate file inputs) and must be reselected.

### DIFF
--- a/app/models/form_answer.rb
+++ b/app/models/form_answer.rb
@@ -11,7 +11,20 @@ class FormAnswer < ApplicationRecord
   belongs_to :form_field, optional: true
   belongs_to :form_submission
 
+  # A file-upload answer stores its file on a polymorphic Asset (the same
+  # attachment/validation/display machinery story ideas use). Text answers leave
+  # this nil; submitted_answer still holds the filename so every text-only
+  # display and export shows something readable.
+  has_one :asset, as: :owner, dependent: :destroy
+
   def name
     "#{question_name_when_answered.presence || form_field&.name}: #{submitted_answer}"
+  end
+
+  # The attached upload, when this answer is a file-upload answer with a file on
+  # file. nil for text answers or an unanswered file question.
+  def uploaded_file
+    file = asset&.file
+    file if file&.attached?
   end
 end
--- a/app/models/form_field.rb
+++ b/app/models/form_field.rb
@@ -118,7 +118,8 @@ class FormField < ApplicationRecord
     :no_user_input,
     :multi_select_checkbox,
     :group_header,
-    :single_select_dropdown
+    :single_select_dropdown,
+    :file_upload
   ]
 
   enum :input_type, [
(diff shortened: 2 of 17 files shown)


maebeale:
@jmilljr24 i'm building up to supporting the post-event survey and the daily surveys via callouts. they ask for people to upload a photo of their work, so that's why i'm angling to change our form answers to support uploads.

jmilljr24 (approved):
Cool

This should probably get some extra user testing in staging.

PR #2140, opened 2026-08-09
Auto-fill org type/website/address when linking from a registration, and keep a record of it

maebeale:
🤖 suggested review level: 5 Inspect 🔬 four POROs, shared controller helpers, a new column, one XSS fix, and address-matching semantics that decide when two saved addresses are the same place

## Why
When an admin links an org to a registration, the submitted org profile answers (type, website, address) weren't applied — and there was no signal about what happened or whether the form's answers matched the org. Admins couldn't tell that a form said "School" while the org is saved as "501c3", or that the submitted address differs from the one on file.

## What
Both link paths — **Create and link** *and* **Link existing** (`{create,select}_organization`):

- **Fill blanks only** for `agency_type`, `website_url`, and the work address (street/city/state/ZIP/country) from the submission — never clobbering curated values.
- **Anchor the affiliation to the org's sole address** when the submission carried no address.
- **Flash notice** naming only what was really written: `"Acme linked. Saved from the form: type, website, and work address in Austin."`, or `"ZIP and country on the Austin work address"` when an existing one was filled, or nothing at all when the submission changed nothing.
- **Flash warning** when a submitted answer differs from a value already on the org (kept-vs-form values named), so the admin can reconcile.
- **Two persistent per-org notes** on the linking page, since the flash is gone by the next page load: what a form submission *filled* on an org that already existed (new `event_registration_organizations.form_filled_fields`, written by the public flow too — that's where most fills happen), and the type/website/address answers that *differ* from the org's saved profile.

### Semantics
- **Whose answers apply to which org** (`submission_entry_for`): the submission whose typed org name matches the org, else — only when there's exactly one submission and one linked org — that sole entry, which covers a registrant who named no org and an admin resolving a typo'd "Acme Inc" to the saved "Acme Corporation". An extra org an admin links by hand matches nothing, so another org's answers are never written onto it or reported against it.
- Type/website/address are **fill-blanks** in the admin flow (public registration keeps its latest-wins overwrite via the new `overwrite:` flags on `SyncProfile`/`UpsertAddress`).
- An org **created** from the submission records no "filled" note — everything on it came from the form, so there's no change to flag.
- **Address matching** (`AddressMatcher`) tries same city + state, then same city + street, then — across cities — **same street + same ZIP**. The last one reunites an address whose city was respelled between submissions ("St. Louis" / "Saint Louis") without merging two real offices that share a street name in different towns. City and state on file are never flipped: the saved value stays and the difference is flagged for an admin.
- An address the matcher doesn't recognize isn't a discrepancy — it's added as an additional work address.
- Address fields are **trimmed on write and compared trimmed**, so a padded answer (browsers don't trim text inputs) isn't stored verbatim and then read back as a discrepancy against itself.

### Security
Flash messages render with `html_safe`, so registrant-typed form answers folded into the linking warning reached admins as raw HTML. All dynamic values in the notice and warning are now escaped.

### New / changed
- `event_registration_organizations.form_filled_fields` (json) + `EventRegistrationOrganization#form_filled_labels` / `#record_form_fills` — persistent provenance for what a submission wrote onto an org.
- `OrganizationServices::SyncProfile` — fill-blanks type/website sync; returns filled labels. Extracted from `PublicRegistration`.
- `OrganizationServices::ProfileDiff` — read-only comparison → `Discrepancy` list (type + website + address).
- `OrganizationServices::AddressMatcher` — shared street/ZIP-aware address matcher.
- `OrganizationServices::UpsertAddress` — gains `overwrite:`, matcher-based lookup, and a `Result` (`address`, `created`, `filled`, `saved_label`) so callers report what was actually saved.

The linking actions memoize the form-submission entries they read repeatedly: 66 → 41 queries per `select_organization`.

## Tests
- Service specs for `SyncProfile`, `ProfileDiff` (incl. address + city), `AddressMatcher` (cross-city ZIP fallback, and that a differing ZIP does *not* merge), and `UpsertAddress` (`overwrite: false`, dedup, what `filled` reports, skipped street/ZIP/state).
- Request specs: linking survives an address with no ZIP; both link paths fill blanks / preserve curated values / emit notice & warning; the fills are recorded and shown on the page afterwards; an org created from the submission records none; an extra linked org is neither seeded nor reported against; sole-address anchoring; a script payload in a form answer never renders unescaped; the notice doesn't claim an address save that didn't happen.
- `PublicRegistration` spec: a registrant's answers are recorded on the registration's org link.
- `EventRegistrationOrganization` model spec: `form_filled_labels` / `record_form_fills` — the merge-and-de-dup contract, and that a resubmission of what's already recorded issues no write.

## Anything else to add?
`addresses.street_address` and `zip_code` are `NOT NULL` with no default, and `city`/`state` also have presence validations — but the registration form's address answers are individually optional. A registrant who named an org and gave a city but skipped the ZIP took down the whole public registration with a `NotNullViolation` (and the admin linking action with it). This predated the branch; fixed here since the flow is being reworked anyway:

- Skipped street/ZIP are stored blank.
- State can't be papered over the same way, so a submission without one saves no *new* address rather than failing the registration. A matching address is still updated, the answers stay on the form submission, and the linking page still shows them for an admin to enter.


app/controllers/event_registrations_controller.rb
```diff
@@ -464,12 +485,77 @@ def submitted_agency_address(registration)
...
+    if conflicts.any?
+      descriptions = conflicts.map { |conflict| "#{conflict.label} (form: “#{flash_safe(conflict.submitted)}”, saved: “#{flash_safe(conflict.saved)}”)" }
+      flash[:warning] = "Some form answers differ from #{flash_safe(organization.name)}’s saved profile and were not applied: #{descriptions.to_sentence}. Edit the organization if they should change."
+    end
+
+    saved = filled + [ address_saved_label(address_result) ].compact
+    notice = "#{flash_safe(organization.name)} #{verb}."
+    notice += " Saved from the form: #{saved.to_sentence}." if saved.any?
+    notice
+  end
```
maebeale, line 565: 🤖 _From Claude:_ Flash messages render with `html_safe`, so anything registrant-typed that lands in one has to be escaped here — a form answer containing a script tag would otherwise run in the admin browser.

app/services/organization_services/address_matcher.rb
```diff
@@ -0,0 +1,43 @@
...
+      street = street_address&.strip
+      return if street.blank?
+
+      by_street = in_city.find { |address| address.street_address.to_s.strip.casecmp?(street) }
+      return by_street if by_street
+
+      zip = zip_code&.strip
+      return if zip.blank?
+      addresses.find do |address|
+        address.street_address.to_s.strip.casecmp?(street) && address.zip_code.to_s.strip.casecmp?(zip)
```
maebeale, line 42: 🤖 _From Claude:_ This last-resort match crosses city boundaries, so it needs the ZIP as corroboration — street alone would merge two real offices on "1 Main St" in different towns. Requiring both means it only fires for a genuinely respelled city.

app/services/organization_services/upsert_address.rb
```diff
@@ -1,61 +1,112 @@
...
+    # between the form and the org's saved address is kept and surfaced by
+    # ProfileDiff instead of being silently replaced. City and state are always
+    # fill-only (never flipped) so a street/ZIP match can supply what we were
+    # missing without rewriting what's already on file.
+    def field_updates(existing)
+      updates = {}
+      updates[:city] = @city if existing.city.blank?
+      updates[:state] = @state if @state.present? && existing.state.blank?
+      updates[:street_address] = @street_address if @street_address.present? && (@overwrite || existing.street_address.blank?)
+      updates[:zip_code] = @zip_code if @zip_code.present? && (@overwrite || existing.zip_code.blank?)
```
maebeale, line 91: 🤖 _From Claude:_ State is the one address field that can't be defaulted — Address validates its presence — so a submission without one saves no new address instead of failing the whole registration. A matching address is still updated, and the raw answers stay on the form submission for an admin to enter by hand.

maebeale:
@jmilljr24 this was on the spreadsheet: "Can the organization type and website url be automatically populated when we create and link the org?"

app/models/event_registration_organization.rb
```diff
@@ -3,4 +3,20 @@ class EventRegistrationOrganization < ApplicationRecord
...
 
   validates :organization_id, uniqueness: { scope: :event_registration_id }
+
+  # Labels of the org values written from this registration's form answers
+  # ("website", "type", "ZIP on the Austin work address"). Recorded when the org
+  # is linked so the linking page can persistently say what the submission
+  # changed on an org that already existed — the flash notice saying so is gone
+  # by the next page load.
+  def form_filled_labels
+    Array(form_filled_fields)
```
jmilljr24, line 13: The column name `form_filled_fields` was a bit confusing on first read (I assumed it was saving the filled out answers). Disregard this comment if I'm wrong but it appears the column is storing the labels and this method just ensures an empty array is returned if the column is null.

Could the column be named `form_filled_labels` to be more clear its purpose? Then this method can still be called to return the empty array vs nil.


```suggestion
  def form_filled_labels
    Array(super)
```

jmilljr24 (approved):
Just one thought

app/models/event_registration_organization.rb
maebeale: 🤖 _From Claude:_ You read it right, and the rename is in (`7663a6cb9`) — the column stored the names of what changed ("website", "ZIP on the Austin work address"), never the submitted answers, which stay on the form submission. It is now `form_filled_labels`, and the method survives as an override of the generated reader (`Array(super)`), so a link the form never filled still answers `[]` while the column itself stays NULL.

One thing worth a look: I did it as a separate `rename_column` migration rather than editing the original `add_column`, since that one has already run in dev DBs on this branch. If you would rather this ship as a single clean migration, say the word and I will fold it into the original.

maebeale: 🤖 _From Claude:_ Correcting my earlier reply — the name landed further along than what I said. It is now `form_autofill_changes`, not `form_filled_labels`: "label" collides with a form question’\s display text, which is the same ambiguity you flagged, so it reads `changes` instead. The payload also changed shape while we were in there — each entry is now `{ field, label, value, change_type: "new"|"update", previous_value, scope }` rather than a bare string, so the linking page can show what a submission set a field *to*, and what it displaced. The reader (`form_autofill_changes`) still does the nil-to-`[]` job your comment was about. All folded into the single `add_form_autofill_changes_…` migration, since none of it has shipped.

PR #2142, opened 2026-08-09
Keep roster column toggles selected across search filters

maebeale:
🤖 suggested review level: 3 Read 📖 small client-side Stimulus change to persist toggle state

## What
The registrants roster column toggles (Linked org, CE, Scholarship, Payment method, Date registered, Attendance, User account, Certificate) reset to their server defaults whenever a search filter is applied.

## Why
The toggles and the table share one Turbo frame (`registrants_results`). Applying a filter re-renders that frame, and `column_toggle_controller` kept no persisted state — the freshly-connected switches came back at their hard-coded ERB defaults, wiping the admin's choices.

## Fix
- Persist each toggle's on/off state in `sessionStorage` and reapply it in `connect()`, so selections survive the frame re-render (and other Turbo visits within the tab session). Falls back gracefully when storage is unavailable.
- Namespace the stored state per event via `data-column-toggle-scope` on the toggle root, so each event's roster keeps its own selections and generic group names can't collide if the controller is reused elsewhere.

## Notes
- State lives in browser `sessionStorage` (per-tab, client-side — not the Rails session). It survives in-tab navigation/reload and resets when the tab is closed.

### DIFF
--- a/app/frontend/javascript/controllers/column_toggle_controller.js
+++ b/app/frontend/javascript/controllers/column_toggle_controller.js
@@ -6,13 +6,28 @@ import { Controller } from "@hotwired/stimulus"
 // table can host several independent column toggles (e.g. "User confirmation",
 // "CE status"). Columns live outside the switch's element, under a shared
 // [data-column-toggle-root] ancestor.
+//
+// The chosen state is remembered per group in sessionStorage and reapplied on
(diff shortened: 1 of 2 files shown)


maebeale:
@jmilljr24 sorry! this was the one i meant

jmilljr24 (approved):

PR #2143, opened 2026-08-09
Shared registrant filter bar: prefill bulk emails, collapsible More filters, new filters

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new SQL scopes on a hot roster query + shared partial across two filtering backends

- Unify the registrants roster and the bulk-emails (reminder) recipient picker behind **one shared filter bar**, so the two stay in sync by construction.
- Carry the roster's active filters into the picker via the "Send bulk emails (filtered)" link.
- Expand what both pages can filter by.

- **Shared partial** `events/_registrant_filters` (+ `_filter_text`, sibling of `_filter_select`): a "Filters" heading, one visible row (Progress · Attendance · Payment status) plus a text-style **More filters** toggle (`filter_toggle` Stimulus controller) and Clear; the rest collapse, auto-opening when a collapsed filter is active. Per-page differences via locals.
- **Progress** column + filter (was the ambiguous "Status"; param/service stay `readiness`).
- **New filters, wired into both backends** (roster SQL scopes in `EventsController#registrants`; picker `ReminderRecipientFilter`): Funder (scholarship→grant donor), City, Comment text (topic+body), Submission status (none/has/multiple form submissions), and scholarship **Agreed**.
- Collapsed dropdowns ordered by how a registration is processed.

### Anything else to add?
- New model scopes: `grantor`, `registrant_city`, `comment_text`, `submission_status`, scholarship `agreed` — covered by model specs; roster + picker covered by request specs.
- `filter_toggle` is a small toggle (drawer already opens server-side; JS only powers manual open/close). Alternatives discussed: URL-param and pure-CSS `:has()` — kept the controller for layout + `aria-expanded`.

---
**Rebased onto main (2026-08-09)** — reconciled with #2147 (grants `donor`→`funder` rename). The funder-name free-text search now uses param `funder_name` (distinct from the awbw/external funding-source `funder` scope) consistently on both the roster and the picker, so the shared filter bar can't collide the two on the roster.


maebeale:
@jmilljr24 this was on the spreadsheet. would love to ship this if you're ok w it.

app/frontend/javascript/controllers/filter_toggle_controller.js
jmilljr24, line 1: This seems eerily similar to the dropdown controller.

jmilljr24 (approved):
I can't seem to find what you are referring to in the spreadsheet so I guess I will defer to your judgement.

maebeale:
> I can't seem to find what you are referring to in the spreadsheet so I guess I will defer to your judgement.

umm that's bc i meant to put this comment on another pr! the one re sticky column show/hide sliders.

PR #2175, opened 2026-08-12
Automate Tailwind class ordering + tokenize sub-xs text sizes and social brand colors

maebeale:
🤖 suggested review level: 3 Read 📖 wide but low-risk — mechanical class reordering plus token migrations across ~75 view files, one dev script/config, and a vendored agent skill

Adopts practices from [Evil Martians — "5 best practices for preventing chaos in Tailwind CSS"](https://evilmartians.com/chronicles/5-best-practices-for-preventing-chaos-in-tailwind-css): **automated class ordering**, **semantic design tokens**, and the article's **companion agent skill**.

### What & why

- **Class ordering (rustywind).** The article's tool (`prettier-plugin-tailwindcss`) can't parse ERB, where ~all our classes live. `rustywind` is the ERB-native equivalent: reorders only class-attribute tokens, leaving `<%= … %>` and markup untouched. Added `ai/tw-sort` (defaults to files changed vs `main`; `--all` / `--check`) + `npm run format:classes` / `lint:classes`.
- **Text tokens — all arbitrary sizes eliminated.** Replaced every `text-[…px/rem]` with a named scale in even 2px steps below `text-xs` (12px): `text-2xs` (10px), `text-3xs` (8px), `text-4xs` (6px). Oddballs collapse to the nearest step (11px/0.7rem → 2xs; 5px/7px → 4xs), per Justin's call to accept the sub-pixel tradeoff. Compiled output is `font-size` only — no line-height change.
- **Social brand color tokens.** The same hexes were hand-typed across 4 view files (`#0A66C2`, `#1877F2`, `#1DA1F2`, `#FF0000`, and the 3-stop Instagram gradient). Now named once in `@theme`.
- **Redundant utilities.** Collapsed pairs the ordering pass surfaced (`mt-1 mb-1`, `pr-3 pl-3`, `mb-6 mt-6`) and dropped two no-ops in the resource card (`flex-row` is the default direction; `md:w-40` restates `w-40`). `py-4 pb-3` → `pt-4 pb-3` so padding no longer depends on stylesheet order to resolve.
- **Vendored the article's agent skill** at `.claude/skills/tailwind-best-practices/` (unmodified, MIT; attribution in skills `THIRD_PARTY_NOTICES.md`).

### Bug found and fixed mid-review

`TopicSubscriptionDecorator` used `text-[0.65rem]`, which **only ever compiled as a side effect** of the 13 view files that happened to use the same arbitrary value. Converting those to `text-2xs` silently killed it — no build error, the Unsubscribed badge icon just lost its size.

Root cause is that `@source` listed `views`, `javascript`, and `helpers` but not `decorators` — which CLAUDE.md actively steers presentation logic toward. Added the `@source` line so decorator classes can't silently drop out again, and documented the scanning limits in AGENTS.md. Verified by build probe: a class in `app/decorators/` compiles now and provably did not before.

### Notes for reviewer

- The vendored skill's practice #4 advises against `@apply`, which `components/buttons.css` uses (24 sites). Not changed here — flagging the tension for a later call.
- `shared/_badge` takes an open `classes:` string across 69 call sites, which is practice #5's anti-pattern (design-system component accepting arbitrary classes). A fixed variant map is the right fix but wants its own PR.
- No CI gate on class order yet: `ai/tw-sort --check` is clean on changed files, but `--all` still reports 484 unsorted legacy files, so a whole-tree gate can't land until convergence finishes.
- Verified: full suite green (6851 examples, 0 failures); `bin/vite build` green; all 3 text tokens and all 7 color tokens compile; `ai/tw-sort --check` clean; zero `text-[…]` sizes remain anywhere in `app/`.

---

### Scorecard — where the codebase stands vs the article's 5 practices

| # | Practice | Status | Notes |
|---|---|---|---|
| 1 | Use fewer utility classes | 🟡 Partial | `components/buttons.css` extracts `.btn*` and CLAUDE.md pushes presentation into decorators. **This PR** collapses the redundant pairs it touched; `shared/_badge`'s open `classes:` prop remains. |
| 2 | Group & name design tokens semantically | 🟢 Strong | `@theme` has semantic colors; **this PR** adds the sub-xs text scale + social brand colors and removes all arbitrary font sizes. |
| 3 | Keep class ordering | 🟢 Now automated | Was 🔴 (no sorter). **This PR** adds rustywind + `ai/tw-sort`. |
| 4 | Prevent inconsistencies when overriding/extending | 🟢 Good | Rails-adapted: extracted `.btn*` variants + CLAUDE.md rules against interpolated classes. |
| 5 | Minimize build size | 🟢 Automatic | Tailwind v4 minifies via Lightning CSS in the Vite prod build. |


maebeale:
@jmilljr24 this is in response to the tailwind article. any reservations w adding rustywind?

jmilljr24:
I don't see the `SKILL.md` from the article?

jmilljr24:
>text-[0.7rem], text-[11px], text-[7px], text-[5px] (11 total) left as-is — genuine one-offs / distinct sizes, out of the consolidated cluster.

I agree to not to make new ones for everything but can this oddballs be changes to either 1. fit the classes we have or 2. add one or two more class that are close enough so we can eliminate all the customs sizes. Like the difference between 11 and 10 px is so small, can we accept that tradeoff and just use 10?

jmilljr24 (approved):

maebeale:
🤖 _From Claude:_ Both addressed 👇

**SKILL.md** — added. Vendored the article's companion skill unmodified from [`evilmartians/agent-skills`](https://github.com/evilmartians/agent-skills) to `.claude/skills/tailwind-best-practices/SKILL.md` (MIT; attribution in skills `THIRD_PARTY_NOTICES.md`). Heads-up: its practice #4 pushes back on `@apply`, which our `components/buttons.css` uses — left as-is, flagged in the PR notes.

**Oddballs** — all eliminated, went with option 2. There is now a small named scale in even 2px steps below `text-xs` (12px): `text-2xs` 10px, `text-3xs` 8px, `text-4xs` 6px (added one new token, `4xs`). Mappings, taking the 11→10 tradeoff you called: `11px`/`0.7rem` → `text-2xs`, `5px`/`7px` → `text-4xs`, `0.65rem`/`0.625rem` → `text-2xs`. Zero `text-[…px/rem]` sizes remain anywhere in the tree.

jmilljr24 (approved):

PR #2178, opened 2026-08-12
Scholarship decline flow with agreement status and history

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new agreement state model + column replacement + history model that change financial totals across dashboards, grant budgets, and reports

### Why

- Recipients can decline a scholarship (with a reason) from their scholarship page, not just agree.
- A declined award drops out of every total, so budgets and reports stop counting money nobody took.
- Agreement state was two hand-synced fields that could contradict each other.

### State model

- `agreement_signed_at` → tri-state `agreement_response_status` (pending / accepted / declined).
- No backfill — there are no signed agreements in production.
- Predicates (`agreement_signed?` / `agreement_declined?`) unchanged for existing callers.
- New `ScholarshipAgreementResponse`: append-only row per transition (status, reason, responder, amount at the time). The responded-at date and decline reason are read from the latest row, not stored on the scholarship.
- Logged on create as well as update, so an award saved with the agreement toggle already on keeps its signed date.

### Decline

- Native `<details>` reason box beside **Agree**; no JS.
- Recording a decline zeroes the allocation, so allocation-based totals self-correct.
- Admin re-offer is an explicit button on the declined banner — editing the amount no longer reinstates an award.
- Excluded from grant budgets, the funding split, `EventDashboard`, `AttendeesBreakdowns`, `EventScholarshipFigures`, `EventRevenueFigures`, and the index groupings.

### What a decline means downstream

- Onboarding readiness reads **Award declined**, ranked above the payment gap the decline itself opened — answering it is the actual next step.
- Tasks stop gating on an award nobody took, so a decline can't hold the certificate open.

### On the recipient's page

- The balance box prices the choice while the agreement is pending: *"Accept and you'll owe $50 of the $100 registration cost." / "Decline and $100 is due."*
- Declining is priced off that award alone, so anything already paid stays paid.
- Decline confirmation, plus an admin-only agreement history, afterwards.

### Visible everywhere

- One `ScholarshipDecorator#agreement_status_badge` renders the pill on every surface: scholarships index, registration card, grant scholarships table, event recipient card, registrant roster, onboarding row, admin show/edit.
- Struck-through amount wherever a declined award sits in a money column.
- Agreement history timeline on the edit page, admin-only on the recipient-facing callout.
- Seeded into Features & tips (`config/features.yml`).

### Not here

- All automated email is a stacked follow-up PR.


maebeale:
@jmilljr24 this is in response to kat's mention that this back and forth for scholarship agreements is a lot of work and impacts a large portion of scholarship recipients, who are large portion of registrants

maebeale:
@jmilljr24 we could get out of the extra model for logging history if we send autoemails for decline, reoffer, accept. this one autoemail for decline fyi feels acceptable to send as it requires staff action. this is the thing kat was asking about (i commented that on an earlier email version we decided to punt on. this is that other per less the email/reminder setup.)

jmilljr24:
> @jmilljr24 we could get out of the extra model for logging history if we send autoemails for decline, reoffer, accept. this one autoemail for decline fyi feels acceptable to send as it requires staff action. this is the thing kat was asking about (i commented that on an earlier email version we decided to punt on. this is that other per less the email/reminder setup.)

Yes, an action based email makes sense.(User clicks decline -> email sent). My previous reservations were on time based emails.

jmilljr24 (approved):

app/models/event_registration.rb
```diff
@@ -623,9 +623,16 @@ def registration_subject_noun
     scholarship_requested? ? "event scholarship registration" : "event registration"
   end
 
+  # A declined award carries no tasks, so it can't hold the certificate or the
+  # readiness checklist open.
   def scholarship_tasks_met?
-    return true if scholarships.empty?
-    scholarships.all?(&:tasks_completed?)
+    live = scholarships.reject(&:agreement_declined?)
```
maebeale, line 628: 🤖 _From Claude:_ Worth knowing this reaches past the readiness checklist: `scholarship_tasks_met?` also gates `certificate_available?` and the CE certificate callout, so a registrant who declined and attended can now get their certificate instead of being held on tasks they never took on. That looked right, but it is a behaviour change in files this diff does not touch.

PR #2183, opened 2026-08-13
Rename event Manage→Registrations; one shared filter bar across roster, picker & attendees

maebeale:
🤖 suggested review level: 5 Inspect 🔬 one shared filter bar across roster / bulk-reminder picker / cross-event attendees, plus new EventRegistration scopes and cross-event "any registration" filtering

- "Registrations" reads clearer than "Manage" across the event subnav/links; the dashboard tab shows on the dashboard page.
- Reports **Breakdowns** links drill into the attendees index (charts open, scrolled to `#breakdowns`) instead of the roster; reports→attendees eyebrows return to the reports hub. "Full report" → "Details" on all three cards.
- **One shared filter bar** across the registrants roster, the bulk-reminder picker, and the cross-event attendees index.

- Extracted `events/_filter_bar` (Filters heading with the applied-chip row pinned right + primary row + optional More-filters drawer). All three pages render one shared field list, gated by visibility:
  - **Everywhere:** search, Attendance, Payment status/method, Scholarship, Funder, CE status, User account, Comment status/text, City, State, County, Sector, Program status, Affiliation status, Topic subscription.
  - **Cross-event only:** Event type, Event, Year.
  - **Single-event only:** Submission status, Organization linking, Progress (all event/form- or in-memory-computed).
- **Cross-event matching:** a registration-level filter matches a person if **any** of their attended registrations qualifies — reusing the `EventRegistration` scopes against `attendee_registrations`. New scopes: `registrant_topic_subscription`; `person_address_ids` gains `city`.
- **Negative values invert instead**, since "any registration matches" would let them through on a technicality: **No scholarship** and **No comments** ask the positive scope and exclude, so someone who was funded (or commented) on a *different* attended registration isn't counted. Scoped to one event both readings agree.
- **Address filters** (city/state/county) read the person's own address data directly, no registration required.
- Scholarship is one dropdown everywhere (recipient sub-statuses + new **No scholarship**, backed by `without_scholarship`).
- **CE status reads as certificate state:** Issued/Not issued → **Certificate issued** / **Certificate outstanding** (they overlapped — one person could match both), plus a new **No CE**. That absence case was previously the chip-only `ce` param, so the CE pie now drills into the dropdown and `ce` is retired; `registered` and `none` are exact complements because the license join is on a NOT NULL column.
- Every param the attendees index narrows on now has a visible control or an `AttendeesActiveFilters` chip — otherwise it shrinks the list silently and is dropped by the form's next auto-submit. A spec asserts a control exists for each `ATTENDEE_REGISTRATION_FILTERS` key. `return_to` rides the form so the results frame's own links keep the eyebrow's origin.
- Panel Show/Hide controls became text toggles with a rotating chevron; the section-heading anchor uses `reveal-section` so `#breakdowns` scrolls even inside a lazy Turbo frame.

### Anything else to add?
- Full suite green (6405 examples, 0 failures, 17 pending). Rebased on main (folds in #2179's topic-subscription filter).

- **Clear filters** resets Attendance and Event type to the page's default population, not to the widest option they lead with — they render pre-selected, so index 0 isn't neutral for them (`clear_to:` on `events/_filter_select`).
- City and state ask **one** address subquery, so they describe the same address the way the roster's shared join does.
- The four option lists both bars render (scholarship, CE, account, comment status) are now `EventRegistration::*_FILTER_OPTIONS` constants alongside the four that already existed, so the two bars can't drift.
- **The filter bar sits outside the results frame**, so two things it renders had to be kept honest across frame-only submits:
  - The applied-chip row is replaced out-of-band from the results frame (alongside the count) — its remove links are built from the request's params and otherwise still described the URL the page loaded with, dropping every filter set since.
  - `reveal-section` drops the `#breakdowns` hash once it has scrolled there; a frame submit never touches the address bar, so the hash kept matching and pulled the page back on every filter change.

### DIFF
--- a/app/frontend/javascript/controllers/collection_controller.js
+++ b/app/frontend/javascript/controllers/collection_controller.js
@@ -65,8 +65,12 @@ export default class extends Controller {
   clearAndSubmit(event) {
     event.preventDefault();
 
+    // Clear to empty rather than calling form.reset(): a page reached with its
+    // filters already in the URL renders them as value/selected attributes, which
+    // are exactly what reset() restores — so resetting puts back the filters we
+    // just cleared and "Clear filters" re-submits them.
     this.element
-      .querySelectorAll('input[type="text"], input[type="search"]')
+      .querySelectorAll('input[type="text"], input[type="search"], input[type="date"]')
       .forEach((input) => {
         input.value = "";
       });
@@ -81,11 +85,10 @@ export default class extends Controller {
         }
         input.checked = false;
       });
-    this.element.reset();
 
-    // TomSelect (remote-select) renders its own UI and ignores selectedIndex and
-    // form.reset(), so clear each enhanced control explicitly. Silent clear avoids
-    // a redundant submit before the one below.
+    // TomSelect (remote-select) renders its own UI and ignores selectedIndex, so
+    // clear each enhanced control explicitly. Silent clear avoids a redundant
+    // submit before the one below.
     this.element.querySelectorAll("select").forEach((select) => {
       if (select.tomselect) select.tomselect.clear(true);
     });
--- a/app/frontend/javascript/controllers/collection_controller.js
+++ b/app/frontend/javascript/controllers/collection_controller.js
@@ -74,8 +74,16 @@ export default class extends Controller {
       .forEach((input) => {
         input.value = "";
       });
+    // Index 0 is the neutral state only for a select with a placeholder first
+    // option. One rendered pre-selected (no include_blank) leads with a real value,
+    // so it says what it clears to via data-clear-to.
     this.element.querySelectorAll("select").forEach((select) => {
-      select.selectedIndex = 0;
+      const { clearTo } = select.dataset;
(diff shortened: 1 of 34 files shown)


app/frontend/javascript/controllers/collection_controller.js
```diff
@@ -65,8 +65,12 @@ export default class extends Controller {
   clearAndSubmit(event) {
     event.preventDefault();
 
+    // Clear to empty rather than calling form.reset(): a page reached with its
```
maebeale, line 68: 🤖 _From Claude:_ Widest blast radius in this PR from the smallest diff — `clearAndSubmit` backs the Clear filters button on 26 index pages, not just the event ones. Worth noting the fix removes `reset()` rather than reordering it: the checkbox branch below keys its label-class bookkeeping off `input.checked` at loop time, so resetting first would leave the classes wrong for a box the user checked whose default is unchecked. Full system suite passes (161 examples).

PR #2192, opened 2026-08-13
MAYBE: Drop benign SolidQueue shutdown connection errors from Honeybadger

maebeale:
🤖 suggested review level: 3 Read 📖 small before_notify filter with one guard condition, plus unit spec

### What & why
- A terminating worker pod raises `ActiveRecord::ConnectionNotEstablished` (`TRILOGY_CLOSED_CONNECTION`) in `at_exit` while SolidQueue deregisters its `Process` record after the DB pool has closed. Pages on every deploy/restart, 0 users affected, nothing lost.
- Adds a Honeybadger `before_notify` hook that halts these — but **only when raised outside any request** (`component` blank), so a genuine mid-request/job DB outage still reports.

### Notes
- Kept narrow on purpose: requires the closed-connection sentinel *and* no component. Infra-side fix (pod `terminationGracePeriod` / stop ordering) left out as riskier; suppression first.
- Spec exercises the real registered hook: drops the shutdown case, keeps in-request + "connection refused".

### DIFF
--- a/config/initializers/honeybadger.rb
+++ b/config/initializers/honeybadger.rb
@@ -1,0 +1,23 @@
+# Suppress benign shutdown noise. When a pod is terminated (deploy, restart,
+# scale-down) the DB connection pool closes while SolidQueue's supervisor is
+# still trying to deregister its Process record, so that final query fails with
+# a closed connection. There's no request, no user, and nothing is lost — it's
+# just restart noise, so we drop it rather than page on it.
+#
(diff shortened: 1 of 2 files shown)


jmilljr24 (commented):
Same comment as before on the other error. This is the first seeing it. I don't know what this code does but I know honeybadger is wrapped around basically everything so if this is incorrect it could take the whole site down.

I'll accept a tiny bit of noise on the hb notifications.

PR #2195, opened 2026-08-14
Reconcile facilitator affiliations against attendance (admin bulk action)

maebeale:
🤖 suggested review level: 5 Inspect 🔬 changes what reconciliation writes to affiliation history, which the grant figures read

Brings facilitator affiliations in line with who actually attended a training, as an admin bulk action with a preview-and-confirm step.

## What reconciliation does, by attendance status

| Status | Outcome |
|---|---|
| `attended` | Keep active |
| `no_show`, `cancelled`, `incomplete_attendance` | **Delete** the row this training minted; **end** any older row the day before |
| `transferred_out` | **Move** the open row to the destination event and re-point its provenance |
| `registered` | **Report only** — an unfilled roster isn't an outcome |

`transferred_in` isn't a status (it's an FK), so those reconcile on whichever status they hold.

## Why deleting, not deactivating

A row minted by this training's registration records an *assumption*: that the person would become a facilitator on the training date. If they didn't attend there's no period to preserve, and a zero-length row is a worse way of saying it never happened.

Anything **older** — hand-entered, or from an earlier training — is only ever end-dated. Deleting those would erase years of real history and move the organization's program status at every training in between. That distinction is the core of this PR.

## Why the dates matter

Program status (New / Ongoing / Reinstated) is anchored to each training's date and feeds the annual grant report. A reconciliation that rewrites an affiliation's dates rewrites those figures retroactively — an organization that read Ongoing at its 2026 training could start reading Reinstated afterwards. The end date is the day *before* the training, matching `ApplyScenarioEndDating`, so a row ending as another starts can't double-count the person.

## Also here

- **Active / Inactive tabs** on the person and organization affiliation editors — server-rendered, no JavaScript (`:has()` on two detached radios)
- **`inactive` as a deliberate override** — an explicitly supplied flag survives a later date edit, and an end date of today or earlier ticks it for you
- **Admin → Data health** — consistency checks spanning every record, two of them repairable
- **ADR-0003** — the two relationships one table carries, historical vs current reads, and the rules above

## Notes for review

- `docs/adr/0003-affiliations-as-the-record-of-two-relationships.md` is the fastest way in; D6/D6a–D6d are the reconciliation rules
- Organizations with a stale "Active" status column and no active affiliation will start rendering as unpublished — that's ADR-0001 D3a's drift, surfaced rather than introduced


app/services/affiliation_services/create_from_registration.rb
```diff
@@ -90,7 +90,7 @@ def create_affiliation(title, start_date:)
     end
 
     def facilitator_start_date
-      (@training_date || Date.current).to_date.beginning_of_month
+      (@training_date || Date.current).to_date
```
maebeale, line 97: 🤖 _From Claude:_ Intentional shift toward accurate dates: this starts the facilitator affiliation on the actual training date rather than the first of its month (changing code that just landed in #2194). Reconciliation works either way; flagging since it edits a freshly-merged file.

maebeale:
@jmilljr24 this is bc it's going to be harder for them to go deleting affiliations for people who don't complete the training. this gives them a view into which affiliations would be deleted (or inactivated) if they complete the action.

jmilljr24 (approved):

jmilljr24 (approved):

PR #2207, opened 2026-08-15
Person History card + Ahoy person filter

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new service + controller filter aggregating many associations, plus admin view changes

Give admins one entry point to a person's full audit trail. A new **History** card on person edit surfaces a count and links to the Ahoy activities index, scoped to that person.

- **`Analytics::PersonActivityEvents`** aggregates every Ahoy event for the person, their user account (lifecycle **and** `auth.*`, matched via the JSON `record_id` those events use), continuing-education registrations, and all associated data. Comments reuse **`PersonCommentAggregator`** so History and the comments page stay in lockstep.
- **`AhoyActivitiesController#index`** gains a `person_id` filter (expanded server-side by the service), a **`Person: <name>` chip** in the applied-filters subheader, and a **back-to-person eyebrow** (instead of the generic Admin link) when person-scoped.
- **Not-ahoy-tracked data surfaced directly** on the person-scoped activities view, since ahoy would be redundant *and* incomplete for both:
  - **Communications** — notifications are already a timestamped log; read from the notifications table (reusing `notification_row`).
  - **Attendance time entries** — self-service sign-ins happen on a login-free public callout (no `Current`), so ahoy would only capture staff edits; read the entries directly.
- **Associated-records cards**: History card fills the empty bottom-right grid cell (admin-only); added **Memberships** (gated on `Membership.enabled?`) and a **CE registrations** card (links to the admin CE index added in #2208, filtered by person); relabeled **"Subscriptions" → "Topic subscriptions"**.

### Anything else to add?
- Rebased on main; CE card links to #2208's browse index; History card's zero-count shows "None" to match #2211.
- The History link opens the index with `time_period=all_time` and all audiences, since the index otherwise defaults to *past month* and *excludes admins* — hiding most staff-made edits.
- **Membership ≠ TopicSubscription**: Membership is the annual paid-membership model; the "Topic subscriptions" card is `topic_subscriptions`.
- Full system suite not yet run (targeted specs green).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -192,6 +192,7 @@ action, or `authorize! :workshop, to: :summary?`).
 - `Analytics::LifecycleBuffer` — Thread-safe event buffer for batch tracking
 - `Analytics::EventBuilder` — Constructs analytics event payloads
 - `Analytics::AhoyTracker` — Coordinates ahoy event tracking
+- `Analytics::PersonActivityEvents` — Aggregates Ahoy events for a person, their user, and associated data (powers the person edit History card + `person_id` filter on the Ahoy activities index)
 
 ### Business Logic
 
--- a/app/controllers/admin/ahoy_activities_controller.rb
+++ b/app/controllers/admin/ahoy_activities_controller.rb
@@ -83,6 +83,15 @@ module Admin
           @person_communications = email.present? ?
             Notification.email(email).includes(:noticeable, sender: :person).order(created_at: :desc).limit(10) :
             Notification.none
+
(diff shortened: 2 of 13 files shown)


maebeale:
@jmilljr24 this adds a "History" button to footer of person edit that goes to ahoy analytics, but merges in notifications (rather than adding ahoy to communications which will bloat ahoy and basically serve same function as notifications)

jmilljr24 (approved):

PR #2216, opened 2026-08-16
Default payments index to a recent time period

maebeale:
🤖 suggested review level: 3 Read 📖 small filter scope + view grid change, model spec covers the behavior

- The payments index scanned the full history on every visit; day-to-day work is on recent payments.
- Adds a **Time period** filter defaulting to the **past month** so the first search shows recent payments — **All time** is an intentional opt-in.

- `Payment`: `PERIOD_OPTIONS` + `DEFAULT_PERIOD` constants and an `in_period` scope on `created_at`; `search_by_params` applies the default when no `period` param is present (so first load and "Clear filters" both default to the past month).
- Moved the filter fields into a 12-column grid so the narrow amount / payment-method / time-period controls share rows, with **Time period** last on the amount row.
- Shrank the metadata-search **placeholder** to 13px to match the TomSelect remote fields (typed text stays normal size/color).

### Anything else to add?
- Default is a **rolling** 30-day window (`1.month.ago`), matching the admin activities filter — not a calendar month.
- Also tightens the AI files' comment guidance (default to no comment) per team preference — unrelated to the filter but bundled here.


jmilljr24:
@maebeale I thought we talked about this a bit but I guess we didn't go into detail.

>The payments index scanned the full history on every visit; day-to-day work is on recent payments.

That is incorrect. It is paginated on most recent.

I don't see a huge value in adding day/week/month queries. If anything, we should just have 'start_date' and 'end date' search fields. Keep it simple. You don't gain anything by doing a "year" search if its paginated anyways.

PR #2217, opened 2026-08-16
Harden the file-upload answer type against forged, dropped, and oversized uploads

maebeale:
🤖 suggested review level: 5 Inspect 🔬 security + data-integrity fixes on the unauthenticated registration form, incl. an authz change and an STI type change

Follow-ups to #2136. Everything here is reachable from the public registration form, which is why it's worth a close read.

## Why

- **Documents the form advertises were rejected.** `assets.type` defaults to `"PrimaryAsset"`, so the bare `build_asset` produced one — narrowing `ACCEPTED_CONTENT_TYPES` to five image types while the field advertised Asset's full list. Uploading a PDF, DOC, DOCX, ODT, or even a GIF failed the whole registration with "File type not accepted". Now a named `FormUploadAsset`, with the view advertising that same class so the two can't drift.
- **Forged upload param 500'd a public endpoint.** A file-upload value that isn't a live signed id reached Active Storage's `find_signed!`, whose `InvalidSignature` isn't in `PublicRegistration`'s rescue list and has no `rescue_from`. Now resolved leniently and surfaced as a form error.
- **Re-registering wiped the stored filename.** `submitted_answer` was cleared before the blank guard. Active Storage's JS only disables file inputs that hold a file, so an untouched one still posts `""` — which blanked the filename while leaving the file attached, and `submitted_answer.present?` checks (`ScholarshipApplication#answered?`, the scholarship card's `answered` guard) then read the question as unanswered.
- **A validation error elsewhere on the form discarded an already-uploaded file.** File inputs can't be repopulated, and every other answer type round-trips its `value`. The signed id now comes back in `public_registration[retained_uploads][<field_id>]` — a separate param name, because a blank file input would win on duplicate keys — and the controller falls back to it.
- **No size cap.** `Asset` validated content type but not bytes, and the form is public. Capped at 25 MB, with a client-side guard so an oversized file isn't direct-uploaded before the server rejects it.
- **`person_id` on the registration show page was ungated.** Person ids are sequential, so unlike the registration slug that branch is not a capability — it's the admin-side fallback for registrations with no slug, and it now exposes uploaded files. It requires `EventPolicy#form_submissions?` (or being that person); slug access stays public.

## Notes for the reviewer

- `submitted_answer` is only a cache of the attachment's filename — the `Asset` is authoritative. `FormAnswer#sync_uploaded_filename!` is the single derivation, with a spec pinning the invariant across upload / replace / blank re-submit.
- `FormUploadAsset` is deliberately absent from `Asset::TYPES` — that list drives the author-facing asset-type picker and these are never chosen there.
- The size validation runs on every `Asset` save, matching the existing `file_type` validation — an already-oversized asset would block edits, same as an already-unaccepted content type does today.
- `file_preview_controller`'s new `maxSize` value defaults to `0` (no limit), so its other callers are unaffected.
- Existing `GET show` request specs now sign in as an admin, since they exercise the `person_id` path.

## Not in scope

Direct upload stores the blob before submit, so abandoned forms and rejected files leave unattached blobs and nothing reclaims them. Tracked separately in #2236.

### DIFF
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -33,7 +33,7 @@ module Events
       @scholarship_form = @event.scholarship_form if @scholarship
       @continuing_education_form = @event.continuing_education_form
 
-      all_params = params.dig(:public_registration, :form_fields)&.to_unsafe_h || {}
+      all_params = merge_retained_uploads(params.dig(:public_registration, :form_fields)&.to_unsafe_h || {})
       registration_params, scholarship_params, continuing_education_params = split_form_params(all_params)
 
       @field_errors = validate_required_fields(registration_params)
@@ -96,6 +96,12 @@ module Events
         person = registration.registrant
       elsif params[:person_id].present?
         person = Person.find(params[:person_id])
+        # Unlike the registration slug — an unguessable token handed to the
+        # registrant — person ids are sequential, so this branch is not a
+        # capability. It exists as the admin-side fallback for registrations with
+        # no slug, so gate it on the same access as the event's other submission
+        # views (or on the viewer being that person).
+        authorize! @event, to: :form_submissions? unless current_user&.person_id == person.id
         registration = @event.event_registrations.find_by(registrant: person)
       else
         redirect_to event_path(@event), alert: "Registration not found."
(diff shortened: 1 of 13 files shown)


app/controllers/events/public_registrations_controller.rb
```diff
@@ -96,6 +96,12 @@ def show
         person = registration.registrant
       elsif params[:person_id].present?
         person = Person.find(params[:person_id])
+        # Unlike the registration slug — an unguessable token handed to the
+        # registrant — person ids are sequential, so this branch is not a
+        # capability. It exists as the admin-side fallback for registrations with
+        # no slug, so gate it on the same access as the event's other submission
+        # views (or on the viewer being that person).
+        authorize! @event, to: :form_submissions? unless current_user&.person_id == person.id
```
jmilljr24, line 104: I don't understand what this means but if its related to authorization, can it be moved to the policy?

jmilljr24: I think we skip authorization in other places, so I'm not against this. I just don't understand it. So your judgement.

PR #2226, opened 2026-08-16
HOLD UNTIL WE MESS W RHINO EDITOR: Make rich-text editor previews match the published page (WYSIWYG)

maebeale:
🤖 suggested review level: 1 Skim 👀 CSS-only; no Ruby, JS, or markup changes

- The rhino editor didn't preview like the published page, so editing was guesswork.
- Published rich text renders inside `prose` (`app/views/layouts/action_text/contents/_content.html.erb`). The editor sits in the same `prose` wrapper, but `trix.css`'s blanket `.trix-content *` reset zeroes every margin and padding — so the editor showed collapsed blocks, unindented lists, and oversized tables.

- Restate, scoped to `custom-rhino-editor .trix-content`, every value the reset strips: paragraph/list margins, list indent and item spacing, blockquote, `h1`–`h4`, `figure`, `hr`, `table` font-size + cell padding, and the grid cell's own `p-3`.
- Values mirror `@tailwindcss/typography`'s em defaults, so the two stay in step.
- Higher specificity than `.trix-content *` is required, hence the repetition rather than leaning on `prose`.
- Drop the old editor-only `table` padding and zeroed cell margins — the published page gives paragraphs inside a cell full prose margins.
- Restate `h1`'s font-size: `trix.css` shrinks it to `1.2em`, which would halve the em-based margin below it.

### UI Testing Checklist
- [ ] Story body: paragraphs, heading, list, blockquote — editor spacing matches the published story.
- [ ] Event header/description: images and grid cells preview at published size.
- [ ] Table in a story: cell padding and paragraph spacing match the published page.
- [ ] Grid cells still show the dashed editing outline, with published padding.

### Anything else to add?
- Scope note: earlier `centered:` / `full_width_images:` options on `rhino_editor` were dropped — centering already comes from the content's own inline `text-align`, so no per-form styling is needed. This PR is now CSS-only.
- Editor-only chrome that stays on purpose: dashed grid outlines, table cell borders, attachment filename labels.

### DIFF
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -549,14 +549,14 @@
         <%= f.input :autoshow_title, as: :boolean, label: "Show title on event show page" %>
 
         <div class="form-group <%= 'has-error' if f.object.errors[:rhino_header].present? %>">
-          <%= rhino_editor(f, :header, label: "Header content") %>
+          <%= rhino_editor(f, :header, label: "Header content", centered: true, full_width_images: true) %>
 
           <% if f.object.errors[:rhino_header].present? %>
             <div class="text-danger mt-1 text-sm">Field <%= f.object.errors[:rhino_header].first %></div>
           <% end %>
         </div>
 
-        <div class="form-group mt-12"><%= rhino_editor(f, :description, label: "Description") %></div>
+        <div class="form-group mt-12"><%= rhino_editor(f, :description, label: "Description", full_width_images: true) %></div>
       </div>
     </div>
 
--- a/app/views/events/_form.html.erb
+++ b/app/views/events/_form.html.erb
@@ -549,14 +549,14 @@
         <%= f.input :autoshow_title, as: :boolean, label: "Show title on event show page" %>
 
         <div class="form-group <%= 'has-error' if f.object.errors[:rhino_header].present? %>">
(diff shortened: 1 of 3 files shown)


jmilljr24 (approved):
No required, but some screenshot would be nice. I pulled this down quick and tried to find an example of what that changes but I didn't notice anything.

maebeale:
@jmilljr24 Before and after, where spacing looks in Edit more like how it will look on Show.


Screenshot 2026-08-18 at 5 41 24 AM

Screenshot 2026-08-18 at 5 41 36 AM

app/views/events/_form.html.erb
```diff
@@ -549,14 +549,14 @@
         <%= f.input :autoshow_title, as: :boolean, label: "Show title on event show page" %>
 
         <div class="form-group <%= 'has-error' if f.object.errors[:rhino_header].present? %>">
-          <%= rhino_editor(f, :header, label: "Header content") %>
+          <%= rhino_editor(f, :header, label: "Header content", centered: true, full_width_images: true) %>
```
jmilljr24: I know there isn't really any risk with this because it's just the editor but I don't love the idea of setting different styling for editors in different places. This is making the assumption that every Event will have the header section as a full width image.

I understand the slight discrepancy from the editor to the view, but every editor on the entire internet displays things differently from edit to show.

PR #2234, opened 2026-08-16
Interleave communications into the person activity timeline

maebeale:
🤖 suggested review level: 3 Read 📖 contained change to one admin page: a new merge/paginate service, a view split into row partials, and filter routing

### What & why

- The person-scoped **Activities** page (`/admin/activities?person_id=…`) showed **Communications** (Notification records) in a separate panel above the activity table.
- Notifications aren't ahoy-tracked, so the two were independent streams displayed side by side.
- Now they're **merged into one time-ordered timeline** so a facilitator's emails and their profile edits/views read as a single history.

### How

- **`Analytics::PersonTimeline`** — merges the person's events + communications into one newest-first list of `Entry` records; the controller paginates it in Ruby (`will_paginate/array`).
- Split the activity `<tr>` into `_activity_row`; added `_communication_row` (tinted, bell icon) that reuses `notifications/_notification_row` with a new `show_date: false` local (the table's Time column carries the timestamp).
- Global (non-person) view is unchanged — still DB-paginated events only.

### Filter behavior for communications

- **Person/user scope** and **date range** narrow them.
- **Activity-name** search matches the communication **subject**.
- **Props** search matches **subject + body**.
- Event-only filters (visit, resource, prefixes) don't exclude communications.

### Tests

- `PersonTimeline` service spec (merge order, kinds, sort key, empty).
- Request specs: interleaving + ordering, activity-name→subject, props→subject/body, event-only filter doesn't drop comms.

### DIFF
--- a/app/controllers/admin/ahoy_activities_controller.rb
+++ b/app/controllers/admin/ahoy_activities_controller.rb
@@ -172,11 +172,11 @@ module Admin
     # The person's communications for the merged timeline. Only the filters that
     # make sense for a message narrow them: the person/user scope, the date
     # window, the activity-name search (matched against the subject), and the
-    # props search (matched against subject + body). Event-only filters (visit,
-    # resource, prefixes) don't exclude communications.
+    # props search (matched against subject + body). A visit_id filter excludes
+    # communications entirely — they have no visit to prove membership in one.
     def person_communications
       email = @person.communications_email
-      return Notification.none if email.blank?
+      return Notification.none if email.blank? || params[:visit_id].present?
 
       scope = Notification.email(email).includes(:noticeable, sender: :person).order(created_at: :desc)
       scope = scope.where(created_at: time_range) if time_range.present?
--- a/app/controllers/admin/ahoy_activities_controller.rb
+++ b/app/controllers/admin/ahoy_activities_controller.rb
@@ -94,8 +94,7 @@ module Admin
           .limit(15)
           .decorate
 
-        # Notifications aren't ahoy-tracked, so merge the person's communications
-        # into the activity events for one time-ordered timeline (paginated in Ruby).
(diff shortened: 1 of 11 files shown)


app/controllers/admin/ahoy_activities_controller.rb
```diff
@@ -72,30 +72,32 @@ def index
...
+          .where(event_registration_id: @person.event_registrations.select(:id))
+          .includes(event_registration: :event)
+          .order(signed_in_at: :desc)
+          .limit(15)
+          .decorate
+
+        # Notifications aren't ahoy-tracked, so merge the person's communications
+        # into the activity events for one time-ordered timeline (paginated in Ruby).
+        entries = Analytics::PersonTimeline.new(
+          events: scope.to_a,
```
maebeale, line 99: 🤖 _From Claude:_ To page a merge of two tables as one timeline, this loads the person's full ahoy-event set into memory (`scope.to_a`) rather than DB-paginating. Person history is bounded and this is admin-only, so it's acceptable, but flagging the tradeoff — a very high-activity facilitator would load many rows here.

maebeale:
@jmilljr24 i really don't want to do the ahoy surgery and papertrail update now. but, maybe now's already the time?

jmilljr24:
> @jmilljr24 i really don't want to do the ahoy surgery and papertrail update now. but, maybe now's already the time?

I would think you could add papertrail now on the applicable models and at a later date work through cleaning up what ahoy touches. 

Like just for event registrations, what trail/models do we want to surface? Then slowly add in more as we see what comments that make.

maebeale:
@jmilljr24 you ok w merging this so it's ahoy + communications now. then in new pr i do the papertrail and change the page to be papertrail + ahoy + communications?

jmilljr24 (approved):

PR #2237, opened 2026-08-17
Public pretty-URL endpoint for standalone forms + form-filterable submissions index

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new public unauthenticated endpoint that find-or-creates People + FormSubmissions and sends email; consolidates the answer-persistence used by event registration onto the model

## What
- **Publish a standalone (event-less) `Form` at a public pretty URL** (`/f/:slug`) — anyone can fill it out, no account.
- **View those submissions** via the existing `/form_submissions` index, now filterable by form and lazy-loaded.
- **Email** a confirmation to the submitter and an FYI to the AWBW team on every public submission.

## Public form endpoint
- `Form` gains `slug` (admin-set, URL-normalized, unique) + `published`; `publicly_fillable?` (standalone + published + slug) gates the route and `FormPolicy#public_show?`.
- `PublicFormsController` (`show`/`create`/`thank_you`, unauthenticated) with honeypot + required-field validation; reuses the public-registration field partials.
- `PublicFormSubmission` service: find-or-creates a `Person` from the form's name/email answers (email + last-name match reuses an existing person), records one-time consent, stores a `role: "public"` submission, captures "Other" answers, sends notifications.
- The slug column never holds `""`. Truly blank input becomes `nil` so the unique index tolerates the many forms without a slug; input that `parameterize` strips to nothing (`"!!!"`) is left intact for the format validation to reject, rather than silently blanked into a later "has already been taken" collision.

## Emails
- New notification kinds `form_submission_confirmation` (to submitter) + `form_submission_confirmation_fyi` (to admin), with `NotificationMailer` methods, html/text views, job dispatch, and previews — mirroring the event-registration confirmation/FYI pattern.

## Answer persistence (consolidation)
- Moved `persist_answer` + the hardened file-upload path onto **`FormSubmission#persist_answer`**. Event registration, public forms, **and** bulk payment share it; the interim mixin is gone.

## Submissions index
- Branches on the Turbo frame, filters by `form_id`, paginates, links from the forms index.
- Event-less public submissions show a **"Public form"** pill.
- **Fix:** the View link now breaks out of the results frame (`turbo_frame: "_top"`) — previously it navigated the frame to a frameless show page and errored.
- **Eyebrow:** arriving from the forms index shows a **← Forms** back link anchored to that form's row.
- **Fix:** the View link carries `form_id`, so the trip back keeps the form filter instead of dropping to that person's submissions.
- **Fix:** the View link also carries the index's own `origin`, which the detail page hands back — so the **← Forms** eyebrow survives the full Forms → submissions → detail → back round trip instead of vanishing on the return leg. The filter form carries it too, so changing the filter doesn't strip it from the frame's params.

## Forms index / editor polish
- Event-connected unpublished forms read **"Event form"**, not "Not published".
- Delete link removed from the index; on the editor, Save/Delete placement swapped and a reason shown when a form can't be deleted.
- Editor's Public form card clarifies publish only controls the public link.

## Dev seed
- `db/seeds/dev/public_forms.rb` — two published event-less forms with submissions + answers, wired into `db:seed:dev`. Idempotent.

## Open question
- A form **can** be both an event form and publicly published (standalone + published + slug, regardless of event links). Currently allowed; a guard could restrict publishing to role-less forms if desired.

## Follow-up
- Rate limiting for the public POST endpoints is deliberately out of scope — tracked in #2252. `/f/:slug` is always open, unlike the event endpoints behind `ensure_registerable`; the honeypot is the only current defence.
- #2254 — moving `persist_answer` onto the model gave the bulk-payment path a raise it couldn't produce before. `BulkPayment#call` rescues only `RecordInvalid`, so a `file_upload` field with a forged or stale signed id would 500 rather than re-render. Reachability is narrow today (the bulk-payment view renders no file input), hence a follow-up rather than an inline fix.

## Tests
- New: `public_forms` request, `PublicFormSubmission` service (incl. notifications + "Other" capture), `Form` slug/publish + slug-rejection, `FormSubmission#persist_answer`, `FormPolicy#public_show?`, `NotificationMailer` form-submission emails, forms-index public-link column, submissions eyebrow + frame-escape + form-filter + origin round trip.
- Registration + bulk-payment + notification-job specs pass unchanged.
- Full suite green: 6958 examples, 0 failures.


maebeale:
@jmilljr24 they need public forms for collaboration agreement (for reups outside events), annual survey. and could also use them for the immediate need of post-event survey, day-1-survey, day-2-survey until we get those wired up into event edit selections. any concerns w approach taken in this pr w the slug and published boolean?

jmilljr24 (approved):

app/controllers/public_forms_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  def show
+    authorize! @form, to: :public_show?
+    @form_fields = ordered_fields
+  end
+
+  def create
+    authorize! @form, to: :public_show?
+
+    # Honeypot — a bot that fills the hidden field is silently bounced.
+    if params.dig(:public_registration, :website_url).present?
```
jmilljr24, line 17: Great idea! (and perfect comment).

Any other key we could use instead of website_url? That's a real value we use in other places and could be confusing or conflict at some point.

PR #2241, opened 2026-08-17
Cache recipient_name on notifications to preserve contact names

maebeale:
🤖 suggested review level: 3 Read 📖 small denormalization + a before_create hook and decorator fallback

Cache the contact person's name onto each notification so the communications index and detail views keep showing a real name even after the linked `Person` is nullified or deleted — today they degrade to a raw email once the person is gone.

Added a nullable `recipient_name` column, a `before_create` hook that snapshots the person's name from `recipient_email` (primary or `email_2`, reusing the decorator's lookup) filling only blanks, and a `contact_name` fallback to the cached value before the raw email. Populated going forward only — no backfill.

### UI Testing Checklist

- [ ] Communications index/detail still show the person's live name and email hover when the person is on file.
- [ ] A communication whose person has been deleted shows the cached name instead of a bare email.

### Anything else to add?

Scope is intentionally just `recipient_name` — `sender_name` and the separate `User` `dependent: :nullify` gap were discussed but left out.

### DIFF
--- a/app/decorators/notification_decorator.rb
+++ b/app/decorators/notification_decorator.rb
@@ -42,9 +42,10 @@ class NotificationDecorator < ApplicationDecorator
     end
   end
 
-  # Name of the person when we know them, otherwise the raw email.
+  # Name of the person when we know them, then the cached snapshot taken at send
+  # time (survives a deleted/nullified Person), otherwise the raw email.
   def contact_name
-    contact_person&.name.presence || recipient_email
+    contact_person&.name.presence || recipient_name.presence || recipient_email
   end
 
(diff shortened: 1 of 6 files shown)


maebeale:
@jmilljr24 building up to deleting people, this assumes we keep all notifications after deletion. were you thinking we should dependent destroy?

jmilljr24:
> @jmilljr24 building up to deleting people, this assumes we keep all notifications after deletion. were you thinking we should dependent destroy? 

My gut says yes(haven't looked at it deeply). If the intent of person delete is only for cleaning up tests or mistakes, I would think the notifications should be deleted as well.

I think any other situation would warrant more of an archive

Issue #12, opened 2025-05-22
Support running the application locally without S3 account


carlost:
The application uses paperclip for handling image and document (PDF, word doc) handouts for workshops. The only storage option configured for all envs is S3. We will ultimately replace paperclip with activestorage, but that will likely be a few weeks/months off. It the meantime it would be great if paperclip used the local file storage option for the test and development env.

maebeale:
closing this since i believe we're making progress on moving off paperclip and on to activestorage via @jmilljr24 's work

PR #2245, opened 2026-08-17
Capture what a save actually changed

maebeale:
🤖 suggested review level: 5 Inspect 🔬 `AhoyTrackable` is included in `ApplicationRecord`, so this changes what every save in the app records, and how much it writes

## What a save did to a record's children went unrecorded

Rich text, nested records, and attachments were all invisible to activity tracking. Every `rhino_*` field passed through unrecorded — event descriptions, stories, resources, community news — and no event ever said a nested record was added, edited, or removed, or that a file was attached.

`AhoyTrackable` has collectors for exactly this. None of them ever produced anything, for the same two reasons:

- They looked associations up with a **string**, but the association cache is symbol-keyed — a string returns a fresh, unloaded association whose target is empty, so there was never anything to inspect.
- They read the result **after** the save, by which point autosave has cleared the dirty state they were looking for.

**This is the half that can't wait.** An event that isn't captured today can't be reconstructed later; every admin edit to a description, a phone number, or a license is lost until this lands.

## What's captured now

Everything is read in `before_save`, while it's still pending, and the update event assembles what was captured.

- **Rich text** lands on the record's own event keyed by the attribute (`rhino_description`), as plain text truncated to a 300-character preview — readable in a change log, and no whole articles in `ahoy_events`.
- **Nested records** report added / updated / removed **with their content** — an added comment reads as its body, an edited one as before/after. Keys, timestamps, and secret-shaped columns are left out; the child's own event keeps the full snapshot. The id is filled in after the save, since a record added through nested attributes has none while the parent's event is being assembled.
- **Attachments** come from ActiveStorage's staged changes — the only reliable account of an attach or purge — and carry the filename, which is the part a reader of a change log wants. A staged attachment has no id yet, so the details cell renders it by name and action instead of a dead reference lookup.

Children's own events are unaffected; this is the parent's account of what a save did.

## Category and sector changes, assigned outside dirty tracking

`categories=` / `sectors=` — the `has_many :through` collection setters behind the category / workshop-settings and sector checkboxes — persist immediately, outside the record's own save, so the collectors above never saw them and a category changed on the edit form was invisible in the log. `TagAssignable#assign_associations` now diffs the memberships around the reassignment and folds added / removed entries onto the record's own update event, reading like the nested edits do. Covers every `assign_associations` consumer (people, organizations, workshops, resources, stories, events, …), not just people.

## Seeing it on the record

`application/_activity_log` puts a record's own events on its page as a **Change log** dropdown, under the audit line already there. Rendered on the pages admins correct records from:

`people` · `organizations` · `affiliations` · `scholarships` · `continuing_education_registrations` · `topic_subscriptions` · `topic_subscription_types` · `event_registrations` · `memberships` · `membership_invoices` · `professional_licenses` · `forms` · `features` · `category_types` (all `#edit`)

Same position on every one: directly under the created/updated line, so it reads as the rest of that record's edit history. Eight of those pages had no created/updated line at all — they have one now.

CE registrations and staff tags get it on their **show** page — that's the page an admin lands on for those, not the form.

A record with no tracked activity reads "Change log empty" as plain text; the control with its chevron appears only when there's something to expand. Leaving it out entirely gave no way to tell *nothing has changed* from *this page hasn't got one*, which matters while tracking is new and most records predate it.

Browsing stays out of it: `Ahoy::Event.mutations` drops `view.`, `print.`, `search.`, `filter.`, and `download.` events, which remain on the admin activities index where browsing is the point.

`payments#show` deliberately keeps its existing PaperTrail change log (#1737) rather than switching — not an oversight.

A shared example (`spec/support/shared_examples/change_log_page.rb`) keeps each honest about both halves: the log appears when the record has activity, and stays out of the way when it doesn't.

The before/after rendering isn't rebuilt — #2316's details cell is extracted from `_activity_row` into `admin/ahoy_activities/_event_details`, rendered by both, so a change reads the same inline as it does in the admin table. That cell also gained the detail an association entry carries: it linked the record and dropped its `changes`/`attributes`, so both pages showed "added Comment" and nothing else. Admin-gated by `Admin::AhoyActivityPolicy`, matching the activity links in `shared/_audit_info`.

Each log resolves its referenced records in one batch per type (`Analytics::EventReferenceLoader`, already used by the admin table) rather than one query per reference.

## A model fix this uncovered

`SectorableItem#title` had its guard inverted — everything that *wasn't* a `WorkshopLog` fell into the branch that calls `sectorable.title`, and an organization hasn't got one. Nothing reached it while association changes were never recorded; once they were, the org's change log referenced its sector taggings and `/organizations/:id/edit` raised `NoMethodError`.

The guard is fixed (a tagging reads as the sector it applied), and two callers are hardened, because a model's label shouldn't be able to take down a page that merely mentions the record:

- `Ahoy::EventDecorator#reference_row` falls back to `SectorableItem #12` instead of raising.
- `Analytics::EventBuilder.safe_title`'s rescue was calling the same raising method again, so `create.sectorable_item` events were **being dropped and logged** rather than recorded. That was already happening on `main`.

## Comments and communications read as their own domain

- Person edit was the only comments card in the app hard-coded blue. It is now the standard card with the comments (purple) icon chip, matching the communications card directly above it.
- A comment in edit mode had no box at all, so opening the editor looked like reading it. It now gets the comments-theme card, like a new comment and an open communication already did.
- The `admin-only` blue wash moves off the card and onto the heading, so the section still reads as privileged.

Listed on **Features & tips** (`config/features.yml`) so facilitators find it.

## How to test

Ahoy events are written during a **signed-in request**, so console and rake edits produce nothing to look at.

1. `ai/server`, sign in as an admin.
2. Edit an event's description (`/events/:id/edit`) — previously recorded nothing at all; now `update.event` carries a `Rhino description` before/after. Check `/admin/activities/events`.
3. Edit a person's nested records (`/people/:id/edit` — add a phone, change a license, remove one) or replace their avatar. The person's own event now reports each child change, with the filename for the avatar.
4. On any of the pages above, change a field, save, reload, and open **Change log** under the created/updated line. On `/people/:id/edit`, give the person a staff tag — the log reports the tagging, since staff tags are assigned through nested attributes on that form.
5. On `/people/:id/edit`, tick or untick a category or workshop-settings box (or a sector on `/organizations/:id/edit`) and save — the **Change log** now reports the category/sector added or removed.

### DIFF
--- a/app/controllers/concerns/tag_assignable.rb
+++ b/app/controllers/concerns/tag_assignable.rb
@@ -9,6 +9,8 @@ module TagAssignable
     selected_category_ids = Array(params[key][:category_ids]).reject(&:blank?).map(&:to_i)
     selected = Category.where(id: selected_category_ids).to_a
 
+    categories_before = record.categories.to_a
+
     if params[key].key?(:managed_category_type_ids)
       # The form only edits certain category types (e.g. age ranges + workshop
       # settings). Preserve taggings of every other type the form never shows so
@@ -20,10 +22,14 @@ module TagAssignable
     else
       record.categories = selected
     end
+    categories_after = record.categories.to_a
 
-    if params[key].key?(:sector_ids)
+    sectors_changed = params[key].key?(:sector_ids)
+    if sectors_changed
+      sectors_before = record.sectors.to_a
       selected_sector_ids = Array(params[key][:sector_ids]).reject(&:blank?).map(&:to_i)
       record.sectors = Sector.where(id: selected_sector_ids)
+      sectors_after = record.sectors.to_a
     end
 
     record.save!
@@ -34,5 +40,20 @@ module TagAssignable
     if params[key].key?(:primary_age_category_ids) && record.respond_to?(:apply_primary_age_groups!)
       record.apply_primary_age_groups!(Array(params[key][:primary_age_category_ids]))
     end
+
+    # These memberships change outside the record's dirty tracking, so hand the
+    # diff to the change log directly (a no-op when nothing actually moved).
+    return unless record.respond_to?(:track_membership_changes)
+
+    record.track_membership_changes(
+      categories: membership_delta(categories_before, categories_after),
+      sectors: (membership_delta(sectors_before, sectors_after) if sectors_changed)
+    )
+  end
+
+  def membership_delta(before, after)
+    before = Array(before)
+    after = Array(after)
+    { added: after - before, removed: before - after }
   end
 end
--- a/app/decorators/ahoy/event_decorator.rb
+++ b/app/decorators/ahoy/event_decorator.rb
@@ -64,7 +64,7 @@ module Ahoy
     end
 
     def hash_rows(hash, label, depth)
-      return [ reference_row(label, hash, depth) ] if reference?(hash)
+      return reference_rows(label, hash, depth) if reference?(hash)
       return [ { label: label, value: "(empty)", depth: depth } ] if hash.empty?
 
       rows = label ? [ { label: label, value: nil, depth: depth } ] : []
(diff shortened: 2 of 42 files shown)


maebeale:
@jmilljr24 looks like this is what would be needed to properly track registration lifecycles (with an eye to person timelines and addition of subject_person_id)

jmilljr24:
@maebeale check my slack message

jmilljr24 (approved):

PR #2258, opened 2026-08-18
Add "Copy form" button to duplicate a form with its fields and options

maebeale:
🤖 suggested review level: 3 Read 📖 small additive feature — one service, one action, two view links; core risk is what the deep-copy includes/omits

## Why
Admins rebuild near-identical forms by hand. A one-click **Copy** seeds a new editable form (`COPY of [Name]`) from an existing one so they can tweak rather than start from scratch.

## What
- **`FormCopyService`** — transactional deep-copy of a `Form`.
  - Copies every `form_field` (order preserved) and each field's `form_field_answer_options`, **re-pointed at the shared `AnswerOption` records** (not duplicated).
  - Leaves `slug` nil and `published` false — a published form requires a unique slug.
  - Does **not** copy submissions, event links, `user_forms`, or reports.
  - Keeps `role` — nothing selects a standalone form by role expecting uniqueness (event-level lookups go through `event_forms`), and role stays editable in the form editor.
- **`POST /forms/:id/copy`** → `FormsController#copy` (admin-gated), redirects to the new form's editor.
- **Copy** link on the forms index row and the show-page action bar, each with a confirm.
- **Features & tips seed** — new `Copy a form` entry in `config/features.yml`.

## Tests
- `FormCopyService` spec — naming, field/option copy, shared AnswerOption reuse, slug/published reset, submissions/event links excluded, original untouched.
- `POST /forms/:id/copy` request spec — admin success + redirect, non-admin denied.

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -1,5 +1,5 @@
 class FormsController < ApplicationController
-  before_action :set_form, only: %i[show edit update destroy reorder_field reorder_fields edit_sections update_sections]
+  before_action :set_form, only: %i[show edit update destroy copy reorder_field reorder_fields edit_sections update_sections]
   before_action :set_dashboard_event, only: %i[show edit edit_sections update update_sections]
 
   def index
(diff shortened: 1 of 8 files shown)


jmilljr24 (approved):
Great idea! Thanks

PR #2259, opened 2026-08-19
Denormalize the facilitator affiliation flag off the title

maebeale:
🤖 suggested review level: 5 Inspect 🔬 denormalized flag on a core model, plus a migration and a post-deploy backfill that gates org Art Program status

- "Is this a facilitator affiliation?" was a raw, collation-sensitive `BINARY TRIM(title) = 'Facilitator'` SQL scope.
- The same rule was re-encoded in Ruby and copy-pasted across three JS controllers.
- Four hand-written copies of one rule, one of them unindexable.

**Model**

- New `affiliations.facilitator` boolean (`null: false, default: false`).
- `sync_facilitator_from_title` (`before_validation`) keeps it in step with the title: exactly `"Facilitator"`, trimmed, case-sensitive.
- Title stays the input and the single source of truth — a retitle re-syncs the flag.
- `.facilitators` reads the flag; `#facilitator?` is the column's own reader.
- Invariant: never write `title` via `update_columns`/`update_all`, which skips the callback and lets the flag drift.

**JS**

- Rule extracted to `lib/affiliation.js` (`isFacilitatorTitle`); the three affiliation editors share it for their live preview.
- Nothing facilitator-related is submitted from the form — the server owns the flag.

### ⚠️ Required post-deploy step

- Run `rake affiliations:backfill_facilitator` immediately after deploy.
- Until it runs, every existing row reads `facilitator: false`, so `.facilitators` matches nothing.
- In that window `Affiliation#sync_organization_status_with_affiliations` (an `after_save`) sees no active facilitators and flips the org to Inactive via `update_column`, logging `autochange.organization`. Any affiliation save triggers it, including public event registration.
- Open for review: folding the two `UPDATE`s into the migration's `up` closes the window.

### Anything else to add?

- Earlier revisions of this branch modeled the split as STI (`FacilitatorAffiliation` / `JobAffiliation`). Reverted: the subtypes needed `model_name`, `policy_class`, and an Ahoy `event_resource_types` indirection purely to keep behaving like one `Affiliation`.


maebeale:
@jmilljr24 as i work more w the affiliations and org's art program status and org's active/inactive status, it'd really help to have the affiliation itself identified by more than the Title string on affiliation. you ok w adding this boolean?

i still want to do the same js style changes based on the title field, so this also includes drying up that logic bc it's used in more than one place rn.

jmilljr24 (approved):
I agree, this makes sense. Some of this will be surfaced again when membership gets fully  built but I think a boolean to capture this on affiliation has been knocking on our door for a while.

maebeale:
@jmilljr24 i talked myself out of this bc i'd still want to keep it simple for staff and therefore need that callback reconciliation on title. at that point, the boolean is just row-level caching and doesn't buy us much. def open to re-opening.

PR #2296, opened 2026-08-21
Pad unpadded back-nav eyebrows to match the standard eyebrow

maebeale:
🤖 suggested review level: 3 Read 📖 adds `px-2 py-1` to 128 back-nav eyebrow links so they match the standard; verified by view specs

Follow-up to #2292 (merged, added `eyebrow_link_class`).

### Goal
- 128 back-nav / eyebrow links used the shared `eyebrow_link_class` color but omitted the `px-2 py-1` the standard eyebrows carry, so they sat slightly out of line.
- Pad them so every back-nav eyebrow is visually identical — same helper, same class string.

### Scope
- Only the plain unpadded back-nav eyebrows (`text-sm #{eyebrow_link_class}` at the end of the class list) — confirmed nav links ("← Subscriptions", "Home", back links).
- **Left distinct components alone**: `text-xs` "View all" links, edit/close toggle buttons (`<summary>`, search close-X, edit toggles), and a few eyebrows with special margins.

### Verification
- Full view-spec suite green (494 examples, 0 failures). No new Tailwind classes, so no build change.

### Note
- Visual consistency, not new single-control — fully single-controlling eyebrow size/padding would be the two-helper split discussed on #2292.

### DIFF
--- a/app/views/admin/ahoy_activities/charts.html.erb
+++ b/app/views/admin/ahoy_activities/charts.html.erb
@@ -2,7 +2,7 @@
 <div class="<%= DomainTheme.bg_class_for(:admin_only) %> rounded-xl border border-gray-200 p-6 shadow">
       <div class="mb-2 flex flex-wrap items-center gap-y-2">
         <% if allowed_to?(:index?, with: Admin::HomePolicy) %>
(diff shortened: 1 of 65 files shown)


maebeale:
@jmilljr24 working on getting the site brandable by consolidating classes in general. had this as a tailwind apply and then converted to a helper.

jmilljr24 (approved):

PR #2297, opened 2026-08-21
Introduce button_classes helper (move buttons off @apply)

maebeale:
🤖 suggested review level: 3 Read 📖 new helper + reference conversion of the event-flow buttons; rest of the app unchanged

## Why
- Evil Martians best practice #4: express reusable button styles as utilities in a helper, not `@apply` component classes in CSS. `@apply` hides where styles change and bloats the bundle.

## What
- New `ButtonHelper#button_classes(variant, size:, extra:)` — single source of truth for base + all variants + sizes. Tailwind already scans `app/helpers/**/*.rb`, so the utility strings generate normally.
- Converted the event registration-flow buttons as the reference pattern: registration section, event show, event card, scholarship callout.
- Helper spec added.

## Deliberately incremental
- `buttons.css` (`.btn` / `.btn-*` `@apply`) **stays** — the ~200 remaining call sites still use it, so helper + `@apply` coexist with zero visual change. Convert-on-next-touch, per the same Evil Martians guidance against big-bang `@apply` rewrites.

## Gotcha this design handles
- Today `.btn`'s padding lives in the `components` layer, so a call site's `px-10`/`text-2xl` (in the `utilities` layer) overrides it. Once styles are inlined they're all utilities in one class list, where conflicts resolve by Tailwind's **generation order, not markup order**. So `size:` is explicit: `:md` (default), `:sm`, or `nil` when the call site supplies its own padding/text-size via `extra:`.

## Open question
- Full sweep of the remaining ~200 files now (large diff) vs. incremental next-touch? See PR thread.


maebeale:
@jmilljr24 curious if you think this work it to move off tailwind apply

jmilljr24 (dismissed):
I'm all for avoiding @apply where we can but I'm confused with this PR. If we had a custom button we used in many places, I think this pattern in the PR is sound. However, @apply isn't being used for buttons. `btn` is a built in tailwind class and `btn-primary` uses @theme which is what we want to do. This change seems to just duplicate what tailwind already gives us for buttons.

maebeale:
@jmilljr24 i wanted to apply the custom branding info and success colors from the branding sheet they sent to us. so that would've been an apply.

jmilljr24:
Just for clean up of my comments. I accidentally edited you comment instead of creating a new one. I meant to comment this:

OOOOH, totally forgot we built the button. I was looking at the tailwind docs for their components(paid feature we don't use)

Sorry. This is great!

PR #2304, opened 2026-08-22
Split the collaboration agreement into 3 scenario forms with per-scenario affiliation processing

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new affiliation-mutating actions (end-dating, org linking), a notification kind, and index filter scopes across forms/people/submissions

## Why
- One collaboration agreement form served three intake scenarios — on-demand training, reinstatement, job change (see the AWBW walkthrough videos) — so scenarios couldn't be counted and each needs different handling of existing affiliations.
- Splitting into three purposed forms + processing tooling replaces the FileMaker-era manual flow: see submissions, reconcile affiliation data, grant portal access.

### Scenario forms — expressed through form roles (no stored purpose)
- No new column: a **standalone public registration-role form is the on-demand agreement** (the registration flow, timestamped at submission instead of event start), and **`new_job`** / **`reinstatement`** are new form roles. `FormSubmission#linking_scenario` derives the scenario per submission (registration-role counts only on the standalone public path). The forms admin copies the collab form into carry these roles.
- (Earlier iterations of this PR used a `forms.purpose` column; it was added and later removed — the two migrations cancel out.)

### On-demand agreement = registration for the on-demand training
- `Event.current_on_demand_facilitator_training` picks "this year's" published on-demand training (most recently started, else next upcoming). Submitting the standalone on-demand agreement form auto-registers the person for it (idempotent via the unique registrant+event index) and stamps the event on the submission — so the fallback `/f/` path and the event registration form land in the same place.

### Send the right link from a person's page
- Person edit gains an "Email form links" section (collapsed by default): a **Registration** group nests the current **on-demand training** and each **upcoming scheduled facilitator training (next 12 months)** — Send emails that event's **public registration form** (the full registration pipeline; the standalone `/f/` on-demand form is only a fallback and isn't offered). New job / reinstatement send their `/f/` links. Every send is recorded as a `form_link_request` notification (new kind + mailer with the standard button + fallback-URL footer) with per-row send counts and last-sent-by.

### Processing a submission
- Purposed submissions show an admin processing panel: scenario guidance, submitted organization, the person's affiliations with a **one-click End** (job change only, dated to the submission — new `POST /affiliations/:id/end`), and portal-account status with Invite/Create user.

### Submissions index as a processing queue
- Extends main's #2310 grid filter bar (form, role, event/person/org remote-selects, date range) with: name/email **search**, **Agreement scenario**, **Organization linking** (`FormSubmission.org_link_status`), and **User account** (same vocabulary as the registrants roster) — all through `FormSubmission.search_by_params`, with the carryover helper threading them through a submission's detail page and back.
- New **Organization** and **User account** columns; the org chip opens the linking editor, the account cell invites inline.

### Org-linking editor for submissions (shared core with the registration editor)
- `GET/POST /form_submissions/:id/{link_organization,select_organization,create_organization}` — the event-less sibling of the registration editor, styled the same and sharing the "Add an organization" block (`shared/_link_organization_actions`) and the affiliation pills partial.
- **Both editors now run one linking core**, `OrganizationServices::LinkSubmittedOrganization`: fill-blank profile/address sync (conflicts flagged not overwritten), affiliation creation (`CreateFromRegistration` — standing Facilitator affiliation only for agreement forms / training events), conflict diff, and the flash notice/warning text. Each flow keeps only its wrapper (registration link row + pins + autofill notes vs. submission metadata link).
- **One five-value linking scenario vocabulary across both flows** (ADR-0002 D2), always derived at call time — from the form's role for a submission, from the event for a registration (never stored as its own column; `events.facilitator_training` stays canonical).

  | Scenario | Confers Facilitator | End-dating |
  |---|---|---|
  | `on_demand` (standalone public registration-role submission, **or** an on-demand facilitator-training event) | yes — dated to submission / event start | none |
  | `reinstatement` (form role) | yes — dated to submission, only when no active one exists | none |
  | `new_job` (form role) | yes — dated to submission; the job affiliation is start-dated too | other orgs' affiliations |
  | `facilitator_training` (scheduled facilitator-training event) | yes — dated to event start | none |
  | `non_facilitator_training` (any other event) | no — job affiliation only | none |
- **Per-scenario affiliation handling at linking time** (`AffiliationServices::ApplyScenarioEndDating`, see **ADR-0002**): a **new job** ends the person's active affiliations at other orgs (job + facilitator) and starts both new ones on the submission date; a **reinstatement** reconciles registration-style — creating affiliations only where no active one exists, ending nothing; **on-demand** ends nothing; event-registration linking never auto-ends. Ends are dated the day before the submission so rows never overlap and the duplicate-skip can't swallow the reinstatement; the panel's manual End button uses the same date. Everything a scenario ends is recorded on the submission and flagged on the panel ("Ended by this agreement" + edit link) so a wrongly-ended row — a multi-org facilitator changing only one job — is visible and correctable.
- "Linked" = an explicit submission→org link (`metadata.linked_organization_ids`) or an active affiliation matching the submitted name. The explicit link is written by the submission editor **and back-applied by the event registration editor** onto the submission its pinning rules pair with the org — so a typo'd name resolved to a differently-named org reads as Linked on the submissions index/panel instead of staying amber forever.

### Dev data
- Seeds create the three agreement forms with submissions (incl. a job-changer with an old active affiliation and a returner with an ended one).

## Notes
- Rebased on main three times; reconciled with #2310's parallel filter bar (main's grid kept as base, this branch's filters folded in) and adapted to the dropped `affiliations.position`/`user_id` columns.
- Auto-processing of affiliations on public submission is deliberately out of scope — staff review each submission before changing anything (per the videos).
- Column show/hide toggles on the submissions index deferred by request ("not yet").
- No screenshots — implemented headless; `ai/seed` + the seeded `/f/collab-agreement-*` forms make the flow easy to eyeball locally.


app/models/form_submission.rb
```diff
@@ -55,9 +55,11 @@ def resolved_event
   end
 
   # Answers keyed by their field's identifier. Bulk payment (and similar) forms
-  # address fields by identifier rather than position.
+  # address fields by identifier rather than position. Reuses an already-loaded
+  # association so per-row calls on a preloaded index don't requery.
   def answers_by_identifier
-    form_answers.includes(:form_field).each_with_object({}) do |answer, map|
+    answers = form_answers.loaded? ? form_answers : form_answers.includes(:form_field)
+    answers.each_with_object({}) do |answer, map|
```
maebeale, line 188: 🤖 _From Claude:_ Behavior nuance: when `form_answers` is already loaded but `form_field` wasn't preloaded, this now lazy-loads each field individually instead of one batched `includes` query. The submissions index preloads both, and other callers hit the unloaded branch (unchanged), but a caller that preloads only `form_answers` would regress to N+1.

app/controllers/form_submissions_controller.rb
```diff
@@ -21,4 +24,189 @@ def show
...
+      organization: organization, overwrite: false, **entry[:address]
+    )
+
+    AffiliationServices::CreateFromRegistration.call(
+      person: @form_submission.person,
+      organization: organization,
+      job_title: entry[:position],
+      training_date: @form_submission.created_at.to_date,
+      organization_address: address_result.address || sole_address(organization),
+      facilitator_training: @form_submission.form.purpose?
```
maebeale: 🤖 _From Claude:_ Load-bearing decision: only agreement-scenario (purposed) forms mint the standing Facilitator affiliation here — linking from any other form's submission creates just the job-title affiliation. Being a facilitator is conferred by training/agreement, not by any org-carrying form (mirrors the `facilitator_training:` rule in the event flow).

maebeale:
@jmilljr24 i'm not wedded to this option. just is one way to handle the different `scenarios` and to make more explicit the affiliation updates being done by org-linking (to event reg on events, and to form submission for non-event-related forms)

jmilljr24 (approved):

app/services/public_form_submission.rb
```diff
@@ -44,6 +45,21 @@ def call
 
   private
 
+  # A standalone registration-role form is the on-demand agreement (ADR-0002),
+  # and submitting it IS registering for the current on-demand facilitator
+  # training — so mint that event registration (idempotent via the unique
+  # registrant+event index) and stamp the event on the submission. A quiet
+  # no-op when no current on-demand training exists.
+  def register_for_on_demand_training(submission)
```
maebeale, line 53: 🤖 _From Claude:_ Two intake paths now converge here: the standalone /f/ on-demand form auto-registers for `Event.current_on_demand_facilitator_training` (most recently started published on-demand training, else next upcoming), while the person-page panel sends the event's public registration form directly. If two on-demand trainings ever overlap, "most recently started" picks the newer one.

PR #2342, opened 2026-08-22
Guard DisplayImagePresenter against undeclared image variants

maebeale:
🤖 suggested review level: 3 Read 📖 small contained guard in one service, covered by specs

`Person#avatar` declares only a `:thumbnail` variant, so rendering a person avatar at full width through the shared `display_image` partial asked ActiveStorage for the undeclared `:card` variant and raised `ArgumentError: Cannot find variant :card` (seen on staging via `people/_show_card`).

Added a `variant_defined?` guard in `DisplayImagePresenter` that checks the attachment's declared named variants before requesting `:card`/`:thumbnail`, degrading to a declared variant or the original file instead of raising. Wrote a failing spec reproducing the crash first, then the fix; the avatar now serves its `:thumbnail` (appropriate for its small display).

### UI Testing Checklist
- [ ] Person show card / list renders avatars without error at all sizes

### Anything else to add?
Fix lives in the generic presenter rather than adding a wasteful large `:card` variant to avatars, so any attachment lacking a variant degrades gracefully.

### DIFF
--- a/app/services/display_image_presenter.rb
+++ b/app/services/display_image_presenter.rb
@@ -89,9 +89,9 @@ class DisplayImagePresenter
       resolve_previewable
     elsif @variant == :hero
       @file
-    elsif large_display?
+    elsif large_display? && variant_defined?(:card)
       @file.variant(:card)
-    elsif use_thumbnail?
+    elsif use_thumbnail? && variant_defined?(:thumbnail)
       @file.variant(:thumbnail)
     else
       @file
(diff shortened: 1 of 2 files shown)


jmilljr24:
What was actually requesting the `:card` size to begin with? This adds the guard which is good but if a view wants a card size image and you give it a thumbnail is going to look bad. Not saying person.avatar needs a card size, just where is this image going to be displayed and how?

maebeale:
@jmilljr24 https://app.honeybadger.io/projects/137790/faults/133771270

jmilljr24:
> @jmilljr24 https://app.honeybadger.io/projects/137790/faults/133771270

I don't have access to honeybadger. Remember we had to drop it down to one account?

Regardless, I'm trying to find out what view made the call to render the persons avatar at the card size. If some code was changed to display the avatar at a card size, there's more to this issue than just guarding against it.

PR #2346, opened 2026-08-22
Move person_id submission-view authorization into EventPolicy

maebeale:
🤖 suggested review level: 3 Read 📖 small authz refactor moving one inline controller check into the policy, no behavior change

Follow-up to #2217, addressing [Justin's review comment](https://github.com/rubyforgood/awbw/pull/2217#discussion_r3792637502). The public registration `show` action gated its `person_id` fallback with an inline exception in the controller (`authorize! @event, to: :form_submissions? unless current_user&.person_id == person.id`) — authorization logic that belongs in the policy.

Added `EventPolicy#person_form_submission?`, which captures the whole rule (event submission managers, or the viewer being that person), and pass the person as authorization context — mirroring the existing `authorize :slug` pattern in `FormSubmissionPolicy`. Behavior is unchanged; policy and request specs pin the manager, self-view, other-person, and signed-out cases.

### Anything else to add?

Purely a refactor prompted by review feedback — no user-facing change.

### DIFF
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -99,9 +99,9 @@ module Events
         # Unlike the registration slug — an unguessable token handed to the
         # registrant — person ids are sequential, so this branch is not a
         # capability. It exists as the admin-side fallback for registrations with
-        # no slug, so gate it on the same access as the event's other submission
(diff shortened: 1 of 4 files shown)


jmilljr24 (approved):
Great!

Issue #1900, opened 2026-06-23
Revert "mailing lists consent"


jmilljr24:
#1885 adding a column for mailing list consent. This raising a number of questions.

E-mail
1. I don't believe this was requested from the stakeholders - Katie asked if they currently had a consent list
2. The consent is applied if they are granted a user account. In order to registered -> attended -> get account the registration form requires consent. The user account is manually created because they consented. Additional code is unnecessary. 

The code suggest this is related to "e-mail" but the naming is confusing because physical mailing addresses are also collected. RE physical mailing address: Stakeholders are not requesting permission to mail as its only used on special occasions.

maebeale:
🤖 _From Claude:_ Related PR #2358 hides the editable **Consented to mailing list** flag from the person edit form (and drops it from the permitted params) while keeping the DB column and consent-capture paths — an incremental step toward the revert discussed here.

maebeale:
🤖 _From Claude:_ Scope of #2358 expanded to a full revert: it now drops the `mailing_list_consent_at`/`_source` columns, the virtual accessor + strong param, and the capture paths in `PublicFormSubmission`/`PublicRegistration`. The person form links to the person's **News** topic subscriptions in place of the old checkbox, so mailing-list interest lives in topic subscriptions going forward (ties into #2079).

jmilljr24:
#2358

PR #2348, opened 2026-08-23
Reword blog-contributor and mailing-list consent hints

maebeale:
🤖 suggested review level: 1 Skim 👀 view-only copy + alignment tweaks to two form hints

## Why
- The admin panel (`lg:w-auto`) is sized to its widest content; the long blog-contributor hint forced it wide and squeezed the adjacent timezone card (`lg:flex-1`). A terser hint lets the blue panel narrow and the timezone card grow back.

## Changes
- Blog-contributor hint: "Adds a blog contributor badge to this person's profile." → "Adds a badge to their profile."
- Mailing-list no-consent hint: "No consent on file. Check to record consent." → "Has agreed through form or email."
- Left-align the blog-contributor column (`sm:flex-1`) to match its mailing-list neighbor instead of right-aligning it.

### DIFF
--- a/app/views/people/_form.html.erb
+++ b/app/views/people/_form.html.erb
@@ -474,7 +474,7 @@
                       label: "Consented to mailing list",
                       hint: f.object.mailing_list_consent_at.present? ?
                               "Consented #{f.object.mailing_list_consent_at.to_date.to_fs(:long)} (#{f.object.mailing_list_consent_source}). Uncheck to withdraw." :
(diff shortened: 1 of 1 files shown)


jmilljr24:
@maebeale this reminded me of #1900 

We 100% early on decided with Rudy that consent was captured by the fact they agree when signing up and attending. I think the consent flag was added when Katie brought up the question and we never finished the conversion and she wasn't part of the original.

PR #2349, opened 2026-08-23
Serve thumbnail variant in tiny people/tagging cards + gate taggings people group (PR #2342 follow-up)

maebeale:
🤖 suggested review level: 3 Read 📖 image-variant param across two card partials + a policy gate on one search group, covered by specs

Follow-up to #2342 (which added the `variant_defined?` guard). Related: #1204 introduced the person `:thumbnail` variant.

## Review comments addressed
| Comment | Resolved by this PR? |
|---|---|
| [#2349 — "why not specify `thumbnail` at the source instead of more fallback?"](https://github.com/rubyforgood/awbw/pull/2349#issuecomment-5385783997) | ✅ Yes — added an explicit `image_variant:` selector; both cards now declare `image_variant: :thumbnail`. [Reply](https://github.com/rubyforgood/awbw/pull/2349#issuecomment-5387897000) |
| [#2342 — "what actually requested `:card`? if a view wants card and gets thumbnail it'll look bad"](https://github.com/rubyforgood/awbw/pull/2342#issuecomment-5381368081) | ✅ Yes — the requesters were `people/_show_card` + `taggings/_tagged_item_card` via `width: "full"`; their boxes are 48–72px, so thumbnail is correct (not a downgrade). |
| [#2342 — "what view rendered the avatar at card size? there's more here than just guarding"](https://github.com/rubyforgood/awbw/pull/2342#issuecomment-5381567487) | ✅ Yes — #2342 guarded the symptom; this fixes the source by not requesting `:card` in tiny boxes. |

## Image sizing
- `people/_show_card` (~48–64px box) and `taggings/_tagged_item_card` (~72px box) passed `width: "full"`, so `DisplayImagePresenter` treated them as a large display and fetched the **1200×1200 `:card`** variant into a tiny box.
- The presenter's `variant:` is a *layout* mode (`:hero`/`:gallery`/`:index`), not the image variant — size was inferred from `width`. Added an explicit **`image_variant:`** selector so a caller names the variant at the source.
- Both cards now pass `image_variant: :thumbnail` → **256px** across the board. The `large_display?`/`use_thumbnail?` inference only runs when size is unspecified. Guarded by `variant_defined?` (handles missing-png strings / PDFs). Views that genuinely want card (workshop banner, hero) are untouched.

## Public exposure of person avatar + name
- The public `taggings#index` rendered a **people group** (avatar + name) to logged-out visitors: `TaggingPolicy#index?` is public and `PersonPolicy`'s `relation_scope` exposes searchable people (that scope is for the pending public-profile feature, not this browse page).
- Now the people group renders **only for viewers allowed to see the People index** (admins); everyone else gets an empty group, mirroring how `grants` already behaves here. `PersonPolicy#relation_scope` is left untouched.
- Event "meet the staff" pages (`events#staff`) stay intentionally public — unchanged.

## Cleanup
- Dropped the **shadowed** first `PersonDecorator#default_display_image` (a second definition of the same method — returning the avatar — always won, so the first was dead code).

### DIFF
--- a/app/decorators/person_decorator.rb
+++ b/app/decorators/person_decorator.rb
@@ -8,10 +8,6 @@ class PersonDecorator < ApplicationDecorator
     length ? text&.truncate(length) : text
   end
 
-  def default_display_image
-    "missing.png"
-  end
-
   def primary_asset
     avatar
   end
--- a/app/services/display_image_presenter.rb
+++ b/app/services/display_image_presenter.rb
@@ -22,7 +22,7 @@ class DisplayImagePresenter
   end
 
   def initialize(resource: nil, item: nil, file: nil, field_name: :primary_asset,
-                 variant: :gallery, width: "32", height: nil,
+                 variant: :gallery, width: "32", height: nil, prefer_thumbnail: false,
                  link: nil, link_to_object: false, display_open_pdf_link: false,
                  idx: 0, item_type: "PrimaryAsset", extra_image_classes: "",
                  view_context:)
@@ -32,6 +32,7 @@ class DisplayImagePresenter
     @variant = variant
     @width = width.to_s
     @height = (height || width).to_s
+    @prefer_thumbnail = prefer_thumbnail
     @link_to_object = link_to_object
     @link = link.nil? ? link_to_object : link
     @display_open_pdf_link = display_open_pdf_link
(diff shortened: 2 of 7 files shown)


jmilljr24:
"people/_show_card (~48–64px box) and taggings/_tagged_item_card (~72px box) passed width: "full", so DisplayImagePresenter treated them as a large display and fetched the 1200×1200 :card variant into a tiny box."

Why are we not specifying `thumbnail` as the variant in `tagged_item_card`? This change again just as more failback instead of starting at the source. Maybe I'm missing something.

maebeale:
🤖 _From Claude:_ Good call. The snag was that the presenter's `variant:` param is a *layout* mode (`:hero`/`:gallery`/`:index`) that only drives CSS — image size was inferred from `width` — so `variant: :thumbnail` was a no-op. Added an explicit `image_variant:` selector so the card declares `image_variant: :thumbnail` at the source; the `large_display?`/`use_thumbnail?` inference now only runs when size is left unspecified. (Still guarded by `variant_defined?` since `display_image` can be a missing-png string or a PDF.) Pushed in 75be975.

jmilljr24 (approved):

Issue #2251, opened 2026-08-18
PaperTrail timelines: subject-tagging architecture for person/org/event/registration/scholarship/CE/license/membership


maebeale:
Umbrella for the PaperTrail-backed **timeline** feature. Tracks the architecture; #2250 is a child (changeset normalization + actor rendering).

## Goal

Give nonprofit admins per-subject timelines ("what happened, when, by whom") built from PaperTrail `versions`. We already version the financial/membership/registration/user models and stamp `subject_person_id` via meta.

## The core insight: timelines are a graph, not a list

One version usually belongs to **several** timelines at once — a scholarship payment is on the **person**, **scholarship**, and **event** timelines; an `event_registration` is the root of the registration timeline *and* a member of person/event; a payment surfaces under person, event, scholarship, membership, and CE. So the real problem is **tagging each version with the *set* of subjects it rolls up to**, not adding one key per timeline.

## Mechanism decision: polymorphic subject-tag table

Two forces rule out both naive options:

- **Read-time association traversal** (union the aggregate's descendants at query time) blows up for **deep/wide aggregates**: to render page 1 of an **event** you must enumerate the entire descendant id set in Ruby (registrations → payments → checklists → attendance), build an unbounded `IN (...)`, and sort without an index. Unbounded memory, no DB pagination.
- **A `subject_*_id` column per timeline** means ~8 nullable columns + 8 indexes + a backfill migration each, on the fastest-growing table — and every *shared* record must compute and store *all* roots it belongs to (combinatorial).

**Decision — one polymorphic tag table:**

```
version_subjects(version_id, subject_type, subject_id)
  index (subject_type, subject_id, version_id)
```

- On version create, a hook asks the versioned record "which subjects do you roll up to?" and inserts one tag row per subject. (Same subject computation we already do for `subject_person_id` meta — now a *set*, to a table.)
- Any timeline = one indexed join with `LIMIT`: no descendant traversal, no giant `IN`, DB-paginated, bounded memory.
- New timeline = zero schema change; teach the subject-computation the new root.
- Cost: an `after_create` on the version model + a few association lookups per write + a handful of extra rows per version. Bounded write-time work replacing unbounded read-time traversal — right trade for a read-heavy feature.
- **Migration:** fold the existing `subject_person_id` meta into this table (backfill person tags). Early enough in the branch to unify now rather than accrete columns.

## Timelines to support (each = a subject tag)

**Requested**
- **Person** — cross-cutting.
[...]

PR #2354, opened 2026-08-23
Reach the person history through records that hang off a person

maebeale:
🤖 suggested review level: 3 Read 📖 contained: ten resource types added to one query service, no schema change

Part of #2251.

## Reach, not tracking

`AhoyTrackable` already covers every model, so a person's history isn't missing events — it's missing **routes to them**. `Analytics::PersonActivityEvents` maps `person → ids` per `resource_type`, and records that connect to a person only through a parent were absent from that map.

Nothing needs backfilling: the mapping is read-time, so it covers everything already recorded.

## Ten types added

- **Registration children** — `EventRegistrationChecklistCompletion`, `EventRegistrationOrganization`, `EventAttendanceTimeEntry`
- **Staffing** — `EventStaff`
- **Money** — `Allocation` (through registration / CE registration / membership invoice), `Refund` (recipient, or reversing one of their payments), `MembershipInvoice`
- **Forms** — `FormAnswer`, plus the `Asset` uploads those answers own
- **Scholarships** — `ScholarshipAgreementResponse`

Repeated subqueries (registration / submission / payment / scholarship ids) are memoized rather than rebuilt inline.

## How to test

1. `ai/server`, sign in as an admin.
2. Act on something that reaches a person through a registration — add a payment/allocation, tick an onboarding checklist step, upload a form answer file.
3. Open `/people/:id/edit` → *Associated records* → **History**, or `/admin/activities/events?person_id=:id&time_period=all_time&audience[]=visitors&audience[]=users&audience[]=staff`.
4. Look at the **Activity** column for `create.allocation`, `update.event_registration_checklist_completion`, `create.form_answer`, `create.asset`. On `main` none of those can reach a person's history no matter how many exist.

## Scope: all ten, decided

Keeping the full set, including the mechanical ones (`create.allocation`, `update.event_registration_checklist_completion`, `update.event_attendance_time_entry`). The history is meant to be complete; filtering for readability belongs in the UI, not in what reaches the page.

## Note on the approach

This reaches the records at read time — the relation is an OR of ~33 subqueries, bounded per person. #2251 sketches a denormalized tag table instead; that's a different mechanism for a wider problem (event and organization timelines), and not a blocker here.

## Test guard

- Added a spec asserting every `resource_type` key in the map resolves to a real `ApplicationRecord` — most mapped types have no inclusion example, so a rename/typo would otherwise drop that type from the history silently.

### DIFF
--- a/app/services/analytics/person_activity_events.rb
+++ b/app/services/analytics/person_activity_events.rb
@@ -2,8 +2,10 @@ module Analytics
   # Collects every Ahoy event that belongs to a person's history: events about
   # the person record itself, about their user account (both lifecycle and
   # `auth.*` events), and about the associated records surfaced on the person's
-  # "Associated records" panel. Powers the person edit "History" card and the
-  # `person_id` filter on the admin Ahoy activities index.
+  # "Associated records" panel — plus the records that only reach a person
+  # through a parent (registration children, the money ledger, form answers and
+  # their uploads, agreement responses). Powers the person edit "History" card
+  # and the `person_id` filter on the admin Ahoy activities index.
   class PersonActivityEvents
     def initialize(person)
       @person = person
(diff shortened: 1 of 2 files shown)


maebeale:
🤖 _From Claude:_ Scope decided — keeping all ten resource types rather than trimming to the money/forms subset. Completeness over scannability; filtering belongs in the UI. The read-time traversal question against #2251 is still open.

PR #2364, opened 2026-08-23
Role-agnostic form sections + de-dup nested-form questions

maebeale:
> **NOTE — parking the follow-ups.** Review surfaced two known limitations (see the [review comment](https://github.com/rubyforgood/awbw/pull/2364#issuecomment-5389291161)): the submission page hides answers captured on a nested form for a de-duplicated question, and the CE fallback can't fire when only the registration form asks. We're waiting until users actually hit these before reworking — merging as-is.

🤖 suggested review level: 5 Inspect 🔬 substantive logic in public-registration rendering/validation and the form section builder

Two form improvements: a copied single-section form (e.g. scholarship) can now be built out with any section, and a scholarship/CE form nested in an event's registration form no longer re-asks questions the registration form already covers — so registrants aren't asked their name/email/org twice.

- **Edit sections:** `FormBuilderService.editable_sections` now offers every built-in section regardless of the form's `role` (present sections stay checked); removed the now-unused role gate.
- **Nested de-dup:** `PublicRegistrations` renders a nested form through `nested_visible_fields`, dropping any field whose identifier (aliases included) already appears on the registration form, plus emptied section headers. Validation matches, so a suppressed-but-required duplicate can't block submission. Standalone `/f/:slug` forms are untouched and still show everything.

### Where the de-dup needed a second read
Both fixed here, each with a failing-first spec.

- **CE opt-in.** A registration form carrying the built-in CE section shares every identifier with a nested CE form, so the nested copy collapses to nothing. `PublicRegistration` read the opt-in and license details only from the CE form's params, so answering "Yes" created no `ContinuingEducationRegistration`. `ce_field_value` now falls back to the registration form's answer — nested copy wins when it's the one shown.
- **Empty cards.** Same collapse left the submission page's scholarship/CE cards with a heading and "Submitted on …" and no questions. Both are now gated on having fields, matching the public form.

### UI Testing Checklist
- [ ] Copy a scholarship form → Edit sections offers all sections.
- [ ] Attach a scholarship/CE form (with person/org fields) to an event → those questions don't duplicate on the public registration page.
- [ ] Registration form with the CE section + a nested CE form → answering "Yes" still issues CE credit.
- [ ] Open the same form standalone → all fields show.

### Anything else to add?
Person/org data is always captured from the registration form, so suppressing the nested copies changes no persisted data. Added `config/features.yml` entries for both.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -474,7 +474,15 @@ module EventRegistrationServices
       ce_field_value(CE_LICENSE_EXPIRES_ON_IDENTIFIER).presence
     end
 
+    # A CE question can be asked on the nested CE form, on the registration form's
+    # built-in continuing_education section, or both — and the nested copy is
+    # suppressed as a duplicate when both carry it — so read whichever copy the
+    # registrant was actually shown.
     def ce_field_value(key)
+      nested_ce_value(key).presence || field_value(key)
+    end
+
+    def nested_ce_value(key)
       field = @continuing_education_form.form_fields.find_by(field_identifier: key)
       return nil unless field
 
--- a/app/controllers/events/public_registrations_controller.rb
+++ b/app/controllers/events/public_registrations_controller.rb
@@ -17,6 +17,7 @@ module Events
       @scholarship = scholarship_mode?
       @scholarship_form = @event.scholarship_form if @scholarship
       @continuing_education_form = @event.continuing_education_form
+      assign_nested_form_fields
       @event = @event.decorate
     end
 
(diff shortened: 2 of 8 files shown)


app/services/event_registration_services/public_registration.rb
```diff
@@ -474,7 +474,15 @@ def ce_license_expires_on
       ce_field_value(CE_LICENSE_EXPIRES_ON_IDENTIFIER).presence
     end
 
+    # A CE question can be asked on the nested CE form, on the registration form's
+    # built-in continuing_education section, or both — and the nested copy is
+    # suppressed as a duplicate when both carry it — so read whichever copy the
+    # registrant was actually shown.
     def ce_field_value(key)
```
maebeale, line 481: 🤖 _From Claude:_ Precedence matters here: the nested CE form wins when it actually rendered the question, and we fall back to the registration form only when the nested copy was suppressed as a duplicate (which is all-or-nothing — both forms come from the same builder, so a reg form with the CE section collapses every nested CE field). Without the fallback, that config silently created no `ContinuingEducationRegistration`.

maebeale:
🤖 _From Claude:_ Review notes on the nested de-dup half. Not blocking — recording them so they aren't lost if we revisit.

**Two behaviors worth knowing about**

- **Read-side suppression hides answers that were already captured.** `load_scholarship_submission` / `load_continuing_education_submission` filter through `nested_visible_fields`, so an answer stored on the nested submission for a duplicated question stops rendering — and the registration card's twin renders blank, since `@responses` only covers the registration submission. Verified with the fixture at `spec/requests/events/public_registrations_spec.rb:882`: the question label renders, the stored answer text is absent from the body. The registrants roster still links those registrants to `#scholarship-application`, an anchor the page no longer emits.
- **The new CE fallback can't fire when only the registration form asks.** `ce_credit_requested?` and each `ce_license_*` reader open with `return … unless @continuing_education_form`, so `ce_field_value`'s fallback is unreachable in that case. Reproduced: registration form built from `sections: %i[person_identifier continuing_education]`, no nested CE form, answer "Yes" → registration created, `ContinuingEducationRegistration.count` stays 0.

**Suggested shape if we rework**

- Keep the de-dup on the input side (`new`/`create`) only. On `show`, render every field and resolve answers by identifier across submissions the way `ScholarshipApplication#answer_for` already does — that also retires the `@scholarship_fields.any?` card gates.
- Extract `nested_visible_fields` + `registration_field_identifiers` into a PORO that both the controller and `PublicRegistration` consume, so the service knows which copy was shown instead of guessing via `nested_ce_value || field_value`.
- Consider surfacing full collapse to admins (a nested form contributing zero questions for an event is a misconfiguration, not something to hide) — `BuiltinCalloutCards` already emits messages of this kind.

**One strategic note:** the Edit sections half partly obsoletes the de-dup, since it's what lets an admin trim duplicate sections out of a copied nested form. The de-dup earns its keep mainly when one nested form is shared across events with differing registration forms.

PR #2365, opened 2026-08-23
Credit and search the submitter when content names no author

maebeale:
🤖 suggested review level: 5 Inspect 🔬 changes what `AuthorCreditable` credits (display + search + sort + role filters) across 6 models

Follows up #2312 / #2335. Routes author search through the credited author — and makes an author-less record credit its submitter, since `author_id` shipped nullable with no backfill on most of these models.

## Why
- Pointing author search and the **Workshop log authors** role filter at `author_id` alone matched nothing for every log on file.
- `credited_to_person` (what the profile lists) already falls back to `created_by`, so the two disagreed — a profile showed logs the person wasn't findable by.
- Workshop variations and monthly reports had the fallback for *display* only, via a `legacy_author_name_text` override that SQL can't reach — so their author search and role filter missed author-less rows.
- Some lists displayed something other than the credit they filter on.

### `credits_creator` (AuthorCreditable)
- Opt-in per model: an author-less record credits the submitting account's person, through **that person's** `display_name_preference`, suppressed by their anonymity latch.
- On for `WorkshopLog`, `WorkshopIdea`, `StoryIdea`, `WorkshopVariationIdea`, `WorkshopVariation`, `MonthlyReport`. Off elsewhere — on `Story`/`Workshop`/`Resource` the creator is an entry clerk, not the author.
- One seam drives display (`author_credit`, `author_credit_person`), search (a `creator_person` join alias), and sort (`order_by_author` coalesce).
- The creator clause is gated on `author_id IS NULL`, so a named author always outranks the submitter.
- `WorkshopVariation` / `MonthlyReport` drop their duplicate `legacy_author_name_text` overrides.

### Search by credited author
- **Workshop log** / **Monthly report** — `search` gains `author_name`.
- **Workshop idea** — `.author_name` searches the credited author instead of joining `created_by`'s user/person + email.
- **Story idea** / **Workshop variation idea** — `search_by_params` gains `author_name`.
- **Workshop variation** — its existing `author_name` box now reaches author-less rows.

### Person role filters
- New `Person.credited_on(model)` derives a role filter from what the model credits: named authors, plus the submitter where `credits_creator` is on.
- `workshop_authors`, `workshop_variation_authors`, `workshop_log_authors` all go through it.

### UI
- "Author name" box on the story-idea, workshop-variation-idea, workshop-log and monthly-report index search forms (workshop idea and workshop variation already had one).
- Story ideas index "Author" column used `created_by.name` raw, bypassing credit/anonymity rules → now `credited_author_link`.
- Workshop logs index "Person" column and monthly reports index "Submitter" column → **Author** (credited author, falling back to the submitter); both `created_by_id` filters are relabeled **Submitted by** so the two read distinctly.
- Workshop log and monthly report show pages gain "Author credit:", matching the idea show pages.
- The author-picker note explains a creator-based credit instead of going silent on it.
- `:author` added to the story-idea, workshop-log and monthly-report `includes` to keep those columns off an N+1.

### Docs
- `/features` entry; `AuthorCreditable` row in AGENTS.md.

## Tests
- `author_creditable` shared example gains a `credits_creator:` mode: creator credit, their profile preference, anonymity suppression, and "a named author outranks the creator" in search. `credits_creator_legacy:` is replaced by `anonymous_when_unattributed:`, which is what it actually controlled.
- Model specs for each `author_name` path incl. the creator fallback; `person_spec` covers the log and variation role filters' fallback and their no-person case; `monthly_report_spec` covers its new `.search` filter.
- Divergence query: an author-less variation now surfaces in the creator section, where an admin can assign the author.

### DIFF
--- a/app/models/concerns/author_creditable.rb
+++ b/app/models/concerns/author_creditable.rb
@@ -194,8 +194,7 @@ module AuthorCreditable
 
     private
 
-    # Only an explicit author is ever credited, so search and sort reach that person
-    # alone — and nobody at all on the idea models, which have no author_id.
+    # Only an explicit author is ever credited, so search and sort reach that person alone.
     def credited_person_aliases
       explicit_author? ? [ "credited_author" ] : []
     end
--- a/app/models/concerns/author_creditable.rb
+++ b/app/models/concerns/author_creditable.rb
@@ -9,6 +9,10 @@ module AuthorCreditable
 
   ANONYMOUS = "anonymous"
 
+  # Join aliases for the people a record credits, in display precedence order.
+  CREDITED_AUTHOR_ALIAS = "credited_author".freeze
+  CREATOR_PERSON_ALIAS = "creator_person".freeze
+
   # The generic credit AWBW puts on content with no named credit. Content AWBW produces
   # itself — news, workshops, resources — reads as "AWBW Staff"; everything else,
   # including facilitator submissions and the idea forms, reads as "AWBW Facilitator".
@@ -41,6 +45,11 @@ module AuthorCreditable
     # content follows the facilitator label; org-produced models flip it to the org
     # label with `credits_to_org`.
     class_attribute :unattributed_author_label, instance_writer: false, default: FACILITATOR_AUTHOR_LABEL
+
+    # Whether the submitting account's person is credited when no explicit author is
+    # on file. Off by default — on most models the creator entered the content rather
+    # than writing it. Turn it on with `credits_creator`.
+    class_attribute :creator_credited, instance_writer: false, default: false
 
     before_create :snapshot_author_credit_preference
     # Blank means "follow the profile"; nil keeps it off the divergence worklist.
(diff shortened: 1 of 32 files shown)


app/models/concerns/author_creditable.rb
```diff
@@ -162,6 +183,14 @@ def credits_to_org
       self.unattributed_author_label = ORG_AUTHOR_LABEL
     end
 
+    # Content whose submitter is its author — the idea forms and workshop logs. An
+    # author-less row (anything predating the author column) then credits, links,
+    # searches and sorts under the creating account's person, through that person's
+    # own credit preference.
+    def credits_creator
```
maebeale, line 190: 🤖 _From Claude:_ Deliberately not turned on for `WorkshopVariation` and `MonthlyReport` — they do the same creator fallback through a `legacy_author_name_text` override, which display honors but `by_credited_person_name` can't reach, so their author search (and `Person.workshop_variation_authors`) still misses author-less rows. Migrating them onto this seam would unify both; left out of scope here.

PR #2377, opened 2026-08-25
Add spreadsheet importer for event registrants

maebeale:
🤖 suggested review level: 5 Inspect 🔬 bulk data creation — people, registrations, form submissions, org links, and facilitator affiliations

## What
Admin-only importer that bulk-adds attended registrants to an event from a **CSV**, **simulating a real registration** so the existing reconciliation flow takes over.

### Per row (`EventRegistrationImporter`)
- **Person** — matched on `PublicRegistration`'s rule (email + last name, tolerant of a first/legal-name swap); created (attributed to the importing admin) only on a miss. No duplicates.
- **EventRegistration** — found or created, forced to `attended` (a real reg would be `registered`; these people attended). Existing non-attended reg is promoted.
- **FormSubmission (role registration) + FormAnswers** for the row's name/email/organization — the import's **own** submission, marked `imported_from` in metadata and created *alongside* (never in place of) a real registrant's. This is what makes an **unmatched** org appear in the admin **link organization** reconciliation queue, identically to a live reg. Idempotent: re-running reuses that marked submission.
- **Matched org** → linked (`EventRegistrationOrganization`, pinned to the import submission) and, on a **facilitator-training** event, a `Facilitator` affiliation via `AffiliationServices::CreateFromRegistration`.
- **No emails**, never creates organizations, no tags/addresses/CE (not in the sheet).

### Flow
`new` (pick event + upload CSV) → `create` (dry-run preview, nothing written) → `confirm` (real import). File stashed as an ActiveStorage blob between steps, mirroring `StoryImports`. Blocks events without a registration form + org field (`.importable?`).

## Why
Staff have rosters of training attendees (e.g. "2026 On-Demand Trainees") and need them in the portal as attended registrations — matched to existing people, organizations reconciled the same way as normal registrations — without hand-entry, duplicates, or emailing historical attendees.

## Notes
- CSV only, parsed with the stdlib `csv` (no third-party parser).
- Gated by `EventRegistrationPolicy#import?` (admins only). Idempotent: re-running the same file creates no duplicate submissions, links, or affiliations.


app/controllers/event_registration_imports_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    return render_new_error("Choose a spreadsheet file to import.") if file.blank?
+    return render_new_error("That file must be a .xlsx, .csv, or .xls.") unless supported?(file)
+
+    @filename = file.original_filename
+    @extension = extension_for(file)
+    begin
+      @blob = ActiveStorage::Blob.create_and_upload!(
+        io: file.open, filename: @filename,
+        content_type: file.content_type.presence || "application/octet-stream"
+      )
```
jmilljr24: I think this is just duplicating what active storage does under the hood.
```suggestion
```

jmilljr24 (changes requested):
I don't think we should be relying on or passing around the file extension. We should be able to get the file type from active storage `blob.content_type` which is actually checking the file contents instead of using an extension which is easily changed on a file by a user (intentional or not)

If the content_type isn't available yet from the blob or you need the type somewhere else, active storage uses `Marcel::MimeType` to generate the content_type.

app/services/event_registration_importer.rb
```diff
@@ -0,0 +1,230 @@
...
+    "name" => :last_name,
+    "organization" => :organization,
+    "organisation" => :organization,
+    "org" => :organization,
+    "email" => :email,
+    "e-mail" => :email,
+    "email address" => :email
+  }.freeze
+
+  SUPPORTED_EXTENSIONS = %w[xlsx csv xls].freeze
```
jmilljr24:
```suggestion
  SUPPORTED_EXTENSIONS = %w[xlsx csv].freeze
```
I think we need the addition `roo-xls` gem for `xls` to be supported

jmilljr24 (approved):

PR #2384, opened 2026-08-27
Deliver a form inline on a callout card

maebeale:
🤖 suggested review level: 5 Inspect 🔬 inline form-delivery flow + a new registrants-index column set — a person-first submit page, a submission service, a generalized shared form partial, and per-callout completion columns

Lets staff attach a form to any registration-ticket callout, deliver it **inline** to the registrant (they fill it out from their ticket and see their responses on return), and **track completion on the registrants index**. Pulled out of #2135 (post-event surveys) as the generic "form on a callout" mechanism — none of the survey-specific readiness/roles/profile-sync.

## What
- **Attach a form** to any callout via an "Inline form" picker in the callout editor (`form_id` FK). Offers all standalone forms.
- **Person-first form page** — `registration/:slug/forms/:callout_id` (`Events::CalloutsController#callout`), alongside the other registrant callout pages. Reuses the `_form_field` partial (generalized with an optional `param_scope`) for full field-type support.
- **Submit + store** — `submit_callout` records answers via `EventRegistrationServices::CalloutFormSubmission`: one `FormSubmission` per `(registrant, form, event)`; re-submitting edits in place. After submitting, the page shows the answers read-only with an "Edit responses" affordance.
- **Registrants-index completion columns** — one column per form-bearing callout, revealed together by a single **"Ticket forms"** toggle (off by default, only shown when the event has form callouts). Each cell is a dated check linking to that registrant's responses, or a dash. One preloaded query.

## Design notes
- **Keyed by the callout, not a form slug** — forms only carry a slug when published (callout forms often aren't), and one form can back several callouts, so the callout is the unambiguous anchor.
- **Submission role mirrors the form's own role** (`.role_for`) so a callout submission looks like any other submission of that form (uniform stats/lookups; #2135's readiness can find them). Falls back to `"callout"` only for a role-less or reserved-role form. The index columns look up completion by the callout's `(form, event, role)` tuple, so they work either way.

## Not here
- Survey-specific behavior from #2135 (readiness gating, profile write-through, per-resource "clarity" fan-out, staff FYI email) — targeted follow-ups.

## Testing
- Service spec (records, edits in place, role logic), request specs (renders for a registrant, no-form redirect, submit stores + shows answers, event-scoped preview note, registrants-index completion column), model spec (`page_content?` counts a form).
- Existing form-partial callers (public registration, public forms) reverified green. Rubocop clean.


jmilljr24 (changes requested):
Linked resource do not display if the callout includes a form.

maebeale:
@jmilljr24 yeah i thought that was ok but yeah better to make this work now. done!

jmilljr24 (approved):

PR #2397, opened 2026-08-28
Retry public registration on InnoDB deadlock

maebeale:
🤖 suggested review level: 3 Read 📖 contained retry/error-handling change in one service, covered by specs

Retries the public event-registration transaction when two concurrent submissions from the same registrant deadlock in InnoDB — so a double-click or browser retry no longer 500s a would-be registrant.

A Honeybadger `ActiveRecord::Deadlocked` fired from `PublicRegistration` on this unauthenticated form. Concurrent submissions from the same registrant contend on one large multi-table transaction and deadlock, costing us the registration.

- Extracted the transaction into a private `run` and wrapped `call` in a bounded deadlock retry (`MAX_DEADLOCK_RETRIES = 2`); a deadlock fully rolls back the loser, so the retry re-runs and takes the idempotent "existing" path.
- On exhausting retries, returns a friendly failure Result instead of a raw 500.

### Anything else to add?
Trade-off: the exhausted-retry case is swallowed into a friendly Result, so a *persistent* deadlock (surviving 3 attempts) no longer reaches Honeybadger — happy to re-raise instead if you'd rather keep that signal.

### DIFF
--- a/app/services/event_registration_services/public_registration.rb
+++ b/app/services/event_registration_services/public_registration.rb
@@ -2,6 +2,13 @@ module EventRegistrationServices
   class PublicRegistration
     Result = Struct.new(:success?, :event_registration, :form_submission, :errors, keyword_init: true)
 
+    # This public form runs one large, multi-table transaction. Two concurrent
+    # submissions from the same registrant (a double-click or browser retry) can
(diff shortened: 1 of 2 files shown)


maebeale:
@jmilljr24 maybe we ignore this, but...

PR #2427, opened 2026-08-30
Convert all buttons to button_classes helper; remove buttons.css

maebeale:
🤖 suggested review level: 5 Inspect 🔬 wide sweep across ~200 view files plus deletion of the shared button stylesheet — mechanical, but worth a scan for missed or restyled buttons

Finishes the migration #2297 started: every button now gets its classes from `ButtonHelper#button_classes` instead of the old `.btn*` `@apply` component classes, and `buttons.css` is deleted. One Ruby source of truth for button styling; no more `@apply` indirection.

## What changed
- **~200 view files + 3 helpers + scaffold generator templates** converted from `class="btn btn-primary …"` to `button_classes(:primary, …)`.
- **`buttons.css` deleted** and its `@import` removed. Verified the compiled CSS still emits the button color utilities (`bg-primary`, etc.) and no longer emits any `.btn*` rule.
- **Helper gained `variant: nil`** — base (+ size) with no color, for the handful of custom-color buttons that still want `BUTTON_BASE` as the single source of truth (e.g. the lock/unlock toggle's dynamic orange).

## Worth a look
- **Legacy report toggles** (`monthly_reports/*`, `reports/*`, `shared/_sortable_fields_btn`) keep a bare `btn` + `active` class as a **jQuery hook** (`$('.btn-group .btn.active')`), with styling moved to the helper. Their undefined decorative classes (`btn-blue`, `btn--small`, `btn-fs-medium`, …) were dead CSS and are dropped.
- **`btn-secondary-default`** count chips (staff_tags / organization_statuses / category_types index) were base-only with no color; converted faithfully to `button_classes(nil)`. If they were meant to be colored, that's a easy follow-up — flag it.
- Class order in touched files was re-sorted by `tw-sort` (project convention), which accounts for some of the diff noise.

## Verification
- All changed ERB templates parse; RuboCop clean on changed Ruby; `button_helper_spec` passes (added a `variant: nil` case); `vite build` succeeds.


app/helpers/button_helper.rb
```diff
@@ -36,8 +36,12 @@ module ButtonHelper
     utility_outline: "border border-gray-200 text-gray-600 hover:bg-gray-200 hover:text-gray-800"
   }.freeze
 
+  # `variant: nil` yields the shared base (+ size) with no color, for the few
+  # call sites that supply their own colors (e.g. a dynamic/custom fill) yet
+  # still want BUTTON_BASE as the single source of truth rather than inlining it.
```
jmilljr24: Good comment. I would have missed understanding the option for `variant = nil`.

Is that the same case for size or can `if size` be removed if size is always present

jmilljr24 (approved):

app/helpers/button_helper.rb
maebeale: same case, so I updated comments accordingly

without the guard, size: nil would hit BUTTON_SIZES.fetch(nil) and raise KeyError

PR #2430, opened 2026-08-30
Resolve resources' legacy agency name to an organization link

maebeale:
🤖 suggested review level: 3 Read 📖 one committed data task + spec; no schema/app-behavior change, run by hand

Converts the free-text `resources.agency` into a real `organization_id` link, so the legacy column can be dropped in a follow-up. The `organization_id` column + optional `belongs_to :organization` already shipped (#2421); this fills it in.

## What it does
`bin/rails data:resolve_resource_organizations` (dry-run by default; `DRY_RUN=false` to write). For each resource with an `agency` name but no org linked:
- **exact, case-insensitive name match** → link the existing org
- **no match** → create the org (status "Unknown", since org status is required but legacy) and link it
- **more than one match** → ambiguous, skip and report for hand reconciliation

Idempotent — only touches rows where `organization_id IS NULL`, so it's safe to re-run and safe to run while the `agency` column still exists.

## Notes
- **Not a data migration** — matches this repo's other one-off conversions (`data:migrate_sectors`, `data:convert_age_ranges`): a committed, reviewable, hand-run task with a dry run. Promotes the ad-hoc `.context` script into version control.
- **Follow-up:** once this has run in prod, a separate PR drops the `agency` column and removes the two dev-seed `agency:` writes.
- Spec covers link / create / dedupe-within-run / ambiguous-skip / already-linked / dry-run.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -532,7 +532,7 @@ RuboCop linting on PRs and pushes to main.
 
 ## Rake Tasks
 
-Located in `lib/tasks/` (12 files):
+Located in `lib/tasks/` (13 files):
 - `dev.rake` — Development database seeding from XML/CSV
 - `rhino_migrator.rake` — Rich text editor migration
 - `attachment_report.rake` — Attachment reporting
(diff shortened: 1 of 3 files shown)


maebeale:
@jmilljr24 we can just run this and not have to commit the rake

PR #2442, opened 2026-08-31
Selectable display templates for the event show page

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new per-event template system: schema + model + show-page refactor into shared partials + admin gallery, plus a widened `show?` rule, across ~30 files

Lets an admin or event owner pick a **display template** per event so the public event page can match the new AWBW brand (navy/cream/gold, Fraunces) — without hand-editing views. Facilitators get on-brand event pages, and every existing event keeps today's look (default `none`).

## What you get
- **Layout template** picker on the event form (Event show page section), with a **Preview all templates** link.
- A read-only **gallery** (`/events/templates_gallery`) rendering all templates with a real event.
- Five templates: **none** (current), **centered**, **editorial**, **sidebar**, **hero**.

## How it works
- The show page is refactored into shared content partials under `events/show/` (header, title, details, actions, description) that **centralize the `autoshow_*` show/hide logic once**. Each template in `events/templates/` just frames those pieces.
- Same content, same toggles, same `?preview` unpublished-preview flow — only the frame changes.
- The show page now runs **full-width** so templates control their own measure and can run backgrounds edge-to-edge.

## Notes for review
- **`show?` now allows `owner?`.** Owners previously could not view their own unpublished or past event page at all, so the gallery's live-preview links were dead for drafts. This widens public-page access for owners only, on their own events. `page_bg_class` marker moved to `admin-or-owner-or-public-or-authpublished` to match.
- **`templates_gallery?` is its own rule, not an alias of `new?`.** The action authorizes without a record, so `owner?` has no Event to check and any `manage?`-derived rule collapses to admin-only — locking out the owners who reach it from the form. Sample event comes from the `:reportable` scope, so an owner only previews their own.
- `none` is a faithful copy of the current layout (frozen legacy frame), **including its orange/green buttons** — CTA styling is read off `event.template` so a Turbo re-render of the registration section matches the page it lands in.
- Gallery renders in `sample: true` mode so it never builds live registration routes or duplicate `registration_section` dom_ids across the stacked templates.
- `template` column: `default "none", null: false`; constant + inclusion validation; permitted via `EventPolicy`.

## Tests
- Model: default + inclusion validation. Decorator: CTA styling per template.
- Policy: owner sees their own draft/past event.
- Request: every template renders on `/show`; gallery allows admins and owners, scopes an owner to their own events, forbids a user who owns none.
- Added the gallery view to `page_bg_class_alignment_spec`.

Screenshots to follow.

### DIFF
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -188,6 +188,7 @@ class EventPolicy < ApplicationPolicy
                   :public_registration_enabled,
                   :signed_in_one_click_enabled,
                   :autoshow_registration_details,
+                  :template,
                   :hint_dates,
                   :hint_times,
                   :hint_registration_cost,
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -23,6 +23,16 @@ class EventPolicy < ApplicationPolicy
     admin? || owns_events?
   end
 
+  # The display-template gallery, reached from the event form's layout picker.
+  # Its own rule rather than an alias of new?: the controller authorizes without a
+  # record, so owner? has no Event to check and any manage?-derived rule would
+  # collapse to admin-only — locking out the event owners who reach it from the
+  # form. Gated on "has events to preview", with the :reportable scope narrowing
+  # the sample event to their own.
+  def templates_gallery?
+    admin? || owns_events?
+  end
+
   # A single-event slice of those reports — reached from the per-event Reports and
   # Roster tabs, which pass event_id — is visible to that event's owner too. The
   # controller resolves event_id to the Event and authorizes against it here, so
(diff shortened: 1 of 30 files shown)


app/policies/event_policy.rb
```diff
@@ -42,8 +52,11 @@ def event_reports?
     relation.where(created_by_id: user.id)
   end
 
+  # Owners see their own event's page unconditionally — same as an admin — so they
+  # can preview a draft (unpublished, not yet publicly visible) or a past event
+  # before anyone else can, which is what the layout-template previews rely on.
   def show?
-    return true if admin?
+    return true if admin? || owner?
```
maebeale, line 59: 🤖 _From Claude:_ This widens the public event page for owners: previously an owner could not view their own unpublished or ended event at all, so the templates gallery preview links were dead for drafts. Scope is owner-on-own-event only, but it is the one authorization change in this PR — worth a deliberate look.

PR #2451, opened 2026-08-31
Sort Tailwind classes across the whole tree

maebeale:
🤖 suggested review level: 1 Skim 👀 formatting-only Tailwind class reordering, no logic or copy changes

Runs `bin/tw-sort --all` once over the whole tree so canonical Tailwind class order stops arriving one file at a time and future PRs don't mix reordering noise into their real diffs.

- 197 files (`app/views` + `app/frontend/javascript`) — the ones still unsorted; ~830 were already canonical.
- Class order inside an attribute has no CSS effect. Nothing else changed: every added/removed line is inside a `class` attribute.
- rustywind also collapses the multi-line class attributes it rewrites, which is why the line count drops. Attributes containing an ERB or JS interpolation keep their shape, sorted around the interpolation.
- `bin/tw-sort --all --check` is now clean, so `ai/tw-sort` on a feature branch only ever flags classes that branch actually wrote.

Not included: the ~40 helpers/decorators that build class strings in Ruby. rustywind doesn't scan `.rb`, and those need a custom regex — worth a separate look if we want them canonical too.

### DIFF
--- a/app/frontend/javascript/controllers/event_staff_bio_controller.js
+++ b/app/frontend/javascript/controllers/event_staff_bio_controller.js
@@ -50,7 +50,7 @@ export default class extends Controller {
     const link = editPath
       ? `<a href="${editPath}" class="text-xs font-medium text-blue-600 hover:text-blue-800">Edit on profile</a>`
       : ""
-    return `<div class="flex items-center justify-between mb-1"><span class="text-sm font-medium text-gray-700">Profile bio</span>${link}</div>`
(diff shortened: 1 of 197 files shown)


maebeale:
@jmilljr24 this gets us in line w tw-sort -- i eyeballed all files and spot checked in the ui

jmilljr24:
@maebeale this will hopefully be fine. Just a reminder, we did agree the plan was to only do files that a pr touched.

maebeale:
@jmilljr24 i didn't remember that. i just saw it kept getting skipped on files touched bc it was going to create too big of a diff, so i ripped off the bandaid

PR #2461, opened 2026-08-31
Break lazy-frame away-links out to _top so they stop Oopsie-ing

maebeale:
🤖 suggested review level: 3 Read 📖 explicit `_top` breakout on lazy-frame away-links across a few views + guard specs

Facilitators clicking a name, org, byline, or icon inside several filterable tables hit the "Oopsie!" box instead of the page they wanted — the same Turbo frame-missing bug the bulk-payments page had, now swept across every lazy `*_results` frame.

## What was broken
An audit of all ~35 lazy results frames found **12 away-navigating links** in 7 frames that loaded their destination *into* the results frame (no `_top` breakout) → frame-missing → Oopsie:
- **Registrant roster** (cross-event attendees): name, organization, scholarship, CE, and events-attended links.
- **Comment feed** source chip (drives two frames: `comments_results`, `person_comments_results`).
- **Author byline** (`credited_author_link`) on the workshop-variations and monthly-reports tables.
- **Admin activity/visit** tables: user links, the visit→activities count link, and the details link.

## Fix — explicit `_top` at each call site that needs it
Every away-link that renders inside a lazy frame carries `data: { turbo_frame: "_top" }` at the call site, matching the codebase's existing convention. Helpers stay plain — no magic default baked into `person_profile_button` / `person_edit_button` / `user_button` / `organization_profile_button` / `credited_author_link`; they just forward whatever `data:` a caller passes.
- Fixed the shared `_registrant_roster` partial so every away-link breaks out uniformly wherever it's reused.
- Fixed the comment source chip, the two author bylines, and the admin activity/visit one-offs.

## Tests
- Request-level guard specs assert the in-frame away-links render with the `_top` breakout — fast and deterministic (no browser), and they'd have caught all 12 (roster, comment chip, ahoy, payments, people, professional licenses, and the workshop-variation + monthly-report bylines).
- Helper specs assert each button/byline helper forwards a caller's `data-turbo-frame` onto the link.
- Filled three missing mode-1 (frame-500) request specs: `payments`, `users`, `video_recordings`.

## Follow-up (not in this PR)
The deeper root is that Turbo traps in-frame links *by default* and the repo relies on remembering `_top` at each call site. A safer design would be `target="_top"` on the frame itself (safe-by-default), already used by `events/bulk_payments`. Converting the rest is a deliberate, repo-wide call (pagination/sort drivers need re-tethering), so it's flagged as a separate follow-up rather than done here.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -11,11 +11,13 @@ module ApplicationHelper
   # anonymous and legacy free-text credits never link to a profile.
   def credited_author_link(record, **link_options)
     person = record.author_credit_person
-    if person&.profile_is_searchable
-      link_to record.author_credit, person_path(person), **link_options
-    else
-      record.author_credit
-    end
+    return record.author_credit unless person&.profile_is_searchable
+
+    # Break out of any lazy results frame this byline renders in — a no-op on a
+    # full page, but it stops the person link from loading into (and "Oopsie"-ing)
+    # a `*_results` frame. Callers can still override via their own data: hash.
+    link_options[:data] = { turbo_frame: "_top" }.merge(link_options[:data] || {})
+    link_to record.author_credit, person_path(person), **link_options
   end
 
   # The credited author as an admin edit-person card, for the workshop log /
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -32,8 +32,7 @@ module ApplicationHelper
     person_edit_button(person,
                        display_name: record.author_credit,
                        compact: true,
-                       width_class: "inline-flex w-fit",
-                       data: { turbo_frame: "_top" })
+                       width_class: "inline-flex w-fit")
   end
 
   # The person an author picker should show. Only the record's own author counts —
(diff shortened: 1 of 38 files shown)


maebeale:
@jmilljr24 as far as i can tell the issue w bulk payments is fixed and there isn't an easy revert option. 
this pr is going a step further to try and find any other oopsie's anywhere that are missing `top`.

jmilljr24:
This needs a deeper look so please wait.

jmilljr24 (changes requested):
I get the intent behind this but I really don't think its the smart solution across the board. Rails is convention of configuration and this throws that out the window. `_top` should be used explicitly when needed. I know you've said turbo isn't your strong suit so we can chat through this if you want but at this point my vote is no.

maebeale:
@jmilljr24 cool i'll remove it from the helpers and make this add any missing top's

maebeale:
@jmilljr24 i updated this pr to remove the top from any helpers, and added a few missing top's w tests

jmilljr24 (approved):

PR #2462, opened 2026-08-31
Add a slider (0–100) form-field type, and use it for the evaluation

maebeale:
🤖 suggested review level: 3 Read 📖 new answer type + its rendering/validation, slider-aware results rollup, and the seed conversion

Adds a **"Slider (0–100)" question type** and wires it through the annual evaluation: the survey's percentage questions become drag-to-set sliders, and the results page summarizes them as percentages.

Builds on the merged numeric-results rollup (#2477) and the merged seed (#2479), both now on `main`.

### What / why
- **Slider answer type** (`:slider`, appended to the enum): a range handle with a live value bubble, 0–100, stored as a whole number, with server-side range validation. Renders on the public fill page + admin preview; selectable in the form-builder type dropdown; a `slider` Stimulus controller keeps the handle and value in sync.
- **Results rollup** (`FormResponseAggregator`): sliders join number fields as numeric questions; a slider report is flagged `percent`, so `forms/_numeric_card` headlines the average with a `%` and draws a 0–100 fill bar. Count questions stay plain totals.
- **Seed:** the evaluation's percentage questions are now sliders (`ae_slider!`); the yearly-reach counts stay number fields (`ae_number!`).

### Notes for reviewer
- A slider stores the same 0–100 whole number a number field would, so the seed conversion is presentation-only — no data or rollup change beyond the `%` framing.
- Bounds are fixed 0–100 for now; per-field min/max/step is a clean follow-up.

### Anything else to add?
Some percentage sliders inherit narrow widths (third/quarter) from when they were number inputs — they work but read tight; happy to widen them if you'd like.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -410,7 +410,7 @@ end
 - `searchable_checkbox` — TomSelect checkbox-style multi-select
 - `searchable_select` — Tom Select autocomplete
 - `share_url` — URL sharing/copying
-- `slider` — Keeps a slider form field's range handle and number box in step (and fills the track); backs the `:slider` answer type
+- `slider` — Syncs a slider form field's range handle to its hidden input, shows the picked number in a bubble above the handle, and fills the track; backs the `:slider` answer type
 - `sortable` — Drag-drop sorting (SortableJS); persists order via a per-row PUT (used by categories index and the registration ticket callouts editor)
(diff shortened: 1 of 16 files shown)


maebeale:
@jmilljr24 afaik this is the last remaining feature for them to have parity w their current annual reports

jmilljr24 (approved):

PR #2473, opened 2026-08-31
Nest organizations under a parent org

### LINKED ISSUE #508: Build a way to nest Organizations
They used to make two Projects for organizations that did both Adult and Children's Windows, but now want them all under one Organization roof. We maybe should build a "demote" button to convert it from a Project to a ProjectHistory record or something?

maebeale:
🤖 suggested review level: 3 Read 📖 new self-referential nesting on Organization — a migration + a few model/view additions, no wide blast radius

Organizations can now sit under a parent "roof" org, so the FileMaker import can pull the historically-split Adult/Children program records back under one organization — staff wanted them grouped, not duplicated.

Closes rubyforgood/awbw#508

### What's here
- **Model** — one nullable `parent_id` self-FK (`belongs_to :parent` / `has_many :children`), mirroring FileMaker's Organization `ParentID`. Arbitrary depth, with `#ancestors` / `#descendants` / `#root` helpers.
- **Structural only** — a parent does *not* aggregate its children's affiliations/sectors/workshops/etc. yet. Deliberate first cut; roll-ups are a later change (ADR-0004 D2).
- **Safe by construction** — cycle guard (can't parent yourself or a descendant), and deleting a parent un-nests its children (`dependent: :nullify`) rather than deleting real programs.
- **UI** — admin sets the parent from a "Parent organization" picker on the org edit form (reuses the `remote-select` pattern, excludes self + descendants). The profile shows the relationship both ways: a "Part of …" link up and a "Nested organizations" list down.

### Notes for the reviewer
- Decision record: `docs/adr/0004-nested-organizations.md`.
- Added a Features & tips entry (admin-facing) and an AGENTS.md line for the new model capability.


app/views/organizations/_form.html.erb
```diff
@@ -25,6 +25,33 @@
...
+                      collection: parent ? [ [ parent.remote_search_label[:label], parent.id ] ] : [],
+                      include_blank: "None",
+                      hint: "Nest under a parent “roof” org",
+                      input_html: {
+                        class: "w-full rounded-md border-gray-300 shadow-sm focus:ring-blue-500 focus:border-blue-500",
+                        data: {
+                          controller: "remote-select",
+                          remote_select_model_value: "organization",
+                          remote_select_exclude_value: ([ f.object.id ] + f.object.descendants.map(&:id)).compact.join(","),
+                          remote_select_truncate_chars_value: 22
```
jmilljr24: I can see why you added this. My personal suggestion would be to just place the search box under the Org name input so they would be the same width. This search box is tiny in this position (so not a great experience typing) and drives the need for this extra remote-select code added in the PR. 

I don't love added extra things to the remote-select controller if we can help it because it is used in so many places. Keeping it simple will help keep it maintainable.

This is an example of my concern about design driving the code instead of taking a design compromise with simpler code.

maebeale: @jmilljr24 was thinking same to back out of remote-select change anyway just to not touch it. still, having it be an expandable box that wraps text isn't a great ui everywhere it is, so was considering this signal enough to make the needed change, not this was driven by a short width placement.

maebeale:
@jmilljr24 since you've already reviewed, i'll take out the remote-select change and would love to merge and not wait.

jmilljr24 (approved):
Sounds good. I thought the title had something to do with changes in staging.

This looks good matching the FM pattern. Would love to know your thoughts some other time about the structure and if it's something we would eventually want to move away with or if its a good db pattern.

PR #2483, opened 2026-09-01
Show outgoing email sender as "AWBW Programs"

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained mailer change — display name only, covered by updated specs

Recipients now see **AWBW Programs** as the email sender instead of the bare "programs" — a friendlier, on-brand From name after our first successful bulk send.

- Display name only: the sending address and domain are unchanged, so **deliverability / SPF / DKIM / DMARC are untouched** — no spam risk.
- Centralized in `ApplicationMailer.sender`; `DeviseMailer` and `EventMailer` share it, so every generic-mailbox email is consistent.
- Attributed admin invites still come from the generic mailbox as "AWBW Programs" — never under the admin's personal name (existing behavior preserved).
- If `REPLY_TO_EMAIL` is set in prod it must stay a **bare address** (the code adds the name); a value that already includes a display name would double-wrap.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -2,7 +2,15 @@ class ApplicationMailer < ActionMailer::Base
   helper ApplicationHelper
   include Rails.application.routes.url_helpers
 
-  default from: ENV.fetch("REPLY_TO_EMAIL", "programs@awbw.org")
+  FROM_NAME = "AWBW Programs".freeze
+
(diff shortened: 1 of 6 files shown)


maebeale:
@jmilljr24 this was a request from before. i did more research and this will not hurt and might help w spam detection.

jmilljr24 (approved):

PR #2489, opened 2026-09-01
Deliver contact form emails in the background

maebeale:
🤖 suggested review level: 3 Read 📖 small delivery-path change on one controller plus a shared error list

Contact form submissions were 500ing when SMTP was slow — a `Net::ReadTimeout` fired mid-request and the sender's message was lost (a facilitator reported it). Now the emails send in the background so the request can't hang on SMTP.

- `POST /contact_us` was the only mail flow still using `deliver_now`; switched both emails to `deliver_later`, matching how every other mail is sent.
- Rendered email bodies are still captured on the notifications (persist via the in-process `.message`, no SMTP).
- Mailer args now go through a permitted symbol-keyed hash so ActiveJob can serialize them.
- Added `Net::ReadTimeout`/`Net::OpenTimeout` to the shared `EmailDeliveryErrorHandler` list — the exact class that slipped through — as a safety net for the Devise flows that still deliver synchronously.

### DIFF
--- a/app/controllers/contact_us_controller.rb
+++ b/app/controllers/contact_us_controller.rb
@@ -36,13 +36,16 @@ class ContactUsController < ApplicationController
 
     user = current_user if user_signed_in?
     noticeable = user&.person || user
+    contact_us = contact_us_params
 
-    # Send emails
-    confirmation_mail = ContactUsMailer.confirmation(params[:contact_us], user).deliver_now
-    admin_mail = ContactUsMailer.hello(params[:contact_us], user).deliver_now
+    # Deliver in the background so a slow/hung SMTP server can't time out the request.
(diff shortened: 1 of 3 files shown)


maebeale:
@jmilljr24 claude suggested we route contactus through same job handling and deliver_later we use for all the other emails, and i think that's the right move. seems reasonable to add these errors to the EmailDeliveryErrorHandler, too.

jmilljr24:
> @jmilljr24 claude suggested we route contactus through same job handling and deliver_later we use for all the other emails, and i think that's the right move. seems reasonable to add these errors to the EmailDeliveryErrorHandler, too.

~Sounds good. Can you merge this into the production branch as well so I can deploy as a hot fix?~

Nevermind. It was a mandrill server issue. This change can go with the normal deploy.

jmilljr24 (approved):

PR #2501, opened 2026-09-02
Record payments during registrant import

maebeale:
🤖 suggested review level: 3 Read 📖 contained addition to one service + its preview view; money math (payment + comping discount) capped at remaining cost, covered by specs

Roster imports now mark registrations paid, so facilitators don't have to re-enter payments by hand after importing a spreadsheet of attendees.

## What changed
- **New optional CSV columns**: `AmountPaid`, `PaymentType` (cash / check / card, defaults to cash), `CheckNumber` (for checks).
- Each paying row **finds-or-creates a payment**, then an **allocation** applied to the registration.
- **Partial amounts are comped**: when `AmountPaid` is less than the event cost, a `Discount` allocation covers the difference (e.g. $50 paid on a $750 event → $700 discount), so the registration reads **Paid**. A full payment leaves no discount.
- The preview shows a **Payment** column (with the comping discount) and **Payments / Amount / Discounts / Comped** tiles before anything is saved.

## Why it's safe with money
- The payment match is keyed by the import's own metadata **+ person + amount + type + event id**, so re-running the import adds nothing and imports into different events never steal each other's payments.
- Both allocations are capped at the registration's `remaining_cost`; **free events ignore the amount** (nothing to allocate against).
- It reuses only its *own* import-created payments — it does not attach unrelated existing payments. (Matching pre-existing FileMaker/Stripe payments from the payments index is a possible follow-up — flagged as riskier since the sheet has no reliable key to match those on.)

## Notes
- **The comp always tops up to the full cost.** Every underpaid row is treated as fully settled (paid + comped). If some imported registrants genuinely still owe a balance, that's a different rule — say the word and I'll gate it behind a column.
- Bad amount, unknown payment type, and a check row missing its number are surfaced in the preview and simply record no payment (the registration is still created).

### DIFF
--- a/app/controllers/event_registration_imports_controller.rb
+++ b/app/controllers/event_registration_imports_controller.rb
@@ -93,6 +93,7 @@ class EventRegistrationImportsController < ApplicationController
       "#{", #{result.registrations_promoted} promoted to attended" if result.registrations_promoted.positive?}" \
       "#{", #{result.organizations_to_reconcile} orgs to reconcile" if result.organizations_to_reconcile.positive?}" \
       "#{", #{result.payments_recorded} payments (#{helpers.dollars_from_cents(result.payments_amount_cents)})" if result.payments_recorded.positive?}" \
+      "#{", #{result.discounts_recorded} discounts (#{helpers.dollars_from_cents(result.discounts_amount_cents)})" if result.discounts_recorded.positive?}" \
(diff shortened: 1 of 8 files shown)


maebeale:
@jmilljr24 this is what's going to be talked about tonight

jmilljr24 (approved):

PR #2503, opened 2026-09-02
Greet email recipients by first name only

maebeale:
🤖 suggested review level: 3 Read 📖 copy-only change across email greetings, plus 3 spec assertions and a features.yml entry

Portal emails now open with just the recipient's first name ("Hello Jane,") instead of their full name — warmer, less form-letter.

- **Greetings only.** 18 templates (HTML + text) across event and notification mailers now use `@person.first_name` (validated present) or `@user.first_name_or_email` (falls back to email when a user has no linked person).
- **Staff FYI/notification emails unchanged** — they keep full name + email so staff can still tell two people apart.
- Devise, contact-us, and scholarship-acceptance emails already greeted by first name; left as-is.
- Updated 3 mailer specs that asserted the full name appeared in the greeting.
- Added a `config/features.yml` entry.

### DIFF
--- a/app/views/event_mailer/bulk_payment_confirmation.html.erb
+++ b/app/views/event_mailer/bulk_payment_confirmation.html.erb
@@ -2,7 +2,7 @@
 
 <div style="margin-top: 36px; text-align: left;">
   <p>
-    Hello <strong><%= @person.full_name %></strong>,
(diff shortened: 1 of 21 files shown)


maebeale:
@jmilljr24 here's their request, waiting for answers in slack

jmilljr24 (approved):

PR #2518, opened 2026-09-03
Add a Close Program form role that deactivates affiliations

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new agreement role with affiliation-deactivation processing across the form builder, submission pipeline, and admin panel

Adds a "Close program" form so a facilitator or organization can tell us their AWBW program is ending, and have their affiliations at that organization closed automatically — instead of an admin hunting down and end-dating each affiliation by hand.

Built `close_program` as the mirror of the existing agreement roles (`registration`/`new_job`/`reinstatement`): those create affiliations, this one deactivates them, reusing the same processing-panel and org-linking plumbing. On submission, an exact organization-name match auto-processes (end-dates the Facilitator affiliation to the given date, plus the job affiliation when they say they're also leaving, with a "Program closure" comment); an ambiguous name waits for an admin to link the org on the processing panel, which runs the same `AffiliationServices::CloseProgram` service. Deactivation is end-dating only — it keeps the historical row and comment, and the org's status closes to Inactive on its own via existing callbacks, so there's no data migration.

### UI Testing Checklist

- [ ] In the form builder, create a form with role **Close program**; confirm the section shows organization name, effective date, reason, and "Are you also leaving your job?".
- [ ] Publish it, submit at its `/f/` link with an organization name that exactly matches an org where the submitter has a Facilitator affiliation; confirm that affiliation is end-dated with a "Program closure" comment.
- [ ] Submit with a non-matching / ambiguous org name; confirm nothing changes and the submission's processing panel prompts you to link the organization, which then deactivates.
- [ ] Answer "Yes" to "Are you also leaving your job?" and confirm the job affiliation is end-dated too.

### Anything else to add?

Seeded by `CloseProgramFormSeeder` (+ `data:seed_close_program_form` rake task) as an unpublished template — a super-admin gives it a slug and publishes it before use; the dev seed publishes it with a sample submission to eyeball.


maebeale:
@jmilljr24 this is FM related so they can capture someone closing out a program

jmilljr24 (approved):

PR #2536, opened 2026-09-11
Redirect create actions with 303 See Other so Turbo advances

maebeale:
🤖 suggested review level: 3 Read 📖 one-line status change on 6 create redirects + a request spec each

Fixes a Turbo bug where a successful create didn't advance the page: the record saved but the user stayed on the form, so re-clicking Save produced **duplicate records**.

- **Why:** Turbo Drive needs a **303 See Other** (not 302) to navigate after a form POST. These `create` actions redirected with a plain 302, so Turbo didn't move on. The `update` actions already used `:see_other` — this just aligns `create` to match.
- Added `status: :see_other` to the create-success redirect in **community_news, stories, events, workshops, resources, video_recordings**, each with a request spec asserting 303.

### Scope
- The same 302-after-POST gap exists on other create/destroy actions across the app (e.g. organizations, grants, people, features, sectors…). This PR fixes the content-form controllers where the duplicate-record symptom was reported. Happy to do a broader sweep in a follow-up if wanted — a blanket change is avoided here because some of those redirects are guard/error paths where 303 isn't the right call.

### DIFF
--- a/app/controllers/community_news_controller.rb
+++ b/app/controllers/community_news_controller.rb
@@ -80,7 +80,7 @@ class CommunityNewsController < ApplicationController
 
     if success
       redirect_to @community_news,
-                  notice: "Community news was successfully created."
+                  notice: "Community news was successfully created.", status: :see_other
     else
       @community_news = @community_news.decorate
       set_form_variables
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -563,7 +563,7 @@ class EventsController < ApplicationController
 
     respond_to do |format|
       if success
-        format.html { redirect_to @event, notice: "Event was successfully created." }
+        format.html { redirect_to @event, notice: "Event was successfully created.", status: :see_other }
         format.json { render :show, status: :created, location: @event }
       else
         set_form_variables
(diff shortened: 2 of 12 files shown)


jmilljr24 (approved):
I did a good bit a digging and finally found an [explanation](https://github.com/hotwired/turbo/issues/84) that made sense to me.

This pr would be considered "best practice" according to the turbo docs. A browser may handle other 3xx responses fine but the official support if for a `redirect_to` to set 303 with see_other.

Couldn't tell you if that was actually the root cause or it was a weird timing/server issue but I'm good with add `see_other` on this. Thanks

spec/system/community_news_create_redirect_spec.rb
jmilljr24, line 1: @maebeale Did you intend to leave this in?

maebeale: ack, i merged before seeing this comment. ok w you to leave in?

jmilljr24: I commented after the merge so not on you. Yeah, no biggie. Just brought it up in case you had different intentions for it and forgot. I'm not concerned with it.

PR #2540, opened 2026-09-16
Track invoice page views with Ahoy

maebeale:
🤖 suggested review level: 3 Read 📖 small, contained tracking calls in two controllers plus specs

Records who opens an invoice so facilitators/admins can see invoice activity in analytics — the two invoice pages previously emitted no Ahoy event.

- Emits `view.invoices` from the registrant invoice (`Events::RegistrationsController#invoice`) with `event_id` + `registration_id`.
- Emits `view.invoices` from the admin/bulk-payment invoice (`Events::InvoicesController#show`) with `event_id` (+ `submission_id` for a bulk-payment invoice).
- Uses the existing string-based `track_view` pattern (`EventInvoice` is a presenter, not an AR model, so resource-based tracking doesn't apply).

## Download tracking — not included
"Download PDF" is a client-side `window.print()` with no server hit, so there's nothing to track server-side today. Capturing *who downloaded* would need a client-side beacon (new Stimulus controller + a tracking endpoint) — flagged for a decision rather than added here.


maebeale:
@jmilljr24 i'd like to turn this on now, and we can figure out if we want to do something different later

jmilljr24 (approved):

PR #2541, opened 2026-09-16
check invited status via welcome_sent column

jmilljr24:

It is still possible to gain portal access before being invited if the user account is created but no invite is set. The user could use the password reset flow.

I know we fixed some other gap a few months ago but I can't remember what exactly is was. This change would require a db backfill but is minor.

Use `welcome_sent_at` column to verify that the user was invited. The alternative I think would be to set the account at `locked` (devise built in) when the account is created. This decouples our custom code and keeps things consist with devise but I haven't looked at how complex that would be.
### Anything else to add?

```
User.where.not(confirmed_at: nil)
    .where(welcome_instructions_sent_at: nil)
    .where.not(super_user: true)
    .find_each { |u| u.update_columns(welcome_instructions_sent_at: u.confirmed_at) }
```


app/models/user.rb
```diff
@@ -253,8 +253,7 @@ def set_welcome_instructions_token!
   def clear_welcome_instructions_token!
     update_columns(
       welcome_instructions_token: nil,
-      welcome_instructions_created_at: nil,
-      welcome_instructions_sent_at: nil
+      welcome_instructions_created_at: nil
```
maebeale, line 256: Why did you keep created_at instead of sent at? I'm a little fuzzy and don't remember why we had both, but I'm pretty sure there was something in the flow and handling resends and resetting of stale tokens. Maybe it was only for the launch blast situation? I think your change to active for authentication removes the reason we had to nil out before?

I think the difference is moot for the detection purposes, but, curious if this was an educated choice to use created_at? Pretty sure we need this backfill for that filter I showed them tonight to work too -- do you know if that's true? We should update for this the field that keys off. Or maybe with this backfill we update both values?

maebeale (approved):

spec/factories/users.rb
```diff
@@ -15,6 +15,21 @@
...
+    # A factory user is "invited" by default so User#active_for_authentication?
+    # lets them sign in. Written after create (update_column - no callbacks) so it
+    # does not fire the real invite side effect, after_commit :start_membership,
+    # which would otherwise mint a Membership for every factory person. Pass
+    # invited: false for a genuinely uninvited user.
+    transient do
+      invited { true }
+    end
+
+    after(:create) do |user, evaluator|
```
maebeale, line 27: Doesn't this change our tests such that no factories will be able to stimulate the first time flow?

maebeale (approved):

app/models/user.rb
jmilljr24: I assumed sending was different than just creating the welcome.

Yes it would need a backfill and is a lot more heavy handed than used devise locked. Devise locked would only need to be on the handful of accounts in the november training that may be created but not invited. We can assume the state of every other current user account is correct. With using `welcome` we need to go through every account we think should have access and set this value in a backfill.

PR #2542, opened 2026-09-16
user device lock on user creation

jmilljr24:

Alternative to #2541 using devise locked.

### Anything else to add?

```
User.where(confirmed_at: nil, welcome_instructions_sent_at: nil, locked_at: nil)
    .where.not(super_user: true)
    .update_all(locked_at: Time.current)
```


app/services/event_registration_services/process_confirmation.rb
```diff
@@ -56,7 +56,8 @@ def create_user_account
         password_confirmation: password,
         person: @person,
         created_by: @current_user,
-        updated_by: @current_user
+        updated_by: @current_user,
+        locked: true
```
maebeale: I don't love this. I think we should go the other route w welcome instructions bc that's accurate to the account state whereas this muddies things, alongside manual locks, autolocks, etc.

PR #2543, opened 2026-09-18
Profile change requests with an admin review queue

maebeale:
🤖 suggested review level: 5 Inspect 🔬 new model + migration, admin apply-on-approve logic, notification wiring, and owner submission flow

Facilitators can't edit their primary email, organization name, or affiliation details — this gives them a way to request those corrections and gives admins an in-app queue to act on, instead of the changes living only in an email.

### Owner side
- On their own profile, each locked field has a "Request a change" link → a small per-field form (shows the current value, takes the requested value and/or a note).
- Submitting creates a `ProfileChangeRequest` **and** the admin + submitter notification email thread (kinds `profile_change_requested` / `_fyi`), now pointing at the structured record.

### Admin side
- New **Changes requested** queue (`/profile_change_requests`), linked from the admin home (with a pending count) and surfaced on each person's edit page.
- Four actions: **Approve** (auto-applies where it can — primary email via the existing `ProcessEmailChange` confirmation flow, org rename — via `ProfileChangeRequests::Apply`), **Update manually** (jumps to the right editor), **Mark as resolved**, **Decline**.
- `status` = pending / resolved / declined; `resolution_method` = approved / manual. The row shows **Resolved** whether approved or manually resolved.

### Notes for reviewers
- Owner reachability still rides the staged `OWNER_PROFILE_EDIT` flag (owners can only see the profile edit page — and thus the request links — when it's on). The request model/queue/apply logic is independent of the flag.
- Auto-apply is intentionally limited to the two structured fields; affiliation requests are free-form and resolved by a manual edit.
- Org rename applies to the person's `primary_organization` and ripples to everyone affiliated — flagged since an admin should be aware before approving.
- Server-side field strip for owner self-edit still covers only email + affiliations (see prior note); broader allowlist remains a follow-up.


app/controllers/people_controller.rb
```diff
@@ -644,6 +645,18 @@ def reject_locked_license_changes!(attrs)
     end
   end
 
+  # Server-side backstop for owner self-service profile editing: primary email
+  # and affiliations are rendered read-only for owners, who can only *request*
+  # changes to them. Drop those from a non-admin owner's submission so a crafted
+  # request can't slip past. No-op for admins, who edit them directly.
+  def reject_owner_locked_changes!(attrs)
```
jmilljr24: Can we use action policy scoped params?

app/controllers/profile_change_requests_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+    end
+  end
+
+  def edit
+    authorize! @request, to: :update?
+    @person = @request.person
+  end
+
+  def update
+    authorize! @request, to: :update?
```
jmilljr24:
```suggestion
    authorize! @request
```

app/controllers/profile_change_requests_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+      redirect_to edit_person_path(@person, anchor: "affiliations"), status: :see_other,
+                  notice: "Thanks — we've sent your change request to AWBW staff."
+    else
+      @person = @request.person
+      render :new, status: :unprocessable_content
+    end
+  end
+
+  def edit
+    authorize! @request, to: :update?
```
jmilljr24:
```suggestion
    authorize! @request
```
I'd either add a policy action for edit or use alias in the policy.

jmilljr24: Same comment for all the controller actions in this file.

app/controllers/profile_change_requests_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+  def respond_with_updated_row(notice)
+    respond_to do |format|
+      format.turbo_stream do
+        render turbo_stream: turbo_stream.replace(
+          ActionView::RecordIdentifier.dom_id(@request),
+          partial: "profile_change_requests/request",
+          locals: { request: @request }
+        )
+      end
+      format.html { redirect_back fallback_location: profile_change_requests_path, notice: notice }
```
jmilljr24:
```suggestion
      format.html { redirect_back_or_to profile_change_requests_path, notice: notice }
```
`redirect_back` is soft deprecated.

app/controllers/profile_change_requests_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  def respond_with_updated_row(notice)
+    respond_to do |format|
+      format.turbo_stream do
+        render turbo_stream: turbo_stream.replace(
+          ActionView::RecordIdentifier.dom_id(@request),
+          partial: "profile_change_requests/request",
+          locals: { request: @request }
+        )
+      end
```
jmilljr24:
```suggestion
      format.turbo_stream 
```
If you create a `turbo_stream.erb` `dom_id is available without calling `ActionView....` and its a bit more idiomatic

jmilljr24: Also you will want to add `flash.now` if you want the same notice that you are using for html response

app/models/person.rb
```diff
@@ -92,6 +93,13 @@ class Person < ApplicationRecord
   validates :email_type, inclusion: { in: %w[work personal] }, allow_blank: true
   validates :email_2_type, inclusion: { in: %w[work personal] }, allow_blank: true
 
+  # Owner self-service profile editing is staged behind this flag so it can be
+  # trialed in staging before PersonPolicy#edit? flips from admin-only to
+  # admin-or-owner at profile launch. Default off; set OWNER_PROFILE_EDIT=true.
+  def self.owner_editing_enabled?
```
jmilljr24: I'm not apposed to using an env but just a thought.

We have `!Rails.env.production?` for `enable?` on membership. We could do the same for this. On less thing to do on DO

jmilljr24 (commented):
I like how this is setup.

The change request flag is not showing on affiliation change like the others. The validation does work though if a person goes in and tries to make another change when one is still pending.
Image
No route for this noticiable.
Image

